PR #7, opened 2015-09-30
Update namespace_changes_t10.dart

whesse:
The self-reference in the test was not updated when the name was changed.


whesse:
@alsemenov

PR #70, opened 2016-08-26
Remove 'dart:mirrors' import.

floitschG:
dart:mirrors seems to be unused.
Since mirrors are unsupported in the fast-startup emitter and usually slow down compilation I removed it.


sgrekhov:
It's a safe change. Merged

whesse:
Thanks.  I'll now roll this new version of co19, with all recent fixes, into the Dart checkout.

Issue #165, opened 2018-10-10
Test LibTest/io/Directory/stat_A01_t02.dart is flaky


mkustermann:

mkustermann:
The change in https://github.com/dart-lang/co19/pull/166 should fix it.

PR #166, opened 2018-10-10
Wait with deleting a directory until the test finished, only call asyncEnd() as the very last step

mkustermann:
Fixes https://github.com/dart-lang/co19/issues/165


mkustermann:
Not sure why the CLA is not shown as accepted. I'll try another PR.

PR #176, opened 2018-10-26
Use sandbox for IO tests. Link, Directory and FileSystemCreateEvent fixed

sgrekhov:
This is the first bunch of the chenges in IO tests. Each tests uses its own sandbox to play with and deletes it in the end


LibTest/io/Directory/Directory_A02_t03.dart
```diff
@@ -21,11 +21,12 @@ import "../../../Utils/expect.dart";
 import "../file_utils.dart";
 
 main() {
-  File file = getTempFileSync();
+  Directory sandbox = getTempDirectorySync();
```
mraleph: This pattern repeats a lot - I wonder if it is worth introducing helper like so:

```dart
// sandbox.dart
Future<void> inSandbox(dynamic testMain(Directory sandbox)) async {
  Directory sandbox = getTempDirectorySync();
  try {
    await testMain(sandbox);
  } finally { 
   sandbox.delete(recursive: true);
  }
}
```

```dart
import '../sandbox.dart';

main() => inSandbox(_main);
_main(Directory sandbox) {
  File file = getTempFileSync(parent: sandbox);
  Directory dir = new Directory(file.path);
  Expect.equals(file.path, dir.absolute.path);
}
```

LibTest/io/Directory/Directory_A02_t03.dart
```diff
@@ -21,11 +21,12 @@ import "../../../Utils/expect.dart";
...
+  File file = getTempFileSync(parent: sandbox);
```
mraleph: Note that by running in sandbox you don't need to generate temp file names anymore for the nested files.

LibTest/io/Directory/watch_A01_t01.dart
```diff
@@ -21,15 +21,14 @@ import "../file_utils.dart";
 main() {
   Directory dir = getTempDirectorySync();
   asyncStart();
-  StreamSubscription s = dir.watch().listen((FileSystemEvent event) {
```
mraleph: there is a lot of duplicated code related to handling of watch stream subscriptions. Consider factoring it out into a helper.

LibTest/io/Directory/Directory_A02_t03.dart
sgrekhov: I'd like to left is as is. Pattern repeats, but tests became simpler and much more obvious for the reader. If we need to change it "Find in path" option will hepl

mraleph: ok

LibTest/io/Directory/Directory_A02_t03.dart
sgrekhov: getTemp....() is preferrable anyway. If, for example, file is errorneously created outside of the sandbox it won't be deleted. But file/directory/link, created by getTemp...() starts with tests name (like lastModified_A01_t01_7ecf3194-db60-11e8-95d5-bcaec534d2ad ) and this will help to identify the tests which doesn't clean up

Issue #196, opened 2018-11-26
co19_2/LibTest/io/FileSystemCreateEvent/isDirectory_A01_t06 flaked on Mac


mraleph:
It is currently unclear if this is a test issue or a VM issue.

I am going to add printing to verify.

```
FAILED: dartk-vm product_x64 co19_2/LibTest/io/FileSystemCreateEvent/isDirectory_A01_t06
Expected: Timeout Pass
Actual: RuntimeError

--- Command "vm" (took 833ms):
DART_CONFIGURATION=ProductX64 xcodebuild/ProductX64/dart --sync-async --ignore-unrecognized-flags --packages=/b/s/w/ir/.packages /b/s/w/ir/tests/co19_2/src/LibTest/io/FileSystemCreateEvent/isDirectory_A01_t06.dart

exit code:
255

stdout:
unittest-suite-wait-for-done

stderr:
Unhandled exception:
Expect.isFalse(true) fails.
#0      _fail (file:///b/s/w/ir/tests/co19_2/src/Utils/expect.dart:21:5)
#1      Expect.isFalse (file:///b/s/w/ir/tests/co19_2/src/Utils/expect_common.dart:48:5)
#2      _main.<anonymous closure> (file:///b/s/w/ir/tests/co19_2/src/LibTest/io/FileSystemCreateEvent/isDirectory_A01_t06.dart:33:16)
#3      testFileSystemEvent (file:///b/s/w/ir/tests/co19_2/src/LibTest/io/file_utils.dart:53:7)
#4      _RootZone.runUnary (dart:async/zone.dart:1379:54)
#5      _FutureListener.handleValue (dart:async/future_impl.dart:126:18)
#6      Future._propagateToListeners.handleValueCallback (dart:async/future_impl.dart:639:45)
#7      Future._propagateToListeners (dart:async/future_impl.dart:668:32)
#8      Future._completeWithValue (dart:async/future_impl.dart:483:5)
#9      Future.timeout.<anonymous closure> (dart:async/future_impl.dart:737:16)
#10     _RootZone.runUnary (dart:async/zone.dart:1379:54)
#11     _FutureListener.handleValue (dart:async/future_impl.dart:126:18)
#12     Future._propagateToListeners.handleValueCallback (dart:async/future_impl.dart:639:45)
#13     Future._propagateToListeners (dart:async/future_impl.dart:668:32)
#14     Future._completeWithValue (dart:async/future_impl.dart:483:5)
[...]

PR #197, opened 2018-11-26
Make io/FileSystemCreateEvent/isDirectory_A01_t06.dart fail verbosely.

mraleph:
Issue #196


mraleph:
This test has failed on the bots flakily but unfortunately the failure was not informative. 

This PR tries to make it more verbose.

Issue #193, opened 2018-11-19
LibTest/io/Stdin/first_A04_t01 should be made more robust


dcharkes:
This test has the same issue as https://github.com/dart-lang/co19/issues/182: 

> Using `stdin.last.then(...)` in  `run_process()` is not right, since `stdin` is `Stream<List<int>>`, i.e. a stream of bytes. The bytes can be delivered in any sizes, e.g. each individual byte can be devlivered on it's own. So it cannot be assumed that the bytes obtained in `stdin.last` is  the last line.

Same for `stdin.first.then()` and `stdin.elementAt(0).then()`.

From [LibTest/io/Stdin/first_A04_t01.dart](https://github.com/dart-lang/co19/blob/master/LibTest/io/Stdin/first_A04_t01.dart):

```dart
run_process(String arg) {
  if(arg == "first") {
    stdin.first.then((List<int> l) {
      print(new String.fromCharCodes(l));
    });
  } else {
    stdin.elementAt(0).then((List<int> l) {
      print(new String.fromCharCodes(l));
    });
  }
}

run_main() async {
  String executable = Platform.resolvedExecutable;
  String eScript = Platform.script.toString();
  List<int> res_first, res_elementAt;

  await Process.start(executable, [eScript, "first"], runInShell: true).then(
      (Process process) async {
        process.stdin.writeln("12345");
        await process.stdout.toList().then((out) { res_first = out[0]; });
    });

  await Process.start(executable, [eScript, "elementAt"], runInShell: true).then(
      (Process process) async {
    process.stdin.writeln("12345");
    await process.stdout.toList().then((out) { res_elementAt = out[0]; });
  });

  Expect.listEquals(res_elementAt, res_first);
}
```

We see the following flakiness on our build infrastructure:

```dart
FAILED: dartk-vm release_simdbc64 co19_2/LibTest/io/Stdin/first_A04_t01
Expected: Pass
Actual: RuntimeError

--- Command "vm_compile_to_kernel" (took 535ms):
[...]

Issue #184, opened 2018-11-05
co19_2/LibTest/io/Stdin/readLineSync_A03_t02 flaky failures on Mac


sjindel-google:
Here is an example:

https://ci.chromium.org/p/dart/builders/luci.dart.ci.sandbox/vm-kernel-mac-product-x64/704

mkustermann:
@sjindel-google When filing a co19 issue it is very much beneficial to actually look at the test and see why it is flaky and determine if the flakiness is actually caused by the test itself!

It appears to me from [LibTest/io/Stdin/readLineSync_A03_t02](https://github.com/dart-lang/co19/blob/master/LibTest/io/Stdin/readLineSync_A03_t02.dart#L29):
```dart
void run_process(String filename, Encoding enc) {
  String res = stdin.readLineSync(encoding : enc);
  File fl = new File(filename);
  fl.openWrite();
  fl.writeAsBytes(res.codeUnits);
}
```

Two issues stick out here:

* The `fl.openWrite()` opens the file and returns an `IOSink`, this returned sink is never used for anything!
* The `writeAsBytes` method is asynchronous and the process does not await it's return value.

I would suggest using one of these two methods and waiting for the operation to be completed:
```dart
runProcess(String filename, Encoding enc) async {
  String res = stdin.readLineSync(encoding : enc);
  File fl = new File(filename);

  // Either
  final sink = fl.openWrite();
  sink.add(res.codeUnits);
  await sink.close();

  // Or
  await fl.writeAsBytes(res.codeUnits);
}
```

iarkh:
FIXED

Issue #182, opened 2018-11-01
Make LibTest/io/Stdin/last_A02_t01.dart more robust


mkustermann:
From [LibTest/io/Stdin/last_A02_t01.dart](https://github.com/dart-lang/co19/blob/master/LibTest/io/Stdin/last_A02_t01.dart):

```dart
run_process() {
  stdin.last.then((List<int> l) { print(new String.fromCharCodes(l)); });
}

run_main() async {
  String executable = Platform.resolvedExecutable;
  String eScript = Platform.script.toString();
  int called = 0;

  await Process.start(executable, [eScript, "test"], runInShell: true).then(
      (Process process) async {
    process.stdin.writeln("Testme");
    process.stdin.close();
    await process.stderr.toList().then((errors){
      Expect.isTrue(errors.isEmpty);
    });
    await process.stdout.toList().then((out) {
      String res = systemEncoding.decode(out[0]);
      // Get rid from possible new line symbols here
      Expect.equals("Testme", res.trimRight());
      called++;
    });
  });
  Expect.equals(1, called);
```

There are at least two issues here:

* Using `stdin.last.then(...)` in  `run_process()` is not right, since `stdin` is `Stream<List<int>>`, i.e. a stream of bytes. The bytes can be delivered in any sizes, e.g. each individual byte can be devlivered on it's own. So it cannot be assumed that the bytes obtained in `stdin.last` is  the last line.

=> Instead something like `stdin.transform(utf8.decoder).transform(const LineSplitter()).last.then((lastLine) { ... });` will ensure to split the incoming bytes by line endings and gets the last line.

* Using `systemEncoding.decode(out[0])` to get the entire output of the subprocess is also not quite right, since it only looks at a prefix of the output.

=> Instead it should do something like `await process.stdout.transform(systemEncoding.decoder).join('')`.

We see the following flakiness on our build infrastructure:
```


FAILED: dartk-vm release_simdbc64 co19_2/LibTest/io/Stdin/last_A02_t01
Expected: Pass
Actual: RuntimeError

--- Command "vm_compile_to_kernel" (took 660ms):
[...]

PR #199, opened 2018-11-27
Fix for Stdin issues

iarkh:
Fix for issues #193, #182, #184


LibTest/io/Stdin/first_A01_t01.dart
```diff
@@ -27,10 +27,8 @@ run_main() async {
     await process.stderr.toList().then((errors){
       Expect.isTrue(errors.isEmpty);
     });
-    await process.stdout.toList().then((out) {
-      String res = systemEncoding.decode(out[0]);
-      // Get rid from possible new line symbols here
-      Expect.equals("Testme", res.trimRight());
+    await process.stdout.transform(systemEncoding.decoder).join().then((out) {
```
mkustermann: Isn't `Stream<T>.join()` returning a  `Future<String>` ?

Then this should be rather:
```dart
final String stdout = await process.stdout.transform(systemEncoding.decoder).join();
Expect.isTrue(stdout.startsWith("Testme"));
called++:
```

LibTest/io/Stdin/first_A04_t01.dart
```diff
@@ -43,7 +43,10 @@ run_main() async {
     await process.stdout.toList().then((out) { res_elementAt = out[0]; });
   });
 
-  Expect.listEquals(res_elementAt, res_first);
+  Expect.isTrue("12345\n\n".startsWith(String.fromCharCodes(res_first).
```
mkustermann: Please use `utf8.decode()` / `ascii.decode()` instead of `String.fromCharCodes`, ideally using the system encoding.

LibTest/io/Stdin/first_A04_t01.dart
```diff
@@ -43,7 +43,10 @@ run_main() async {
     await process.stdout.toList().then((out) { res_elementAt = out[0]; });
```
mkustermann: Same problem as above, `out` is a `List<List<int>>` there are no guarantees that `out[0]` will contain all the bytes.

LibTest/io/Stdin/first_A01_t01.dart
iarkh: Regarding the result check:
`Expect.equals("Testme", res.trimRight());` assumes that I get exactly `Testme` string (maybe with the trimmed line ending symbols) in the stdout, and `stdout.startsWith("Testme")` does not check that I don't have any unexpected symbols at the result string. 
Going to change to `Expect.isTrue("Testme\n\n".startsWith(stdout.replaceAll("\r\n", "\n")));` fimally.

iarkh: New fix is on review

PR #200, opened 2018-11-28
Make error information more informative or first_A02_t02 test.

sjindel-google:


LibTest/io/Stdin/first_A02_t02.dart
```diff
@@ -24,17 +30,23 @@ run_process() async {
...
+  await process.stdin.flush();
+
+  try {
+    await process.stdin.close();
+    int code = await process.exitCode;
+    Expect.equals(0, code);
+  } catch (e) {
+    print("Subprocess failed or exited prematurely!");
+    print("Exit code: ${await process.exitCode}");
+    print("stdout: " + await readOutput(process.stdout));
```
mkustermann: If the subprocess happens to print a lot of output (e.g. crash information, ...)  this test can timeout: After the kernel buffer of the pipe is full, the subprocess might block on a `write()`. 

Because of this situation it's always good to start draining stdout/stderr right after starting the process (and not waiting until it has quit)

LibTest/io/Stdin/first_A02_t02.dart
```diff
@@ -15,6 +15,12 @@
 import "../../../Utils/expect.dart";
 import "dart:io";
 
+Future<String> readOutput(Stream<List<int>> output) async {
+  List<List<int>> result = await output.toList();
+  if (result.isEmpty) return "";
+  return String.fromCharCodes(result.reduce((x, y) => x + y));
```
mkustermann: Normally we use e.g.
```dart
return output.transform(io.systemEncoding.decoder).join('');
```
(or `utf8.decoder`)

Which allows us to decode unicode characters as well.

LibTest/io/Stdin/first_A02_t02.dart
```diff
@@ -24,17 +30,23 @@ run_process() async {
...
-  await Process.start(executable, [eScript, "test"], runInShell: true).then(
-      (Process process) async {
-    process.stdin.writeln("1");
-    await process.exitCode.then((int code) async {
-      Expect.equals(0, code);
-        called++;
-    });
-  });
-  Expect.equals(1, called);
+  Process process = await Process.start(executable, [eScript, "test"], runInShell: true);
```
mkustermann: Could be `final`

PR #203, opened 2018-12-12
Add information to aid debugging to 'run_main'

sjindel-google:


LibTest/io/file_utils.dart
```diff
@@ -181,3 +181,14 @@ String getTempDirectoryPath({Directory parent, String name}) {
...
     entity.path.substring(entity.path.lastIndexOf(Platform.pathSeparator) + 1);
+
+void expectSuccess(ProcessResult result) {
+  if (result.exitCode != 0) {
+    String details = """
+EXIT CODE: ${result.exitCode}
+STDOUT: ${result.stdout}
+STDERR: ${result.stderr}
+""";
+    Expect.equals(result.exitCode, 0, details);
```
iarkh: First of all, Expect.equals itself prints expected and actual result (it's result.exitCode in our case), so I believe there is no need to add exit code value to the details variable here.

Secondly, result.exitCode is actually checked two times here, in the line 186 and implicitly in the line 192.

I would suggest just to just call 
    Expect.equals(results.exitCode, 0,
        """STDOUT: ${results.stdout}, STDERR: ${results.stderr}""");

in test.lib.dart and don't add additional expectSuccess function in file_utils.dart.

PR #207, opened 2018-12-17
Wrong File.openRead() test rewritten

sgrekhov:
Test was rewrittent to test that returned stream can be listened to

### DIFF
--- a/LibTest/io/File/openRead_A02_t01.dart
+++ b/LibTest/io/File/openRead_A02_t01.dart
@@ -29,12 +29,16 @@ main() async {
 
 _main(Directory sandbox) async {
   File file = getTempFileSync(parent: sandbox);
-  file.writeAsBytesSync([1, 2, 3]);
+  List<int> content = [3, 1, 4, 1, 5, 9, 2, 6];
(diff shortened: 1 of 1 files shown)


LibTest/io/File/openRead_A02_t01.dart
```diff
@@ -30,9 +31,13 @@ _main(Directory sandbox) async {
   File file = getTempFileSync(parent: sandbox);
   file.writeAsBytesSync([1, 2, 3]);
   asyncStart();
-  await file.openRead().listen((data) {
-    Expect.throws(() {file.deleteSync();});
-  }).onDone(() {
+  final c = new Completer();
+  StreamSubscription<List<int>> s = null;
+  s = await file.openRead().listen((data) async {
+    Expect.listEquals([1, 2, 3], data);
```
mkustermann: The callback given to `listen((data) {... })` can be called multiple times, each with a subset of the bytes. There is no  guarantee that the very first time we get all 3 bytes, consider changing this to expect that `data` is a prefix of `[1, 2, 3]`

PR #243, opened 2019-02-12
Fix wait for window.onLoad event

whesse:
This test was timing out because window.onLoad was not triggering after a listener was added.

Also, addEventListener is preferred to onLoad = , because it does not overwrite a previous handler.


whesse:
Other web tests that are timing out might have this problem as well.

sgrekhov:
We'll change the other tests that use window.onLoad. Thank you

ngolovleva:
The last commit replace usage of window.onLoad.listen with addOnLoadListener with parameter from window.onLoad.listen. As concern asyncStart() it was in the listener, and now it is replaced from listener before this listener.

PR #353, opened 2019-03-29
Update scoping_A02_t01.dart

whesse:
Add a check that the separate scope variables are live and can be modified.

### DIFF
--- a/LanguageFeatures/Control-flow-collections/scoping_A01_t01.dart
+++ b/LanguageFeatures/Control-flow-collections/scoping_A01_t01.dart
@@ -36,4 +36,11 @@ main() {
       for (var j = i; j < 4; j++) i + j,
   ];
   Expect.listEquals([2, 3, 4, 4, 5], list3);
+  
+  var x = "a";
+  List<String> list4 = [
+    for (var x in ["b", "c"]) x,
+    x
+  ];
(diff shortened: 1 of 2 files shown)


whesse:
This is more what I had in mind.  Just a check that the variables in closures are also settable.
Also noticed that the scope check could also check that a scope ends by stopping the shadowing of an external scope.  Maybe there should be a multitest that checks for a compile-time error when it isn't in scope.

sgrekhov:
A very good change, thank you!

LanguageFeatures/Control-flow-collections/scoping_A01_t01.dart
```diff
@@ -36,4 +36,11 @@ main() {
       for (var j = i; j < 4; j++) i + j,
   ];
   Expect.listEquals([2, 3, 4, 4, 5], list3);
+  
+  var x = "a";
+  List<String> list4 = [
+    for (var x in ["b", "c"]) x,
+    x
+  ];
+  Expect.listEquals(["b", "c", "a"]);
```
whesse: Should be 
Expect.listEquals(["b", "c", "a"], list4);

Please fix at your earliest convinience, maybe added to another PR. Thanks.

PR #1074, opened 2021-04-12
Fixes #1073. Wait for isolate start before pausing it

### LINKED ISSUE #1073: Incorrect LibTest/isolate/Isolate/pause_A01_t01.dart
The `LibTest/isolate/Isolate/pause_A01_t01.dart` test look like this
```dart
entryPoint(SendPort sendPort){
  Random random = new Random();
  int s = 0;
  while (true){
    s = -s + random.nextInt(100);
    sendPort.send(s);
  }
}

test() async {
  ReceivePort receivePort = new ReceivePort();
  ReceivePort onExit = new ReceivePort();
  Isolate isolate = await Isolate.spawn(
      entryPoint,
      receivePort.sendPort, // message
      onExit:onExit.sendPort,
      errorsAreFatal:true
  );
  isolate.pause();
  // check that messages are received from paused isolate
  int count = 0;
  await for (var _ in receivePort){
    if (count++==1000000){
      break;
    }
  }
```

This test spawns and pauses an isolate and then expects to have already received messages.

This flakily times out, because the isolate might be paused even before the `entrypoint` is invoked, therefore not allowing it to send any messages, therefor making the main isolate wait forever to receive messages.

sgrekhov:


LibTest/isolate/Isolate/pause_A01_t01.dart
```diff
@@ -22,34 +22,37 @@ import "dart:math";
 import "../../../Utils/expect.dart";
 
 // indefinitely running isolate
-entryPoint(SendPort sendPort){
+entryPoint(List<SendPort> sendPorts) {
+  sendPorts[0].send(42);
```
mkustermann, line 26: After this send, the other side can theoretically receive the `42` message and pause this isolate **immediately**. 

Depending on the timing, the pause might happen before we even enter the loop below. The main isolate then would wait indefinitely to receive messages on `sendPorts[1]`.

The only effect of pausing an isolate is that it temporarily stops executing. Once resumed it will continue where it left off. So the only way to really measure if it was paused is to detect if there was a long pause (or use VM's `vm-service` protocol to ask for the isolate's state, but that's probably not a good idea for co19 tests).

The isolate itself can detect that with reasonably high probability, e.g. via something like this
```dart

const Duration pauseTime = const Duration(seconds: 10);

entryPoint(SendPort port) {
  final sw = Stopwatch()..start();
  bool notified = false;
  while (true) {
    if (!notified) {
      port.send('in-loop');
      notified = true;
    }
    if (sw.elapsed.inSeconds >= pauseTime.inSeconds) {
      // We were probably paused, otherwise this loop would not stop executing
      // for 10 seconds.
      port.send('detected-pause');
      break;
    } else {
      sw..reset()..start();
    }
  }
}

test() async {
  final onExit = ReceivePort();

  final rp = ReceivePort();
  final si = StreamIterator(rp);

  final isolate = await Isolate.spawn(entryPoint, rp.sendPort, onExit: onExit.sendPort);
  
  Expect.equals(true, await si.moveNext());
  Expect.equals('in-loop', si.current);

  final cap = isolate.pause();
  await Future.delayed(pauseTime);
  isolate.resume(cap);

  Expect.equals(true, await si.moveNext());
  Expect.equals('detected-pause', si.current);

  si.cancel();
  rp.close();

  await onExit.first;
}
```

sgrekhov:
@mkustermann thank you! I rewrote the tests in your way. Please review.
Note that I had to add a small dalay after `sw..reset()..start();`. It didn't work without it. Can it be an issue?

LibTest/isolate/Isolate/pause_A01_t01.dart
```diff
@@ -13,43 +13,56 @@
  * response to, e.g., timers or receive-port messages. When the isolate is
  * resumed, it handles the already enqueued events.
  *
- * @description Check that pause() does not stop the isolate to execute code.
```
mkustermann, line 16: Actually I think I misunderstood this test and what `Isolate.pause()` was supposed to do: Pausing should take only effect when coming back to the event loop. So this test was supposed to test that pausing will actually not pause the `entryPoint` function (which is synchronous and therefore doesn't go back to event loop)

So I must apologize and the last change you made, was correct. Maybe you could revert to that revision?

sgrekhov:
No problem. Reverted

mkustermann (approved):

sgrekhov:
@mkustermann thank you. What about this note?

> After this send, the other side can theoretically receive the 42 message and pause this isolate immediately.
> 
> Depending on the timing, the pause might happen before we even enter the loop below. The main isolate then would wait indefinitely to receive messages on sendPorts[1]

Is the probability of such scenario reasonably low? May be it makes sense additionally add some delay after getting the first event? Say
```dart
  await receivePort1.first;
  await Future.delayed(Duration(millisecinds: 100));
  isolate.pause();
```

mkustermann:
> > After this send, the other side can theoretically receive the 42 message and pause this isolate immediately.
> > Depending on the timing, the pause might happen before we even enter the loop below. The main isolate then would wait 
> > indefinitely to receive messages on sendPorts[1]

> Is the probability of such scenario reasonably low? May be it makes sense additionally add some delay after getting the first 
> event? Say

My comment was not correct, because `isolate.pause()` only effects the event loop processing not synchronous code. So just disregard my initial comment.

PR #1079, opened 2021-04-15
Fixes #1004. Add tests that confirms that tight-looped isolate is not killed with Isolate.beforeNextEvent priority

### LINKED ISSUE #1004: Have a test that would confirm tight-looped isolate won't get killed with Isolate.beforeNextEvent priority
https://github.com/dart-lang/co19/blob/master/LibTest/isolate/Isolate/kill_A01_t01.dart confirms that isolate will get killed even if it runs tight loop.
Complementary test that would confirm that tight-looped isolate won't get killed if killed with beforeNextEvent priority would be useful as there seems to be (an erroneous) possibility for such an isolate to still get killed.

sgrekhov:
@aam please review. Does this test do what you want?


aam (approved):
Thank you!

PR #1233, opened 2021-11-02
Fix `stat` conformance tests to resolve links

brianquinlan:
Note that I replaced `Link/stat*_A01_t01.dart` with `Link/stat*_A01_t04.dart` (since they test the same scenario) and moved `Link/stat*_A01_t05.dart` to `Link/stat*_A01_t04.dart` to fill the gap.


sgrekhov:
Thank you!

PR #1242, opened 2021-12-02
Avoid sender making more progress than the receiver in Isolate/pause_…

rmacnak-google:
…A01_t01/2.

In these tests, one isolate continuously sends messages to another without back pressure. If the system is loaded, the sender may send messages at a faster rate then the receiver can process them and the test can time out before receiver processes enough messages to end the test. Instead of processing a large number of messages to ensure some of them would happen after the pause request, use synchronous sleeps.


sgrekhov:
Thanx!

PR #1312, opened 2022-03-09
Correct an enhanced-enum test to allow setters named values

chloestefantsova:
We had a discussion with Erik about the static `values` setter in enums, and the conclusion is that they should be allowed. This PR adjusts a test accordingly.

### DIFF
--- a/LanguageFeatures/Enhanced-Enum/grammar_A10_t04.dart
+++ b/LanguageFeatures/Enhanced-Enum/grammar_A10_t04.dart
@@ -10,7 +10,8 @@
 /// declaration.
 ///
 /// @description Check that it's a compile-time error if the enum declaration
-/// contains a static member declaration with the name `values`
+/// contains a static setter with the name `values`, and its argument is not a
+/// supertype of the generated member `values`.
 /// @author sgrekhov@unipro.ru
 /// @issue 48387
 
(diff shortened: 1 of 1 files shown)


LanguageFeatures/Enhanced-Enum/grammar_A10_t04.dart
```diff
@@ -25,9 +26,6 @@ enum E1 {
   e3;
 
   static void set values(int value) {}
```
eernstg, line 28: I'd think this is still an error, because of the mismatch between the parameter type of this setter and the return type of the corresponding (implicitly induced) getter.

eernstg (approved):
LGTM, except that I think we should keep one of those errors.

LanguageFeatures/Enhanced-Enum/grammar_A10_t04.dart
chloestefantsova: I agree. I accounted for that error in the header to the enum itself, a few lines above. The getter `values` is generated, so its location is the location of the overall enum. Do you think changing the location to the type of the argument of the setter would be better?

eernstg: Oh, I'm sorry, I did not notice that this was handled by text that hasn't changed (so it wasn't shown in the github diff view until I pressed a button).

So, I'll close this thread.

eernstg (commented):
Thread resolved.

PR #1329, opened 2022-03-29
Remove unused and uninitialized non-nullable field from a test

chloestefantsova:
The code of the test contains a field that isn't required for the purpose of the test, and yet it causes an additional compile-time error because it is uninitialized and non-nullable.


johnniwinther (approved):

eernstg (approved):
LGTM!

PR #1330, opened 2022-03-29
Add expected CFE errors on concrete 'index' members

chloestefantsova:
This PR adds expected errors that the CFE reports on concrete members named `index` in implementers of the `Enum` interface. The discrepancy between the CFE and the Analyzer in this case is that the Analyzer doesn't report the error on `index` if the implementer is concrete and, therefore, already erroneous.


johnniwinther (approved):

eernstg (approved):
LGTM!

I'd expect the analyzer to report the errors about the `index` declarations as well, but I guess it's up to the analyzer team to make similar adjustments for the analyzer.

PR #1331, opened 2022-03-29
Remove extra expected CFE errors that are no longer reported

chloestefantsova:
This PR removes some of the extra expected CFE errors that are no longer reported by the CFE. The removal of the errors brings the behavior of the CFE in agreement to that of the Analyzer.


eernstg (approved):
LGTM

PR #1336, opened 2022-04-06
Update expected errors after recent CFE udpates in Enhanced Enums

chloestefantsova:
This PR updates the expectations of the errors in some of the Enhanced Enums tests to match those emitted by the CFE after some recent updates in it.


eernstg (approved):
LGTM, thanks!

PR #1346, opened 2022-07-01
Fixes #1344. Wait for entryPoint is invoked before doing the test

### LINKED ISSUE #1344: `co19/LibTest/isolate/Isolate/kill_A01_t04` is racy
Fundamentally the same issue as described in https://github.com/dart-lang/co19/issues/1095.

After `await Isolate.spawn()` has returned the isolate was created but the entrypoint hasn't necessarily run yet, so an `isolate.kill()` may kill it before the entrypoint is run.

I'd suggest making the isolate communicate back to the main isolate that it started running the entrypoint (something similar to https://github.com/dart-lang/co19/commit/eccdd5c326f20544b941bc88b09b3f9a7152093d)

sgrekhov:


mkustermann (approved):

eernstg (approved):
LGTM

PR #1357, opened 2022-07-07
Fixes #1355. Expect that `Process.killPid` returns `false` only if process is already dead

### LINKED ISSUE #1355: killPid_A01_t01,t02 expect false from killPid on Windows
https://github.com/dart-lang/co19/blob/master/LibTest/io/Process/killPid_A01_t02.dart#L49, https://github.com/dart-lang/co19/blob/master/LibTest/io/Process/killPid_A01_t01.dart#L46 both expect `Process.killPid` to always return `false`. Why?

https://github.com/dart-lang/sdk/blob/main/runtime/bin/process_win.cc#L934 might return false if the process is dead already, but I don't think this was the intention of the test.

sgrekhov:


eernstg:
@aam, I guess you would be the best person to review this PR?

aam (approved):
thanks for the fix!

eernstg:
Thanks! I'll then land the PR.

PR #1358, opened 2022-07-07
Fixes #1354. Horizontal inference tests added

### LINKED ISSUE #1354: Add tests for Horizontal inference feature
Add tests for [Horizontal inference](https://github.com/dart-lang/language/blob/master/working/0731%20-%20horizontal%20inference/feature-specification.md) language feature

sgrekhov:


LanguageFeatures/Horizontal-inference/bounds_t01.dart
```diff
@@ -0,0 +1,29 @@
...
+  Expect.equals(T, C3); // Without horizontal inference it'd be C1
+}
+
+void f2<T extends C2>(T Function() a, void Function(T) b) {
+  Expect.equals(T, C3); // Without horizontal inference it'd be C2
+}
+
+main() {
+  f1((v){}, C3());
+  f2(() => C3(), (x) {});
```
eernstg, line 28: I actually can't make it anything other than C3 in any version down to 2.9.0-21.0.dev (tried about 5 versions).

Looking a bit more at Paul's example:

```dart
void main() {
  Iterable<int> values = [1, 2];
  var largestValue = values.fold(0, (a, b) => a < b ? b : a); // Compile-time error.
  // ...
}
```

It is crucial that the invocation of `fold` does not have a context type. If we declare `int largestValue` then the context type of `int` will guide the inference to choose `int` as the actual type argument to the invocation of `fold`, and then everything just works.

So one thing to be aware of in these tests is that they should not provide any context types when we expect to see an improvement, or at least: We must be very careful if we provide a context type, because it might very well eliminate the need for the horizontal inference altogether.

I can see that the occurrence of `C3()` causes the inference to infer the actual type argument `C3` both with and without the `inference-update-1` feature, so we do have a situation here where `inference-update-1` is not being tested.

The effect of `inference-update-1` is actually not to change the inferred type argument of `f1` and `f2`, at least not in an example like this one. The change amounts to using the inferred type argument `C3` in the context type provided for the function literal that gets to be inferred later (the one in stage 1, not 0). This means that `inference-update-1` makes a difference if we're using the type of the parameter of the function literal.

So let's change `C3` such that we can see whether a given object is a `C3` or something less than that:

`class C3 extends C2 { void m() {}}`

With that, we can use `f1((v) => v.m(), C3())` and `f2(() => C3(), (x) => x.m())`.

I believe that would change the test such that it does test `inference-update-1`, rather than just testing a behavior which is already present in older versions of the language.

eernstg (changes requested):
I've drilled down into the first test, and I noticed that it doesn't actually test `inference-update-1` as written.

It was somewhat complex to tease out why it is so, and I made the choice to just stop there because that is an issue which needs to be addressed in any case.

So please perform some checks on all these tests, such that we can see that they succeed with the feature and fail without the feature:

```
> dart --enable-experiment=inference-update-1 sometest.dart
> dart --enable-experiment=no-inference-update-1 sometest.dart
```

When that's working then let's take another round in the review.

LanguageFeatures/Horizontal-inference/bounds_t01.dart
sgrekhov: Erik thank you. I fixed the tests. Please re-review. Note that a couple of the tests produce the same result with and without horizontal inference (with `no-inference-update-1` and without it). These are "negative" tests, that checks cases when there are no inference. For example `explicit_type_arguments_t01.dart` checks that if type arguments specified then no horizontal inference happened.

eernstg (changes requested):
I think this one needs some adjustments. ;-)

LanguageFeatures/Horizontal-inference/async_t03.dart
```diff
@@ -0,0 +1,37 @@
...
+/// types of the functions to function parameters types. Test async* functions
+/// @author sgrekhov22@gmail.com
+
+import "../../Utils/expect.dart";
+
+typedef expectedC = Future<String> Function(Object? o);
+typedef expectedD = Future<int> Function(Object? o);
+
+Stream<int> f<T, U, V>(void Function(T, U) a, T Function(V) b,
+    U Function(V) c) async* {
```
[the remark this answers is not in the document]
eernstg: (That's weird, the PR looked like it only contained changes to a single file, that is, '[LanguageFeatures/Horizontal-inference/async_t03.dart](https://github.com/dart-lang/co19/pull/1358/files/5e9d2fe6e603595c8f5d47da48f58890a7f25502..74200e44b739de8a4b4cd747906bb667d00043cd#diff-754868d2c4951cd74e45d0d2a1b938fa4060b8d70c5f9600be734f7a473ff652)').

I'll take another look tomorrow.

sgrekhov:
@eernstg 

> I think this one needs some adjustments. ;-)

Why? I don't see any problems with it. I can compile the test without any changes. Just retested it on the edge version of the SDK. Please check that you are trying to run the test on the those version of the SDK which does have `inference-update-1` enabled. 

> It creates some issues that the return type of a is void. Is that intended? Looks like Stream<T> for some T would be expected.

Indeed, it was not intended that function `a` has a return type `void`, but if it is an error? It's possible to change it to `T Function(T, U) a` it will work as well.

> How did you conclude that the type arguments should be inferred as Stream<String>, Stream<int>, Object??

From arguments `b` and `c` which return `T` and `U` accordingly. For example `b` is
```dart
(v) async* {
      yield "Lily was here";
}
```
This function literal is marked as `async*` and this mean that its return type is `Stream<X>` and `yield "Lily was here";` tell us that `X` is `String`. Likewise `c` inferred `Stream<int>`

> (That's weird, the PR looked like it only contained changes to a single file, that is, '[LanguageFeatures/Horizontal-inference/async_t03.dart](https://github.com/dart-lang/co19/pull/1358/files/5e9d2fe6e603595c8f5d47da48f58890a7f25502..74200e44b739de8a4b4cd747906bb667d00043cd#diff-754868d2c4951cd74e45d0d2a1b938fa4060b8d70c5f9600be734f7a473ff652)'). But I can see lots of files now.

Probably you are looking at a particular commit, not on all commits in the PR

eernstg:
I didn't get very far at all, then there was an interview and some meetings. Tomorrow! ;-)

LanguageFeatures/Horizontal-inference/optional_t01.dart
```diff
@@ -0,0 +1,29 @@
...
+/// positioned parameters
+/// @author sgrekhov22@gmail.com
+
+import "../../Utils/expect.dart";
+
+typedef expectedB = String Function();
+
+void f<T, U>(void Function(T, U) a, [T Function()? b, U? c]) {
+  Expect.equals(String, T);
+  Expect.equals(int, U);
```
eernstg, line 20: Should check the run-time type of `a`.

LanguageFeatures/Horizontal-inference/optional_t01.dart
```diff
@@ -0,0 +1,29 @@
...
+/// from the type of other arguments.
+///
+/// @description Checks that type inferred from non-function types and return
+/// types of the functions to function parameters types. Test optional
+/// positioned parameters
+/// @author sgrekhov22@gmail.com
+
+import "../../Utils/expect.dart";
+
+typedef expectedB = String Function();
```
eernstg, line 16:
```suggestion
typedef ExpectedB = String Function();
```

LanguageFeatures/Horizontal-inference/optional_t02.dart
```diff
@@ -0,0 +1,29 @@
...
+/// from the type of other arguments.
+///
+/// @description Checks that type inferred from non-function types and return
+/// types of the functions to function parameters types. Test optional named
+/// parameters
+/// @author sgrekhov22@gmail.com
+
+import "../../Utils/expect.dart";
+
+typedef expectedB = String Function();
```
eernstg, line 16:
```suggestion
typedef ExpectedB = String Function();
```

LanguageFeatures/Horizontal-inference/optional_t02.dart
```diff
@@ -0,0 +1,29 @@
...
+
+void f<T, U>(void Function(T, U) a, {T Function()? b, U? c}) {
+  Expect.equals(String, T);
+  Expect.equals(int, U);
```
eernstg, line 20: Should test the run-time type of `a`.

LanguageFeatures/Horizontal-inference/optional_t03.dart
```diff
@@ -0,0 +1,29 @@
...
+/// from the type of other arguments.
+///
+/// @description Checks that type inferred from non-function types and return
+/// types of the functions to function parameters types. Test required named
+/// parameters
+/// @author sgrekhov22@gmail.com
+
+import "../../Utils/expect.dart";
+
+typedef expectedB = String Function();
```
eernstg, line 16:
```suggestion
typedef ExpectedB = String Function();
```

eernstg (changes requested):
The most tricky part is that we want to test the inferred type parameter types for a lot of different function literals. In general, the inferred type could be `dynamic`, in which case there would not be any type errors at the location where the function literal is written (`i.isOdd` in the body of the function literal is just fine when `i` has type `dynamic`), and this would also work at run time (passing a function of type `void Function(dynamic)` is just fine when a `void Function(int)` is expected).

So we need to test the values of the actual type arguments, but we _also_ need to test the run-time type of the given function object, to make sure it isn't `dynamic` all over the place.

This is not easy, because we can't assume that the function object has any specific precise type (an implementation is allowed to use special "compiler created" classes as the run-time types of function objects), they just have to be subtypes of the specified function type.

So we could have a function object `(int i) => i` with type `MySpecialInternalType` which is a subtype of `int Function(int)`, and then we can't rely on the run-time type of the function object to be _equal_ to `int Function(int)`.

This means that it is inconvenient to test that the type of a function object is as expected.

In reality, I think it _does_ work to do `Expect.equals(myFunction.runtimeType, typeOf<int Function(int)>())`, but we should at least be aware of the fact that it might not work. Perhaps it stops working in some new version of one of the compilers tomorrow.

So I'm not sure what we can do about this, but I suppose we could just pretend that the `equals` will work, and then we will probably be able to grep all locations where this unsafe trick is used, if it turns out to be a problem.

Do you have a better idea?

Next, I'm still pretty sure the test async_t03.dart has several issues that need to be fixed. The main reason would be that some code in that test was never executed.

So let's take yet another round with this one! ;-)

LanguageFeatures/Horizontal-inference/bounds_t01.dart
```diff
@@ -0,0 +1,31 @@
...
+import "../../Utils/expect.dart";
+
+class C1 {}
+class C2 extends C1 {}
+class C3 extends C2 {
+  void c3() {}
+}
+
+void f1<T extends C1>(void Function(T) a, T b) {
+  Expect.equals(T, C3);
```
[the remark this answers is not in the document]
sgrekhov: @eernstg did you mean 
```dart 
Expect.isTrue(a is void Function(C3));
``` 
here? Because
```dart
void f1<T extends C1>(void Function(T) a, T b) {
  Expect.isTrue(a is void Function(T));
```
will always be true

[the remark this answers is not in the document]
eernstg: That's true, `a is void Function(T)` should be guarantee by soundness already, so we wouldn't expect any type inference bugs to be able to violate that.

What we actually need is `Expect.equals(typeOf<void Function(T)>(), a.runtimeType)`, because that would fit the expected behavior when the call site enables transfer of information from the actual argument for `b` to a function literal which is the actual argument for `a`.

(And I guess we're relying on `.runtimeType == someFunctionType` in so many ways that this extra case makes no difference.)

LanguageFeatures/Horizontal-inference/horizontal_inference_t01.dart
```diff
@@ -0,0 +1,42 @@
...
+typedef expectedD = bool Function(Object? o);
+
+typedef Str = String;
+typedef Int = int;
+typedef Boolean = bool;
+
+void f<T, U, V>(void Function(T, U) a, T b, U Function(V) c, V Function(U) d) {
+  Expect.equals(String, T);
+  Expect.equals(int, U);
+  Expect.equals(expectedC, c.runtimeType);
```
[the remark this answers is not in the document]
sgrekhov: @eernstg For now I'm unable to invent the trick that will allow us to test the actual runtime type and won't fail in future. I think it's the same problem as
```dart
  Expect.equals(List, [].runtimeType); // Ok
  Expect.equals(Map, {}.runtimeType); // Expect.equals(expected: <Map<dynamic, dynamic>>, actual: <_InternalLinkedHashMap<dynamic, dynamic>>) fails.
```
If implementation changes we may have some amount of failing tests that tests instances of `Lists` runtime type. I did a search (searched for `Expect.equals\((.)*.runtimeType, (.)*` and `Expect.equals\((.)*,(.)*.runtimeType`) and didn't find any additional checks for function runtime type. So I suggest to live is as is and if there will be failures, then change the tests and replace `Expect.equals(Expected, f.runtimeType)` by check like `Expect.isTrue(f is  Function(String, int));`

LanguageFeatures/Horizontal-inference/async_t01.dart
```diff
@@ -0,0 +1,34 @@
...
+/// (specifically function literals) to be type inferred in a context derived
+/// from the type of other arguments.
+///
+/// @description Checks that type inferred from non-function types and return
+/// types of the functions to function parameters types. Test async functions
+/// @author sgrekhov22@gmail.com
+
+import "../../Utils/expect.dart";
+
+typedef expectedB = Future<String> Function(Object? o);
```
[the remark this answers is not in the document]
sgrekhov: Changed to `typeOf<...>()` everywhere

LanguageFeatures/Horizontal-inference/horizontal_inference_t04.dart
```diff
@@ -0,0 +1,41 @@
...
+
+/// @assertion Horizontal inference allows certain arguments of an invocation
+/// (specifically function literals) to be type inferred in a context derived
+/// from the type of other arguments.
+///
+/// @description Checks that horizontal inference doesn't work in case of
+/// the circular dependency
+/// @author sgrekhov22@gmail.com
+
+void f<X, Y>(X Function(Y) a, Y Function(X) b) {}
```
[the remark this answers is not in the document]
sgrekhov: Added `circular_dependency_t02.dart`

sgrekhov:
@eernstg thank you for the detailed review! It helped much. I did all of the requested changes. Please rereview

LanguageFeatures/Horizontal-inference/bounds_t01.dart
eernstg: All great, thanks!

sgrekhov:
@eernstg thanx! Please, merge this PR then I'm unable to do it

PR #1364, opened 2022-07-13
Add initial specparser workflow

athomas:


athomas:
Take a look at https://github.com/dart-lang/co19/runs/7323638867?check_suite_focus=true to see how this work (this is the spec parser attempting to parse the GitHub actions yaml file and determining that it's definitely not valid Dart).

@sgrekhov FYI, this should help catch some mistakes using the spec parser tool.

eernstg (approved):
Amazing! LGTM.

The spec_parser will fail to handle certain things because it does not take language versions into account, but I guess we can use status files (or something similar, whatever is appropriate today) in order to handle this.

sgrekhov:
Great! Thank you!

Issue #1363, opened 2022-07-12
Update tests for `yield`/`yield*` in `async*` functions and `await for` loops.


sgrekhov:
Update existing co19 tests according to the [new feature specification](https://github.com/dart-lang/language/blob/master/accepted/future-releases/async-star-behavior/feature-specification.md)

PR #1370, opened 2022-07-14
#1363. Tests for yield updated to the current specification

sgrekhov:


Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A01_t01.dart
```diff
@@ -0,0 +1,63 @@
...
+/// the stream `u` associated with `m` has been paused, then execution of `m` is
+/// suspended until `u` is resumed.
+///
+/// @author a.semenov@unipro.ru
+/// @author sgrekhov22@gmail.com
+
+import 'dart:async';
+import '../../../../Utils/expect.dart';
+
+List<int> readyToSent = [];
```
alexmarkov, line 30: Nit: readyToSend?

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A01_t03.dart
```diff
@@ -0,0 +1,65 @@
...
+      Expect.listEquals([1], sent);
+      Expect.listEquals([1, 2], readyToSent);
+      ss.pause();
+      await Future.delayed(Duration(milliseconds: 100));
+      Expect.listEquals([1], sent);
+      Expect.listEquals([1, 2], readyToSent);
+      await ss.cancel();
+    }
+  });
+  await Future.delayed(Duration(seconds: 1));
```
alexmarkov, line 60: Such timeouts can cause flaky failures on the bots with high loads, as it is not guaranteed that generator advanced and listen callback completed while waiting for 1 second. Consider creating a Completer, completing it in the listen callback and awaiting its future in main to ensure the listen callback is completed. Alternatively you can probably call asyncEnd() earlier and then check expectations.

alexmarkov (commented):
The new tests lgtm (modulo a couple of comments).

However, I think it would be good if @eernstg or @lrhn would review and approve these tests.

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A01_t03.dart
sgrekhov: Added `Completer`. Thanx

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A01_t08.dart
```diff
@@ -0,0 +1,72 @@
...
+  ss = s.listen((int i) async {
+    received.add(i);
+    if (i == 2) {
+      ss.pause();
+      await Future.delayed(Duration(milliseconds: 100));
+      Expect.listEquals([1], sent);
+      Expect.listEquals([1, 2], readyToSend);
+      Expect.isTrue(sc.isPaused);
+      ss.resume();
+      await Future.delayed(Duration(milliseconds: 100));
```
eernstg, line 56: Same issue as before: In principle, we can't know that other microtasks have actually been scheduled just because there was a delay.

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A03_t01.dart
```diff
@@ -0,0 +1,60 @@
...
+  Completer c = Completer();
+  List log = [];
+  Stream<int> s = generator();
+  late StreamSubscription<int> ss;
+  ss = s.listen((int i) {
+    log.add(i);
+    ss.cancel();
+    if (!c.isCompleted) {
+      c.complete();
+    }
```
eernstg, line 51: Should it not be safe to replace line 49,50,51 by `c.complete()`? How would it fail?

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A03_t02.dart
```diff
@@ -0,0 +1,60 @@
...
+  Completer c = Completer();
+  List log = [];
+  Stream<int> s = generator();
+  late StreamSubscription<int> ss;
+  ss = s.listen((int i) async {
+    log.add(i);
+    await ss.cancel();
+    if (!c.isCompleted) {
+      c.complete();
+    }
```
eernstg, line 51: Same as previous test: Why not have only `c.complete()`?

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A03_t03.dart
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+test() async {
+  generatorAlive = false;
+  Stream<int> s = generator();
+  late StreamSubscription<int> ss;
+  ss = s.listen((int x) async {
+    if (!generatorAlive) {
+      Expect.fail("Generator function must be terminated");
```
eernstg, line 46: Probably `Expect.fail("The generator function is unexpectedly not alive")` would be easier to understand.

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A03_t03.dart
```diff
@@ -0,0 +1,66 @@
...
+    }
+    if (x == 5) {
+      // let generator to work some time
```
eernstg, line 49: Perhaps `// Allow the generator to work for some time.`

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A03_t03.dart
```diff
@@ -0,0 +1,66 @@
...
+      Expect.equals(true, generatorAlive);
```
eernstg, line 50: Why not `Expect.isTrue(generatorAlive)`?

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A04_t01.dart
```diff
@@ -0,0 +1,51 @@
...
+  while (!finish) {
+    yield i++;
+  }
+}
+
+test() async {
+  finish = false;
+  Stream<int> s = generator();
+  s.listen((int x) async {
+    finish = (x > 10); // let generator to work some time
```
eernstg, line 42: `// Allow the generator...`.

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A04_t02.dart
```diff
@@ -0,0 +1,51 @@
...
+  while (!finish) {
+    yield i++;
+  }
+}
+
+test() async {
+  finish = false;
+  Stream<int> s = generator();
+  s.listen((int x) {
+    finish = (x > 10); // let generator to work some time
```
eernstg, line 42: `// ...`

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A05_t01.dart
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+
+main() async {
+  List log = [];
+  Stream<int> s = generator();
+  late StreamSubscription<int> ss;
+  ss = s.listen((int i) async {
+    // This 'await' returns control back to event loop and generator goes on
+    await Future.delayed(Duration(milliseconds: 100));
```
eernstg, line 45: I tried to replace this by `await null; await null; await null;`, and it works fine. ;-)

Replacing it by `await null; await null;` does not work, and this illustrates that it is directly connected to the number of other tasks we allow the scheduler to handle (except that we're not supposed to assume that we can control the scheduler in any precise manner by doing these things).

eernstg (approved):
LGTM, but please consider the following as well:

A couple of things are worth considering: In several situations there is an `ss.resume()` followed by code which is intended to check that certain lists contain the expected elements. However, it looks like there should be a suspension point before those checks, because otherwise the generator doesn't get a chance to do anything wrong. So maybe `await null;` should be added?

Another thing to consider is that delays will make the trybots/buildbots run more slowly. I suspect that it will work just as well to have a simple `await null;` in several situations than it does to have a delay of 100 milliseconds. Waiting for an entire second is of course even more disruptive to the bot execution. We could have several `await null;`, too.

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A01_t02.dart
```diff
@@ -0,0 +1,63 @@
...
+  ss = s.listen((int i) async {
+    received.add(i);
+    if (i == 1) {
+      ss.pause();
+      await Future.delayed(Duration(milliseconds: 100));
+      Expect.listEquals([], sent);
+      Expect.listEquals([1], readyToSend);
+      ss.resume();
+      await Future.delayed(Duration(milliseconds: 100));
+      Expect.listEquals([1, 2, 3], sent);
```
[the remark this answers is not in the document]
sgrekhov: Changed to 
```dart
      // `sent` may contain not all elements of the stream
      sent.forEach((element) {
        Expect.isTrue([1, 2, 3].contains(element));
      });
```

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A03_t01.dart
sgrekhov: Indeed. Fixed. Thank you

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A03_t03.dart
sgrekhov: Changed

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A03_t03.dart
sgrekhov: Changed everywhere

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A03_t03.dart
sgrekhov: Fixed

eernstg (approved):
LGTM!

I didn't understand that the checks after `resume` were intended to check the situation without any asynchronous scheduling events. When that's the purpose it is fine.

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A01_t08.dart
lrhn: The delay is a timer-delay, so we know that all microtask-based events have been processed.
I'm not sure we *promise* that buffered non-sync-controller events are delivered using microtasks, but if they're not, it'd probably be zero-timer events.
So, this *likely* works.

lrhn (commented):
drive-by comments.

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A01_t03.dart
```diff
@@ -0,0 +1,65 @@
...
+      Expect.listEquals([1], sent);
+      Expect.listEquals([1, 2], readyToSend);
+      ss.pause();
+      await null;
+      Expect.listEquals([1], sent);
+      Expect.listEquals([1, 2], readyToSend);
+      await ss.cancel();
+    }
+  });
+  await Future.delayed(Duration(seconds: 1));
```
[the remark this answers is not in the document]
sgrekhov: The latest version of this test doen't contain this long delay and is based on a `Completer`. Please see https://github.com/dart-lang/co19/blob/master/Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A01_t03.dart

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A01_t02.dart
```diff
@@ -0,0 +1,65 @@
...
+    if (i == 1) {
+      ss.pause();
+      await null;
+      Expect.listEquals([], sent);
+      Expect.listEquals([1], readyToSend);
+      ss.resume();
+      await Future.delayed(Duration(milliseconds: 100));
+      // `sent` may contain not all elements of the stream
+      sent.forEach((element) {
+        Expect.isTrue([1, 2, 3].contains(element));
```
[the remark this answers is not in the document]
sgrekhov: Indeed. Thanks! Please see https://github.com/dart-lang/co19/pull/1390

Language/Statements/Yield_and_Yield_Each/Yield/execution_async_A05_t03.dart
```diff
@@ -0,0 +1,59 @@
...
+  asyncStart();
+  List received = [];
+  Stream<int> s = generator();
+  late StreamSubscription<int> ss;
+  ss = s.listen((int i) async {
+    received.add(i);
+    if (i == 2) {
+      Expect.listEquals([1], sent);
+      Expect.listEquals([1, 2], readyToSend);
+      ss.resume();
```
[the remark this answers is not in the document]
sgrekhov: Description of this test 

> Check that resuming of the [Stream] which was not paused doesn't affect anything

So yes, it's testing that it's no-op

Issue #673, opened 2020-05-20
co19_2 html tests should not use document.registerElement


srujzs:
`document.registerElement` (or `document.register` in the `dart:html` library) has been deprecated and removed in Chrome 80. In order to replicate the same behavior, `window.customElements.define` is preferred instead. See https://developer.mozilla.org/en-US/docs/Web/API/Document/registerElement for more details.

This currently affects the following tests:

co19_2/LibTest/html/Element/attributeChanged_A01_t01
co19_2/LibTest/html/Element/enteredView_A01_t01
co19_2/LibTest/html/Element/leftView_A01_t01
co19_2/LibTest/html/Element/tagName_A01_t03
co19_2/LibTest/html/IFrameElement/IFrameElement.created_A01_t01
co19_2/LibTest/html/IFrameElement/tagName_A01_t03

The error looks like:
```
TypeError: receiver.registerElement is not a function
TypeError: receiver.registerElement is not a function
    at HtmlDocument.registerElement2  org-dartlang-sdk:///sdk_nnbd/lib/html/dart2js/html_dart2js.dart 17723:5
    at registerElement(inlined)       org-dartlang-sdk:///sdk_nnbd/lib/html/dart2js/html_dart2js.dart 10392:5
    at register(inlined)              org-dartlang-sdk:///sdk_nnbd/lib/html/dart2js/html_dart2js.dart 17729:12
```

sigmundch:
We may want to remove some of these tests or switch to run the tests with the web_components polyfill that was used for browsers that didn't implement the API before it was spec-ed.

For context, CustomElements were implemented for Dart when the spec was at version 0.5. It was later noticed that the newer APIs required defining ES6 classes in a way that was not easily compatible with Dart, so we are not fully supporting the new API in Dart at this time.

I am not sure, but I don't believe `window.customElements.define` will work from Dart at this time.

sigmundch:
Here is the issue describing our decision on not adding support for custom-elements v1 (https://github.com/dart-lang/sdk/issues/27445)

sgrekhov:
I'm trying to check if `window.customElements.define` will work or not, but unable to call it in a Dart code. This function signature looks like
```dart
void define (
String name,
Object constructor,
[Map options]
)
```
The second argument is a constructor. But you cannot tear-off a constructor in Dart. Closure also doesn't work

```dart
class Foo extends HtmlElement {
  Foo.created() : super.created();
...
  }
}

main() {
[...]

PR #1381, opened 2022-07-29
#673. IFrame tests that don't actually test IFrame removed

sgrekhov:
When investigatint #673 found that these two test have nothing to do with `IFrame` and can be safely removed. Seems that these tests are result of copy-paste from `Element`


eernstg:
@zanderso, I do not know the detailed workings of 'dart:http', so it would be great if you can take a look at this PR.

sgrekhov:
Ups. I committed findProxy tests in a wrong branch. I'll fix it.

eernstg:
Aha, thanks!

@zanderso, this means that there's no need to ask for your input on this PR after all!

sgrekhov:
@eernstg fixed. Sorry for mistake

eernstg (approved):
LGTM!

PR #1382, opened 2022-07-29
#673. IFrame tests that don't actually test IFrame removed

sgrekhov:
When investigatint #673 found that these two test have nothing to do with `IFrame` and can be safely removed. Seems that these tests are result of copy-paste from `Element`


eernstg (approved):
LGTM.

eernstg:
The spec parser was invoked with some arguments that it does not expect. This might be because it expects at least one argument, but in this case there were no changed files, only deleted ones, and the empty list of arguments is considered as a request for help. I'll think about it.

PR #1383, opened 2022-07-29
Fixes #1129. Don't expect any proxy routine in case of DIRECT connection

### LINKED ISSUE #1129: findProxy tests timeout
The tests
co19/LibTest/io/HttpClient/findProxy_A01_t01
co19/LibTest/io/HttpClient/findProxy_A01_t02
co19/LibTest/io/HttpClient/findProxy_A02_t01
co19/LibTest/io/HttpClient/findProxy_A02_t02 

timeouts on dartk and dartkp

sgrekhov:


eernstg (commented):
This makes sense to me.

However, @zanderso, it would be great if you could take a look, too!

zanderso:
I'm not the right reviewer. Maybe @brianquinlan or @aam ?

LibTest/io/HttpClient/findProxy_A01_t01.dart
```diff
@@ -51,12 +52,17 @@ test() async {
           .set(HttpHeaders.proxyAuthenticateHeader, 'Basic, realm=realm');
       request.response.statusCode = HttpStatus.proxyAuthenticationRequired;
       request.response.close();
+      // Shutdown server. We are not expecting any additional requests in case
+      // of DIRECT connection
+      Future.delayed(Duration(milliseconds: 100)).then((_) {
```
aam, line 57: Embedding assumptions about timings seems to be unfortunate - they can lead to flaky results when run on systems under heavy load for example.
Is there a way to change the test to get rid of `Future.delayed`?
What if client sends two requests, can server then assert that by the time second request arrives findProxy/authenticaeProxy should definitely be called/not called?

eernstg:
@zanderso, thanks for the redirection; @aam, thanks for the input!

LibTest/io/HttpClient/findProxy_A01_t01.dart
sgrekhov: @aam thank you. Tests rewritten to not to use hardcoded delays. Please rerewiew

LibTest/io/HttpClient/findProxy_A01_t01.dart
```diff
@@ -67,18 +64,21 @@ test() async {
 
   client.authenticateProxy =
       (String host, int port, String scheme, String? realm) {
-        authenticateProxyCalled = true;
-        Completer<bool> completer = new Completer<bool>();
-        completer.complete(true);
-        return completer.future;
+    authenticateProxyCalled = true;
+    Completer<bool> completer = new Completer<bool>();
```
aam, line 68: can this be simplified to `return Future<bool>.value(true);` instead?
Or maybe you can just do `Expect.fail("authenticaterProxy should not be called");`

aam (approved):

LibTest/io/HttpClient/findProxy_A01_t01.dart
sgrekhov: @aam thank you. This makes sense. Code simplified

eernstg (approved):
LGTM.

PR #1385, opened 2022-08-03
#1129. In case of DIRECT connection don't expect any proxy routine called

sgrekhov:
This is a co19_2 version of https://github.com/dart-lang/co19/pull/1383


eernstg (approved):
LGTM

Issue #993, opened 2021-02-03
Create tests for FFI library


sgrekhov:
https://api.dart.dev/dev/2.12.0-279.0.dev/dart-ffi/dart-ffi-library.html

PR #1386, opened 2022-08-03
#993. [FFI] Missed tests added for some new types

sgrekhov:


LibTest/ffi/UintPtr/UintPtr_A01_t01.dart
```diff
@@ -0,0 +1,44 @@
...
+import "package:ffi/ffi.dart";
+import "../../../Utils/expect.dart";
+
+void main() {
+  Pointer<UintPtr> p1 = calloc<UintPtr>();
+  try {
+    p1.value = 42;
+    Expect.equals(42, p1.value);
+    p1.value = -42;
+    Expect.equals(-42, p1.value);
```
sgrekhov, line 24: @dcharkes is it Ok, that `UintPtr` returns a negative value?

LibTest/ffi/Size/Size_A02_t01.dart
```diff
@@ -0,0 +1,21 @@
...
+/// @description Checks that this type represents C size_t type and is big
+/// enough to contain the size of the biggest object the host system can handle
+/// @author sgrekhov22@gmail.com
+
+import "dart:ffi";
+import "package:ffi/ffi.dart";
+import "../../../Utils/expect.dart";
+
+void main() {
+  Expect.equals(sizeOf<IntPtr>(), sizeOf<Size>());
```
sgrekhov, line 20: @dcharkes Is it correct to assume the above?

LibTest/ffi/WChar/WChar_A02_t01.dart
```diff
@@ -0,0 +1,31 @@
...
+
+import "dart:ffi";
+import "package:ffi/ffi.dart";
+import "../../../Utils/expect.dart";
+
+void main() {
+  Pointer<WChar> p1 = calloc<WChar>(2);
+  try {
+    Expect.equals(2, p1.elementAt(1).address - p1.address);
+    Expect.equals(2, sizeOf<WChar>());
```
sgrekhov, line 27: @dcharkes documentation says nothing about the `WChar` size. Is this test correct?

LibTest/ffi/UnsignedLongLong/UnsignedLongLong_A01_t01.dart
```diff
@@ -0,0 +1,43 @@
...
+import "package:ffi/ffi.dart";
+import "../../../Utils/expect.dart";
+
+void main() {
+  Pointer<UnsignedLongLong> p1 = calloc<UnsignedLongLong>();
+  try {
+    p1.value = 42;
+    Expect.equals(42, p1.value);
+    p1.value = -42;
+    Expect.equals(-42, p1.value);
```
sgrekhov, line 27: I believe that it is expected that `UnsignedLongLong` value is negative (because we have no Unsigned 64-bit types in Dart to accomodate `unsigned long long` value). @dcharkes please confirm

sgrekhov (commented):
@dcharkes I have some questions, please see below

LibTest/ffi/UintPtr/UintPtr_A01_t01.dart
dcharkes: I think this is because the underlying representation on x64 is Uint64, in which the upper half of numbers is exposed as dart negative integers.

However, on Uint32 this will not happen, so this test will fail on 32 bit platforms.

So, we should rewrite the test to either accept both outcomes (0x100000000-42 on 32 platforms) or remove the test because we have this behavior covered on the Uint64 and Uint32 tests already.

LibTest/ffi/Size/Size_A02_t01.dart
dcharkes: On all platforms that we currently support yes.

LibTest/ffi/WChar/WChar_A02_t01.dart
dcharkes: No, 2 bytes on Windows, 4 bytes on the rest.

(So I assume you're on Windows.)

LibTest/ffi/UnsignedLongLong/UnsignedLongLong_A01_t01.dart
dcharkes: correct

LibTest/ffi/WChar/WChar_A02_t01.dart
sgrekhov: Yes, I'm. I think, then, it mekes more sense to delete this test rather than do a separete check for Windows

dcharkes: Yeah, we already have tests in the SDK: https://github.com/dart-lang/sdk/blob/main/tests/ffi/c_types_test.dart#L201 (of course copying those to co19 test doesn't help with our test coverage.)

sgrekhov:
@dcharkes thank you. I made the changes the we've discussed. Please review

eernstg (approved):
LGTM!

PR #1389, opened 2022-08-09
#1388. FFI test failures fixed

sgrekhov:


LibTest/ffi/Long/Long_A01_t01.dart
```diff
@@ -29,7 +29,8 @@ void main() {
     p1.value = 0x7FFFFFFF;
     Expect.equals(0x7FFFFFFF, p1.value);
     p1.value = 0x80000000;
-    Expect.equals(-2147483648, p1.value);
+    Expect.isTrue(-2147483648 == p1.value || // 32-bit representation
```
dcharkes, line 32: Shouldn't these comments be flipped?

The same number is returned on 64 bit because the number is wide enough. On 32 bit the number is converted by interpreting the lower bits as a number.

LibTest/ffi/UnsignedLong/UnsignedLong_A01_t01.dart
```diff
@@ -33,7 +33,8 @@ void main() {
     p1.value = 0x80000000;
     Expect.equals(2147483648, p1.value);
     p1.value = -42;
-    Expect.equals(4294967254, p1.value);
+    Expect.isTrue(4294967254 == p1.value || // 32-bit representation
```
dcharkes, line 36: same

dcharkes (approved):

LibTest/ffi/UnsignedLong/UnsignedLong_A01_t01.dart
sgrekhov: If the same number is returned in case of 64-bit, then the comments are correct. `-42` is transformed to `4294967254` in case of 32 bit and remains `-42` in case of 64 bit.

eernstg (commented):
I'll take this opportunity to learn a bit more about how various number types are handled using FFI, so I can't yet comment on the PR as such.

Just one thing: I assume that the platform is the vm (no other platforms support FFI? yet?), and in that case it might be useful to consider `int.unsigned`: `(-42).toUnsigned(32)` yields the 64 bit zero extending interpretation of the bit pattern which is the 32 lowest bits of the representation of the receiver, so that's a way to get hold of (or check) those big numbers that arise because of signed/unsigned/32-bit/64-bit and whatnot conversions.

LibTest/ffi/Long/Long_A01_t01.dart
sgrekhov: @dcharkes probably I'm wrong, but for me comments are ok here. Value here is `0x8000 0000` (`2147483648`). In case of 64-bit representation it remains the same `0x0000 0000 8000 0000 == 2147483648` (I'm writing numbers using big-endian for simplification) but in case of 32-bit representation it is transformed to `0xFFFF FFFF 8000 0000 == -2147483648`. Please correctm me if I'm wrong

LibTest/ffi/UnsignedLong/UnsignedLong_A01_t01.dart
eernstg: (I'd love to hear the answer, too, @dcharkes! ;-)

eernstg:
OK, I think this one is done. Landing.

LibTest/ffi/Long/Long_A01_t01.dart
dcharkes: Ah my bad, I was misreading the diff, thinking the removed line was the setter.

LibTest/ffi/UnsignedLong/UnsignedLong_A01_t01.dart
dcharkes: same

dcharkes:
> Just one thing: I assume that the platform is the vm (no other platforms support FFI? yet?), and in that case it might be useful to consider `int.toUnsigned`: `(-42).toUnsigned(32)` yields the 64 bit zero extending interpretation of the bit pattern which is the 32 lowest bits of the representation of the receiver, so that's a way to get hold of (or check) those big numbers that arise because of signed/unsigned/32-bit/64-bit and whatnot conversions.

That could indeed simplify this test. I learn something new every day!

PR #1404, opened 2022-08-23
Fixes #1403. Added test that `super` cannot be used as an expression

### LINKED ISSUE #1403: Add tests that it is a compile time error to use `super` as an expression
Add regression test for https://github.com/dart-lang/sdk/issues/43168

sgrekhov:


eernstg (approved):
LGTM

eernstg:
We do need to find a way to handle the situation where a test contains a syntax error, because it is correct for the spec parser to report an error here, but it counts as a failed presubmit check.

For now I'll just land it.

@athomas, is there an easy way to adjust the action that runs the spec parser, or should we perhaps just say "we know when to ignore this failure"?

Issue #1405, opened 2022-08-23
Why do we not have test failures on using deprecated entities?


eernstg:
Consider the occurrence of `@proxy` [here](https://github.com/dart-lang/co19/blob/master/LibTest/core/Invocation/isAccessor_A01_t01.dart#L13). This metadata ought to be tested, because the test isn't skipped by status any files.

Still, it counts as a successful test run in all 52 configurations according to https://dart-ci.firebaseapp.com/current_results/#/filter=co19/LibTest/core/Invocation&showAll.

So perhaps the problem is that this failure is a 'hint', and they are ignored?

Perhaps the easiest way to find these tests is to run `dart analyze` on all of co19. But if that yields too much noise then grepping for `@[dD]eprecated` in `$SDK/sdk/**/*.dart` and then grepping for each deprecated thing might be more convenient.

In any case, it would be very useful if all co19 tests are checked for code which is using deprecated entitiea, and such usages are removed before the actual declarations are removed from the SDK.

eernstg:
@mit-mit, do you have any further comments on what we need to do here?

mit-mit:
We should remove all co19 tests of `@deprecated` core library APIs, as they are about to be removed: https://github.com/dart-lang/sdk/issues/49529

mit-mit:
I think the easiest thing might be to configure the analysis options for co19 to upgrade the info `deprecated_member_use` to be an error?

```
   info • fyz.dart:1:2 • 'proxy' is deprecated and shouldn't be used. Try replacing the use of the deprecated member with the replacement. •
          deprecated_member_use
```

mit-mit:
I ran `dart analyze` with this `analysis_options.yaml` placed at the root of this repo:

```
analyzer:
  errors:
    deprecated_member_use: error
```

Running `dart analyze .` from the root produces several thousand errors:
https://gist.github.com/mit-mit/b5a1372c06140eab5e3aac3fe1e9ec74

It looks like the vast majority are in generated files, so perhaps we can fix those in some templates?

eernstg:
The behavior of our tools when they encounter usages of deprecated declarations should of course be tested somehow, but it should not be a problem that co19 doesn't test this behavior, because it isn't a language-defined behavior. So it should be ok to simply change the code in every single case where it uses a deprecated declaration.

sgrekhov:
@mit-mit thank you. I'll update the test shortly

sgrekhov:
[...]

PR #1407, opened 2022-08-24
#1405. Don't use deprecated API in co19 tests. Update generated files

sgrekhov:
```dart
 }, (e) => e is TypeError || e is CastError);
```
replaced by
```dart
  }, (e) => e is TypeError);
```
No any other changes here


eernstg (approved):
LGTM!

Chrome couldn't show all the libraries: About 350 of them were shown as a blank white bar with a 'Load Diff' button, and when I had looked at "many" of them, the Chrome tab crashed. So I looked at the last 100. I think I covered all of them that way. ;-)

Anyway, the only change I saw was `e is TypeError || e is CastError` --> `e is TypeError`, as expected.

Issue #1399, opened 2022-08-19
[Records] Co19 tests - add tests 


itsjustkevin:

itsjustkevin:
@sgrekhov can you add your issue to the project instead of closing it out as a duplicate?  We would still like to track the work being done.

sgrekhov:
@itsjustkevin sorry. I'll go on the work using this issue and close the previous one

eernstg:
@sgrekhov, please give me a heads up here when you've completed writing these tests.

sgrekhov:
If this PR https://github.com/dart-lang/co19/pull/1418 is Ok, then, I believe, I completed with the tests for records

eernstg:
Thanks!

sgrekhov:
Reopened. Records specification was updated, need to update the tests accordingly

sgrekhov:
Records tests updated according too the current specification. Closing this issue

sgrekhov:
Spec changed, `Record.empty` removed. Need to update the tests accordingly

sgrekhov:
Need to add tests for  the new section "Interactions with libraries using older language versions"

sgrekhov:
Spec was changed (Ambiguity with metadata annotations added), need to update tests accordingly

PR #1408, opened 2022-08-25
#1399 [Records] tests. Part 2. Static semantics and members tests added

sgrekhov:


LanguageFeatures/Records/record_type_literals_A01_t01.dart
```diff
@@ -0,0 +1,25 @@
...
+/// used as an expression
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=records
+
+import "../../Utils/expect.dart";
+
+main() {
+  var t1 = (int, String);
+  var t2 = (int i, String s);
```
eernstg, line 22: I think this would be a syntax error, because `int i` is not an expression, and it is not of the form `id: e` where `id` is an identifier and `e` is an expression.

LanguageFeatures/Records/static_semantics_A02_t01.dart
```diff
@@ -0,0 +1,45 @@
...
+  lib.T2 l2 = r2;
+  lib.T3 l3 = r3;
+
+  Expect.isTrue(r1 is lib.T1);
+  Expect.isTrue(r2 is lib.T2);
+  Expect.isTrue(r3 is lib.T3);
+  Expect.isTrue(lib.t1 is R1);
+  Expect.isTrue(lib.t2 is R2);
+  Expect.isTrue(lib.t3 is R3);
+}
```
eernstg, line 45: Please add the newline at the end of the file, I think it's going to cause silly little problems if we don't have it.

eernstg (commented):
Looks good! I suggested a number of typo fixes.
One thing, though: Please cover equality of reified types as well:

typedef T1 = (int, int);
typedef T2 = (int a, int b);

and then expect T1 == T2. [Normalization](https://github.com/dart-lang/language/blob/master/resources/type-system/normalization.md) would also eliminate the difference between top types with the same choice of dynamic/void/Object? and a few other things, so the type (dynamic, FutureOr<Object?>) should be equal to (dynamic, Object?) but different from (void, dynamic); and (FutureOr<Object>, X) should be equal to (Object, Never) in a context where X is a type variable with bound Never.

No need to check normalization as such, but just include 3-4 cases such that we can see that equality on record types does use normalization as it should.

sgrekhov:
@eernstg thank you. Please rereview. Types equality tests added `static_semantics_A04_t0*.dart`

eernstg (approved):
Just one tiny thing: Please double check a file name which seems to be non-standard.

LanguageFeatures/Records/static_semantics_A02_01.lib.dart
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
```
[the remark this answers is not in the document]
sgrekhov: @eernstg thank you for noticing. Typo fixed

PR #1418, opened 2022-09-07
#1399. [Records] Type inference and promotion tests added

sgrekhov:
I believe that this set of tests covers promotion and inference for records. If I missed anything and we need more tests, please, let me know

### DIFF
--- a/LanguageFeatures/Records/type_inference_A01_t01.dart
+++ b/LanguageFeatures/Records/type_inference_A01_t01.dart
@@ -1,0 +1,23 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Type inference and promotion flows through records in much the 
+/// same way it does for instances of generic classes (which are covariant in 
+/// Dart just like record fields are) and collection literals.
+///
+/// @description Checks type inference for records. Test records with positional
+/// fields only
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=records
+
+import "../../Utils/static_type_helper.dart";
+
+main() {
+  var r1 = (42, "");
+  r1.expectStaticType<Exactly<(int, String)>>();
+
+  final r2 = (null, 3.14);
+  r2.expectStaticType<Exactly<(Null, double)>>();
+}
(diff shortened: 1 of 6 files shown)


sgrekhov:
@eernstg please review

LanguageFeatures/Records/type_inference_A01_t01.dart
```diff
@@ -0,0 +1,34 @@
...
+// SharedOptions=--enable-experiment=records
+
+import "../../Utils/static_type_helper.dart";
+
+main() {
+  var r1 = (42, "");
+  r1.expectStaticType<Exactly<(int, String)>>();
+
+  final r2 = (null, 3.14);
+  r2.expectStaticType<Exactly<(Null, double)>>();
```
eernstg, line 21: This is a very interesting case! We have the following:

```dart
void main() {
  var x = null;
  String s = x; // No compile-time error.
}
```

The point is that a variable of type `Null` isn't very useful, so type inference chooses `dynamic` as the declared type when a variable initializer has the type `Null`.

But I can see that we get `Null` as an inferred type argument for normal instance creation inference:

```dart
class C<X, Y> {
  final X x;
  final Y y;
  C(this.x, this.y);
}

void main() {
  var c = C(null, 3.14);
  print(c.runtimeType); // 'C(Null, double)'.
}
```

With that in mind I think it's OK to expect that `null` gives rise to an inferred record field type of `Null`.

LanguageFeatures/Records/type_promotion_A01_t03.dart
```diff
@@ -0,0 +1,54 @@
...
+    r1.$0;
+    r1.s;
+  } else {
+    r1?.$0;
+    r1?.s;
+  }
+
+  R2 r2 = (42, o: "answer");
+  if (r2 is (int? n, {String o})) {
+    r2.n?.isOdd;
```
eernstg, line 33:
```suggestion
    r2.$0?.isOdd;
```
We can't use a name associated with a positional field as a getter on the record.

eernstg (commented):
Looks good!

One thing would be useful: In some cases there is an if condition (in `if (b) ...`, that's `b`) which is expected to evaluate to false, and we only get the static analysis of the body of the if-statement, because it won't be executed at run time. In this case it would be good to have a different record to perform the test on, such that the condition is true and the code gets executed.

(This is certainly true for the cases where a test is performed on `Record.empty`, but I think I noticed a couple of other cases along the way.)

sgrekhov:
Done. Please rereview

eernstg (approved):
LGTM!

Issue #1429, opened 2022-09-14
Remove obsolete experimental flags


sgrekhov:
Remove obsolete experimental flags from co19 tests

PR #1432, opened 2022-09-16
#1429. Remove obsolete experimental flags: constructor-tearoffs

sgrekhov:


eernstg (approved):
LGTM

I can see that it only removes `--enable-experiment=constructor-tearoffs`, but even if there are some other obsolete experiments we can just handle them in additional PRs.

sgrekhov:
Yes, there will be separate PR's for other obsolete flags. Otherwise, the change will be too big

PR #1464, opened 2022-09-20
Fixes #1455. In constant record expression specify collection type

### LINKED ISSUE #1455: co19/LanguageFeatures/Subtyping/static/generated/records_fail_class_member_super_fail_A07_t01
Just like this is a compile-time error:
```dart
final List<int> v = <dynamic>[];
```

It should happen IMHO when `List` is in a record type.
```dart
const t1Default = const (5, "6", b: []);

class ClassMemberSuper1_t02 {
  (int i, String, {List<num> b}) m = t1Default;
...
```

sgrekhov:


eernstg (approved):
LGTM

PR #1467, opened 2022-09-22
#1399. [Records] Private fields tests added

sgrekhov:


eernstg:
Oops, this feature was discussed this week. The language team concluded that record types have no support for private names (that is, the record type `({String _secret})` is a compile-time error, and so is the record literal `(_secret: 'Hello')`).

I'd suggest that we treat private names of positional fields the same (as an error). As usual, those names are just comments, but it makes sense to treat them the same as the actual names of getters, and they can't be private.

We may add support for private names in record types later on, that would be a non-breaking change.

eernstg:
So for now we'd need to test that all those forms of private names are reported as compile-time errors.

sgrekhov:
We do have tests that checks that name started with underscore is a compile-time error (I forgot to delete them :)). 

https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Records/record_expressions_A04_t01.dart
https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Records/record_type_annotations_A06_t01.dart

So, simply close this PR

PR #1481, opened 2022-10-03
#1399. [Records] `Record.empty` replaced by `()`

sgrekhov:


eernstg (approved):
I was surprised that `(,)` isn't allowed, but the record feature spec does say so.

LGTM.

PR #1484, opened 2022-10-04
Fixes #1483. Removed static analysis warnings for different parameter override value

### LINKED ISSUE #1483: Remove all static analysis warnings for different parameter override value
For https://github.com/dart-lang/sdk/issues/49112, we will stop reporting `STATIC_WARNING.INVALID_OVERRIDE_DIFFERENT_DEFAULT_VALUES_POSITIONAL` and `STATIC_WARNING.INVALID_OVERRIDE_DIFFERENT_DEFAULT_VALUES_NAMED` altogether. These should be removed from pre-null safety co19 tests (`co19_2` directory in `sdk` repo).

See failures here when analyzer stops reporting these: <https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8801563216397856865/+/u/test_results/new_test_failures__logs_>

sgrekhov:


eernstg (commented):
Do the tools actually behave in this manner? I would expect legacy libraries to treat different default values in override relationships as they have treated them previously (that is, it's a warning), and only newer language versions would allow them without any diagnostic messages.

If we don't have any evidence to the contrary then I believe we should not change these legacy tests.

sgrekhov:
Description of #1483 contains tryjobs logs, so, I believe, there should be a CL which makes the changes in the tool. @srawlins, please confirm.

srawlins:
Yes, perfect!

eernstg:
Very good! I just wanted to make sure we weren't causing unnecessary churn in legacy tests or tools by changing the expected behavior with legacy code.

PR #1492, opened 2022-10-06
Fixes #1486. Replace empty records by ones with named fields

### LINKED ISSUE #1486: Empty record literals shouldn't be treated as errors
The lines 53 and 68 in https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Records/record_type_annotations_A02_t08.dart should not be considered errors because `()` is the empty record literal of type `()`.

sgrekhov:


eernstg (commented):
Just wondering, why replace the empty records? `()` is both an expression and a type today..

eernstg (approved):
Aha, OK! Thanks! - LGTM.

PR #1496, opened 2022-10-07
Fixes #1490. Missing parameters names added

### LINKED ISSUE #1490: Missing parameter name in a typedef in a records test
According to https://github.com/dart-lang/sdk/issues/50146#issuecomment-1270053884, a name should be given to the parameter of a record type in lines 45 and 49 of https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Records/record_type_annotations_A02_t11.dart.

sgrekhov:


eernstg (approved):
LGTM

However, I think there could be more like 10 of these, based on a run of the spec parser on the currently-rolled-in version of the co19 tests. Perhaps they were already fixed in newer commits? Otherwise you could run the spec parser and see that syntax errors only occur in tests where there should be syntax errors.

(The spec parser was updated to recognize `()` as an expression and as a type yesterday.)

PR #1507, opened 2022-10-18
Fixes #1505. [Records] Added tests for dynamic access to record fields

### LINKED ISSUE #1505: [Records] Add more tests for dynamic access to record fields
While implementing dynamic access to record fields (each field is exposed as a getter, and all getters can be called from `dynamic`), I noticed that there are no corresponding co19 tests for that. However this is a non-trivial part of the records implementation. In general, we should try to ensure that `dynamic` semantics at run time matches semantics of corresponding statically typed code (except extensions and similar static features).

So consider adding tests for accessing record fields dynamically, including function calls via field (`dynamic x = (foo: myFunc); x.foo(42);` should call `myFunc(42)`).

Example of such test in Dart SDK: https://github.com/dart-lang/sdk/blob/main/tests/language/records/simple/dynamic_field_access_test.dart.

sgrekhov:


LanguageFeatures/Records/members_A04_t03.dart
```diff
@@ -0,0 +1,60 @@
...
+  Expect.equals(1, r3.foo(1));
+  Expect.throws(() {r3.foo("");});
+  Expect.throws(() {r3.$0;}, (e) => e is NoSuchMethodError);
+  Expect.throws(() {r3.$100000000000000000000;}, (e) => e is NoSuchMethodError);
+  Expect.throws(() {r3.any;}, (e) => e is NoSuchMethodError);
+
+  Expect.equals("Lily was here", r4.$0("Lily was here"));
+  Expect.throws(() {r4.$0(42);});
+  Expect.equals(1, r4.newC(1).x);
+  Expect.throws(() {r4.newC(42);});
```
alexmarkov, line 54: Why is this code expected to throw? It is similar to the previous line.

alexmarkov (approved):
lgtm

Thanks for adding these tests!

LanguageFeatures/Records/members_A04_t03.dart
sgrekhov: Thank you for noticing! Fixed

sgrekhov:
@chloestefantsova @eernstg this PR was approwed. Please merge it

PR #1514, opened 2022-10-24
Fixes #1428. Subtyping. Unused variables removed from generated tests

### LINKED ISSUE #1428: Update Subtyping tests
- `dynamic/arguments_binding_A01_t02.dart` remove forgetType()
- `dynamic/arguments_binding_A03_t03.dart` use `mixin` keyword
- update generator to delete `<-- NotGenericFunctionType`, remove empty lines,  fix indents in `//    ^^^`,  remove not used variables

sgrekhov:


eernstg (approved):
LGTM.

I could see that it wasn't a big change to 'generator.dart', so it's likely that it was a good trade-off to do this: Readers of this code will have a slightly easier time making sense of everything, and 'generator.dart' will continue to perform this clean-up job each time the code is re-generated. Thanks!

PR #1515, opened 2022-10-24
#1399. [Records] Check corner case with trailing zeros for positional  field getter

sgrekhov:


sgrekhov:
@chloestefantsova please review

chloestefantsova (approved):

Issue #1401, opened 2022-08-19
[Patterns] Co19 tests - add tests


itsjustkevin:

sgrekhov:
@itsjustkevin @eernstg is this a right feature specification link for Patterns? https://github.com/dart-lang/language/blob/master/working/0546-patterns/patterns-feature-specification.md

eernstg:
Yes!

sgrekhov:
Completed

sgrekhov:
Reopened. Spec for [exaustiveness](https://github.com/dart-lang/language/pull/2948) is ready for testing. Need to add appropriate tests

sgrekhov:
Completed

eernstg:
Thump! says the big door as it is closed. Thanks for the great work on creating all those tests!

PR #1520, opened 2022-10-25
#1401. [Patterns] Logical-or tests added

sgrekhov:


LanguageFeatures/Patterns/logical_or_A02_t01.dart
```diff
@@ -0,0 +1,46 @@
...
+      break;
+    default:
+      Expect.equals(expectedArea, s);
+      Expect.equals(expectedType, shape.runtimeType);
+      Expect.isFalse(match);
+  }
+}
+
+main() {
+  test(Circle(1), 3.14, Circle, true);
```
eernstg, line 43: Equality on objects of type `double` is really tricky (e.g., `0.1 + 0.1 + 0.1 == 0.3` evaluates to 'false'). I can't immediately see how the result gets to be `3.14`, but if there is any amount of computation involved we should probably use a different test (absolute value of difference is smaller than 0.0000001 or something like that).

LanguageFeatures/Patterns/logical_or_A02_t02.dart
```diff
@@ -0,0 +1,45 @@
...
+  switch (list) {
+    case [1 | 2, 3]:
+      result = true;
+    default:
+  };
+  return result;
+}
+
+main() {
+  Expect.isFalse([0, 3]);
```
eernstg, line 38:
```suggestion
  Expect.isFalse(matches([0, 3]));
```
as usual.

LanguageFeatures/Patterns/logical_or_A04_t01.dart
```diff
@@ -0,0 +1,67 @@
...
+/// restrictions apply:
+///
+/// The two branches must define the same set of variables.
+///
+/// If the left branch matches, the right branch is not evaluated. This
+/// determines which value the variable gets if both branches would have
+/// matched. In that case, it will always be the value from the left branch.
+///
+/// @description Checks that if the left branch matches, the right branch is not
+/// evaluated
```
eernstg, line 21: I can't immediately see any reason why this test would detect the bug which is that the right branch _is_ evaluated in a situation where the left branch matches. Does it actually detect that situation?

You could say that we get the expected value for `s`, but that's not a safe test: The implementation could still evaluate both branches and then initialize `s` to the value that corresponds to the match that succeeded. Or it could evaluate the right operand first and then the left operand, or it could use some other ordering.

I think it's necessary to (1) have a side effect in a getter such that we can verify that a specific getter was indeed invoked, and (2) check that the getters corresponding to the patterns that should be skipped are not invoked.

An approach that I've seen in many language tests is like this:

```dart
String log = '';
void addToLog(String s) => log = log.isEmpty ? s : '$log, $s';

class C {
  int get g {
    addToLog('C.g');
    return 0;
  }
}
```

LanguageFeatures/Patterns/patterns_lib.dart
```diff
@@ -0,0 +1,40 @@
...
+class Rectangle implements Shape {
+  final double x, y;
+  Rectangle(this.x, this.y);
+
+  double get area => x * y;
+}
+
+class Point implements Shape {
+  Point();
+  Never get area => throw Exception("Point has no area");
```
eernstg, line 39: You could say that the area is zero, but throwing may be useful in order to cover the case where a matching process throws. So: OK!

eernstg (commented):
Thanks! -- I added a few comments to fix some typos.

LanguageFeatures/Patterns/logical_or_A04_t01.dart
sgrekhov: The idea was that if the right branch is evaluated then `Point.area` getter will throw an exception. But now I see that it was a wrong idea. If the right branch is evaluated then an extractor name will be checked first. Runtime type of the variable is not a `Point`, so `Point.area` getter wont't be called.  I'll change the test in the way you've suggested. Thanx!

sgrekhov:
I've updated the tests. Please rereview

eernstg (commented):
Thanks! I think it's useful to take one more step: We need to see that when it says "if the left branch doesn't match, then the right branch is evaluated", some matching activity does actually occur in that left branch, and that could be done by adding a few more tests.

sgrekhov:
@eernstg I updated the tests. Now it logs all `Shape.area` comparison attempts (including the value that we are comparing with), so we are able to check that
```dart
  Shape shape4 = Rectangle(1, 2);
  bool other = false;
  switch (shape4) {
    case Rectangle(area: 3) | Rectangle(area: 1) | Rectangle(area: 42):
      Expect.equals("Rectangle.area=3;Rectangle.area=1;Rectangle.area=42;",
          shape4.log);
      Expect.fail("No branches should match");
      break;
    default:
      other = true;
  }
  Expect.isTrue(other);
```
Please rereview

LanguageFeatures/Patterns/logical_or_A04_t02.dart
```diff
@@ -27,33 +27,46 @@ import "patterns_lib.dart";
...
   }
 
-  Shape shape2 = Shape.withLog();
-  clearLog();
+  Shape shape2 = Shape();
   switch (shape2) {
-    case Square(area: var s) | Rectangle(area: var s) | Shape(area: var s):
-      Expect.equals("Shape.area", shapesLog);
+    case Square(area: 2) | Rectangle(area: 1) | Shape(area: 0):
+      Expect.equals("Shape.area=2;Shape.area=1;Shape.area=0;", shape2.log);
```
eernstg, line 42: I can't immediately see how this would produce a log as stated. Wouldn't it simply write 'Shape.area=0' into the log after having failed already at the type test of the first two patterns (which wouldn't log anything)?

LanguageFeatures/Patterns/logical_or_A04_t02.dart
```diff
@@ -27,33 +27,46 @@ import "patterns_lib.dart";
...
+    default:
+      print("Other");
+  }
+
+  Shape shape4 = Rectangle(1, 2);
+  bool other = false;
+  switch (shape4) {
+    case Rectangle(area: 3) | Rectangle(area: 1) | Rectangle(area: 42):
+      Expect.equals("Rectangle.area=3;Rectangle.area=1;Rectangle.area=42;",
+          shape4.log);
```
eernstg, line 65: Perhaps line 64-65 should simply be deleted?

LanguageFeatures/Patterns/patterns_lib.dart
```diff
@@ -9,67 +9,86 @@
 
 library patterns_lib;
 
-String shapesLog = "";
-
-void logger(String log) {
-  shapesLog = shapesLog.isEmpty ? log : "$shapesLog, log";
-}
+class Area {
```
eernstg, line 12: Nice!

eernstg (commented):
Looks good, just a couple of questions arose!

LanguageFeatures/Patterns/logical_or_A04_t02.dart
eernstg: Oh, line 64-65 could be moved to the default case, right?

sgrekhov: Of course! Thank you for pointing! Fixed.

LanguageFeatures/Patterns/logical_or_A04_t02.dart
eernstg: No, it's the other way around: We're considering an object `o` whose run-time type is `Shape`, and we're testing whether `o` has type `Square`. The answer is "no", because we do not have `Shape <: Square`, we only have `Square <: Shape`.

eernstg (commented):
Just one comment still alive.

sgrekhov:
Thank you for the clarification. Fixed

eernstg (approved):
LGTM!

PR #1524, opened 2022-10-27
#1399. [Records] Tests for ambiguity with metadata annotations added

sgrekhov:


sgrekhov:
@eernstg please review

LanguageFeatures/Records/metadata_annotations_A01_t02.dart
```diff
@@ -0,0 +1,64 @@
...
+/// comments (according to the existing WHITESPACE and COMMENT lexical rules)
+/// between the constructorDesignation and arguments. In other words, for an
+/// argument list to be part of the metadata annotation, the ( must occur
+/// immediately after the last character in the constructorDesignation. The last
+/// character in constructorDesignation may be an identifier or the > in a type
+/// argument list.
+///
+/// @description Checks that if there is a space between @metadata and (a, b)
+/// then (a, b) denotes a record. Test functions with comment between a
+/// metadata and a return type
```
eernstg, line 34: I guess the comment with no space is more important here, because it won't catch a bug where the space triggers "the next thing is not an argument list", but the comment with no spaces (before or after) doesn't. So we should have all the cases with no spaces, and then just a couple of cases where we have both comment and space.

LanguageFeatures/Records/metadata_annotations_A01_t04.dart
```diff
@@ -0,0 +1,64 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=records
+
+import "../../Utils/expect.dart";
+
+const Meta = "meta";
+
+main() {
+  @Meta /* comment */ (int, String) r1 = (1, "2");
```
eernstg, line 44: Again, make sure we cover the no-space cases extensively, and then a couple of comment-with-space cases are fine, too.

LanguageFeatures/Records/metadata_annotations_A01_t01.dart
```diff
@@ -0,0 +1,64 @@
...
+
+@Meta (int, String) foo1() => (1, "2");
+
+@Meta
+(int, String) foo2() => (3, "4");
+
+@Meta () foo3() => ();
+
+@Meta
+() foo4() => ();
```
eernstg, line 51: This one is good to have: The "space" can be any kind of whitespace, and a single newline character (or `\n\r` on Windows) should suffice to prevent `()` from being part of the metadata.

LanguageFeatures/Records/metadata_annotations_A01_t05.dart
```diff
@@ -0,0 +1,98 @@
...
+
+class Meta {
+  const Meta(Type t1, Type t2);
+}
+
+class M {
+  const M();
+}
+
+  @Meta (int, String) foo1() => (1, "2");
```
eernstg, line 46: I understand that it is tempting to use extra indentation here (rather than using the heavy `[error line 46, column 1, length 5]` syntax), but it would probably be better to use a different workaround:

```dart
/**/ @Meta (int, String) ...
```

That syntax will stay in place even in the case where someone does `dart format`.

eernstg (commented):
Looks good!

I think we should make sure that the strictest possible test is used with comments: They should be verified to trigger the "not part of metadata" behavior even in the case where there is no whitespace.

I also suggested that it could be convenient to use a comment to force a non-standard indentation (because the test could so easily be disrupted by `dart format`, and it wouldn't necessarily be obvious to the reader of the code what happened).

sgrekhov:
@eernstg please rereview

eernstg (approved):
LGTM!

PR #1533, opened 2022-11-02
Fixes #1528. Update on clauses ambiguity tests

### LINKED ISSUE #1528: co19/LanguageFeatures/Records/on_clauses_A02_t01
I think this is not an ambiguity.
```dart
main() {
  try {
    print(0);
  } on String {
  } on ([int? i, String? n]) {
//      ^
// [analyzer] unspecified
// [cfe] unspecified
  }
  on();
//^^
// [analyzer] unspecified
// [cfe] unspecified
}
```
Here, we have one required `on` clause after `try`, so we don't have to attempt to parse `on ([int? i, String? n])` as `on RecordType`. And it is not a valid record type, so we parse it as a local function declaration with the name `on`. And the next `on();` invokes it.

sgrekhov:


eernstg (approved):
LGTM.

PR #1534, opened 2022-11-04
#1401. [Patterns] logical-and pattern tests added

sgrekhov:


LanguageFeatures/Patterns/logical_and_A04_t01.dart
```diff
@@ -0,0 +1,45 @@
...
+
+main() {
+  Shape shape = Circle(1);
+  switch (shape) {
+    case Circle(area: var s) & Circle(area: var s):
+//                                              ^
+// [analyzer] unspecified
+// [cfe] unspecified
+      break;
+    case Rectangle(x: var x, y: var width) | Rectangle(:var x, :var y):
```
eernstg, line 33: Probably:
```suggestion
    case Rectangle(x: var x, y: var width) & Rectangle(:var x, :var y):
```

LanguageFeatures/Patterns/logical_and_A05_t01.dart
```diff
@@ -0,0 +1,53 @@
...
+
+main() {
+  Shape shape1 = Circle(1);
+  switch (shape1) {
+    case Circle(area: 2) & Circle(area: 3.14):
+      Expect.fail("Pattern should not match");
+      break;
+    default:
+  }
+  Expect.equals("Circle.area=2;", shape1.log);
```
eernstg, line 34: We can't really see the difference between "the rightmost object pattern wasn't matched" and "the rightmost pattern was matched, but using the cached value for `area`", so we should use two distinct getters and see that the second one isn't called.

LanguageFeatures/Patterns/logical_and_A05_t01.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  Shape shape2 = Square(1);
+  switch (shape2) {
+    case Square(area: 1) & Square(area: 2) & Square(area: 1):
+      Expect.fail("Pattern should not match");
+      break;
+    default:
+  }
+  Expect.equals("Square.area=1;Square.area=2;", shape2.log);
```
eernstg, line 43: Same thing here.

LanguageFeatures/Patterns/logical_and_A05_t01.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  Shape shape3 = Shape();
+  switch (shape3) {
+    case Rectangle(area: 0) & Circle(area: 0) & Square(area: 1) & Shape(area: 0):
+      Expect.fail("Pattern should not match");
+      break;
+    default:
+  }
+  Expect.equals("Shape.area=0;Shape.area=0;Shape.area=1;", shape3.log);
```
eernstg, line 52: We should again use different getters here to avoid confusing matching/not-matching and calling-first-time/using-cached-value.

Also, the test for `Rectangle` fails with a `Shape`, so we should never get to matching in any of the remaining patterns. Better to check for `Shape` every time, or having a subtype and checking for `Shape` plus checking for that subtype, such that it doesn't fail because of the class test.

eernstg (commented):
A few comments.

LanguageFeatures/Patterns/logical_and_A05_t01.dart
sgrekhov: Fixed

LanguageFeatures/Patterns/logical_and_A05_t01.dart
sgrekhov: Fixed

sgrekhov (commented):
Please review once again

eernstg (approved):
LGTM

PR #1538, opened 2022-11-09
#1401. [Patterns] Constant pattern tests added

sgrekhov:


LanguageFeatures/Patterns/constant_A06_t01.dart
```diff
@@ -61,6 +61,18 @@ main() {
...
+// [cfe] unspecified
+    case 1 + 2:
+//       ^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    case 1 & 2:
+//       ^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    case 1 | 2:
```
eernstg, line 73: Given that `1` and `2` are <constantPattern>s, I'd say that this is a <logicalOrPattern>, which should not be an error.

We have many other forms: All constant expressions that cannot be derived by the new `constantPattern` rule.

eernstg (commented):
I think it would be really useful to try to cover all the forms that used to be allowed which are no more derivable in the new grammar.

With `@dart = 2.19` they should be allowed, with the patterns experiment they should be syntax errors.

LanguageFeatures/Patterns/constant_A06_t01.dart
sgrekhov: > Given that 1 and 2 are s, I'd say that this is a , which should not be an error.

Do you mean that this is a "logical-or" pattern which is not an error?

eernstg: Yes, I just edited the text to have the missing backquotes. ;)

sgrekhov:
More forms added. Please rereview

LanguageFeatures/Patterns/constant_A06_t02.dart
```diff
@@ -0,0 +1,89 @@
...
+/// constant's value. We don't allow all expressions here because many
+/// expression forms syntactically overlap other kinds of patterns. We avoid
+/// ambiguity while supporting terse forms of the most common constant
+/// expressions like so:
+/// ...
+/// It is a compile-time error if a constant pattern's value is not a valid
+/// constant expression.
+///
+/// @description Check that it is a compile-time error if a constant pattern's
+/// value is not a valid constant expression. Test expressions which are valid
```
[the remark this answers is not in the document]
eernstg: Another thing is that we _should_ test the case where 'the constant pattern's value is not a valid constant expression'. 

That would be something like `case a:` where `a` is a non-constant variable, `case const C():` where `C` is a class that has a constructor `C()` but it isn't constant, and `case const (C().foo):` where `C` is a class that has a constant constructor and `foo` is a getter (so everything is fine except that `C().foo` is not a constant expression).

Just a few cases such that we can see that there is a check for "is a constant expression". There's no need to cover expressions that aren't constant in general (because we would have other tests covering that, independently of patterns).

LanguageFeatures/Patterns/constant_A06_t03.dart
```diff
@@ -0,0 +1,54 @@
...
+const l = [1, 2, 3];
+
+main() {
+  Object value = Object();
+  switch (value) {
+    case 2 / 1:
+//       ^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    case 1 && 2:
```
eernstg, line 40: We should use type correct expressions: Otherwise we could have a spurious success, because the compiler reports an error for the non-boolean operands, but it could still have a bug which is "parser accepted `case 1 && 2:` even though that's a syntax error".

LanguageFeatures/Patterns/constant_A06_t03.dart
```diff
@@ -0,0 +1,54 @@
...
+//       ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    case 1 || 2:
```
eernstg, line 44: Ditto.

eernstg (commented):
A few comments.

LanguageFeatures/Patterns/constant_A06_t02.dart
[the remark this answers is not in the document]
sgrekhov: > At this point we could also keep it simple and say that this test is concerned with syntax that is a correct <caseHead> in Dart-before-patterns, but is a syntax error in Dart-with-patterns.

That's what I'm doing here. Test description reads "Test expressions which are valid in Dart 2.19 but not valid in Dart-with-patterns".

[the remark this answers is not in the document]
sgrekhov: Another thing is that we should test the case where 'the constant pattern's value is not a valid constant expression'.

The previous test `constant_A06_t01' does it

sgrekhov (commented):
I did requested changes. Please rereview

LanguageFeatures/Patterns/constant_A06_t03.dart
sgrekhov: Indeed. Fixed

LanguageFeatures/Patterns/constant_A06_t01.dart
```diff
@@ -61,6 +61,10 @@ main() {
     case C():
```
eernstg, line 61: If we maintain that this `<caseHead>` has a `<constantPattern>` then this would need to be `case const C():` (or something else, but it can't just be `case C():` because that's a syntax error, not a constant pattern). I think it would be useful to have `case const C()` because it uses the alternative `<constObjectExpression>` in `<constantPattern>`.

Another thing we ought to have is a string literal which isn't constant (like `'x is $x!'`).

LanguageFeatures/Patterns/constant_A06_t01.dart
```diff
@@ -61,6 +61,10 @@ main() {
...
 //       ^^^
 // [analyzer] unspecified
+// [cfe] unspecified
+    case const(C().v):
```
eernstg, line 65: Yes!

Just one personal style nit (probably also enforced by `dart format`): You could say "we must aim to test all possible programs, not just the pretty ones", which is true, but I still prefer a style where keywords are marked as different from expressions denoting functions by having a space before '('. So it's `if (b) S` and `while (b) S`, not `if(b) S` and `while(b) S`; similarly `const (C().v)` rather than `const(C().v)`. So unless you really want to test that the lexer can handle `const(`, `const (` would be nice. ;-)

LanguageFeatures/Patterns/constant_A06_t02.dart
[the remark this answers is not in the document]
eernstg: OK, I had forgotten that `constant_A06_t01` included `case a:` and such.

But then we should adjust the description here: The current description says

> a constant pattern's value is not a valid constant expression

but that description only fits if we're actually looking at a `<constantPattern>`, and that non-terminal cannot derive `1 + 2`, `1 - 2`, etc.

The description could instead be 'Check that a syntax error is emitted for a term which could be derived from `<caseHead>` in Dart 2.19, but cannot be derived from `<caseHead>` when patterns are added to Dart.'

eernstg (commented):
Thanks! Two more comments, and one thread continued.

sgrekhov:
@eernstg I did the changes requested. Additionally I doubled the number of tests. I think we need to tests constant patterns in `if-case` statement as well. Please rereview

LanguageFeatures/Patterns/constant_A03_t02.dart
```diff
@@ -23,87 +23,80 @@
...
-      return "{'1', '2', '3'}";
-    case const <double>{}:
-      return "{}";
-    default:
-      return "default";
+  if (value case const {1, 2, 3}) {
+    return "{1, 2, 3}";
+  } else if (value case const {'1', '2', '3'}) {
+    return "{'1', '2', '3'}";
+  } else if (value case const {}) {
```
eernstg, line 67: I suspect that `{}` could be a `Map<dynamic, dynamic>`.
Cf . https://github.com/dart-lang/language/issues/2637.

For now, we could use `<SomeType>{}` to disambiguate the collection literal, and then we can have a separate test to check that a couple of ambiguous cases using `const {}` as a constantPattern are disambiguated correctly.

LanguageFeatures/Patterns/constant_A03_t02.dart
```diff
@@ -23,87 +23,80 @@
...
-  switch (value) {
-    case const <int>[1, 2]:
-      return "[1, 2]";
-    case const <String>["3", "4"]:
-      return "['3', '4']";
-    case const <double>[]:
-      return "[]";
-    default:
-      return "default";
+  if (value case const [1, 2]) {
```
eernstg, line 35: The type argument is tricky here, cf. https://github.com/dart-lang/language/issues/2637.

LanguageFeatures/Patterns/constant_A03_t02.dart
```diff
@@ -23,87 +23,80 @@
...
+    return "{'1', '2', '3'}";
+  } else if (value case const {}) {
+    return "{}";
+  } else {
+    return "default";
   }
 }
 
 main() {
   Expect.equals("[1, 2]", testList(const [1, 2]));
```
eernstg, line 75: In this case the type argument for the constant list is definitely `dynamic`.

LanguageFeatures/Patterns/constant_A03_t02.dart
eernstg: Similarly for several other locations below with the same structure.

LanguageFeatures/Patterns/constant_A03_t03.dart
```diff
@@ -23,26 +23,87 @@
...
       return "{}";
     default:
       return "default";
   }
 }
 
 main() {
-  const Map m = {};
-  Expect.equals("{}", test(m));
+  Expect.equals("[1, 2]", testList(const [1, 2]));
```
eernstg, line 76: This will definitely fail: The type argument is taken from the context type, yielding `<dynamic>[1, 2]`, and it's compared to `<int>[1, 2]`.

LanguageFeatures/Patterns/constant_A03_t03.dart
```diff
@@ -23,26 +23,87 @@
...
+  Expect.equals("default", testList([1, 2]));
+  Expect.equals("default", testList(const <num>[1, 2]));
+  Expect.equals("['3', '4']", testList(const ["3", "4"]));
```
eernstg, line 79: Same failure, `<dynamic>[....]` vs `<String>[....]`.

LanguageFeatures/Patterns/constant_A03_t03.dart
```diff
@@ -23,26 +23,87 @@
...
+  Expect.equals("default", testList(["3", "4"]));
+  Expect.equals("default", testList(const <Object>["3", "4"]));
+  Expect.equals("[]", testList(const <double>[]));
+  Expect.equals("default", testList(const []));
+  Expect.equals("default", testList(const <num>[]));
+
+  Expect.equals("{1: 2}", testMap(const {1: 2}));
```
eernstg, line 86: Same failure again.

LanguageFeatures/Patterns/constant_A03_t04.dart
```diff
@@ -0,0 +1,106 @@
...
+    return "{'1', '2', '3'}";
+  } else if (value case const <double>{}) {
+    return "{}";
+  } else {
+    return "default";
+  }
+}
+
+main() {
+  Expect.equals("[1, 2]", testList(const [1, 2]));
```
eernstg, line 73: One more failure based on mismatched type arguments. More of them below.

LanguageFeatures/Patterns/constant_A03_t05.dart
```diff
@@ -0,0 +1,58 @@
...
+/// constant's value. We don't allow all expressions here because many
+/// expression forms syntactically overlap other kinds of patterns. We avoid
+/// ambiguity while supporting terse forms of the most common constant
+/// expressions like so:
+/// ...
+/// List literals are ambiguous with list patterns, so we only allow list
+/// literals explicitly marked const. Likewise with set and map literals versus
+/// map patterns.
+///
+/// @description Check empty map and set literals in a constant patterns
```
eernstg, line 26:
```suggestion
/// @description Check empty map and set literals in constant patterns
```

LanguageFeatures/Patterns/constant_A05_t02.dart
```diff
@@ -0,0 +1,127 @@
...
+/// expression forms syntactically overlap other kinds of patterns. We avoid
+/// ambiguity while supporting terse forms of the most common constant
+/// expressions like so:
+/// ...
+/// Other constant expressions must be marked const and surrounded by
+/// parentheses. This avoids ambiguity with null-assert, logical-or, and
+/// logical-and patterns. It also makes future extensions to patterns and
+/// expressions less likely to collide.
+///
+/// @description Check const constructors calls in a constant patterns. Test
```
eernstg, line 27:
```suggestion
/// @description Check const constructors calls in constant patterns. Test
```

LanguageFeatures/Patterns/constant_A06_t05.dart
```diff
@@ -0,0 +1,54 @@
...
+/// A constant pattern determines if the matched value is equal to the
+/// constant's value. We don't allow all expressions here because many
+/// expression forms syntactically overlap other kinds of patterns. We avoid
+/// ambiguity while supporting terse forms of the most common constant
+/// expressions like so:
+/// ...
+/// It is a compile-time error if a constant pattern's value is not a valid
+/// constant expression.
+///
+/// @description Check that it is a compile-time error if a constant pattern's
```
eernstg, line 25: Same fix to the description: "cannot be derived in Dart with patterns".

eernstg (commented):
Wow, a _bunch_ of stuff here! But it's good that we cover a lot of cases, and there are many cases in patterns, no way to deny that.

LanguageFeatures/Patterns/constant_A03_t02.dart
sgrekhov: @eernstg  please clarify how exactly this test should be rewritten? I'm confused because now (`Dart SDK version: 2.19.0-374.0.dev (dev) (Fri Nov 4 12:40:18 2022 -0700) on "windows_x64"`) we have the following behavior.
```dart
String test(List value) {
  switch (value) {
    case const [1, 2]:
      return "[1, 2]";
    default:
      return "default";
  }
}

String testDynamic(List value) {
  switch (value) {
    case const <dynamic>[1, 2]:
      return "<dynamic>[1, 2]";
    default:
      return "default";
  }
}

String testInt(List value) {
  switch (value) {
    case const <int>[1, 2]:
      return "<int>[1, 2]";
    default:
      return "default";
  }
}

main() {
  Expect.equals("[1, 2]", test(const [1, 2])); // Ok
  Expect.equals("<dynamic>[1, 2]", testDynamic(const [1, 2])); // Ok
  Expect.equals("<int>[1, 2]", testInt(const [1, 2])); // Fails
}
```
But in case of `if-case` statement we have
```dart
String test(List value) {
  if (value case const [1, 2]) {
    return "[1, 2]";
  } else {
    return "default";
  }
}

String testDynamic(List value) {
  if (value case const <dynamic>[1, 2]) {
    return "<dynamic>[1, 2]";
  } else {
    return "default";
  }
}

String testInt(List value) {
  if (value case const <int>[1, 2]) {
    return "<int>[1, 2]";
  } else {
    return "default";
  }
}

main() {
  Expect.equals("[1, 2]", test(const [1, 2])); // fail
  Expect.equals("<dynamic>[1, 2]", testDynamic(const [1, 2])); // fail
  Expect.equals("<int>[1, 2]", testInt(const [1, 2])); // fail
}
```
What is wrong and what is correct here?

sgrekhov:
@eernstg thank you for the detailed review. I did the changes, assuming that `switch` and `if-case` should behave the same. Please rereview.

eernstg (approved):
LGTM!

eernstg:
I can see that there are some merging conflicts, so I won't land it yet.

sgrekhov:
Conflicts resolved

eernstg (approved):
LGTM

PR #1542, opened 2022-11-11
Fixes #1541. Fix covariance_t07.dart. Use correct instance for testing

### LINKED ISSUE #1541: The test Super-mixins/covariance_t07.dart needs an adjustment
https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Super-mixins/covariance_t07.dart contains the instance creation `new Y()` in line 43. I believe it's intended to contain `new X()` at that point, which would give rise to the dynamic type error of https://github.com/dart-lang/sdk/issues/35111:

When an `X` is passed (as required by the member named `test` in the superinterface `M<X>` of `MA`), `test` will invoke `bar` with that instance of `X`, but the superinterface `B<Y>` of `MA` has a `bar` implementation that requires a `Y` (subtype of `X`).

As an aside, it might be helpful to use the names `X` and `Y` (and perhaps `T`) for type variables, and names like `A` .. `D` and multi-char words for classes. It is a bit easier to read a test library when the naming convention directly serves to remind the reader that "this is a class", "this is a type variable", etc.

sgrekhov:


eernstg (approved):
LGTM!

Issue #1540, opened 2022-11-11
Test expectations need updates


johnniwinther:
These tests need to be updated to match CFE expectation when [[cfe] Use fullMemberIterator in ClassMembersNodeBuilder](https://dart-review.googlesource.com/c/sdk/+/268946) has landed:

```
co19/Language/Classes/Getters/same_name_method_t01 MissingCompileTimeError (expected Pass)
co19/Language/Classes/Getters/same_name_method_t02 MissingCompileTimeError (expected Pass)
co19/Language/Classes/Instance_Methods/same_name_getter_t01 MissingCompileTimeError (expected Pass)
co19/Language/Classes/same_name_members_t04 MissingCompileTimeError (expected Pass)
co19/Language/Classes/same_name_members_t06 MissingCompileTimeError (expected Pass)
co19/Language/Classes/same_name_members_t08 MissingCompileTimeError (expected Pass)
co19/Language/Classes/same_name_members_t09 MissingCompileTimeError (expected Pass)
co19/Language/Classes/same_name_members_t11 MissingCompileTimeError (expected Pass)
co19/Language/Overview/Scoping/conflicting_names_t19 MissingCompileTimeError (expected Pass)
co19/Language/Overview/Scoping/conflicting_names_t20 MissingCompileTimeError (expected Pass)
co19/Language/Overview/Scoping/conflicting_names_t21 MissingCompileTimeError (expected Pass)
co19/Language/Overview/Scoping/conflicting_names_t45 MissingCompileTimeError (expected Pass)
co19_2/Language/Classes/Getters/same_name_method_t01 MissingCompileTimeError (expected Pass)
co19_2/Language/Classes/Getters/same_name_method_t02 MissingCompileTimeError (expected Pass)
co19_2/Language/Classes/Instance_Methods/same_name_getter_t01 MissingCompileTimeError (expected Pass)
co19_2/Language/Classes/Setters/static_setter_t02 MissingCompileTimeError (expected Pass)
co19_2/Language/Classes/Setters/static_setter_t04 MissingCompileTimeError (expected Pass)
co19_2/Language/Classes/same_name_members_t04 MissingCompileTimeError (expected Pass)
co19_2/Language/Classes/same_name_members_t05 MissingCompileTimeError (expected Pass)
co19_2/Language/Classes/same_name_members_t06 MissingCompileTimeError (expected Pass)
co19_2/Language/Classes/same_name_members_t08 MissingCompileTimeError (expected Pass)
[...]

PR #1544, opened 2022-11-14
#1540. Tests error expectations updated

sgrekhov:


Language/Classes/Setters/static_setter_t02.dart
```diff
@@ -22,6 +22,7 @@ class C {
   static set foo(String s) {
 //           ^
 // [analyzer] unspecified
+    // [cfe] unspecified
```
johnniwinther, line 25: The indentation is inconsistent here.

johnniwinther (approved):

eernstg:
Hi, @sgrekhov! Could you add just a few words about the nature of these changes? I'm not quite sure what to look for when I'm reviewing.

sgrekhov:
> Hi, @sgrekhov! Could you add just a few words about the nature of these changes? I'm not quite sure what to look for when I'm reviewing.

This PR fixes https://github.com/dart-lang/co19/issues/1540. Change https://dart-review.googlesource.com/c/sdk/+/268946 was landed, CFE chander reporting an errors so I'm changing the tests accordingly

eernstg (approved):
I do not understand why the code changed from having try/catch to not having it. However, the test was always expecting a compile-time error, so there's no need for a try/catch construct (unless there is some subtle effect associated with control flow analysis, but I can't see how that could be the case in these tests). So it's fine that the try/catch is gone, but it would be nice to hear just a couple of words about this topic. ;-)

The next thing as that, apparently, the CFE used to report an error on call sites for a name-clashed declaration, and that's no longer the case.

So, LGTM!

sgrekhov:
These `try/catch` preserved here since the times of Dart 1.0. AFAIR  when Dart was not strongly typed, the code which produced a static errors in the analyser, could be run on VM and produced a runtime error there. So, these `try/catch` were for VM meaning "runtime error is Ok here"

eernstg:
OK, thanks!

PR #1545, opened 2022-11-14
Avoid Function.apply in utility code

askeksa-google:
The dart2wasm compiler does not support `Function.apply` and may not do so in the foreseeable future. For this reason, we should avoid using `Function.apply` outside of code that specifically tests `Function.apply`.

### DIFF
--- a/Utils/async_utils.dart
+++ b/Utils/async_utils.dart
@@ -159,8 +159,8 @@ class AsyncExpect {
         return future;
       },
        onError: (e) {
-        if (error is Function) {
-          Expect.isTrue(Function.apply(error, [e]), msg);
+        if (error is Function(Object)) {
+          Expect.isTrue(error(e), msg);
         } else {
           Expect.equals(error, e, msg);
         }


Utils/async_utils.dart
```diff
@@ -159,8 +159,8 @@ class AsyncExpect {
         return future;
       },
        onError: (e) {
-        if (error is Function) {
-          Expect.isTrue(Function.apply(error, [e]), msg);
+        if (error is Function(Object)) {
```
eernstg, line 162: Is the type test intended to ensure that `error` is a function that accepts one positional argument?

In that case it could give rise to spurious failures that `error` has the run-time type `S Function(T)` where `T` is not a supertype of `Object`.

Wouldn't it be safer to test for `Function(Never)`, and then call it dynamically (`(error as dynamic)(e)`)? This would allow the invocation to proceed as long as the run-time type of `e` is a subtype of the actual parameter type of the function, no matter which types they are.

eernstg (commented):
I think the change in the type test may be breaking.

Utils/async_utils.dart
askeksa-google: My main reason for changing the type test was to avoid doing a dynamic function call. As @sgrekhov mentions, all uses supply a function that can take `Object` (which makes sense, since this is exactly what can occur as the argument to the function passed to `onError`). Hence my suggestion to change the API of `AsyncExpect.error` to require such a function for this case.

If we want to keep the ability of `AsyncExpect.error` to take a function that doesn't accept `Object` (for whatever reason) and thus have to do the function call as a dynamic call, we could just as well keep the type test as `error is Function`.

eernstg: OK, but if it doesn't break things then obviously it's a good idea to ensure that `error(e)` isn't a dynamic invocation.

In that case the type test should test for `Function(Object)`, otherwise it will throw because of the implicitly induced type check `error(e as Never)`.

askeksa-google: Does this mean that we can land the change as is?

eernstg: Sorry, I didn't see that this thread had continued. Just LGTM'd.

askeksa-google: Ping. Can we land this?

PR #1548, opened 2022-11-17
#1401. [Patterns] Parenthesized pattern tests added

sgrekhov:


LanguageFeatures/Patterns/parenthesized_A02_t01.dart
```diff
@@ -0,0 +1,34 @@
...
+/// @description Checks parentheses in a switch expression
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+import "../../Utils/expect.dart";
+
+String test(int value) =>
+  switch(value) {
+    case ((> 0 | == 0) & < 42) => "[0, 42)";
```
eernstg, line 20: Wouldn't this be `&&` today? Similarly for `|` / `||`.

LanguageFeatures/Patterns/parenthesized_A04_t01.dart
```diff
@@ -0,0 +1,37 @@
...
+///
+/// @description Checks parentheses in a if-case statement
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+import "../../Utils/expect.dart";
+
+String test(int value) {
+    if (value case ((> 0 | == 0)  & < 42))
```
eernstg, line 19: Logical pattern operators again.

eernstg (commented):
I can see that there is a test where all initializing expressions are of the form `(e)` and a similar one using `e`. Hence, I do understand that all those parentheses aren't there by accident.

Nevertheless, I don't think we should have both versions, because they aren't testing anything which is relevant to patterns (this is all about the right hand side of `=` in declarations, not the left hand side).

So I guess the best approach is to simply drop LanguageFeatures/Patterns/parenthesized_A01_t01.dart and keep LanguageFeatures/Patterns/parenthesized_A01_t02.dart (and perhaps rename, as if t01 had never been there).

"Generous coverage is good, but in each case we should test the specific feature which is the target, not just arbitrary features".

sgrekhov:

sgrekhov:

eernstg (approved):
LGTM

PR #1551, opened 2022-11-22
#1401. [Patterns] More test cases added to variable and wildcards tests

sgrekhov:


LanguageFeatures/Patterns/variable_A01_t01.dart
```diff
@@ -24,7 +24,7 @@
 /// }
 /// ```
 /// @description Checks a variable pattern in a as a subpattern of a
```
eernstg, line 26:
```suggestion
/// @description Checks a variable pattern as a subpattern of a
```

eernstg: Occurs in variable_A03_t01.dart as well.

LanguageFeatures/Patterns/variable_A02_t01.dart
```diff
@@ -47,7 +47,26 @@ String testRecord(Record r, [bool doTypeTest = false]) {
...
+  return switch (r) {
+    case (var a, final b) => "($a, $b)";
+    default => "default";
+  };
+}
+
+String testRecord3(Record r, [bool doTypeTest = false]) {
+  if (r case (var a, final b)) {
+    if (doTypeTest) {
+      a.expectStaticType<Exactly<int>>();
```
eernstg, line 60: I don't understand this one. Isn't `Record` the type from 'dart:core' which is a supertype of all record types? In that case I don't see any elements of the code where `a` would be associated with the type `int` during static analysis.

I would expect that the fields of a receiver of type `Record` would have type `dynamic` or `Object?`, cf. https://github.com/dart-lang/language/issues/2665.

Similarly for line 61.

LanguageFeatures/Patterns/variable_A02_t01.dart
```diff
@@ -60,7 +79,26 @@ String testList(List l, [bool doTypeTest = false]) {
...
+  return switch (l) {
+    case [var a, final b] => "[$a, $b]";
+    default => "default";
+  };
+}
+
+String testList3(List l, [bool doTypeTest = false]) {
+  if (l case [var a, final b]) {
+    if (doTypeTest) {
+      a.expectStaticType<Exactly<int>>();
```
eernstg, line 92: Same issue as in LanguageFeatures/Patterns/variable_A02_t01.dart.

LanguageFeatures/Patterns/variable_A02_t01.dart
```diff
@@ -73,7 +111,28 @@ String testMap(Map m) {
...
+  return switch (m) {
+    case {1: var a} => "{1: $a}";
+    case {2: final b} => "{2: $b}";
+    default => "default";
+  };
+}
+
+String testMap3(Map m) {
+  if (m case {1: var a}) {
+    a.expectStaticType<Exactly<int>>();
```
eernstg, line 124: Same issue as in LanguageFeatures/Patterns/variable_A02_t01.dart.

LanguageFeatures/Patterns/variable_A02_t02.dart
```diff
@@ -47,20 +47,58 @@ String testRecord(Record r, [bool doTypeTest = false]) {
...
+  return switch (r) {
+    case (a, b) => "($a, $b)";
+    default => "default";
+  };
+}
+
+String testRecord3(Record r, [bool doTypeTest = false]) {
+  if (r case (a, b)) {
+    if (doTypeTest) {
+      a.expectStaticType<Exactly<int>>();
```
eernstg, line 60: Same issue as in LanguageFeatures/Patterns/variable_A02_t01.dart.

LanguageFeatures/Patterns/variable_A02_t02.dart
```diff
@@ -47,20 +47,58 @@ String testRecord(Record r, [bool doTypeTest = false]) {
...
+String testList1(List l, [bool doTypeTest = false]) {
   switch (l) {
     case [a, b]:
-      if (doTypeTest) {
-        a.expectStaticType<Exactly<int>>();
-        b.expectStaticType<Exactly<String>>();
-      }
-      return "[$a, $b]";
+    if (doTypeTest) {
+      a.expectStaticType<Exactly<int>>();
```
eernstg, line 73: Same issue as in LanguageFeatures/Patterns/variable_A02_t01.dart

LanguageFeatures/Patterns/variable_A02_t02.dart
```diff
@@ -47,20 +47,58 @@ String testRecord(Record r, [bool doTypeTest = false]) {
...
+  return switch (l) {
+    case [a, b] => "[$a, $b]";
+    default => "default";
+  };
+}
+
+String testList3(List l, [bool doTypeTest = false]) {
+  if (l case [a, b]) {
+    if (doTypeTest) {
+      a.expectStaticType<Exactly<int>>();
```
eernstg, line 92: Same issue as in LanguageFeatures/Patterns/variable_A02_t01.dart

LanguageFeatures/Patterns/variable_A02_t02.dart
```diff
@@ -73,7 +111,28 @@ String testMap(Map m) {
...
+  return switch (m) {
+    case {1: a} => "{1: $a}";
+    case {2: b} => "{2: $b}";
+    default => "default";
+  };
+}
+
+String testMap3(Map m) {
+  if (m case {1: a}) {
+    a.expectStaticType<Exactly<int>>();
```
eernstg, line 124: OK, please search for `expectStaticType`. ;-)

I think it's a good idea to test for the static type. In this case it's clearly `dynamic` (because `Map` means `Map<dynamic, dynamic>`, by instantiation-to-bound). But I don't know whether it's possible to use `expectStaticType` to detect `dynamic`.

It's testing for subtype relationships in a function type in order to get both "less equal" and "greater equal", but that won't see the difference between `dynamic` and any other top type. The real test for `dynamic` would be that `if (b) myTarget.whatever();` is not a compile-time error, where `b` is a top level variable initialized to `false` (such that we don't get in trouble with `invariant_boolean`, and the compiler doesn't delete the if statement entirely).

LanguageFeatures/Patterns/variable_A02_t05.dart
```diff
@@ -45,8 +45,37 @@ String test(Map m) {
...
+  return switch (m) {
+    case ({1: String b}) => "{1: <String>$a}";
+    case ({1: final int b}) => "{1: <int>$b}";
+    default => "default";
+  };
+}
+
+String test3(Map m) {
+  if (m case ({1: String b})) {
+    a.expectStaticType<Exactly<String>>();
```
eernstg, line 58: Ah, this one would work! I might have overlooked that some of the others would work as well, because there is a declared type directly on the variable.

eernstg (commented):
A bunch of usages of `expectStaticType` seem to expect a type that I do not think static analysis will know about.

sgrekhov:
Updated. Pease rereview

LanguageFeatures/Patterns/variable_A02_t01.dart
eernstg: Actually (and luckily ;-), `Record` is much more static than `dynamic`: It simply does not have any members other than the ones in `Object`.

This means that there is no help from a context type when it comes to the inferred declared type of `a` and `b`.

I think this area is somewhat underspecified, but let's settle for `Object?` based on [this sentence](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#:~:text=the%20required%20type%20of%20p%20is%20a%20record%20type%20with%20the%20same%20shape%20as%20p%20and%20object%3F%20for%20all%20fields.).

LanguageFeatures/Patterns/variable_A02_t01.dart
eernstg: Here I'd expect the elements to have type `dynamic`, based on `l` having the type `List<dynamic>`, plus [this sentence](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#:~:text=else%20if%20m%20implements%20list%3Ct%3E%20for%20some%20t%20then%20e%20is%20t.).

LanguageFeatures/Patterns/variable_A02_t01.dart
eernstg: `dynamic`.

LanguageFeatures/Patterns/variable_A02_t02.dart
eernstg: `dynamic` (I think the list has type `List<dynamic>`, but the github UI doesn't make it easy to check that right now).

LanguageFeatures/Patterns/variable_A02_t02.dart
eernstg: `dynamic`.

eernstg (commented):
Looked into the feature specification and the rules about inference. I think we have a conclusion, based on the step called 'type checking' of a pattern.

eernstg (approved):
LGTM

PR #1552, opened 2022-11-23
#1401. [Patterns] List pattern tests added

sgrekhov:


LanguageFeatures/Patterns/list_A01_t04.dart
```diff
@@ -0,0 +1,67 @@
...
+  a3.val = 3;
+  b3.val = 4;
+  Expect.equals(3, list[0].val);
+  Expect.equals(4, list[1].val);
+
+  Expect.throws(() {
+    var [a, b, c] = [1, 2];
+  });
+  Expect.throws(() {
+    var [a, b, 42] = [1, 2, 3];
```
eernstg, line 59: Why would this one throw?

LanguageFeatures/Patterns/list_A01_t04.dart
```diff
@@ -0,0 +1,67 @@
...
+  });
+  Expect.throws(() {
+    var [a, b, String c] = [1, 2, 3];
```
eernstg, line 62: I believe, cf. https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#type-inference, that we will obtain a pattern type schema of `List<String>` (`a` and `b` do not contribute anything, which is different from explicitly having a type like `dynamic`).

I would then expect type inference on the initializing expression to fail and give rise to a compile-time error.

But you should be able to use `c as String` (whose context type schema is `?`).

LanguageFeatures/Patterns/list_A01_t04.dart
```diff
@@ -0,0 +1,67 @@
...
+  });
+  Expect.throws(() {
+    var [a, b, c] = [1, 2, 3, 4];
+  });
```
eernstg, line 66: And rest patterns used to allow matching just some of the list elements are covered somewhere else? Otherwise it would be nice to cover the case where the match succeeds because we're ignoring some some elements, or the match throws because there are too few elements.

Similarly, an initializing expression of type `dynamic` yielding a `List<T>` where `T` isn't a subtype of the type argument of the list pattern (it's probably best to simply write the pattern type argument explicitly, because we should already have other tests where the pattern type argument is obtained by inference).

eernstg (commented):
Looks good! I've commented on a few things.

LanguageFeatures/Patterns/list_A01_t04.dart
eernstg: Ah, the eyes are the first things to go blind. ;-)

This would be a compile-time error, because it's a refutable pattern (the constant pattern `42`) in a declaration context.

eernstg (commented):
Just one more adjustment needed.

LanguageFeatures/Patterns/list_A01_t01.dart
```diff
@@ -0,0 +1,94 @@
...
+class C {
+  const C();
+}
+
+String test(List list) {
+  return switch (list) {
+    case [1, > 0 || 42] => "logical-or";
+    case [2, > 0 && < 10] => "logical-and";
+    case [3, > 0] => "relational";
+    case [4 as num, 42 as num] => "cast";
```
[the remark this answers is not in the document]
sgrekhov: Thank you. Inded. Fixed

LanguageFeatures/Patterns/list_A01_t04.dart
```diff
@@ -0,0 +1,67 @@
...
+  Expect.equals(1, a1);
+  Expect.equals(3, c1);
+  a1 = 4;
+  c1 = "6";
+  Expect.equals(1, a1);
+  Expect.equals(3, c1);
+
+  final [a2, int _, int c2] = [4, 5, 6];
+  Expect.equals(4, a2);
+  Expect.equals(6, c2);
```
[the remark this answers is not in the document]
sgrekhov: Yes, we have.  Tests `variable_A03_t*` check it

LanguageFeatures/Patterns/list_A01_t04.dart
sgrekhov: I'm working on rest pattern tests now. It turns out that we need to check a lot of patterns after `...`. With and without space, comment etc... So, it will be a separete PR

sgrekhov:
The current grammar doesn't allow `case [1, > 0 || -42]:`. Is this intended?

LanguageFeatures/Patterns/list_A01_t04.dart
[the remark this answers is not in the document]
eernstg: OK, thanks!

LanguageFeatures/Patterns/list_A01_t04.dart
eernstg: Very good!

eernstg (commented):
We still have one issue: LanguageFeatures/Patterns/list_A01_t04.dart would have a compile-time error because it uses a refutable pattern in a declaration context.

sgrekhov:
Updated. Please rereview

eernstg (approved):
LGTM

PR #1559, opened 2022-11-28
#1401. Update logical patterns syntax to && and ||

sgrekhov:


LanguageFeatures/Patterns/logical_or_A03_t01.dart
```diff
@@ -28,19 +28,19 @@ import "patterns_lib.dart";
 main() {
   Shape shape = Circle(1);
   switch (shape) {
-    case Square(area: var s1) | Circle(area: var s2):
+    case Square(area: var s1) || Circle(area: var s2):
```
eernstg, line 31: I know the placement of the "^^....^" marker is ignored for `unspecified` errors, but we might as well keep them consistent: Just a simple `^` right after `//` in order to indicate that the position is ignored, or (the more readable option), a real "^^...^" somewhere on the line which is likely to be the actual location of the error.

I guess it should work to grep the diff for "^" with a context of 1 line to find all these.

LanguageFeatures/Patterns/relational_A03_t03.dart
```diff
@@ -21,20 +21,20 @@ String test(List<num> list) {
   int i = 0;
   final j = 10;
   return switch (list) {
-    case [> i & <= 2] => "case 1";
+    case [> i && <= 2] => "case 1";
 //          ^
 // [analyzer] unspecified
 // [cfe] unspecified
     case [== i++] => "case 2";
 //           ^^^^
```
eernstg, line 29: Maybe even this one should be adjusted ("^^^" rather than "^^^^"?)

LanguageFeatures/Patterns/relational_A04_t01.dart
```diff
@@ -21,28 +21,28 @@ const o = const Object();
 
 String test(List<num> list) {
   return switch (list) {
-    case [> "1" & <= 2] => "case 1";
+    case [> "1" && <= 2] => "case 1";
 //          ^
```
eernstg, line 25: "^" --> "^^^"?

eernstg (commented):
I know, those `^^^` markers are ignored for an `unspecified` error. But it seems reasonable to maintain them, given that they seem to be written so as to indicate the expected error location. So I suggested a few changes.

sgrekhov:
Than's reasonable. Fixed. Updated spec_parser showed a couple of missing places with `|` and `&`. Again, it's very helpful to have such tool!

eernstg (approved):
LGTM, thanks!

PR #1566, opened 2022-12-01
#1401. [Patterns] Constant pattern tests update according to the new spec

sgrekhov:
- Added check for negative numbers in constant patterns
- Symbols now are allowed in  constant patterns
- Added check for constant patterns in a switch expressions

Some tests were renamed and this made diffscomplicated. Sorry, but I don't know the way to avoid it

### DIFF
--- a/LanguageFeatures/Patterns/constant_A03_t03.dart
+++ b/LanguageFeatures/Patterns/constant_A03_t03.dart
@@ -4,8 +4,9 @@
 
 /// @assertion constantPattern ::= booleanLiteral
 ///                   | nullLiteral
-///                   | numericLiteral
+///                   | '-'? numericLiteral
 ///                   | stringLiteral
+///                   | symbolLiteral
 ///                   | identifier
 ///                   | qualifiedName
 ///                   | constObjectExpression
@@ -23,8 +24,8 @@
 /// literals explicitly marked const. Likewise with set and map literals versus
 /// map patterns.
 ///
-/// @description Check list, map and set literals with type argument specified
-/// in constant patterns
+/// @description Check list, map and set literals in constant patterns. Test
+/// switch expression
 /// @author sgrekhov22@gmail.com
 
 // SharedOptions=--enable-experiment=patterns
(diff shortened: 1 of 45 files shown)


LanguageFeatures/Patterns/constant_A03_t03.dart
```diff
@@ -23,117 +24,72 @@
 /// literals explicitly marked const. Likewise with set and map literals versus
 /// map patterns.
 ///
-/// @description Check list, map and set literals with type argument specified
-/// in constant patterns
+/// @description Check list, map and set literals in constant patterns. Test
```
eernstg, line 27: I believe set literals are not supported, they'd have to use the `case const (<int>{}):` approach (which is working because `<int>{}` is a constant expression when in a constant context, but the plain `case const <int>{}:` should be an error).

LanguageFeatures/Patterns/constant_A03_t03.dart
```diff
@@ -23,117 +24,72 @@
...
-      return "<dynamic>{}";
-    default:
-      return "default";
-  }
+  return switch (value) {
+    case const {1, 2, -3} => "{1, 2, -3}";
+    case const {'1', '2', '3'} => "{'1', '2', '3'}";
+    case const {} => "{}";
+    default => "default";
+  };
```
eernstg, line 62: These guys would then have to use `()` (or the use of `const (...)` is already tested elsewhere and doesn't belong here, we just need to check that this other location testing `const (...)` does include cases with sets).

LanguageFeatures/Patterns/constant_A03_t03.dart
```diff
@@ -23,117 +24,72 @@
...
-  Expect.equals("<dynamic>{}", testSet(const {}));
-  Expect.equals("default", testSet(<double>{}));
-  Expect.equals("default", testSet(const {1}));
+  Expect.equals("{1, 2, -3}", testSet(const {1, 2, -3}));
+  Expect.equals("{'1', '2', '3'}", testSet(const {'1', '2', '3'}));
+  Expect.equals("{}", testSet(const {}));
+  Expect.equals("default", testSet({1}));
+  Expect.equals("default", testSet({1, 2, -3}));
+  Expect.equals("default", testSet({'1', '2', '3'}));
+  Expect.equals("default", testSet({}));
```
eernstg, line 94: These statements would be affected as well, if `testSet` is removed.

LanguageFeatures/Patterns/constant_A03_t03.dart
eernstg, line 27: Same issue with sets as in LanguageFeatures/Patterns/constant_A03_t01.dart.

LanguageFeatures/Patterns/constant_A03_t04.dart
```diff
@@ -24,94 +25,80 @@
 /// map patterns.
 ///
 /// @description Check list, map and set literals with type argument specified
```
eernstg, line 27: Same issue with sets as in LanguageFeatures/Patterns/constant_A03_t01.dart.

LanguageFeatures/Patterns/constant_A03_t05.dart
```diff
@@ -23,74 +24,131 @@
 /// literals explicitly marked const. Likewise with set and map literals versus
 /// map patterns.
 ///
-/// @description Check empty map and set literals in constant patterns
+/// @description Check list, map and set literals with type argument specified
```
eernstg, line 27: Same issue with sets as in LanguageFeatures/Patterns/constant_A03_t01.dart.

LanguageFeatures/Patterns/constant_A03_t06.dart
```diff
@@ -0,0 +1,120 @@
...
+/// constant's value. We don't allow all expressions here because many
+/// expression forms syntactically overlap other kinds of patterns. We avoid
+/// ambiguity while supporting terse forms of the most common constant
+/// expressions like so:
+/// ...
+/// List literals are ambiguous with list patterns, so we only allow list
+/// literals explicitly marked const. Likewise with set and map literals versus
+/// map patterns.
+///
+/// @description Check list, map and set literals with type argument specified
```
eernstg, line 27: Same issue with sets as in LanguageFeatures/Patterns/constant_A03_t01.dart.

LanguageFeatures/Patterns/constant_A03_t07.dart
```diff
@@ -0,0 +1,122 @@
...
+/// constant's value. We don't allow all expressions here because many
+/// expression forms syntactically overlap other kinds of patterns. We avoid
+/// ambiguity while supporting terse forms of the most common constant
+/// expressions like so:
+/// ...
+/// List literals are ambiguous with list patterns, so we only allow list
+/// literals explicitly marked const. Likewise with set and map literals versus
+/// map patterns.
+///
+/// @description Check empty map and set literals in constant patterns
```
eernstg, line 27: Same issue with sets as in LanguageFeatures/Patterns/constant_A03_t01.dart.

eernstg (commented):
Looks good, just one thing: The new code includes a bunch of cases involving a constant set. They are not supported using the concise syntax like `case const <int>{}:`, that syntax only covers maps and lists. So we'd need to used `case const (<int>{}):` in order to allow a constant set to be matched, and I suspect that should be covered by a more general test targeting `case const (e):`.

eernstg (approved):
LGTM

PR #1570, opened 2022-12-02
Fixes #1568. Exclude VM-specific test form web-platforms

### LINKED ISSUE #1568: co19/LanguageFeatures/Records/type_inference_A01_t03 and web number semantics
The expectation on line 54 isn't true for the web platforms: https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Records/type_inference_A01_t03.dart#L54

Web/JS semantics treats ints as subtypes of doubles, so this ends up not throwing. Could we mark this case as VM-only?

sgrekhov:


eernstg (approved):
LGTM

Issue #1575, opened 2022-12-07
_FROM_DEFERRED_LIBRARY location


scheglov:
With https://dart-review.googlesource.com/c/sdk/+/274141 we changed the location on which `XYZ_FROM_DEFERRED_LIBRARY` is reported. By doing this I broke following tests:
```
co19/Language/Expressions/Constants/static_constant_t06
co19/Language/Expressions/Constants/static_constant_t07
co19_2/Language/Expressions/Constants/static_constant_t06
co19_2/Language/Expressions/Constants/static_constant_t07
```

PR #1576, opened 2022-12-08
#1575. Update test expectations for analyzer

sgrekhov:


eernstg (approved):
LGTM.

I don't know why the analyzer and CFE treat the list as a whole differently, but it seems OK to simply expect the behavior that the tools have chosen.

PR #1578, opened 2022-12-08
#1401. [Patterns] List's rest pattern tests

sgrekhov:


LanguageFeatures/Patterns/list_A05_t06.dart
```diff
@@ -0,0 +1,81 @@
...
+import "../../Utils/expect.dart";
+
+class C {
+  static const c = 42;
+  const C();
+}
+
+String test1(List list) {
+  return switch (list) {
+    case <int>[1, 2, 3, 4, ... 42] => "case-1";
```
eernstg, line 35: OK, let's do that.

I'm working on an issue that raises several issues, including the fact that `... 42` will _definitely_ never succeed, but for now we'll just assume that it is accepted (and the match will fail at run time, but then we just continue to the next case, so it isn't going to be obvious that it's a "bad" pattern).

LanguageFeatures/Patterns/list_A05_t08.dart
```diff
@@ -0,0 +1,95 @@
...
+  }
+  return "default";
+}
+
+main() {
+  var <int>[a1, b1, ... [c1, ... r1]] = [1, 2, 3, 4, 5];
+  Expect.equals(1, a1);
+  Expect.equals(2, b1);
+  Expect.equals(3, c1);
+  Expect.listEquals([3, 4, 5], r1);
```
eernstg, line 68: Just `[4, 5]`?

eernstg (commented):
Just a couple of locations look like typos.

sgrekhov:
Thanx, fixed. Please rereview

eernstg (approved):
LGTM

PR #1580, opened 2022-12-12
#1401. [Patterns] Record pattern tests added

sgrekhov:


LanguageFeatures/Patterns/record_A02_t08.dart
```diff
@@ -0,0 +1,71 @@
...
+  Expect.equals(42, a5);
+  final (n: (x: a6, b6), (y: c6, d6)) = (n: (x: 42, -1), (y: 1, 0));
+  Expect.equals(42, a6);
+  Expect.equals(-1, b6);
+  Expect.equals(1, c6);
+  Expect.equals(0, d6);
+
+  var ((int,) (a7,),) = ((-1,),);
+  Expect.equals(-1, a7);
+  var (n: ({int x}) (x: a8)) = (n: (x: 42));
```
eernstg, line 63: Same problem here: Can't parse it.

LanguageFeatures/Patterns/record_A02_t08.dart
```diff
@@ -0,0 +1,71 @@
...
+  Expect.equals(-1, a4);
+  final (n: ((x: a5)) = (n: (x: 42));
+  Expect.equals(42, a5);
+  final (n: (x: a6, b6), (y: c6, d6)) = (n: (x: 42, -1), (y: 1, 0));
+  Expect.equals(42, a6);
+  Expect.equals(-1, b6);
+  Expect.equals(1, c6);
+  Expect.equals(0, d6);
+
+  var ((int,) (a7,),) = ((-1,),);
```
eernstg, line 61: I can't parse this one. Perhaps?:
```suggestion
  var ((int,) a7,) = ((-1,),);
```

where `(int,) a7` would be a variable pattern with a declared type.

LanguageFeatures/Patterns/record_A02_t08.dart
```diff
@@ -0,0 +1,71 @@
...
+  Expect.equals(-1, a7);
+  var (n: ({int x}) (x: a8)) = (n: (x: 42));
+  Expect.equals(42, a8);
+  var (n: (int, {int x}) (x: a9, b9), (int z, {int y}) (y: c9, d9))
```
eernstg, line 65: Same thing here.

LanguageFeatures/Patterns/record_A04_t01.dart
```diff
@@ -0,0 +1,84 @@
...
+/// pattern with the field name omitted
+///
+/// @description Checks that it is a compile-time error if a name cannot be
+/// inferred for a named field pattern with the field name omitted
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
+main() {
+  var (:n1) = (x1: 1);
```
eernstg, line 20: I think this test is impossible to write, unless the pattern can be refutable.

(Check out #2701 for a discussion about allowing refutable patterns in a declaration context).

LanguageFeatures/Patterns/record_A04_t01.dart
```diff
@@ -0,0 +1,84 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion
+/// recordPattern         ::= '(' patternFields? ')'
+/// patternFields         ::= patternField ( ',' patternField )* ','?
+/// patternField          ::= ( identifier? ':' )? pattern
+/// ...
+/// It is a compile-time error if a name cannot be inferred for a named field
+/// pattern with the field name omitted
```
eernstg, line 11: This error is concerned with the case where the name of a named field whose existence is implied by a `<patternField>` that includes `:` cannot be taken from the part after the `:`.

For instance, `: 31` uses a `<constantPattern>`, and we can't know what the name of the field should be.

I think this means that there is no way we can have that situation in an irrefutable pattern: They must have a `<variablePattern>` after `:` (possibly with a unary pattern around them), so we can _always_ see a name.

LanguageFeatures/Patterns/record_A04_t02.dart
```diff
@@ -0,0 +1,84 @@
...
+/// pattern with the field name omitted
+///
+/// @description Checks that it is a compile-time error if a name cannot be
+/// inferred for a named field pattern with the field name omitted
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
+main() {
+  var (:_) = (x1: 1);
```
eernstg, line 20: Cool! That's a counter-example to what I just said. ;-) So this test _can_ be written.

eernstg (commented):
A handful of suggestions added.

LanguageFeatures/Patterns/record_A02_t03.dart
```diff
@@ -0,0 +1,70 @@
...
+/// @description Checks record patterns with a null-assert subpattern in a
+/// declaration context
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
+import "../../Utils/expect.dart";
+
+main() {
+  var (a1!,) = (-1,);
```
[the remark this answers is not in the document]
sgrekhov: It does. Changed the test to expect a warning here

LanguageFeatures/Patterns/record_A01_t04.dart
```diff
@@ -0,0 +1,112 @@
...
+import "../../Utils/expect.dart";
+import "patterns_lib.dart";
+
+class C {
+  const C();
+}
+
+String test(Record record) {
+  return switch (record) {
+    case (n1: > 0 || -42) => "logical-or";
```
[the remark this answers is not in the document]
sgrekhov: Yes, I updated the switch expressions

LanguageFeatures/Patterns/record_A02_t08.dart
sgrekhov: Fixed

LanguageFeatures/Patterns/record_A02_t08.dart
sgrekhov: Fixed

LanguageFeatures/Patterns/record_A02_t08.dart
sgrekhov: Fixed

LanguageFeatures/Patterns/record_A04_t01.dart
sgrekhov: I added recorrd_A04_t03.dart where we check that it is a compile-time error if after `:` we have nor variable, nor null check/assert, nor cast or parethesized pattern. But, please, clarify what to do with this test (and the next one)? I updated the description to reflect what test really does, but didn't find in a spec that we really should expect compile-time error in this case. So, may bw these tests should be simply deleted

eernstg (commented):
Just one thing needs to be clarified: Is LanguageFeatures/Patterns/record_A04_t01.dart testing something which is already being tested elsewhere?

sgrekhov:
> Just one thing needs to be clarified: Is LanguageFeatures/Patterns/record_A04_t01.dart testing something which is already being tested elsewhere?

I don't think so. I removed this test from this PR but preserved it locally. There should be somewhere in the spec appropriate statement ("the initializing expression isn't assignable to this required type, hence the error") and I'll use it there

eernstg (approved):
LGTM

PR #1583, opened 2022-12-13
#1401. [Patterns] Switch expressions fixed according to the new syntax

sgrekhov:


LanguageFeatures/Patterns/constant_A03_t06.dart
```diff
@@ -34,42 +34,41 @@ import "../../Utils/expect.dart";
...
-    case const <dynamic, dynamic>{'true': true} =>
-        "<dynamic, dynamic>{'true': true}";
-    case const <String, int>{} => "<String, int>{}";
-    case const <dynamic, dynamic>{} => "<dynamic, dynamic>{}";
-    default => "default";
+    const <int, int>{1: -2} => "<int, int>{1: -2}",
+    const <dynamic, dynamic>{1: -2} => "<dynamic, dynamic>{1: -2}",
+    const <String, num>{'answer': 42} => "<String, num>{'answer': 42}",
+    const <dynamic, dynamic>{'answer': 42} =>
+      "<dynamic, dynamic>{'answer': 42}",
```
eernstg, line 53: Does the formatter handle this code at all? If it doesn't then it is probably safest to use an extra indent of 4 relative to the `=>` line (similarly to `=>` functions, and initializing expressions in declarations, and continued expressions). If it does work, and it uses an extra indent of 2 then that's the way to go, of course.

eernstg (commented):
I've asked Bob by email (cc to you) whether we can trust trailing commas to be ignored with switch expressions.

If they _do_ make a difference then I'd assume that the missing command implies things like "fit on a single line if possible", and in that case we might want to use the trailing comma everywhere except in a couple of selected cases (it's the other way around now).

As before, I've commented on a location where I suspect that the formatter will disagree (I'm assuming it doesn't run on this code, yet). We might as well stick to the standard formatting unless there is a point in not doing it (such as syntax tests, or metadata disambiguation).

LanguageFeatures/Patterns/constant_A03_t06.dart
eernstg: This should be safe, for any big switch expressions the formattor should treat the case with a trailing comma and the case without a trailing comma the same (they differ by that trailing comma, but they don't differ by the indentation level and other, larger things). So let's stop worrying. ;-) Landing.

eernstg (approved):
LGTM

PR #1596, opened 2022-12-15
Fixes #1539. Rename "views" experimental flag to "inline-class"

### LINKED ISSUE #1539: Views tests should use the 'views' experimental flag
The views tests, for instance `co19/LanguageFeatures/Views/syntax_A02_t04.dart`, use the `extension-types` flag:
```
// SharedOptions=--enable-experiment=extension-types
```
instead of the correct `views` flag:
```
// SharedOptions=--enable-experiment=views
```

sgrekhov:


eernstg (commented):
The tests seem to target the status of a word (`view`) as a built-in identifier. So that word should be replaced by `inline` in order to update the tests accordingly. However, the tests should also be inverted: `inline` is _not_ a built-in identifier (because it won't be hard to parse anyway). So we _can_ have `class inline {}`.

So we might want to do this in two steps (and this PR would then _only_ change the flag), but in that case it would be useful to keep a TODO somewhere about the need to perform this rename and expectation inversion.

eernstg:
@sgrekhov, I'm a little bit worried that if any implementation work on inline classes takes place then we will have some spurious test failures. I think it's perfectly OK to keep this PR as it is (doing the flag renaming only), but then I also think that there should be a TODO on some list targeting the issues I mentioned above (rename keyword, reverse expectation about built-in-identifier-ness). What would be the most convenient approach for you?

sgrekhov:
@eernstg, sorry for the late reply. Please don't merge it. I'll fix the tests as well as part of this PR. But first I want to finish with the Object pattern tests

eernstg:
Very good, thanks!

sgrekhov:
@eernstg I've updated inline classes tests according to the new syntax. Please rereview

LanguageFeatures/Inline-classes/syntax_A02_t01.dart
```diff
@@ -2,66 +2,35 @@
...
+import "../../Utils/expect.dart";
 
-view class View4(int id) {
-  factory View4(int id) => View4.named(id);
-//        ^^^^^
-// [analyzer] unspecified
-// [cfe] unspecified
-  View4.named(int x) : id = x;
+inline class IC {
+  final int id;
```
eernstg, line 26: Might as well have a plain `IC(this.id);`, too, given that this is likely to be the most common approach.

LanguageFeatures/Inline-classes/syntax_A02_t03.dart
```diff
@@ -2,54 +2,41 @@
...
-// [cfe] unspecified
-}
+import "../../Utils/expect.dart";
 
-view class View3([int id]) {
-//               ^
-// [analyzer] unspecified
-// [cfe] unspecified
+inline class IC {
+  final int id;
```
eernstg, line 26: This one _must_ have a plain `IC(this.id);`, otherwise the redirections have no target.

LanguageFeatures/Inline-classes/syntax_A02_t05.dart
```diff
@@ -2,52 +2,41 @@
...
 
 import "../../Utils/expect.dart";
 
-view class View(int id) {
-  View.c1(int x, {int y = 0}) : id = x + y;
-  View.c2(int x, [int y = 0]) : id = x + y;
-  View.c3([int x = 0]) : id = x;
-  View.c4({int x = 1}) : id = x;
+inline class IC {
+  final int id;
```
eernstg, line 26: Would be good to have the plain `const IC(this.id);` again here. Do we have an error test (where the constant constructor has an initializer list containing an expression which isn't potentially constant)?

LanguageFeatures/Inline-classes/syntax_A03_t03.dart
```diff
@@ -2,54 +2,58 @@
...
+  IC1(this.id);
 }
 
-view class View2 {
+inline class IC2 {
   int id;
+//^
+// [analyzer] unspecified
+// [cfe] unspecified
   final _name = "Name";
```
eernstg, line 42: It's hard to tell, there could be an error on `_name` because it's the second instance variable declaration: `id` could be considered to be an instance variable declaration because it is ;-), or it could be ignored because it has an error. It's the usual difficulty around derived errors, they aren't specified.

LanguageFeatures/Inline-classes/syntax_A03_t03.dart
```diff
@@ -2,54 +2,58 @@
...
+  IC2(int id) : this.id = id;
+}
+
+inline class IC3 {
+  int id = 42;
 //^
 // [analyzer] unspecified
 // [cfe] unspecified
-  View2(int id) : this.id = id;
+  final name = "Name";
```
eernstg, line 52: Same difficulty here.

LanguageFeatures/Inline-classes/syntax_A04_t01.dart
```diff
@@ -2,42 +2,46 @@
...
+  final int _id = 0;
 }
 
 main() {
-  View1 v1 = View1(3.14);
-  v1.id0.isFinite;
-  View2 v2 = View2(42);
-  v2._x.isOdd;
+  IC1 ic1 = IC1(3.14);
+  ic1.id0.isFinite;
```
eernstg, line 42: It's safer to use `expectStaticType<Exactly<int>>` (the type of `id0` could be `Never`).

eernstg: Same consideration below.

LanguageFeatures/Inline-classes/syntax_A04_t03.dart
```diff
@@ -2,38 +2,34 @@
...
+inline class IC1 {
+  final num id;
+  IC1(this.id);
 }
 
 main() {
-  View1 v1 = View1(1);
-  v1.id0.isOdd;
+  IC1 ic1 = ic1(1);
+  ic1.id.isOdd;
```
eernstg, line 32: It is again safer to use `expectStaticType`.

LanguageFeatures/Inline-classes/syntax_A07_t01.dart
```diff
@@ -2,62 +2,60 @@
...
-/// ```
-/// There are no special rules for static members in views. They can be declared
-/// and called or torn off as usual, e.g., View.myStaticMethod(42).
+/// There are no special rules for static members in inline classes. They can be
+/// declared and called or torn off as usual, e.g., Inline.myStaticMethod(42)
 ///
-/// @description Checks that it is a compile-time error if a view declaration
-/// has <viewPrimaryConstructor> and declares a member with the name equal to
-/// the representation name
+/// @description Checks that inline classes may have static members
```
eernstg, line 19: Perhaps add something like 'Note that static members are declared before instance members (and before the unique instance variable). This is not likely to be the style which is recommended, but it should be used in at least one test.'

eernstg (commented):
Very good! Just a couple of comments.

LanguageFeatures/Inline-classes/syntax_A02_t01.dart
sgrekhov: Done

LanguageFeatures/Inline-classes/syntax_A02_t03.dart
sgrekhov: Indeed! Thank you

LanguageFeatures/Inline-classes/syntax_A02_t05.dart
sgrekhov: Added `syntax_A02_t06.dart` testing it

LanguageFeatures/Inline-classes/syntax_A03_t03.dart
sgrekhov: Yes. Or may be an error on the inline class itself. Will update the tests if needed

sgrekhov (commented):
Please rereview

LanguageFeatures/Inline-classes/syntax_A03_t03.dart
eernstg: OK!

eernstg (approved):
LGTM

PR #1623, opened 2022-12-27
Fixes #1620. Small fixes in patterns tests

### LINKED ISSUE #1620: A test needs both 'patterns' and 'records' experiments enable
The following test has only the `patterns` feature enabled, but the body of the functions also use record literals.

https://github.com/dart-lang/co19/blob/96b727142502dfcc8d26d90579f9e87eff999dcb/LanguageFeatures/Patterns/logical_and_A04_t01.dart#L20-L25

sgrekhov:


LanguageFeatures/Patterns/variable_A02_t02.dart
```diff
@@ -83,13 +83,13 @@ String testList1(List l, [bool doTypeTest = false]) {
...
   return switch (l) {
-    [a, b] => "[$a, $b]",
+    var [a, b] => "[$a, $b]",
     _ => "default"
   };
 }
 
 String testList3(List l, [bool doTypeTest = false]) {
-  if (l case [a, b]) {
+  if (l case final [a, b]) {
```
chloestefantsova, line 92: I'm not sure that the `final` (or `var`) keyword can be applied to the outer pattern. Both the CFE and the Analyzer report compile-time errors in this case, but accept the program if the keyword immediately precedes the variable names.

The following program is rejected by the tools.

```dart
test(dynamic x) {
  if (x case final [a, b]) {
    return a + b;
  }
  return null;
}
```

The following program is accepted by the tools.

```dart
test(dynamic x) {
  if (x case [final a, final b]) {
    return a + b;
  }
  return null;
}
```

chloestefantsova (changes requested):

LanguageFeatures/Patterns/variable_A02_t02.dart
sgrekhov: @chloestefantsova yes, you are right. After rereading the spec I found that the following
```dart
var (a, [b, c]) = ("str", [1, 2]);
```
is allowed in a local variable declaration pattern only. It's not allowed after the `case` statements. I'll update the test

sgrekhov:
@chloestefantsova please rereview

chloestefantsova (approved):
LGTM. Thanks!

PR #1624, opened 2022-12-29
#1401. [Patterns] Switch statement tests

sgrekhov:


LanguageFeatures/Patterns/switch_statement_A04_t01.dart
```diff
@@ -0,0 +1,75 @@
...
+/// Constant constructors. A switch case can be a constant constructor call with
+/// implicit const, like:
+/// ```
+/// case SomeClass(1, 2):
+/// ```
+/// With this proposal, that is interpreted as an object pattern whose arguments
+/// are subpatterns.
+///
+/// @description Checks that constant constructor is not called in case of an
+/// object pattern in a switch statement
```
chloestefantsova, line 27: It seems that in either case, the const constructor and the object pattern, the test will pass. I suggest using a constant expression in the field initializers to make the field values different from the constructor arguments, for example, `this.value = value + 1`.

LanguageFeatures/Patterns/switch_statement_A04_t02.dart
```diff
@@ -0,0 +1,61 @@
...
+  final String name;
+  final int value;
+  const Number(String name, int value)
+      : this.name = name,
+        this.value = value;
+}
+
+String test(Number value) {
+  switch (value) {
+    case Number("one", 1):
```
chloestefantsova, line 44: If it's supposed to be an object pattern then that line should be reported with a compile-time errors because positional fields aren't allowed in object patterns AFAIK: https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#object-pattern

> It is a compile-time error if:
> ...
> * A patternField is of the form pattern. Positional fields aren't allowed.

chloestefantsova (changes requested):

LanguageFeatures/Patterns/switch_statement_A04_t01.dart
sgrekhov: That makes sense. Did the change

LanguageFeatures/Patterns/switch_statement_A04_t02.dart
sgrekhov: Indeed. Deleted this test

chloestefantsova (approved):
LGTM. Thanks!

PR #1632, opened 2023-01-03
Fixes #1628. [Patterns] Type argument added to fix list pattern tests

### LINKED ISSUE #1628: List literals need type arguments in some of the patterns tests
In a patterns test a check is made as follows:

https://github.com/dart-lang/co19/blob/ac4a905460672efd4006f633ea63d978ad577bf1/LanguageFeatures/Patterns/list_A01_t01.dart#L61

The testing procedure and the supposed matching line are as follows:

https://github.com/dart-lang/co19/blob/ac4a905460672efd4006f633ea63d978ad577bf1/LanguageFeatures/Patterns/list_A01_t01.dart#L34-L36

The test currently fails at runtime, and I believe it's because the invocation of the testing procedure should say `test(<num>[1, 1])` instead of `test([1, 1])`.

My reasoning is as follows:

1. Procedure `test` accepts `List<dynamic>`, which is taken as the type context in inferring the type argument for the list literal in `test([1, 1])`. The downwards context wins, and the type arguments is inferred as `dynamic`.
2. In the testing procedure the runtime type of the value is `List<dynamic>`.
3. The first check in matching against the supposed pattern `<num>[1, > 0 || -42]` is testing for the type of the object.
4. `List<dynamic>` is not a subtype of `List<num>`, and the matching fails.

@eernstg, do you agree with my analysis?

A similar failure happens in another test.

https://github.com/dart-lang/co19/blob/ac4a905460672efd4006f633ea63d978ad577bf1/LanguageFeatures/Patterns/list_A04_t05.dart#L77

sgrekhov:


eernstg (approved):
LGTM!

PR #1633, opened 2023-01-03
#1401. [Patterns] Switch expression tests

sgrekhov:


LanguageFeatures/Patterns/switch_expression_A03_t01.dart
```diff
@@ -0,0 +1,102 @@
...
+    "two" => 2,
+    _ => null
+  }?.isEven;
+
+int test5(String s) =>
+  switch (s) {
+    "one" => 1,
+    "two" => 2,
+    _ => 0
+  }++;
```
chloestefantsova, line 77: I don't think the `++` operator can be used here because the operand is not assignable. Similarly to the `--` below.

LanguageFeatures/Patterns/switch_expression_A03_t02.dart
```diff
@@ -0,0 +1,89 @@
...
+/// @description Checks a switch expressions can be used as operands of prefix
+/// operators
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+import "../../Utils/expect.dart";
+
+int test1(String s) =>
+  ++switch (s) {
```
chloestefantsova, line 41: As in another test, it seems that the application of `++` and `--` in this test should result in a compile-time error because the operant isn't assignable.

chloestefantsova (changes requested):

sgrekhov:
@chloestefantsova I changed the tests according your notes. Please rereview

chloestefantsova (approved):
LGTM. Thanks!

PR #1643, opened 2023-01-04
Fixes #1637. Missed 'const' keywords added to some patterns tests

### LINKED ISSUE #1637: COMPILE_TIME_ERROR.MISSING_OBJECT_PATTERN_GETTER_NAME in a few tests
The code is like
```dart
class A {
  final int foo;
  const A(this.foo);
}

void f(x) {
  if (x case {0: A(1)}) {}
}
```
Here `A(1)` is a pattern, and we have to say `A(foo: 1)` to make it a valid object pattern. Alternatively we can use `const A(1)`. I'm not sure what was your intention in this test though.
```
co19/LanguageFeatures/Patterns/map_A01_t01
co19/LanguageFeatures/Patterns/map_A01_t02
co19/LanguageFeatures/Patterns/map_A01_t03
```

sgrekhov:


eernstg (approved):
LGTM

PR #1648, opened 2023-01-05
Fixes #1644. Use constant object expression in relational pattern tests

### LINKED ISSUE #1644: Relational patterns require constant expressions.
It is a compile-time error if `bitwiseOrExpression` is not a valid constant expression.

```
co19/LanguageFeatures/Patterns/relational_A06_t01
co19/LanguageFeatures/Patterns/relational_A06_t02
co19/LanguageFeatures/Patterns/relational_A06_t03
co19/LanguageFeatures/Patterns/relational_A06_t04
co19/LanguageFeatures/Patterns/relational_A06_t05
co19/LanguageFeatures/Patterns/relational_A06_t06
```

sgrekhov:


eernstg (approved):
LGTM.

Note that I raised this question in https://github.com/dart-lang/language/issues/2758: Shouldn't we have a 'constant context' here, such that the explicit `const` keyword can be omitted?

If that gets accepted then we'd presumably simply create a copy of this test (and similar ones, if any) where the `const` keyword is omitted.

PR #1652, opened 2023-01-05
#1401. [Patterns] Object pattern test added

sgrekhov:
See https://github.com/dart-lang/language/issues/2745

### DIFF
--- a/LanguageFeatures/Patterns/object_A07_t05.dart
+++ b/LanguageFeatures/Patterns/object_A07_t05.dart
@@ -1,0 +1,58 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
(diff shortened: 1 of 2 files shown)


eernstg (approved):
Looks good. I considered patterns in a declaration context, and I believe we can have a test for that, too:

An `_` in a declaration context is a wildcard (not a variable pattern introducing a variable named `_`), and it is also possible to use an object pattern in a declaration context. So we should be able to do something like this:

  var SomeClass(v1: _, v2: x) = SomeClass(3, 4);

Yes, it's silly (we don't have to look up the value of `v1`, so we don't really need a mechanism that will allow us to look it up and then ignore it), but it could be used to look up the value of `v1` and perform a type cast on it in order to have a dynamic error if it doesn't have that type. So it doesn't have to be silly in all scenarios to have a wildcard in a declaration.

sgrekhov:
Added LanguageFeatures/Patterns/object_A07_t06.dart

eernstg (approved):
LGTM

PR #1668, opened 2023-01-06
Fixes #1654. Replace || and && by | and & in negative tests

### LINKED ISSUE #1654: LanguageFeatures/Patterns/constant_A06_t07
I think this code is OK and requires to errors.

```dart
    case true && true:
//       ^^^^^^^^^^^^
// [analyzer] unspecified
// [cfe] unspecified
    case false || true:
//       ^^^^^^^^^^^^^
// [analyzer] unspecified
// [cfe] unspecified
```
We switched to using `&&` and `||` are logical-and and logical-or patterns, so this might be why it is valid now.

sgrekhov:


eernstg (approved):
LGTM

PR #1738, opened 2023-01-11
Fixes #1734. Expect a static warning in case of not-necessary null-check/assert

### LINKED ISSUE #1734: Tests with STATIC_WARNING.UNNECESSARY_NULL_ASSERT_PATTERN
With https://dart-review.googlesource.com/c/sdk/+/278687 the following test started passing
```
co19/LanguageFeatures/Patterns/record_A02_t03
```
but these tests started failing
```
co19/LanguageFeatures/Patterns/switch_statement_A07_t02
co19/LanguageFeatures/Patterns/list_A05_t05
co19/LanguageFeatures/Patterns/switch_statement_A07_t01
```

There might be other tests that are currently failing with the analyzer, which also have such unnecessary patterns, but because these tests are already failing, we don't see them yet. I will report them later then.

I believe the [specification](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md) does not describe warnings for unnecessary null-assert and null-check patterns, but I think @munificent reacted positively to implementing them, so they might be added to the spec.

sgrekhov:


eernstg (approved):
LGTM

PR #1750, opened 2023-01-12
Fixes #1742. map_A04_t02.dart. Added overridden `hashCode` getter

### LINKED ISSUE #1742: LanguageFeatures/Patterns/map_A04_t02
This test uses instances of class `C` as keys in the maps. Class `C` has a user-defined `operator ==`, but it doesn't define matching `hashCode`, violating the following contract:

https://github.com/dart-lang/sdk/blob/9a7d8857ec81590f9c940f1e15c6923e3ab3453a/sdk/lib/core/object.dart#L65-L71

So, map lookup for `c1` is not guaranteed to find an element which was put into the map with the key `c2` (c1 == c2, but most likely c1.hashCode != c2.hashCode).

https://github.com/dart-lang/co19/blob/780034a9476ea8e9de6d2eecf33e6e0f02a2c220/LanguageFeatures/Patterns/map_A04_t02.dart#L44-L55

https://github.com/dart-lang/co19/blob/780034a9476ea8e9de6d2eecf33e6e0f02a2c220/LanguageFeatures/Patterns/map_A04_t02.dart#L98

https://github.com/dart-lang/co19/blob/780034a9476ea8e9de6d2eecf33e6e0f02a2c220/LanguageFeatures/Patterns/map_A04_t02.dart#L63

sgrekhov:


LanguageFeatures/Patterns/map_A04_t02.dart
```diff
@@ -52,6 +52,9 @@ class C {
     }
     return false;
   }
+
+  @override
+  int get hashCode => x;
```
alexmarkov, line 57: This doesn't look correct as objects which are equal according to `operator==` should have the same hashCode.

sgrekhov:
Indeed. Fixed. Please review

LanguageFeatures/Patterns/map_A04_t02.dart
eernstg: OK, the new `==` should fix that. But the type of `x` is `dynamic`, and this means that returning `x` is a dynamic error unless `x` turns out to be an `int`. It looks like that's true, but in that case we might as well declare `final int x;` in order to avoid having code that may easily turn buggy during otherwise benign-looking updates in the future.

eernstg (commented):
Just to avoid unnecessary bugs in the future, I suggested one change.

sgrekhov:
Thank you. Changed to `final int x;`

PR #1752, opened 2023-01-12
Fixes #1744. object_A01_t05.dart fixed not to fail in a weak mode

### LINKED ISSUE #1744: LanguageFeatures/Patterns/object_A01_t05
This test doesn't account for running without sound null safety (in weak mode):

https://github.com/dart-lang/co19/blob/780034a9476ea8e9de6d2eecf33e6e0f02a2c220/LanguageFeatures/Patterns/object_A01_t05.dart#L31-L32

From the [spec](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#matching-refuting-and-destructuring):
> If the runtime type of v is not a subtype of the cast type of p then throw a runtime error.

Without sound null safety `Null` is a bottom type (subtype of all types), so cast pattern should not throw.

sgrekhov:


eernstg (approved):
LGTM

PR #1759, opened 2023-01-13
Make record positional field getters start at `$1`.

lrhn:
This PR is ready in case we change the record positional getter names to starting at `$1`.
If we do so, this should be landed and then rolled into the CL that changes the SDK: https://dart-review.googlesource.com/c/sdk/+/278912


LanguageFeatures/Records/members_A04_t04.dart
```diff
@@ -85,9 +85,9 @@ main() {
...
-  Expect.equals(0, r.$0);
-  Expect.equals(101, r.$101);
-  Expect.equals(999, r.$999);
-  Expect.equals(1024, r.$1024);
-  Expect.equals(1025, r.$1025);
+  Expect.equals(0, r.$1);
+  Expect.equals(101, r.$102);
+  Expect.equals(999, r.$1000);
+  Expect.equals(1024, r.$1025);
+  Expect.equals(1025, r.$1026);
```
eernstg, line 92: It seems reasonable to test the treatment of large records. Perhaps there could be a bug in the handling of 4 digit `$` names, or in the treatment of `$N` where `N` is a decimal numeral whose binary representation has more than 10 bits, etc.

With that in mind we should probably cover a couple of values _around_ the potential critical values, like `$999`, `$1000`, `$1001`, `$1023`, `$1024`, `$1025`, `$1026`, and `Expect.throws` on `$1027`.

eernstg (commented):
Looks good. Added a suggestion that we cover a couple of extra cases in one test, namely the one that contains a very long record.

eernstg (commented):
(OK, github seems to think that I did not review this PR, so let's do it again ;-)

athomas (approved):

lrhn:
PTAL

lrhn:
And I should reload before adding more comments. Ignore me :)

PR #1776, opened 2023-01-16
Fixes #1764. Fix failing test

### LINKED ISSUE #1764: LanguageFeatures/Patterns/list_A04_t01
`<dynamic>[1, 2, 3, 4, 5, 6]` doesn't match the 1st pattern in `test1` as expected:

https://github.com/dart-lang/co19/blob/cd1d3fd226fdd6693d6d33986dbdb8416f7e81a2/LanguageFeatures/Patterns/list_A04_t01.dart#L80

https://github.com/dart-lang/co19/blob/cd1d3fd226fdd6693d6d33986dbdb8416f7e81a2/LanguageFeatures/Patterns/list_A04_t01.dart#L31-L34

We should either specify a type argument explicitly (`<int>[1, 2, 3, 4, 5, 6]`) or adjust corresponding pattern.

sgrekhov:
Now the both cases work
```dart
  Expect.equals("[1, 2, 3, 4, ...]", test1(<int>[1, 2, 3, 4, 5, 6]));
...
  Expect.equals("default", test1([1, 2, 3, 4, 5, 6]));
```

### DIFF
--- a/LanguageFeatures/Patterns/list_A04_t05.dart
+++ b/LanguageFeatures/Patterns/list_A04_t05.dart
@@ -26,16 +26,16 @@
 
 import "../../Utils/expect.dart";
 
-String test1(List<int> list) {
+String test1(List list) {
   return switch (list) {
-    <int>[1, 2, 3, 4, ...] => "[1, 2, 3, 4, ...]",
-    [2, ..., 5, 6] => "[2, ..., 5, 6]",
-    <int>[... /* comment */, 6, 7, 8] => "[..., 6, 7, 8]",
-    _ => "default"
-  };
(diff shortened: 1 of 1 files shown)


eernstg (commented):
This one looks funny: The 'before' indentation looks better than the 'after' indentation in the first diff block, and the removal of the type argument on the declared type of the formal parameter seems to re-introduce an issue which was removed by some other PR.

Is this an accidental reversal of some other PR?

sgrekhov:
Indentation fixed. The original issue didn't mention this test, so the change acually was not needed here.

eernstg (approved):
Of course, the list literal has no context type (in particular: it doesn't have `List` as its context type), so it's a `List<int>` already. Thanks!

LGTM

PR #1783, opened 2023-01-18
#1401. [Patterns] Pattern context tests added

sgrekhov:


sgrekhov:
@chloestefantsova please review

LanguageFeatures/Patterns/pattern_context_A05_t03.dart
```diff
@@ -0,0 +1,52 @@
...
+  (_) = 42;
+  Expect.equals(0, _);
+
+  (int _) = 42;
+  Expect.equals(0, _);
+
+  [a, _] = [1, 2];
+  Expect.equals(0, _);
+
+  [a, int _] = [1, 2];
```
eernstg, line 32: Isn't `int _` an error in an assignment context?

eernstg: Found it: https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#:~:text=It%20is%20a%20compile%2Dtime%20error%20if%20a%20variable%20pattern%20appears%20in%20an%20assignment%20context.

LanguageFeatures/Patterns/pattern_context_A05_t03.dart
```diff
@@ -0,0 +1,52 @@
...
+  (_, name: String s) = (1, name: "one");
+  Expect.equals(0, _);
+
+  (int a, name: _) = (1, name: "one");
+  Expect.equals(0, _);
+
+  Square(areaAsInt: _) = Square(2);
+  Expect.equals(0, _);
+
+  Square(areaAsInt: int _) = Square(2);
```
eernstg, line 50: Same issue as line 32.

LanguageFeatures/Patterns/wildcards_A03_t01.dart
```diff
@@ -0,0 +1,69 @@
...
+/// ```
+/// The `_` identifier can also be used with a type annotation when you want to
+/// test a value's type but not bind the value to a name:
+/// ```dart
+/// switch (record) {
+///   case (int _, String _):
+///     print('First field is int and second is String.');
+/// }
+/// ```
+/// @description Check that if an object contains getter named _ the it can be
```
eernstg, line 21:
```suggestion
/// @description Check that if an object contains getter named _ then it can be
```

eernstg (commented):
Looks good, just a couple of places that look like unintended compile-time errors and one typo.

One thing came up, though: Do we already have coverage for the kind of test where a wildcard is used several types in a way that would introduce a new variable if the name had not been `_`, and there are conflicts on the type or finality?

`case [var _] || [final _]:`

I believe this should not be an error, because there is no variable that has a type/finality conflict.

sgrekhov:
I've updated the tests. @eernstg please rereview. As for `case [var _] || [final _]:` we don't have this test right now, but I'm going to add it later. First I want to finish with the main spes, then add some corner cases tests. For the wildcards my todo list is:
- add _? and _!
- add test that _ in `when` is an error
- _ in a cast pattern tests
- `case var _ || dynamic _:`
The last one includes `case [var _] || [final _]`

eernstg:
About corner cases: Sounds good! Just one thing:

I can't find anything about the treatment of `_` in a `when` clause. I would assume that there is nothing special about the use of `_` in the expression after `when`, that is, it would be a private identifier, and it might resolve to a declaration in an enclosing scope.

(That kind of resolution would probably be linted in the future, meaning "don't do this, we want to make `_` a wildcard everywhere", but right now we can have declarations named `_` and they can be looked up as usual.)

sgrekhov:
Yes, I mean the case when there is no `_` variable in the enclosing scope

sgrekhov:
@eernstg can we merge it?

LanguageFeatures/Patterns/pattern_context_A05_t03.dart
eernstg: Resolving: This library does not exist any more, and I can't find `[... int _]` in a declaration context any more.

eernstg:
Certainly, I just hadn't noticed that a few comments had been resolved.

eernstg (approved):
LGTM

PR #1786, opened 2023-01-19
Remove extra `the`

asashour:
On `pre-nnbd` branch. Similar change was previously done on `master`.


eernstg (approved):
Cool, in each case this eliminates the the which is superfluous. ;-)
LGTM

PR #1790, opened 2023-01-23
Remove superfluous words.

asashour:
Please note the change of `operator operator`.

### DIFF
--- a/Language/Classes/Instance_Methods/Operators/syntax_t01.dart
+++ b/Language/Classes/Instance_Methods/Operators/syntax_t01.dart
@@ -4,7 +4,7 @@
 
 /// @assertion
 /// operatorSignature: 
-///   returnType? operator formalParameterList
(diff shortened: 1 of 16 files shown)


sgrekhov:
[Dart specification](https://spec.dart.dev/DartLangSpecDraft.pdf)  10.2.1. Operators

> Operators are instance methods with special names, except for operator ‘[]’ ⋄
> which is an instance getter and operator ‘[]=’ which is an instance setter.
> ⟨operatorSignature⟩ ::=
> ⟨type⟩? operator ⟨operator⟩ ⟨formalParameterList⟩

So `operator operator` change is wrong. It makes sense to add braces, like in a spec but removing the second `operator` is wrong

asashour:
I see, thanks. It is reverted. No braces were added, because the convention in other places don't use it.

sgrekhov:
LGTM

athomas (approved):

PR #1791, opened 2023-01-23
Remove superfluous words

asashour:


athomas:
@asashour I don't think there's value on fixing issues on pre-nnbd. We are likely going to discontinue co19_2 (pre-nnbd) sometime this year as everything will have moved to null safety. Consider only filing PRs against the master branch.

athomas (approved):

PR #1796, opened 2023-01-24
#1401. Pattern uses (static semantics) tests added

sgrekhov:


LanguageFeatures/Patterns/pattern_uses_A02_t02.dart
```diff
@@ -0,0 +1,41 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion
+/// The static type of a switch expression is the least upper bound of the
+/// static types of all of the case expressions.
+///
+/// @description Check that it is a compile-time error to assign a switch
+/// expression to a type which is not the least upper bound of the static types
+/// of all of it's case expressions.
```
eernstg, line 11:
```suggestion
/// of all of its case expressions.
```
("its" is a special exception, there's no apostrophe in the possessive form)

eernstg (commented):
Looks good, just noted one typo.

eernstg (approved):
LGTM

LanguageFeatures/Patterns/pattern_uses_A01_t01.dart
```diff
@@ -0,0 +1,50 @@
...
+///
+/// @description Check that it is a compile-time error if the expression in a
+/// guard clause is not assignable to bool.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+String test1(List<int> v) {
+  switch (v) {
+    case var [a, _] when a - 1:
```
eernstg, line 17: I think the spec parser is right: We can't derive a pattern of the form `var [a, _]` (it should be `[var a, _]`, assuming that we have surely covered the plain `var _` somewhere else).

eernstg (commented):
I think one construct must be changed (`var [a, _]`), just checked the feature spec.

eernstg:
Not landing, an issue came up.

LanguageFeatures/Patterns/pattern_uses_A01_t01.dart
sgrekhov: @eernstg please provide a link to that part of the spec which makes `var [a]` invalid in a matching context. I can't find it

sgrekhov:
Fixed. PTAL

LanguageFeatures/Patterns/pattern_uses_A02_t02.dart
eernstg: I think this one slipped under the carpet: The above text means "of all of it is case expressions", which is probably not intended.

LanguageFeatures/Patterns/pattern_uses_A01_t01.dart
eernstg: Oops, 2 weeks ago? Sorry about the delay. Anyway, here's an answer:

`var [a]` occurs after `case`, so it should be a `<guardedPattern>`. There's no `when`, so it should be a `<pattern>`. A `<pattern>` never starts with 'var' unless it is a `<variablePattern>`, and in that case it should have the form `'var' <identifier>`, which does not include `var [a]`.

So we can't do that.

Anyway, this change has actually been performed since then, so we probably agree on that already.

eernstg (commented):
This one is old, but let's just fix the typo before landing it.

sgrekhov:
The current description of `pattern_uses_A02_t02` is

> /// @description Check that it is a compile-time error to assign a switch
/// expression to a type which is not the least upper bound of the static types
/// of all of its case expressions.

sgrekhov:
I don't see anything else to fix. Please point me if I missed

LanguageFeatures/Patterns/pattern_uses_A02_t02.dart
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion
+/// The static type of a switch expression is the least upper bound of the
+/// static types of all of the case expressions.
+///
+/// @description Check that it is a compile-time error to assign a switch
+/// expression to a type which is not the least upper bound of the static types
```
eernstg, line 10: We should probably say something like 'to a type `T` such that the standard upper bound of the static types of all its case expressions is not assignable to `T`'.

eernstg (commented):
Looks good, just one suggestion about the description in one test.

sgrekhov:
Thank you. Description updated

PR #1798, opened 2023-01-25
#1401. Pattern variable sets tests

sgrekhov:


LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+      return "cast";
+    case final int? v3? || final int v3 when v3 == 3:
+      return "null-check";
+    case final int? v4! || final int v4 when v4 == 4:
+      return "null-assert";
+    case var v5 || var v5 when v5 == 5:
+      return "variable";
+    case (final num? v6) || final num? v6 when v6! == 6:
+      return "parenthesized";
+    case var [v7] || var v7 when v7 == 7:
```
eernstg, line 64: Error, `Object?` vs. `Object`.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    case Square(sizeAsInt: final int v10) || final int v10 when v10 == 10:
+      return "object";
+    default:
+      return "no match";
+  }
+}
+
+String test3(Object x) =>
+  switch (x) {
+    var v1 || [var v1] && var [_] when v1 == 1 => "logical-and",
```
eernstg, line 79: `var [_]` syntax error; `Object` vs. `Object?`.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    final int v2 || final v2 as int when v2 == 2 => "cast",
+    final int v3 || final int? v3? when v3 == 3 => "null-check",
+    final int v4 || final int? v4! when v4 == 4 => "null-assert",
```
eernstg, line 82: Ditto.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    var v5 || var v5 when v5 == 5 => "variable",
+    final num? v6 || (final num? v6) when v6! == 6 => "parenthesized",
+    var v7 || var [v7] when v7 == 7 => "list",
```
eernstg, line 85: Error, `Object` vs. `Object?`.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    var v8 || var {"key1": v8} when v8 == 8 => "map",
```
eernstg, line 86: Ditto.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    final int v9 || final (int v9,) when v9 == 9 => "record",
+    final v10 || Square(sizeAsInt: final v10) when v10 == 10 => "object",
```
eernstg, line 88: Error, `Object` vs. `int`.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    _ => "no match"
+  };
+
+String test4(Object x) =>
+  switch (x) {
+    [var v1] && var [_] || var v1 when v1 == 1 => "logical-and",
```
eernstg, line 94: Syntax error `var [_]`, type discrepancy `Object?` vs. `Object`.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    final v2 as int || final int v2 when v2 == 2 => "cast",
+    final int? v3? || final int v3 when v3 == 3 => "null-check",
```
eernstg, line 96: Error, `int?` vs. `int`.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    final int? v4! || final int v4 when v4 == 4 => "null-assert",
```
eernstg, line 97: Ditto.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    var v5 || var v5 when v5 == 5 => "variable",
+    (final num? v6) || final num? v6 when v6! == 6 => "parenthesized",
+    var [v7] || var v7 when v7 == 7 => "list",
```
eernstg, line 100: Error, `Object?` vs. `Object`.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    var {"key1": v8} || var v8 when v8 == 8 => "map",
```
eernstg, line 101: Ditto.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    final (int v9,) || final int v9 when v9 == 9 => "record",
+    Square(sizeAsInt: final v10) || final v10 when v10 == 10 => "object",
```
eernstg, line 103: Error, `int` vs. `Object`.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    _ => "no match"
+  };
+
+String test5(Object x) {
+  if (x case var v1 || [var v1] && var [_] when v1 == 1)
```
eernstg, line 108: Probably similar to previous function.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -0,0 +1,251 @@
...
+    return "map";
+  if (x case final int v9 || final (int v9,) when v9 == 9)
+    return "record";
+  if (x case final int v10 || Square(sizeAsInt: final int v10) when v10 == 10)
+    return "object";
+  return "no match";
+}
+
+String test6(Object x) {
+  if (x case [var v1] && var [_] || var v1 when v1 == 1)
```
eernstg, line 132: Ditto.

LanguageFeatures/Patterns/pattern_variable_sets_A02_t01.dart
```diff
@@ -0,0 +1,171 @@
...
+///
+/// @description Check that it is a compile-time error if any two sets being
+/// unioned have a variable with the same name
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
+String test1(Object x) {
+  switch (x) {
+    case [var v1] && var [v1, ...]:
```
eernstg, line 27: Syntax error `var [v1, ...]`.

LanguageFeatures/Patterns/pattern_variable_sets_A02_t01.dart
```diff
@@ -0,0 +1,171 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+      return "object";
+    default:
+      return "no match";
+  }
+}
+
+String test2(Object x) {
+  if (x case [var v1] && var [v1, ...])
```
eernstg, line 78: Probably similar to previous function.

LanguageFeatures/Patterns/pattern_variable_sets_A02_t01.dart
```diff
@@ -0,0 +1,171 @@
...
+//                                         ^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    return "object";
+  return "no match";
+}
+
+String test3(Object x) =>
+  switch (x) {
+    [var v1] && var [v1, ...] => "logical-and",
```
eernstg, line 128: Looks similar again.

LanguageFeatures/Patterns/pattern_variable_sets_A03_t01.dart
```diff
@@ -0,0 +1,88 @@
...
+
+// SharedOptions=--enable-experiment=patterns,records
+
+import "../../Utils/expect.dart";
+import "../../Utils/static_type_helper.dart";
+import "patterns_lib.dart";
+
+String test1(int x) {
+  switch (x) {
+    case var v1 when v1 == 1:
```
eernstg, line 31: `case int v1 when v1 == 0:` missing?

eernstg: Similarly for the next switches.

eernstg (commented):
I think the type discrepancy error is computed differently, and this affects a bunch of cases.

sgrekhov:
@eernst thank you for the detailed review! PTAL

LanguageFeatures/Patterns/pattern_variable_sets_A01_t01.dart
```diff
@@ -105,13 +100,8 @@ String test2(Object x) {
 // [analyzer] unspecified
 // [cfe] unspecified
     return "variable";
-  if (x case var _ || final _)
-//                    ^^^^^
-// [analyzer] unspecified
-// [cfe] unspecified
-    return "variable-2";
   if (x case final num? v6 || (var num? v6) when v6! == 6)
```
eernstg, line 103: By the way, I think this would be a syntax error, because `var T id` is never supported. I guess `case final num? v6 || (num? v6)` would do?

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -20,15 +20,15 @@
 import "../../Utils/expect.dart";
 import "patterns_lib.dart";
 
-String test1(Object x) {
+String test1(Object? x) {
```
eernstg, line 23: Ah, smart move! ;-)

eernstg (commented):
Looks good, just one single thing still came up, line 103 of LanguageFeatures/Patterns/pattern_variable_sets_A01_t01.dart.

[Ah, so I could edit it after all, then the next comment is redundant.]

eernstg (commented):
LanguageFeatures/Patterns/pattern_variable_sets_A01_t01.dart line 103 still needs to be considered, and one more case where types differ and the construct isn't expected to be an error.

sgrekhov:
It turns out that there were much more issues than line 103 in `LanguageFeatures/Patterns/pattern_variable_sets_A01_t01.dart`. Please rereview

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -51,201 +51,201 @@ String test2(Object? x) {
...
-    var v7 || var [v7] when v7 == 7 => "list",
-    var v8 || var {"key1": v8} when v8 == 8 => "map",
-    final int v9 || final (int v9,) when v9 == 9 => "record",
-    final v10 || Square(sizeAsInt: final v10) when v10 == 10 => "object",
+    int v1 || [var v1] && [_] when v1 == 1 => "logical-and",
+    final int? v2 || final int? v2? when v2 == 2 => "null-check",
+    final int? v3 || final int? v3! when v3 == 3 => "null-assert",
+    var v4 || var v4 when v4 == 4 => "variable",
+    final num? v5 || (final num? v5) when v5! == 5 => "parenthesized",
+    int v6 || [var v6] when v6 == 6 => "list",
```
eernstg, line 84: Looks like an `int`/`Object?` discrepancy again.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -51,201 +51,201 @@ String test2(Object? x) {
...
+    int v7 || {"key1": var v7} when v7 == 7 => "map",
```
eernstg, line 85: Ditto.

eernstg (commented):
I can see that I overlooked several misplaced `var`s, looks good now! Just one thing looks like an unexpected compile-time error (a type discrepancy).

sgrekhov:
@eernstg please review. It's strange but newly added `null_check_A02_t01` and `null_assert_A03_t01` fail in edge analyzer now. There are no one form the expected errors. Looks like an analyzer issue

LanguageFeatures/Patterns/null_assert_A03_t01.dart
```diff
@@ -0,0 +1,80 @@
...
+//                                      ^
+// [analyzer] STATIC_WARNING.UNNECESSARY_NULL_ASSERT_PATTERN
+// [cfe] The null-assert pattern will have no effect because the matched type isn't nullable.
+      return "match-1";
+    case int? v2 || int? v2!:
+//                         ^
+// [analyzer] STATIC_WARNING.UNNECESSARY_NULL_ASSERT_PATTERN
+// [cfe] The null-assert pattern will have no effect because the matched type isn't nullable.
+      return "match-2";
+    case int v3!:
```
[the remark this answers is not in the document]
eernstg: If this rule is supported by the language team then we'll know that `x` starts off as having type `int?` at the beginning of every case.

LanguageFeatures/Patterns/null_assert_A03_t01.dart
```diff
@@ -0,0 +1,80 @@
...
+String test2(int? x) {
+  if (x case final int? v1! || final int? v1!)
+//                                          ^
+// [analyzer] STATIC_WARNING.UNNECESSARY_NULL_ASSERT_PATTERN
+// [cfe] The null-assert pattern will have no effect because the matched type isn't nullable.
+    return "match-1";
+  if (x case int? v2 || int? v2!)
+//                             ^
+// [analyzer] STATIC_WARNING.UNNECESSARY_NULL_ASSERT_PATTERN
+// [cfe] The null-assert pattern will have no effect because the matched type isn't nullable.
```
eernstg, line 49: I wonder if this is because the inferred type of `x` for the right hand operand of `||` is `Never`? In any case, it makes sense that it can't be null, but it would also make sense to report that the right hand operand of `||` is dead code. ...

LanguageFeatures/Patterns/null_assert_A03_t01.dart
```diff
@@ -0,0 +1,80 @@
...
+    return "match-2";
+  if (x case int v3!)
```
eernstg, line 51: This is again a case where I'd expect the null check to be accepted, assuming #2801

LanguageFeatures/Patterns/null_check_A02_t01.dart
```diff
@@ -0,0 +1,66 @@
...
+
+/// @assertion nullCheckPattern ::= primaryPattern '?'
+///
+/// A null-check pattern matches if the value is not null, and then matches the
+/// inner pattern against that same value. Because of how type inference flows
+/// through patterns, this also provides a terse way to bind a variable whose
+/// type is the non-nullable base type of the nullable value being matched:
+///
+/// @description Check that it is a static warning if matched value of a
+/// null-check pattern cannot ne null
```
eernstg, line 13: Same typos.

LanguageFeatures/Patterns/null_check_A02_t01.dart
```diff
@@ -0,0 +1,66 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+String test1(int? x) {
+  switch (x) {
+    case int? v1 || int? v1?:
```
eernstg, line 20: I think the same considerations apply here.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -24,9 +24,9 @@ String test1(Object? x) {
   switch (x) {
     case int v1 || [int v1] && [_] when v1 == 1:
       return "logical-and";
-    case final int? v2 || final int? v2? when v2 == 2:
+    case final int? v2? || final int? v2? when v2 == 2:
```
eernstg, line 27: Aren't we going to get the same dead code / unnecessary `?` warning here for the second `v2?`?

eernstg: Similarly for several later cases.

eernstg (commented):
This one just keeps going. ;-)
Maybe split out the new tests about UNNECESSARY_NULL_ASSERT_PATTERN and add them in a separate PR, just so we can get the bulk of it landed?

LanguageFeatures/Patterns/null_assert_A03_t01.dart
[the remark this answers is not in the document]
sgrekhov: I hought that we have a warning here just because `int` is not nullable. Why should we here rely on the previous cases?

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
sgrekhov: Moved static warnings tests to another PR.

eernstg: OK!

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
```diff
@@ -51,201 +51,201 @@ String test2(Object? x) {
...
-    final int v2 || final v2 as int when v2 == 2 => "cast",
-    final int? v3 || final int? v3? when v3 == 3 => "null-check",
-    final int? v4 || final int? v4! when v4 == 4 => "null-assert",
-    var v5 || var v5 when v5 == 5 => "variable",
-    final num? v6 || (final num? v6) when v6! == 6 => "parenthesized",
-    var v7 || var [v7] when v7 == 7 => "list",
-    var v8 || var {"key1": v8} when v8 == 8 => "map",
-    final int v9 || final (int v9,) when v9 == 9 => "record",
-    final v10 || Square(sizeAsInt: final v10) when v10 == 10 => "object",
+    int v1 || [var v1] && [_] when v1 == 1 => "logical-and",
```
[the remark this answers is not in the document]
eernstg: I think this change has been made.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
eernstg: I think this change has been made.

LanguageFeatures/Patterns/pattern_variable_sets_A01_t02.dart
eernstg: I think this change has been made.

LanguageFeatures/Patterns/null_assert_A03_t01.dart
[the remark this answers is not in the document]
eernstg: That would be an ordering issue: The `!` pattern contains the `int v3` pattern syntactically, so I'd expect the null check to occur first, and then the `int` type test. You probably wouldn't do this in practice, but I believe the code as written should throw in case of null, rather than testing `int` and then checking that the given int value isn't null.

LanguageFeatures/Patterns/null_assert_A03_t01.dart
eernstg: This was moved to a new test library, IIUC.

eernstg (commented):
Almost there! ;-)

I think the only remaining issue is that `case int v3!:` should throw at run time in case the scrutinee is null, rather than giving rise to a warning about unnecessary checks. The point is that `!` happens first.

sgrekhov:
`case int v3!` moved to #1807

eernstg (approved):
LGTM!

PR #1800, opened 2023-01-26
#1401. [Patterns] Scope tests added

sgrekhov:


LanguageFeatures/Patterns/scope_A05_t01.dart
```diff
@@ -0,0 +1,71 @@
...
+// [cfe] unspecified
+      print(v);
+      break;
+    default:
+  }
+}
+
+void test2() {
+  var z = switch (42) {
+    final int v when (v = 1) => 0 => v,
```
eernstg, line 36: I think this construct is an error (essentially a syntax error), because of the ambiguity around `=>`. In order to make sure the error is the relevant one we'd probably have to use `()` around the function literal.

LanguageFeatures/Patterns/scope_A06_t01.dart
```diff
@@ -0,0 +1,31 @@
...
+/// ...
+/// Switch statement, switch expression, if-case statement, if-case-element:
+/// Each guardedPattern introduces a new case scope which is where the variables
+/// defined by that case's pattern are bound.
+/// ...
+/// The then statement of an if-case statement is executed in a new scope whose
+/// enclosing scope is the case's case scope.
+///
+/// @description Checks that it is a compile-time error if a variable defined in
+/// a case's scope of if-case statement is referenced in its `then` branch
```
eernstg, line 17: Probably?:
```suggestion
/// a case's scope of if-case statement is referenced in its `else` branch
```

LanguageFeatures/Patterns/scope_A06_t02.dart
```diff
@@ -0,0 +1,31 @@
...
+/// ...
+/// Switch statement, switch expression, if-case statement, if-case-element:
+/// Each guardedPattern introduces a new case scope which is where the variables
+/// defined by that case's pattern are bound.
+/// ...
+/// The then element of an if-case element is evaluated in a new scope whose
+/// enclosing scope is the case's case scope.
+///
+/// @description Checks that it is a compile-time error if a variable defined in
+/// a case's scope of if-case element is referenced in its `then` branch
```
eernstg, line 17:
```suggestion
/// a case's scope of if-case element is referenced in its `else` branch
```

eernstg (approved):
A couple of small things came up

LanguageFeatures/Patterns/scope_A02_t01.dart
```diff
@@ -0,0 +1,29 @@
...
+/// declared by the pattern in a pattern-for statement until after the pattern
+/// variable declaration's initializer
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+const c = 1;
+
+main() {
+  for (var [c] = [c] in [[]]) {
```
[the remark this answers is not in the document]
sgrekhov: Yes, indeed! And we cannot reference variable declared in a `<outerPattern` here. Have to rewrite the test expecting the `c == []` in the loop body, not `1`. Thank you.

LanguageFeatures/Patterns/scope_A02_t02.dart
```diff
@@ -0,0 +1,48 @@
...
+/// where the pattern appears:
+/// ...
+/// Pattern-for statement: Scoping follows the normal for and for-in statement
+/// scoping rules where the variable (now variables) are bound in a new scope
+/// for each loop iteration. All pattern variables are in the same scope. They
+/// are considered initialized after the for loop initializer expression.
+///
+/// @description Checks that it is a compile-time error to refer to the variable
+/// declared by the pattern in a pattern-for statement until after the pattern
+/// variable declaration's initializer
```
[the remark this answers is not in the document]
sgrekhov: Changed description text

LanguageFeatures/Patterns/scope_A05_t01.dart
sgrekhov: Yes, it's a typo. There should be `>=`, not a `=>`

LanguageFeatures/Patterns/scope_A06_t01.dart
sgrekhov: Fixed

sgrekhov (commented):
Thank you for a good review. I've updated the tests

LanguageFeatures/Patterns/scope_A05_t01.dart
```diff
@@ -33,7 +33,7 @@ void test1() {
 
 void test2() {
   var z = switch (42) {
-    final int v when (v = 1) => 0 => v,
+    final int v when (v = 1) >= 0 => v,
```
eernstg, line 36: Ah, so it was just a typo! ;-)

LanguageFeatures/Patterns/scope_A02_t01.dart
```diff
@@ -18,12 +18,15 @@
...
+
 const c = 1;
 
 main() {
-  for (var [c] = [c] in [[]]) {
-//                ^
-// [analyzer] unspecified
-// [cfe] unspecified
+  bool visited = false;
+  for (var [c] in [[]]) {
```
eernstg, line 27: I think the only entity that could play the role as the "initializer" for the iteration variable of a for-in statement is the iterable. 

However, the specification of a for-in statement uses a desugaring where there is no name clash, because the iterable is evaluated outside the while statement of the desugared code, and the iteration variable is declared inside that while statement.

In other words, we should probably have

  const c = [2];

  void main() {
    bool visited = false;
    for (var [c] in [c]) {
      Expect.listEquals([2], c);
      visited = true;
    }
    Expect.isTrue(visited);
  }

It looks funny, but the desugaring would actually put the iterable in the scope that makes `c` refer to the top-level variable (or whatever `c` resolves to when it is outside the loop itself).

I wonder if we can find anyone at all who doesn't think that's weird. ;-)

eernstg: Oh, and the description should be adjusted to say that there are no errors here.

eernstg (commented):
I think the for-in statement still needs some adjustment.

eernstg (approved):
LGTM

eernstg (commented):
Two locations look like a syntax error.

LanguageFeatures/Patterns/scope_A04_t02.dart
```diff
@@ -0,0 +1,96 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+      break;
+    default:
+  }
+}
+
+void test2() {
+  var z = switch (42) {
+    var [x] when v > 0 => 1,
```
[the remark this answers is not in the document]
sgrekhov: Thank you. Fixed

eernstg (approved):
LGTM

PR #1804, opened 2023-01-27
Fixes #1801. Improve sandbox objects logging

### LINKED ISSUE #1801: LanguageFeatures/Patterns/logical_and_A06_t01
This test matches

https://github.com/dart-lang/co19/blob/05dd8a2bbe227ab92062e2d923c391ed9ef96adf/LanguageFeatures/Patterns/logical_and_A06_t01.dart#L43

and then expects the following log:

https://github.com/dart-lang/co19/blob/05dd8a2bbe227ab92062e2d923c391ed9ef96adf/LanguageFeatures/Patterns/logical_and_A06_t01.dart#L48

However, when matching value `v` of `area` field against constant `two`, `two.==(v)` is called, according to the spec:

> Evaluate the pattern's value to c.

> The pattern matches if c == v evaluates to true. This is opposite the operand order that relational patterns use. This is deliberate to preserve compatibility with existing switch cases and continue to enable compilers to determine exactly which concrete == method is called in a constant pattern for optimization purposes.

However, `Unit.==` only logs argument of operator== and doesn't log this:

https://github.com/dart-lang/co19/blob/05dd8a2bbe227ab92062e2d923c391ed9ef96adf/LanguageFeatures/Patterns/patterns_lib.dart#L59-L63

As a result, `Circle.area=3.14;` is logged instead of expected `Circle.area=2;`.

sgrekhov:


alexmarkov (approved):

sgrekhov:
@eernstg, @chloestefantsova this PR was approved, please merge

eernstg (approved):
LGTM

PR #1816, opened 2023-02-01
Fixes #1814. Return type changed to void

### LINKED ISSUE #1814: Missing return value in co19/LanguageFeatures/Patterns/list_A01_t03
`test3` in `co19/LanguageFeatures/Patterns/list_A01_t03` doesn't return a value. The return type should probably just be `void`.

sgrekhov:


eernstg (approved):
LGTM

PR #1818, opened 2023-02-01
Fix year in header

asashour:
Another point, sometimes the year is `2020` and sometimes is it `2011-2017`, I guess it should be unified to have a single year instead of a range, as this is the majority of cases.


sgrekhov:
LGTM

eernstg (approved):
LGTM

PR #1825, opened 2023-02-03
#1401. Pattern type schema tests added

sgrekhov:


LanguageFeatures/Patterns/type_inference_A02_t01.dart
```diff
@@ -17,38 +17,9 @@
...
-  try {
-    final ([A? v1, B? v2, v3]!) = [];
-    v1.expectStaticType<Exactly<A>>();
-    v2.expectStaticType<Exactly<B>>();
-    v3.expectStaticType<Exactly<B>>();
-  } catch (_) {}
+  var (v1!) = 42 as num?;
+  v1.expectStaticType<Exactly<num>>();
+  final (v2!) = 42 as num?;
+  v2.expectStaticType<Exactly<A>>();
```
eernstg, line 24: This is tricky, as usual.

Assume `X whatever<X extends Object?>(Object? o) { ... return o as X; }` where `...` will do some logging of the value of `X` (and we avoid worrying about inference yielding `dynamic` when there's no bound, because there isn't no bound ;-).

If we are testing `var (num v1!) = whatever(1);` then the actual type argument will reflect the context type schema (it's the greatest closure), so `X` should get the value `num?`.

I think the tests in line 21 and 23 are less safe, because there's no way for the context type schema to influence the typing of the initializing expression, which means that the declared type of `v1` is decided based on the _third_ step of the static analysis: Type checking and finalizing the pattern, based on the static type of the initializing expression.

It is not even an error for type inference to fail (say, because the context type schema is very different from the static type of the expression under inference), and I believe that it would be possible for the context type schema of `(v1!)` to be `Map<Widget, _>` or basically anything else, and we would still see the same behavior.

So we really need to have a testing set up where the initializing expression supports exploration of the context type schema, rather than a test where the context type schema is basically blocked from having any effect at all.

A function like `whatever` can do that, and the fact that it takes an argument makes it possible for the invocation of `whatever` to complete normally, which means that it's OK to have it in a run-time test. (Surely we can `try`/`catch`, but this is more convenient.)

The next step is that we'd need to test context type schemas where the unknown type `_` occurs in one or more locations. That's even more tricky. So how about aiming for a set of tests where the context type schema has no occurrences of `_`, and then we can see that the actual static analysis behavior matches the specification in those cases.

eernstg: I'd expect that we have a bunch of cases of the general type 'Check that the context type schema for a ... pattern is a context type schema ... where ...'.

They will all fit nicely with this kind of test where we use `whatever` to get hold of `X` from a context type schema which is based on a declared type in a variablePattern, plus some extra stuff around the variablePattern which will change the context type schema in a specified manner.

eernstg (commented):
In this particular case I think there's a need to reorganize several tests. So I've just commented on the core idea in this review.

sgrekhov:
@eernstg thank you. Please take a look at https://github.com/dart-lang/co19/pull/1825/files#diff-6da5c466e048ab941f28ac8767eead76fc04c5820e61c2f590b7d1aebc651316 Do this and similar tests needs to be updated as well?

eernstg:
The magic link ending in 651316 took me to a library named LanguageFeatures/Patterns/type_inference_A07_t03.dart, but I'm not 100% sure 'this [and similar] tests' is a reference to that particular library? I'll assume that we're talking about 'type_inference_A07_t03'.

For line 48++,

```dart
    var [A v1, B v2, v3] = [b, c, c];
    v1.expectStaticType<Exactly<A>>();
    v2.expectStaticType<Exactly<B>>();
    v3.expectStaticType<Exactly<C>>();
```

we could use the same approach: `var [A v1, B v2, v3] = whatever([]);`, and we would then expect `X` to be `List<B>`.

We do have the same issue here: I think `[A v1, B v2, v3]` could have context type schema `String`, and it wouldn't make a difference. We want to detect that kind of situation. So I'd say yes, we do need to reconsider those test cases as well.

sgrekhov:
@eernstg it looks like you are reviewing not the whole PR but the last commit only. Please check the following setting

github

Now about type inference testing. Yes, that's tricky!

> If we are testing `var (num v1!) = whatever(1);` then the actual type argument will reflect the context type schema (it's the greatest closure), so X should get the value `num?`.

I'd expect to test `var (num? v1!) = whatever(1);` and get inferred type `num`. Is there a typo above? How we can gen `num?`?

I tested the following on the freshly-built SDK.
```dart
T getType<T extends Object?>(Object? o, [void Function(String s)? logger = null]) {
  if (logger != null) {
    logger(T.toString());
  }
  return o as T;
}

main() {
  String log = "";
  var (num? v1!) = getType(42, (String s) {log += s;});
  print(log);
}
```
This program is OK in analyzer but on `dartk` it produces an error
```
 Error: The matched value of type 'Object' isn't assignable to the required type 'num?'.
 - 'Object' is from 'dart:core'.
Try changing the required type of the pattern, or the matched value type.
  var (num? v1!) = getType(42, (String s) {log += s;});
            ^
```
Is this a valid error?

I, also have a set of tests that should check that context type schema may contain "holes" _. So, in this case, differern initializing expressions should infer different type. In particular
```dart
class A {}
class B extends A {}
class C extends B {}
class D extends C {}

main() {
  A a = A();
  B b = B();
  C c = C();
  D d = D();
  {
    var [A v1, B v2, v3] = [b, c, c];
    v1.expectStaticType<Exactly<A>>();
    v2.expectStaticType<Exactly<B>>();
    v3.expectStaticType<Exactly<C>>();
  }
  {
    var [A v1, B v2, v3] = [b, c, d];
    v1.expectStaticType<Exactly<A>>();
    v2.expectStaticType<Exactly<B>>();
    v3.expectStaticType<Exactly<D>>();
  }
...
```
Here I expect `v3` to have type `C` or `D`, depending on the initializer. Is it a valid tests? Or type of `v3` is GLB of `A` and `B`, that is `B`?

eernstg:
> I'd expect to test `var (num? v1!) = whatever(1);` and get inferred type `num`.
> Is there a typo above? How we can gen `num?`?

I would expect `num v1` to have the context type schema `num`, and then `num v1!` would have the context schema `num?` (cf. https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#:~:text=of%20the%20branches.-,Null%2Dassert,-%3A%20A%20context%20type), and so would `(num v1!)`, which means that the initializing expression would be inferred with context schema `num?`, and hence the inferred value of `X` should be `num?`. This would test the bullet labeled '**Null-assert:**'.

Considering `var (num? v1!) = whatever(1);`, the pattern `num? v1` would have the context type schema `num?`, so `num? v1!` would also have that context type schema, and so would `(num? v1!)`, and we'd still expect `X` to be `num?`.

We might well want to have both of these tests.

However, it seems more informative to check that `num` is mapped to `num?` when the rule basically says that we're adding a `?`, and less informative to check that `num?` is mapped to `num?`, because the latter test would also succeed if the rule were erroneously implemented to use the same context type schema for `p` and for `p!`.

> ...
> Is this a valid error?

I believe it isn't valid, the context type schema in the given example should still be `num?`, and `getType(...)` should be inferred as `getType<num?>(...)`, and it should compile and run with no errors.

For the tests involving `_`, I think we should settle the simpler tests where the context type schema is a type, and then we can discuss the ones with holes later on.

sgrekhov:

LanguageFeatures/Patterns/type_inference_A08_t01.dart
```diff
@@ -43,10 +45,10 @@ main() {
     v2.expectStaticType<Exactly<A>>();
     v3.expectStaticType<Exactly<A>>();
   }
+  String log = "";
   try {
-    var <String, A>{"key1": A v1, "key2": B v2, "key3": v3} = {};
-    v1.expectStaticType<Exactly<A>>();
-    v2.expectStaticType<Exactly<A>>();
-    v3.expectStaticType<Exactly<A>>();
+    var <String, A>{"key1": A v1, "key2": B v2, "key3": v3} =
```
eernstg, line 50: I think `B v2` is a compile-time error because this is in a declaration context.

LanguageFeatures/Patterns/type_inference_A08_t03.dart
```diff
@@ -51,10 +53,16 @@ main() {
...
+  String log = "";
   try {
-    final {"key1": A v1, "key2": B v2, "key3": v3} = {};
-    v1.expectStaticType<Exactly<A>>();
-    v2.expectStaticType<Exactly<B>>();
-    v3.expectStaticType<Exactly<B>>();
+    var {"key1": A v1, "key2": B v2, "key3": v3} =
+        getType({}, (String s) {log += s;});
+  } catch (_) {}
+  Expect.equals("Map<Object?, B>", log);
```
eernstg, line 61: Is that key type actually going to be `Object?`? Checking the spec: [Yes!](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#:~:text=are%20_.-,Else%20K%20is%20_,-and%20V%20is)

This is slightly surprising (it seems likely that this will give the map itself a less informative run-time type in some cases, including the one given here).

eernstg (commented):
A couple of locations look like unintended compile-time errors.

LanguageFeatures/Patterns/type_inference_A08_t03.dart
eernstg, line 61: I think `getType` is invoked here with a context type of `Map<_, B>`, so the result may well be `Map<dynamic, B>`.

eernstg (commented):
A few issues came up.

LanguageFeatures/Patterns/type_inference_A02_t01.dart
eernstg: I can see that lines 21-24 were changed. As I mentioned, we can hardly claim that we're testing the computation of the type schema of a pattern in the case where the initializing expression type inference will be able to simply ignore the given context type schema.

So what are we actually testing here?

I think we're testing the 3rd phase, type checking and finalization of the pattern. So we have an initializing expression whose type is `num?`, then we're type checking `(v1!)` with matched value type `num?`, i.e., `v1!` with `num?`, i.e., `v1` with `num`, and the result is that `v1` gets the inferred declared type `num`.

I think the `@description` should be adjusted to say that we're testing the type checking phase of patterns, or something like that.

Line 24 should be adjusted to expect type `num` as well.

LanguageFeatures/Patterns/type_inference_A07_t01.dart
```diff
@@ -0,0 +1,68 @@
...
+class C extends B {}
+class D extends C {}
+
+main() {
+  A a = A();
+  B b = B();
+  C c = C();
+  D d = D();
+  {
+    var <A>[A v1, B v2, v3] = [b, c, c];
```
eernstg, line 52: Same compile-time error as line 65.

LanguageFeatures/Patterns/type_inference_A07_t01.dart
```diff
@@ -0,0 +1,68 @@
...
+    v1.expectStaticType<Exactly<A>>();
+    v2.expectStaticType<Exactly<A>>();
+    v3.expectStaticType<Exactly<A>>();
+  }
+  {
+    final <A>[A v1, B v2, v3] = <C>[d, d, d];
```
eernstg, line 58: Same error as in line 65 again.

LanguageFeatures/Patterns/type_inference_A07_t01.dart
```diff
@@ -0,0 +1,68 @@
...
+    v1.expectStaticType<Exactly<A>>();
+    v2.expectStaticType<Exactly<A>>();
```
eernstg, line 60: Expect the type which is actually declared, again.

sgrekhov:
@eernstg indeed, the previous version of this PR tested more p.3 "Calculate the static type of the pattern" than calculating of the type schema. Now the tests are updated and, I believe, really check the type schema. Please review. I think it makes sense to review all the files, as if this PR was new, not a last commit

LanguageFeatures/Patterns/type_inference_A04_t01.dart
```diff
@@ -0,0 +1,49 @@
...
+
+main() {
+  A a = A();
+  B b = B();
+  C c = C();
+  D d = D();
+
+  B v1 = b;
+  String log = "";
+  v1 = getType(d, (String s) {log += s;});
```
eernstg, line 37: This isn't actually a pattern assignment and `v1` isn't actually an identifier pattern, it's just an old-fashioned assignment, and `v1` is an `<assignableExpression>`.

But let's just keep the test, because it tests the context type which is used in a regular assignment, and we might not have a test specifically on that.

On the other hand, a `<patternAssignment>` has an `<outerPattern>` on the left hand side of the `=`, and this means that we can't have an identifier pattern as the entire pattern, it has to be nested inside some other pattern.

However, those cases are presumably being tested in some other libraries in this PR, because they would cover all those outer patterns that we can have.

But we should adjust the `@description` such that it doesn't claim that `v1` in line 37 is an identifier pattern.

LanguageFeatures/Patterns/type_inference_A04_t02.dart
```diff
@@ -0,0 +1,41 @@
...
+
+main() {
+  A a = A();
+  B b = B();
+  C c = C();
+  D d = D();
+
+  String log = "";
+  var (v1,) = getType((a,), (String s) {log += s;});
+  Expect.equals("Object?", log);
```
eernstg, line 36: In this case I'd expect a type schema which is a record type with one positional field of type `_`, and that would become `Object?` when we take the greatest closure (or `dynamic`, as usual, please check!). That would presumably print as `(Object?)` or `(dynamic)`.

LanguageFeatures/Patterns/type_inference_A05_t01.dart
```diff
@@ -0,0 +1,36 @@
...
+  C c = C();
+  D d = D();
+
+  String log = "";
+  var (v1 as B) = getType(d, (String s) {log += s;});
+  Expect.equals("Object?", log);
+
+  log = "";
+  final (v2 as B) = getType(d, (String s) {log += s;});
+  Expect.equals("Object?", log);
```
eernstg, line 35: So how can we test that the type schema is `_` rather than `Object?`?

I'm too tired now. I'll continue tomorrow. ;-)

eernstg (commented):
Added some comments (reached LanguageFeatures/Patterns/type_inference_A05_t01.dart), then decided that I'm tired. I'll comment on the remaining libraries tomorrow.

LanguageFeatures/Patterns/type_inference_A05_t01.dart
sgrekhov: The current analyzer produces an error here. Is this an analyzer issue?
```dart
class A {}
class B extends A {}
class C extends B {}
class D extends C {}

main() {
  A a = A();
  B b = B();
  C c = C();
  D d = D();

  var (v2 as B) = d;
  v2.expectStaticType<Exactly<D>>(); // COMPILE_TIME_ERROR.TYPE_ARGUMENT_NOT_MATCHING_BOUNDS
                                     // 'D Function(D)' doesn't conform to the bound 'B Function(B)' of the type parameter 'R'.
```

eernstg: It looks correct: `v2` gets the declared type `B` because that's the matched value type which is passed on to the subpattern `p` of a cast pattern `p as B`, and there's no error at the initializing expression because the cast pattern has a required type of `Object?`.

sgrekhov: Then how we can test that the context type schema is `_`? Should, then test llooks like below?
```dart
  var (v1 as B) = b;
  v1.expectStaticType<Exactly<B>>();
  var (v2 as B) = c;
  v2.expectStaticType<Exactly<B>>();
  var (v3 as B) = d;
  v3.expectStaticType<Exactly<D>>();
  ```

**Comment by sgrekhov** [comment]
Tests updated. Please go on with the review

File: LanguageFeatures/Patterns/type_inference_A06_t01.dart [hunk 10]
```diff
@@ -0,0 +1,45 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion
+/// The context type schema for a pattern p is:
+/// ...
+/// Parenthesized: The context type schema of the inner subpattern.
+///
+/// @description Check that a context type schema of a parenthesized pattern is
+/// a context type schema of its subpattern
```
eernstg, line 11:
```suggestion
/// the context type schema of its subpattern
```

LanguageFeatures/Patterns/type_inference_A06_t01.dart
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion
+/// The context type schema for a pattern p is:
+/// ...
+/// Parenthesized: The context type schema of the inner subpattern.
+///
+/// @description Check that a context type schema of a parenthesized pattern is
```
eernstg, line 10:
```suggestion
/// @description Check that the context type schema of a parenthesized pattern is
```

LanguageFeatures/Patterns/type_inference_A09_t01.dart
```diff
@@ -0,0 +1,44 @@
...
+    var (A v1, v2, n1: B v3, n2: v4) = getType((), (String s) {log += s;});
+  } catch (_) {}
+  Expect.equals(typeOf<(A, Object?, {B n1, Object? n2})>().toString(), log);
+
+  log = "";
+  try {
+    final (B v1 as C, v2, n1: v3, n2: C v4 as D) =
+        getType((), (String s) {log += s;});
+  } catch (_) {}
+  Expect.equals(typeOf<(B, Object?, {Object? n1, C n2})>().toString(), log);
```
eernstg, line 43: Wouldn't this be `Object? n2`, because the cast pattern widens the requirements to "anything goes"?

eernstg (commented):
Added a few comments.

eernstg (approved):
LGTM! 🎉

PR #1830, opened 2023-02-03
Fixes #1828. Missed experimental flag added

### LINKED ISSUE #1828: co19/LanguageFeatures/Patterns/pattern_assignment_A04_t05 uses records
The test `LanguageFeatures/Patterns/pattern_assignment_A04_t05` uses the `records` feature through the imported `patterns_lib.dart` library but only states `patterns` in its `SharedOptions`.

sgrekhov:


sgrekhov:
@eernstg plese review

eernstg (approved):
LGTM

eernstg (approved):
LGTM.

Fun fact: Line 40 is actually a syntax error, so we don't get to check whether that one is also an error because it (sort of ;-) declares a variable.

I have to think about the consequences of mixing syntax errors and non-syntax errors, so here are a few remarks about that:

I'd like to get syntax errors recognized by the test runner such that we can run the spec parser via the test runner and get the correct interpretation of the situation (that is, the test runner shouldn't say "oh, I thought the spec parser would be happy here, but it does complain" at each syntax error).

However, thinking about it a bit, I don't think it's necessary for this to succeed that we keep syntax errors separate, it should certainly be possible to have several error expectations in one library, one of which is a syntax error expectation.

The only thing which is questionable is having _several_ syntax errors in the same library when that library isn't a multi test. In that case we can have a complex response where the number and location of syntax errors isn't in line with the expectations.

OK, done thinking, there is no problem. ;-)

PR #1832, opened 2023-02-06
#1401. Shared case scope tests

sgrekhov:


eernstg (commented):
Looks good, a few changes suggested.

LanguageFeatures/Patterns/shared_case_scope_A04_t01.dart
```diff
@@ -0,0 +1,169 @@
...
+    case Shape(sizeAsInt: var a, areaAsInt: final n):
+      return a.toString();
+    default:
+      return "default";
+  }
+}
+
+main() {
+  Expect.equals("1", testLogicalOr([1, 0]));
+  Expect.equals("2", testLogicalOr([-1, 2]));
```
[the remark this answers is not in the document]
sgrekhov: Cool! Fixed

sgrekhov (commented):
@eernstg thank you for noticing of so tiny errors! Please review once again.

LanguageFeatures/Patterns/shared_case_scope_A03_t01.dart
```diff
@@ -0,0 +1,57 @@
...
+  captured = closure;
+  return true;
+}
+
+main() {
+  String log = "";
+  switch (['before']) {
+    case [String a] when capture(() {log = a;}):
+    case [_, String a]:
+    a = 'after';
```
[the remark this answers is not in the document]
eernstg: OK!

LanguageFeatures/Patterns/shared_case_scope_A04_t01.dart
```diff
@@ -0,0 +1,169 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion In a switch statement, multiple cases may share the same body.
+/// ...
+/// At runtime, we initialize all of the shared variables in the body of the
+/// case with the values of the corresponding case variables from the matched
+/// case.
+///
+/// @description Checks that at a runtime all of the shared variables in the
```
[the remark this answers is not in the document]
eernstg: Thx!

eernstg (commented):
Just one thing remains: I don't think there's any promotion going on in the test where the description says that it's testing promotion. The description could be adjusted, or the test, or maybe I just got it wrong. ;-)

sgrekhov:
Fixed. Please review once again

LanguageFeatures/Patterns/shared_case_scope_A02_t02.dart
```diff
@@ -38,24 +38,26 @@
...
 }
 
-String test2(int x) {
+String test2(num x) {
   switch (x) {
-    case var a when a == 42:
+    case var a when a is int:
     case int a when a == 0:
-    case var a:
+    case var a when a is Never: // this breaks the promotion
```
eernstg, line 59: I'm not sure about that comment: The flow analysis of this branch would presumably mark it as stopped: There is no way to continue executing the program such that the previous step was a successful `is Never` test. This just means that it doesn't contribute to the merge of the branches, and that again means that `a` is promoted by the other branches and this one is ignored. So `a` is promoted to `int`.

That doesn't sound like 'breaks the promotion' ... ?

LanguageFeatures/Patterns/shared_case_scope_A02_t02.dart
```diff
@@ -65,9 +67,10 @@ String test2(int x) {
 
 String test3(num? x) {
   switch (x) {
-    case var a! when a == 0:
-    case num a when a == 1:
-    case num a? when a == 2:
+    case var a? when a == 0:
```
eernstg, line 70: I think flow analysis is going to keep `a` on this branch at `num?`. The reason why it won't promote `a` to `num` is that an equality test is a rather weak witness of anything (because it could in principle be an arbitrary piece of code that says 'true' when you ask whether a particular object is equal to 0, or indeed anything).

You could say that this is special because `num?` admits only null, `int`s, and `double`s, and they all have equality which is computable at compile time (except that we don't know the value of `a`, of course). So we could in principle rule out null, because `null == 0` will definitely yield false.

However, I just don't think the flow analysis will do any of these things. Here is a concrete example yielding a strong hint that the flow analysis doesn't do it:

```dart
void main() {
  int? a = 1 as dynamic;
  if (a == 0) {
    a.isEven; // Error: `a` could be null.
  }
}
```

eernstg (commented):
I think one case needs to be adjusted.

sgrekhov:
Tests updated. Please review.

I don't think that we here have tests covering [this](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#:~:text=If%20any%20of%20the%20corresponding%20variables%20in%20vs%20are%20promoted%2C%20calculate%20the%20promoted%20type%20of%20the%20variable%20in%20s%20based%20on%20all%20of%20the%20promoted%20types%20of%20n%20in%20the%20cases%20in%20the%20same%20way%20that%20promotions%20are%20merged%20at%20join%20points.) statement 

> If any of the corresponding variables in vs are promoted, calculate the promoted type of the variable in s based on all of the promoted types of n in the cases in the same way that promotions are merged at join points.

If you have any ideas how to test it, plese advise

eernstg:
About testing "merged at join points": That shouldn't be hard, all we need is to have different promotions, and then we'll get a resulting promotion which is the standard upper bound of them all.

For example, we could have an `a` of type `num?`, and then we have `when a is num` in one case and `when a is int` in another case. We would then have a promotion of `a` to `num` because that's the standard upper bound of `num` and `int`.

```dart
num? x = ...;
switch (x) {
  case var a when a is int: // `a` has declared type `num?`, promoted type `int`.
  case var a when a is num: // `a` has declared type `num?`, promoted type `num`.
    // `a` has declared type `num?`, promoted type `num`.
    a.isEven; // Error.
    a.isNegative; // OK.
    a = null; // OK, and demotes `a` to `num?`.
}
```

LanguageFeatures/Patterns/shared_case_scope_A02_t01.dart
```diff
@@ -87,6 +88,19 @@ String test4(int x) {
   }
 }
 
+String test5(num? x) {
+  switch (x) {
+    case num? a? when a is Never:
+    case num? a when a != null:
+    case num? a! when a == 1:
+      a.expectStaticType<Exactly<num?>>();
```
eernstg, line 96: Wouldn't this be `num`? The first branch is a dead end, so that's ignored. The second branch promotes `a` from `num?` to `num` because we're in the true continuation of `a != null`. The third branch _also_ promotes `a` to `num` because it did not throw at `!`.

eernstg: Can this be tested at this point? It would be good to know that `a != null` actually promotes `a` in a `when` clause, but it _does_ promote in some simpler cases:

```dart
void main() {
  num? a = 1 as dynamic;
  if (a != null) a.isNegative; // OK.
}
```

eernstg (commented):
Looks good, just one location seems to need a different expected type.

LanguageFeatures/Patterns/shared_case_scope_A02_t01.dart
sgrekhov: @eernstg  I've updated [shared_case_scope_A02_t01.dart](https://github.com/dart-lang/co19/pull/1832/files/ea871d40c740e654e874ea643394786acbb6b16b..ef51d98032d4266c130fededc75c2406c9607e5e#diff-ab7a3c9dc83dc25bd1f27978bdd65d5f656edcd15132e6b307bd929eec96faef) to expect `num` instead of `num?`, and added "merged at join points" test. Both tests (updated and the newly added) fail now in analyzer

Testing that `...  when a != null` actually promotes `a` I'll add later together with other type promotion tests in another PR

eernstg: Sounds good!

eernstg (approved):
LGTM!

eernstg:
I'm still a bit confused by the comment 'this breaks the promotion', but I'm landing this now because that's a tiny thing.

sgrekhov:
Thank you! There's no 'this breaks the promotion' comment in this PR. Probably it was an outdated diff

eernstg:
Ah! When I click on the link [LanguageFeatures/Patterns/shared_case_scope_A02_t02.dart](https://github.com/dart-lang/co19/pull/1832/files/fd621e604191c35be5ea34ecb7c312091866219e..ea871d40c740e654e874ea643394786acbb6b16b#diff-de1826515b7bca6346a676ade8839ae4ab98a0ecfa636fe92da4dd71939ac3d9) in the comment above, right below 'I think one case needs to be adjusted.', github shows me a library that contains this comment.

However, that is indeed showing a single commit, and the text is gone today.

I keep clicking on those comment boxes on the 'Conversation' tab, because the 'Files changed' tab, when enabling 'All commits', constantly hides a lot of comment threads that weren't resolved, and I don't know any other way to find those old threads.

Anyway, this clarifies what happened. Thanks! Nothing is broken, except github. ;-)

PR #1834, opened 2023-02-07
Fixes #1833. Fix null-assert pattern tests to not produce false warnings

### LINKED ISSUE #1833: Patterns tests have redundant `!`s
There are new test failures on [Flow analysis: add support for constant patterns.](https://dart-review.googlesource.com/c/sdk/+/280861).

The tests
```
co19/LanguageFeatures/Patterns/switch_expression_A02_t01 StaticWarning (expected Pass)
co19/LanguageFeatures/Patterns/switch_statement_A06_t01 StaticWarning (expected Pass)
co19/LanguageFeatures/Patterns/switch_statement_A06_t03 StaticWarning (expected Pass)
```
are failing on configuration
```
analyzer-asserts-linux
```

Failure logs are here: https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8790057905485670161/+/u/test_results/new_test_failures__logs_

These tests are failing because they contain redundant null checks.  For example, in `co19/LanguageFeatures/Patterns/switch_statement_A06_t01`, the function `testNullAssert` looks like this:
```dart
String testNullAssert(int? value) {
  switch (value) {
    case var a1! when a1 > 0:
      return "null-assert-1";
    case var a2!:
      return "null-assert-2";
    default:
      return "default";
  }
}
```
If `null` is passed to this function, the null-assert pattern (`!`) in the first switch case will cause an execption to be thrown.  Therefore the null-assert pattern (`!`) in the second case is unnecessary.

`co19/LanguageFeatures/Patterns/switch_expression_A02_t01` and `co19/LanguageFeatures/Patterns/switch_statement_A06_t03` also have functions called `testNullAssert` which fail for similar reasons.

sgrekhov:


eernstg (approved):
OK, I think we'll just take https://github.com/dart-lang/language/issues/2801#issuecomment-1419397350 as the specification of the intended behavior when it comes to these warnings.

LGTM.

PR #1841, opened 2023-02-08
Fixes #1836. Don't use _ in a guard clause

### LINKED ISSUE #1836: co19/LanguageFeatures/Patterns/null_assert_A03_t02
```dart
String test3(int? value) {
  switch (value) {
    case var _! when _ == 0:
      return "null-assert-1";
    case var _! when _ == 42:
//            ^
// [analyzer] STATIC_WARNING.UNNECESSARY_NULL_ASSERT_PATTERN
// [cfe] The null-assert pattern will have no effect because the matched type isn't nullable.
      return "null-assert-2";
    default:
      return "default";
  }
}
```

We cannot use `_ == 0`, because `var _` does not define a variable.

sgrekhov:


eernstg (approved):
LGTM

PR #1850, opened 2023-02-12
Fixes #1846. Update tests to avoid unintended warnings

### LINKED ISSUE #1846: Some switch tests will start reporting warnings soon
There are new test failures on [Flow analysis: fix reachability for cast patterns.](https://dart-review.googlesource.com/c/sdk/+/281740).

The tests
```
co19/LanguageFeatures/Patterns/switch_statement_A07_t01 StaticWarning (expected Pass)
co19/LanguageFeatures/Patterns/switch_statement_A07_t02 StaticWarning (expected Pass)
```
are failing on configurations
```
analyzer-asserts-linux
```

Here's an excerpt from `switch_statement_A07_t01` that illustrates the problem:
```dart
String test(Object? o) {
  String toReturn = "";
  switch (o) {
    ...
    case var c as Object when c == 11: // (1)
      toReturn = "cast";
    case var a? when a == 12: // (2)
      toReturn = "null-check";
    case final b! when b == 13: // (3)
      toReturn = "null-assert";
    ...
  }
  return toReturn;
}
```
The analyzer is reporting the warning `UNNECESSARY_NULL_CHECK_PATTERN` on the line marked (2) and `UNNECESSARY_NULL_ASSERT_PATTERN` on the line marked (3).  This is correct--the null-check and null-assert patterns will have no effect.  This is because when a cast pattern fails to match, an exception is thrown; control doesn't just proceed to the case that follows.  So regardless of whether the match at (1) succeeds, after that line, the matched value is known to have type `Object` (i.e. it's not null).

The easiest way to avoid the analyzer warnings would probably be to move the "cast" case after the "null-check" and "null-assert" cases.

sgrekhov:


eernstg (approved):
LGTM

PR #1855, opened 2023-02-14
#1401. Type checking and pattern required type tests

sgrekhov:
It turns out that this PR is big. I'll try not to make such big PR's in the future. Sorry


eernstg:
Sounds good! ;-) I'll take a look tomorrow.

LanguageFeatures/Patterns/type_inference_A29_t15.dart
```diff
@@ -0,0 +1,82 @@
...
+///     is K.
+///   c. Else if M is dynamic then K and V are dynamic and C is _.
+///   d. Else K and V are Object? and C is _.
+/// ii. Type-check each key expression using C as the context type.
+/// iii. Type-check each value subpattern using V as the matched value type.
+/// vi. The required type of p is Map<K, V>.
+///
+/// @description Check that each value subpattern is type checked using V as the
+/// matched value type. Test the refutable context and the case when p has no
+/// type arguments <K, V> and M is dynamic
```
eernstg, line 23: This one should be easy: Use a `var a` pattern as the value and test that `a` has declared type `dynamic`.

(But do we have that test already?)

LanguageFeatures/Patterns/type_inference_A29_t15.dart
```diff
@@ -0,0 +1,82 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+import "../../Utils/expect.dart";
+import "../../Utils/static_type_helper.dart";
+
+String test1() {
+  switch ({"key1": 1 as num} as dynamic) {
+    case {"key1": int  a}:
```
eernstg, line 33: The declared type `int` isn't helpful: It just forces `a` to have the declared type `int`, and then we can't see that the type that we're using as the 'matched value type' is `dynamic`.

We won't get an error, even in the case where the matched value type is `String` and the declared type of the pattern variable is `bool`, so I believe that we're learning exactly nothing when `a` has a declared type.

LanguageFeatures/Patterns/type_inference_A29_t16.dart
```diff
@@ -0,0 +1,43 @@
...
+///   c. Else if M is dynamic then K and V are dynamic and C is _.
+///   d. Else K and V are Object? and C is _.
+/// ii. Type-check each key expression using C as the context type.
+/// iii. Type-check each value subpattern using V as the matched value type.
+/// vi. The required type of p is Map<K, V>.
+///
+/// @description Check that each value subpattern is type checked using C as the
+/// context type. The case when p has no type arguments and M is dynamic. Test
+/// that in irrefutable context it is a run-time error if value subpattern fails
+/// a type check.
```
eernstg, line 24: I don't understand how it could be relevant to check for a run-time error when we're testing how a pattern is type checked and finalized.

LanguageFeatures/Patterns/type_inference_A29_t17.dart
```diff
@@ -0,0 +1,94 @@
...
+///     is K.
+///   b. Else if M implements Map<K, V> for some K and V then use those, and C
+///     is K.
+///   c. Else if M is dynamic then K and V are dynamic and C is _.
+///   d. Else K and V are Object? and C is _.
+/// ii. Type-check each key expression using C as the context type.
+/// iii. Type-check each value subpattern using V as the matched value type.
+/// vi. The required type of p is Map<K, V>.
+///
+/// @description Check that if p has no type arguments and M sdoesn't implement
```
eernstg, line 21:
```suggestion
/// @description Check that if `p` has no type arguments and `M` doesn't implement
```

LanguageFeatures/Patterns/type_inference_A29_t17.dart
```diff
@@ -0,0 +1,94 @@
...
+/// Map<K, V> and is not dynamic then K and V are Object? and C is _.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+import "../../Utils/static_type_helper.dart";
+import "../../Utils/expect.dart";
+
+String test1() {
+  switch ({}) {
```
eernstg, line 31: Needs to be `{} as Object` or `{} as Object?`, I think. The static type of `{}` will _always_ implement `Map<K, V>` for some `K` and `V`.

eernstg: I'm afraid the key context is untestable, but a `var a` as a value should allow us to test that `a` gets the type `Object`.

LanguageFeatures/Patterns/type_inference_A30_t01.dart
```diff
@@ -0,0 +1,63 @@
...
+/// if type test fails in refutable context
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
+import "../../Utils/expect.dart";
+
+String test1() {
+  switch ((3.14, name: "pi")) {
+    case (int v1, name: String n1):
```
eernstg, line 33: An obvious way to detect that a field is type checked with a specific context type (not schema, that's harder), is to use a plain `var v` and then check the declared type of `v`.

I think you could use `(String v1, name: bool n1)` and you wouldn't see a difference, because it's OK to test for types that you can't expect based on the static type of the matched value.

LanguageFeatures/Patterns/type_inference_A30_t02.dart
```diff
@@ -0,0 +1,39 @@
...
+///
+/// @description Check that each field of the record pattern is type checked.
+/// Test the case when M is a record type with the same shape as p. It is a
+/// compile-time error if type test fails in irrefutable context
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
+main() {
+  var (int v1, name: String n1) = (3.14, name: "pi");
```
eernstg, line 30: Yes! For once, the type check of `int v1` will not be able to ignore the matched value type.

LanguageFeatures/Patterns/type_inference_A30_t03.dart
```diff
@@ -0,0 +1,69 @@
...
+///     where the field is present, so it's safe to just assume the field exists
+///     when type checking here.
+///   d. Type-check s using F as the matched value type.
+///
+/// ii. The required type of p is a record type with the same shape as p and
+///   Object? for all fields.
+///
+/// @description Check that each field of the record pattern is type checked.
+/// Test the case when M is dynamic. Match fails if runtime type test fails in
+/// refutable context
```
eernstg, line 24: We can test that the type check has context `dynamic` because it will give a plain variable the declared type `dynamic`.

The run-time semantics should not be relevant here.

LanguageFeatures/Patterns/type_inference_A30_t04.dart
```diff
@@ -0,0 +1,46 @@
...
+///     where the field is present, so it's safe to just assume the field exists
+///     when type checking here.
+///   d. Type-check s using F as the matched value type.
+///
+/// ii. The required type of p is a record type with the same shape as p and
+///   Object? for all fields.
+///
+/// @description Check that each field of the record pattern is type checked.
+/// Test the case when M is dynamic. It is a run-time error if type test fails
+/// in irrefutable context
```
eernstg, line 24: Same as previous.

LanguageFeatures/Patterns/type_inference_A31_t02.dart
```diff
@@ -0,0 +1,38 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
+import "patterns_lib.dart";
+import "../../Utils/expect.dart";
+
+main() {
+  var Square(type: x) = Square<Centimeter>(2);
+  Expect.equals("Centimeter", x);
```
eernstg, line 37: This one would surely work with a `Type get type => T` and testing for `Centimeter`.

LanguageFeatures/Patterns/type_inference_A31_t03.dart
```diff
@@ -0,0 +1,67 @@
...
+///     tear-off a method. When X is dynamic or Never then X has all properties
+///     and their types are likewise dynamic or Never unless the property is
+///     defined on Object, in which case it has its usual type.
+///
+///   b. Type check f using G as the matched value type to find its required
+///     type.
+/// iii. The required type of p is X.
+///
+/// @description Check each field subpattern of p is type checked. Test that
+/// match fails if type test fails in refutable context
```
eernstg, line 27: 'match fails': Sounds like the dynamic semantics again?

LanguageFeatures/Patterns/type_inference_A31_t04.dart
```diff
@@ -0,0 +1,43 @@
...
+///     tear-off a method. When X is dynamic or Never then X has all properties
+///     and their types are likewise dynamic or Never unless the property is
+///     defined on Object, in which case it has its usual type.
+///
+///   b. Type check f using G as the matched value type to find its required
+///     type.
+/// iii. The required type of p is X.
+///
+/// @description Check each field subpattern of p is type checked. Test that
+/// it is a compile-time error if type test fails in irrefutable context
```
eernstg, line 27: I don't understand this. A type test is usually `x is T`, and this is about type checking.

So, probably: `int v1` is type checked with matched value type `T` (what is the type of `Square.area` by the way?), and `T` is not assignable to `int`, and there is no coercion which will turn the `T` into an `S` such that `S` is assignable to `int`.

That should be testable, too!

eernstg (commented):
_A28_ is cool!

I've suggested adding a bunch of back quotes. The main issue is that HTML-ish text tends to disappear (so `List<int>` is shown as `List`, somewhere, sometimes). So it's probably
a good habit to have them whenever we're talking directly about code.

OK, this one was very long, and there are lots of things to consider. Lots of comments about that, take a look. ;-)

LanguageFeatures/Patterns/type_inference_A29_t08.dart
```diff
@@ -0,0 +1,82 @@
...
+///     is K.
+///   c. Else if M is dynamic then K and V are dynamic and C is _.
+///   d. Else K and V are Object? and C is _.
+/// ii. Type-check each key expression using C as the context type.
+/// iii. Type-check each value subpattern using V as the matched value type.
+/// vi. The required type of p is Map<K, V>.
+///
+/// @description Check that each key expression is type checked using C as the
+/// context type. Test the refutable context and the case when p has no type
+/// arguments and M implements Map<K, V>
```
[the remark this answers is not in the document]
sgrekhov: Ok, I'll delete it but first I want the updated version of the previous test to work

LanguageFeatures/Patterns/type_inference_A29_t09.dart
```diff
@@ -0,0 +1,38 @@
...
+///   c. Else if M is dynamic then K and V are dynamic and C is _.
+///   d. Else K and V are Object? and C is _.
+/// ii. Type-check each key expression using C as the context type.
+/// iii. Type-check each value subpattern using V as the matched value type.
+/// vi. The required type of p is Map<K, V>.
+///
+/// @description Check that each key expression is type checked using C as the
+/// context type. Test that in irrefutable context it is a compile-time error if
+/// key expression fails a type check. The case when p has no type arguments and
+/// M implements Map<K, V>
```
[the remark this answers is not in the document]
sgrekhov: I'll delete it

LanguageFeatures/Patterns/type_inference_A29_t10.dart
```diff
@@ -0,0 +1,82 @@
...
+///     is K.
+///   c. Else if M is dynamic then K and V are dynamic and C is _.
+///   d. Else K and V are Object? and C is _.
+/// ii. Type-check each key expression using C as the context type.
+/// iii. Type-check each value subpattern using V as the matched value type.
+/// vi. The required type of p is Map<K, V>.
+///
+/// @description Check that each value subpattern is type checked using V as the
+/// matched value type. Test the refutable context and the case when p has no
+/// type arguments <K, V> but M implements Map<K, V>
```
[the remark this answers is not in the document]
sgrekhov: It makes sense. I'll delete it. Just waiting for a couple of the previous test to work

LanguageFeatures/Patterns/type_inference_A29_t11.dart
```diff
@@ -0,0 +1,38 @@
...
+/// context type. Test that in irrefutable context it is a compile-time error if
+/// value subpattern fails a type check. The case when p has no type arguments
+/// and M implements Map<K, V>
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+main() {
+  var {"key1": int a1} = <String, num>{"key1": 1};
+//                       ^
```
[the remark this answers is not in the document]
sgrekhov: As the previous ones

LanguageFeatures/Patterns/type_inference_A29_t13.dart
```diff
@@ -0,0 +1,82 @@
...
+///     is K.
+///   c. Else if M is dynamic then K and V are dynamic and C is _.
+///   d. Else K and V are Object? and C is _.
+/// ii. Type-check each key expression using C as the context type.
+/// iii. Type-check each value subpattern using V as the matched value type.
+/// vi. The required type of p is Map<K, V>.
+///
+/// @description Check that each key expression is type checked using C as the
+/// context type. Test the refutable context and the case when p has no type
+/// arguments nad M is dynamic
```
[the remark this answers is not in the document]
sgrekhov: Let's delete it then?

LanguageFeatures/Patterns/type_inference_A29_t14.dart
```diff
@@ -0,0 +1,43 @@
...
+///   c. Else if M is dynamic then K and V are dynamic and C is _.
+///   d. Else K and V are Object? and C is _.
+/// ii. Type-check each key expression using C as the context type.
+/// iii. Type-check each value subpattern using V as the matched value type.
+/// vi. The required type of p is Map<K, V>.
+///
+/// @description Check that each key expression is type checked using C as the
+/// context type. The case when p has type no arguments and M is dynamic. Test
+/// that in irrefutable context it is a run-time error if key expression fails a
+/// type check.
```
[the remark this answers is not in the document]
sgrekhov: Let's delete it then?

LanguageFeatures/Patterns/type_inference_A29_t15.dart
sgrekhov: Updated

LanguageFeatures/Patterns/type_inference_A29_t16.dart
sgrekhov: Updated

LanguageFeatures/Patterns/type_inference_A29_t17.dart
sgrekhov: Fixed

sgrekhov (commented):
Thabk you for review. Guess it was hard :) Tests updated but there are a couple of questions regarding checking of map pattern key type. After these questins are resolved I'll delete excessive tests

LanguageFeatures/Patterns/patterns_lib.dart
```diff
@@ -115,6 +115,7 @@ class Shape<T extends MetricUnits> {
   List<num> get areaAsList => [areaAsDouble];
   Map<String, num> get areaAsMap => {"area": areaAsDouble};
   ({num area}) get areaAsRecord => (area: areaAsDouble);
+  String get type => T.toString();
```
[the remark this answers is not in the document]
eernstg: Sounds good!

LanguageFeatures/Patterns/type_inference_A20_t03.dart
```diff
@@ -0,0 +1,40 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion
+/// To type check a pattern p being matched against a value of type M:
+///
+/// Logical-or and logical-and: Type check each branch using M as the matched
+/// value type.
+///
+/// @description Check that it is a compile-time error if any branch of a
+/// logical-and pattern fails a type check in irrefutable context
```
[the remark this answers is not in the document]
eernstg: There are about 6 occurrences of 'in irrefutable', I think it will read better as 'in an irrefutable'

LanguageFeatures/Patterns/type_inference_A21_t01.dart
```diff
@@ -0,0 +1,110 @@
...
+/// @description Check that it is a compile-time error if the operator is a
+/// comparison (<, <=, >, or >=) and M does not define that operator
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+String test1() {
+  String s = "";
+  switch(s) {
+    case < 0:
```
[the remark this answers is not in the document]
eernstg: Sounds good!

LanguageFeatures/Patterns/type_inference_A24_t03.dart
```diff
@@ -0,0 +1,85 @@
...
+/// ...
+/// Variable:
+///
+/// If the variable has a type annotation, the required type of p is that type,
+/// as is the static type of the variable introduced by p.
+///
+/// Else the required type of p is M, as is the static type of the variable
+/// introduced by p
+///
+/// @description Check that if a variable pattern has no type annotation, then
```
eernstg, line 16:
```suggestion
/// @description Check that if a variable pattern `p` has no type annotation, then
```

We know that `p` must be that variable pattern, but it's better style to introduce every name which is used.

LanguageFeatures/Patterns/type_inference_A25_t01.dart
```diff
@@ -0,0 +1,77 @@
...
+/// If the variable has a type annotation, the required type of p is that type,
+/// as is the static type of the variable introduced by p.
+///
+/// Else the required type of p is M, as is the static type of the variable
+/// introduced by p
+///
+/// @description Check that if a variable pattern `p` has a type annotation,
+/// then the required type of `p` is that type. Test that there is no match in
+/// a refutable context if the matched value type is not assignable to the type
+/// of the variable pattern
```
eernstg, line 19: OK. This is a property of the dynamic semantics ("there is a match / there is no match" is a question that we can only answer at run time). That seems to be out of topic here, cf. the title 'Type checking and pattern required type tests'. However, the situation where the matched value type is not assignable to the required type of the variable pattern _is_ relevant to compile time: That's not an error in a refutable context, but it is an error in an irrefutable context.

Perhaps it's just a typo in line 17: 'Test that there is no compile-time error in ...'?

LanguageFeatures/Patterns/type_inference_A25_t01.dart
```diff
@@ -0,0 +1,77 @@
...
+///
+/// If the variable has a type annotation, the required type of p is that type,
+/// as is the static type of the variable introduced by p.
+///
+/// Else the required type of p is M, as is the static type of the variable
+/// introduced by p
+///
+/// @description Check that if a variable pattern has a type annotation, then
+/// the required type of `p` is that type. Test that there is no match in
+/// refutable context if matched value type is not assignable to the type of the
```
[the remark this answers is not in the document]
eernstg: OK, this explanation fits perfectly with the comment I just wrote on line 19 of LanguageFeatures/Patterns/type_inference_A25_t01.dart. So I'll resolve this one.

LanguageFeatures/Patterns/type_inference_A26_t01.dart
```diff
@@ -0,0 +1,41 @@
...
+/// @description Check that in an assignment context, the required type of p is
+/// the unpromoted static type of the variable that p resolves to
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
+import "../../Utils/static_type_helper.dart";
+
+main() {
+  num a = 0;
```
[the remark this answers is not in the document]
eernstg: Very good! The analyzer should fix that.

The resulting type should be `num`: `double` is not a supertype of the promoted type of `a` at the beginning of the pattern assignment line (and it was never a type of interest for `a`), so we can't demote it to `double`.

So the static type `num` must be the correct expectation.

LanguageFeatures/Patterns/type_inference_A26_t01.dart
```diff
@@ -0,0 +1,41 @@
...
+  a = 42;
```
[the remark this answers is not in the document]
eernstg: Thx!

LanguageFeatures/Patterns/type_inference_A26_t02.dart
```diff
@@ -0,0 +1,67 @@
...
+/// type of the variable that p resolves to.
+///
+/// In a matching context, the name refers to a constant. Type check the
+/// constant identifier expression in context type M.
+///
+/// In a declaration context, the required type of p is M, as is the static type
+/// of the variable introduced by p.
+///
+/// @description Check that in a matching context identifier pattern is type
+/// checked in context type M
```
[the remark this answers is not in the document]
eernstg: Looks good!

LanguageFeatures/Patterns/type_inference_A26_t03.dart
```diff
@@ -0,0 +1,36 @@
...
+/// @description Check that in a declaration context, the required type of p is
+/// M, as is the static type of the variable introduced by p.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
+import "../../Utils/static_type_helper.dart";
+
+main() {
+  num a = 0;
```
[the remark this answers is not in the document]
eernstg: OK, I can see how this can happen: With a context type of `_` it can be very difficult to verify that we have `_` and not `Object?`:

```dart
class A<X> {
  const A();
}

X whatever() => throw 0;

void main() {
  switch (whatever()) { // Context type schema should be `_`.
    case const A(): ...; // There is no expression that denotes the constant A, how can we test its type argument?
    case 1: ... // Doesn't care about the context type schema.
    case {"key": 5}: ... // Can't distinguish `Object?` and `_`.
    ... // And so on.
  }
}
```

In the case where we really can't check some specific property directly we may have to check it indirectly, and there are cases where we can do this (e.g., we can check the actual type arguments of a list pattern because it will be used as the declared type of a `var x` pattern field).

So we can't expect to get rid of all redundant test cases.

But it would still be a good idea to try to avoid redundancy whenever possible, especially in the cases where the given test case doesn't actually test precisely the topic of the given test.

(Like a run-time semantics test occurring in a test PR named 'Type checking' ;-)

LanguageFeatures/Patterns/type_inference_A29_t06.dart
```diff
@@ -0,0 +1,95 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+import "../../Utils/expect.dart";
+import "../../Utils/static_type_helper.dart";
+
+String test1() {
+  switch (<String, int>{"key1": 1}) {
+    case <Object, num>{"key1": var a}:
```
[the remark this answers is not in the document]
eernstg: (Damned github, I just wrote a rather detailed response, and it disappeared. Perhaps it isn't gone, and then you may see two very similar answers. ;-)

I think you'd need a `class D<X> extends C<X> { const D(); }` such that you can use a key in the pattern of the form `const D()`, which should then be inferred as `const D<B>()`.

The type arguments of the map pattern should still be `<C<B>, num>` such that `D<B>` is obtained from the context type `C<B>` in a way which involves a tiny bit of non-trivial type inference.

The initalizing expression should probably have static type `Object`, (`switch (<...>{...} as Object)`), such that we have the best possible reason to conclude that the key in the pattern was subject to type inference which is based on the actual type arguments of the map pattern.

And then the initializing expression should have a key `const D<B>()`, such that we will get a match at run time if and only if the inferred type argument of the pattern key is actually `B` as it should be.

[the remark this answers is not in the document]
eernstg: And it _should_ match!!! (The current version doesn't match because the initializing expression has an actual first type argument `C<A>`, which is too general when the pattern says `C<B>`, so we never reach the part where we're trying to match on a key.)

LanguageFeatures/Patterns/type_inference_A29_t07.dart
```diff
@@ -0,0 +1,104 @@
...
+/// `Map<K, V>` then value's value type is `V` and key context is `K`
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+import "../../Utils/expect.dart";
+import "../../Utils/static_type_helper.dart";
+
+class C<T> {
+  const C();
```
eernstg, line 31: Github refuses to show me the comment from you on this line:

> Again, function below returns "no match". Wha't wrong?

Looks like a bug, the pattern key (line 36) should infer as `const C<String>()`.

LanguageFeatures/Patterns/type_inference_A29_t13.dart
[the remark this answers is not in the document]
eernstg: OK!

LanguageFeatures/Patterns/type_inference_A29_t14.dart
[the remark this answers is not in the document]
eernstg: OK!

eernstg (commented):
Almost there, thanks for all the hard work!

LanguageFeatures/Patterns/type_inference_A29_t06.dart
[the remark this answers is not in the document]
sgrekhov: Thank you. I've changed the test

LanguageFeatures/Patterns/type_inference_A29_t07.dart
sgrekhov: Ok

LanguageFeatures/Patterns/type_inference_A31_t03.dart
sgrekhov: Yes. But here we check that it is no compile-time error to have an object pattern with required type which doesn't match

sgrekhov (commented):
Thank you for your hard review! Please take another look

LanguageFeatures/Patterns/type_inference_A29_t06.dart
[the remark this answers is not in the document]
eernstg: Thx!

PR #1857, opened 2023-02-16
#1401. Patterns execution. Assignment and switch statement tests

sgrekhov:


LanguageFeatures/Patterns/execution_switch_statement_A03_t02.dart
```diff
@@ -0,0 +1,83 @@
...
+main() {
+  test([1, 1]);
+  Expect.equals("guard1(true);shared body;", log);
+  test([2, 2]);
+  Expect.equals("guard2(true);shared body;", log);
+  test([3, 3]);
+  Expect.equals("guard3(true);shared body;", log);
+  test([4, 4]);
+  Expect.equals("guard4(true);not shared body;", log);
+  test([1, 42]);
```
eernstg, line 73: I think the test cases until line 72 are sufficient: The test cases in line 73-82 are, as far as I can see, precisely duplicating the cases in the previous library (because they never run any code that differs from that earlier test). It's better to avoid this redundancy such that the test bots are used to get the greatest possible coverage with given amount of power.

LanguageFeatures/Patterns/patterns_legacy_lib.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @description Defines data used by Patterns tests to check interaction with
+/// legacy
+/// @author sgrekhov22@gmail.com
+
+// @dart=2.9
```
eernstg, line 9: We shouldn't do this, because it will not be supported when patterns are part of the language, it's very very short-lived.

LanguageFeatures/Patterns/execution_switch_statement_A05_t01.dart
```diff
@@ -0,0 +1,57 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+import "patterns_legacy_lib.dart";
+import "../../Utils/expect.dart";
+
+bool test(bool b) {
+  switch (b) {
+    case true:
```
eernstg, line 46: I think it's better to use a `when` clause to ensure that the run-time error occurs.

LanguageFeatures/Patterns/patterns_legacy_lib.dart
eernstg: Actually, I'd be surprised if it will ever be possible to run this: Isn't it already a compile-time error to request unsound-null-safety execution when patterns or records are enabled?

Anyway, it's too short-lived to be useful.

LanguageFeatures/Patterns/execution_pattern_assignment_A01_t01.dart
```diff
@@ -0,0 +1,63 @@
...
+    dynamic pi = 3.14;
+    (r1, r2) = (0, pi);
+  });
+  Expect.equals(42, r1);
+  Expect.equals(42, r2);
+
+  int o1 = 42;
+  int o2 = 42;
+  Expect.throws(() {
+    Circle(sizeAsInt: o1, areaAsDynamic: o2) = Circle(1);
```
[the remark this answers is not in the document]
eernstg: Ah, `areaAsDynamic` will actually deliver a `double`. Surprising, but now I understand that it should work.

Maybe rename it to `areaAsDynamicDouble`, just so the next reader gets a hint?

eernstg (commented):
Looks good! Added a few comments.

LanguageFeatures/Patterns/patterns_legacy_lib.dart
sgrekhov: I can't run it manually, but via `test.py` this test works. Just to confirm, we are going to delete this test and leave the statement below (and the similar statemwent for switch-expression) untested (as almost not-actual). Correct?

> If the static type of v is an always-exhaustive type, no case matches, and there is no default clause, then throw a runtime error. This can only occur when null or a legacy typed value flows into this switch statement from another library that hasn't migrated to [null safety](https://dart.dev/null-safety). In fully migrated programs, exhaustiveness checking is sound and it isn't possible to reach this runtime error.

LanguageFeatures/Patterns/execution_switch_statement_A05_t01.dart
sgrekhov: Deleted this test at all

LanguageFeatures/Patterns/patterns_legacy_lib.dart
sgrekhov: Deleted

sgrekhov (commented):

LanguageFeatures/Patterns/execution_pattern_assignment_A01_t01.dart
```diff
@@ -0,0 +1,63 @@
...
+/// @description Check that if matching fails no any temporary values assigned
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
+import "../../Utils/expect.dart";
+import "patterns_lib.dart";
+
+main() {
+  var v1 = 42;
```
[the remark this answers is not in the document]
eernstg: Sounds good!

LanguageFeatures/Patterns/patterns_legacy_lib.dart
eernstg: Yes, there's no need to test unsound null safety semantics with patterns, they will never be available to the public together (and it doesn't matter that we can have that combination right now, basically by accident).

LanguageFeatures/Patterns/execution_switch_statement_A05_t01.dart
eernstg: OK!

eernstg (approved):
LGTM

PR #1859, opened 2023-02-17
#1401. Type inference. The last part of the tests

sgrekhov:


LanguageFeatures/Patterns/type_inference_A32_t01.dart
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion
+/// To type check a pattern p being matched against a value of type M:
+/// ...
+/// If p with required type T is in an irrefutable context:
+/// - It is a compile-time error if M is not assignable to T.
```
eernstg, line 9: This seems to be an alternative notion of 'assignable', cf. https://github.com/dart-lang/language/issues/2845#issuecomment-1434351830.

So we should understand the above spec line as "Perform applicable coercions on the matched value. It is a compile-time error if the static type of the coerced matched value is not assignable to `T`".

Consider the dynamic type check applied to an object which is obtained when the matched value type is `dynamic` and the required type is not a top type: The type check which is performed in this case is not strictly considered to be a coercion, because it is "included" in the notion of assignability ("`S` is assignable to `T` if `S <: T` or `S` is `dynamic`).

LanguageFeatures/Patterns/type_inference_A32_t01.dart
```diff
@@ -0,0 +1,36 @@
...
+/// - Else if M is not a subtype of T then an implicit coercion or cast is
+///   inserted before the pattern binds the value, tests the value's type,
+///   destructures the value, or invokes a function with the value as a target
+///   or argument.
```
eernstg, line 13: At this point the "real" coercions (generic function instantiation and int-to-double) have already taken place, and we only need to perform the dynamic type check where the matched value type is `dynamic`.

LanguageFeatures/Patterns/type_inference_A32_t01.dart
```diff
@@ -0,0 +1,36 @@
...
+  final (num v2) = "42";
+//                 ^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  int v3 = 0;
+  (v3) = "42";
+//       ^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
```
eernstg, line 36: Good, this should work!

LanguageFeatures/Patterns/type_inference_A32_t02.dart
```diff
@@ -0,0 +1,44 @@
...
+
+// SharedOptions=--enable-experiment=patterns
+
+main() {
+  var <int>[x1, num x2] = <num>[1, 2];
+//                        ^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  final [x3, num x4] = ["42", true];
```
eernstg, line 27: We need to use the simplest and most firm way to obtain a specific required type. I believe this would be achieved by using explicit type arguments on the pattern. So please use explicit types on the patterns everywhere in this test.

For a list and a map it's the type arguments at the top level, for a record it's specified for each field, so that could just be a declared type of a variable pattern. We could have `var <num>[int i] = ...;`, but that's a different error (it's an error for a variable pattern to have a declared type which is not statically known to be a supertype of the matched object type). So we don't need to test that here, either.

We have other tests where it is tested how the required type of the pattern is computed when it is not simply specified as explicit type arguments, and there's no need to duplicate this kind of testing here.

For the initializing expression I actually think it makes sense to write all types explicitly as well. The point is that (presumably) there is no way to obtain a different set of matched value types at all the levels that the pattern explores, other than those which can be expressed using explicit types: `["42", true]` will get one or the other type argument based on the pattern type schema and/or the elements, and we can always write an explicit type argument which is the same as that which would arise because of some complex interaction between elements and the pattern schema. No need to do that, we just need to force the type to be something specific, in the most firm way we can find.

This means that there are fewer cases to test here, but do consider `var <S>[...] = <T>[...]` with all relevant relationships between `S` and `T`: Same type, subtype, supertype, `dynamic`. I think this test does not depend on the elements in the initializing expression: They can be wrong (but that's a different error), and they can be correct (and that's what we want), but they can't change the matched value type anywhere in the pattern. So no tricks are needed in order to make those elements explicitly typed.

LanguageFeatures/Patterns/type_inference_A32_t03.dart
```diff
@@ -0,0 +1,46 @@
...
+///   inserted before the pattern binds the value, tests the value's type,
+///   destructures the value, or invokes a function with the value as a target
+///   or argument.
+///
+/// @description Check that it is a compile-time error if `M` is not assignable
+/// to `T` in irrefutable context. Test map patterns
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
```
eernstg, line 20: The comments in LanguageFeatures/Patterns/type_inference_A32_t02.dart should be applicable here as well.

LanguageFeatures/Patterns/type_inference_A32_t04.dart
```diff
@@ -0,0 +1,41 @@
...
+///   inserted before the pattern binds the value, tests the value's type,
+///   destructures the value, or invokes a function with the value as a target
+///   or argument.
+///
+/// @description Check that it is a compile-time error if `M` is not assignable
+/// to `T` in irrefutable context. Test record patterns
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
```
eernstg, line 20: The comments in LanguageFeatures/Patterns/type_inference_A32_t02.dart should be applicable here as well.

LanguageFeatures/Patterns/type_inference_A32_t05.dart
```diff
@@ -0,0 +1,33 @@
...
+///   inserted before the pattern binds the value, tests the value's type,
+///   destructures the value, or invokes a function with the value as a target
+///   or argument.
+///
+/// @description Check that it is a compile-time error if `M` is not assignable
+/// to `T` in irrefutable context. Test object patterns
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
```
eernstg, line 20: The comments in LanguageFeatures/Patterns/type_inference_A32_t02.dart should be applicable here as well.

LanguageFeatures/Patterns/type_inference_A33_t01.dart
```diff
@@ -0,0 +1,38 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
+import "../../Utils/expect.dart";
+
+main() {
+  var list = [1 as dynamic];
+  Expect.throws(() {
+    var [String _] = list;
```
eernstg, line 26: I think there is no other type argument for the list pattern that will work than a top type (because the initializing expression has type `List<dynamic>`). It probably has to be `dynamic`, because otherwise it's a compile-time error for the pattern field to have declared type `String`.

So the test will be safer and clearer if we have that explicit type argument on the pattern. It would be even safer if `list` had an explicitly declared type.

eernstg: So, in spite of the fact that this is different (it's about cast from `dynamic`), the comments in LanguageFeatures/Patterns/type_inference_A32_t02.dart seem to be applicable here as well.

LanguageFeatures/Patterns/type_inference_A33_t02.dart
```diff
@@ -0,0 +1,46 @@
...
+
+// SharedOptions=--enable-experiment=patterns,records
+
+import "../../Utils/expect.dart";
+
+T foo<T>(T t) => t;
+
+main() {
+  var list = [foo];
+  var [int Function(int) f1] = list;
```
eernstg, line 27: You could say that this is silly, but I think `var <int Function(int)>[f1] = ...` is better. The point is that we're consistently using the same mechanism to obtain a firmly and explicitly typed pattern.

Note that the type argument of the list pattern could be `dynamic`, so that's one thing which is less firm with `[int Function(int) f1]` than it is with `<int Function(int)>[f1]`.

We already have tests confirming that the declared type of `f1` in this case is `int Function(int)`, so it should not be necessary to declare the type of `f1` explicitly when the type argument of the list pattern is given explicitly.

eernstg (commented):
Added a few comments. The main idea that I'm pushing for is that the test should be very clearly focused on testing the property which is mentioned in `@description`, and this is achieved by specifying the types of all elements in the test in the most firm and simple way we have, which is typically achieved by giving all list/map patterns explicit type arguments, and using variables with an explicitly declared type in record patterns, etc.

LanguageFeatures/Patterns/type_inference_A33_t02.dart
sgrekhov: Changed. Please note that updated test fails in bot analyzer and dartk
```
unexpected analysis errors in type_inference_A33_t02.dart:
- Line 27, column 7: COMPILE_TIME_ERROR.PATTERN_TYPE_MISMATCH_IN_IRREFUTABLE_CONTEXT
  The matched value of type 'List<T Function<T>(T)>' isn't assignable to the required type 'List<int Function(int)>'.
```

sgrekhov (commented):
Tests updated. Please teake another look

LanguageFeatures/Patterns/type_inference_A32_t05.dart
eernstg: The matched value type in line 24 is `Shape<T>` for a `T` which is obtained by type inference.

Given that we're not testing phase 1 (computation of the pattern type schema) and phase 2 (inference on the initializing expression), it's better if the initializing expression has a firm type: All type arguments should be specified in the initializing expression.

(We should already have other tests testing phase 1 and 2.)

For the pattern, when testing "matched value type isn't assignable", we should also have a firm type, because then there's no doubt at all about (1) the matched value type and the pattern required type can be seen in the source code directly, and (2) from that we can easily see whether the assignability holds or not, so we can easily choose the correct expectation.

With object patterns we can have nested cases as well as top-level cases.

```dart
var Square(...) = Shape(); // OK, no type arguments after all, this must fail no matter what. ;-)
var Square<T1>(...) = Square<T2>(); // Top-level case, assignability determined by `T1`/`T2`.
var Square<T3>(area: T5 x1) = throw 0; // `area` has some return type, let's call it `T4`.
```

The third case gives rise to a matched value type from the return type `T4` of the getter `area`, which might again depend on the value of `T3`. We must then test that `T5` causes an error unless `T4` is assignable to `T5`. The initializing expression has type `Never` because it does not matter (and we just want it to be type correct and not interfere with the relationship between `T4` and `T5`).

This is again the shortest possible test: From `T3` to `T5`, we avoid inferring `T3` from the initializing expression, and we avoid inferring `T5` from the return type of `area` (which wouldn't allow us to test that the compile-time error is actually reported, so that's another reason why we don't allow `x1` to have an inferred type).

eernstg: The main thing here is that the nested case is missing.

LanguageFeatures/Patterns/type_inference_A33_t02.dart
eernstg: Of course, that's an error! ;-)

So we need to consider one more time how we can test the matched value type against the required type of the pattern. I think it isn't possible to obtain the generic function instantiation for the list pattern as such: We don't have a list-level coercion (like, "transform the elements one at a time").

This means that we'll just have to give up on the list (which means that it doesn't actually matter that the thing we're doing happens in a list, we could just as well have done it at the top level or in a record, etc.).

```dart
var <T Function<T>(T)>[int Function(int) f1] = list;
```

This will actually yield a situation where the matched value type is `T Function<T>(T)` and the pattern required type is `int Function(int)`, and I expect the spec wording to cover this case (so we should get a coercion, not a compile-time error).

Again, we could leave out the type argument on the pattern, but it's more readable to have it because it shows exactly the thing which is being tested: The relationship between the list pattern type argument and the declared type of the element, and the actions taken when they differ in this specific way.

Fun thing to try: :smile_cat:

```dart
var <T Function<T>(T)>[int Function(int) f1int && String Function(String) f1String] = list;
```

eernstg (commented):
A couple of threads are still open.

sgrekhov:
Thank you. Updated.

eernstg (commented):
Just one comment, I think we need to treat the map pattern in a way which is similar what we did to the list pattern.

LanguageFeatures/Patterns/type_inference_A33_t02.dart
```diff
@@ -0,0 +1,57 @@
...
+  Expect.throws(() {
+    fInt("1" as dynamic);
+  });
+  Expect.equals("42", fString("42"));
+  Expect.throws(() {
+    fString(1 as dynamic);
+  });
+
+  final map = {"key1": foo};
+  final <String, int Function(int)>{"key1": f2} = map;
```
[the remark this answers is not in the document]
eernstg: Is it true that the existing version of line 45 just gives rise to a compile-time error because the initializing expression has a static type which is not assignable to the required type of the map pattern?

eernstg (commented):
I think there's a need to make a change in this one before it proceeds.

eernstg:
I wonder why this one seems to have two copies of the comment that I just wrote. I wonder why this one seems to have two copies of the comment that I just wrote. ;-)

LanguageFeatures/Patterns/type_inference_A33_t02.dart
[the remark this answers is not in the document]
sgrekhov: Yes, this is true that the existing version of the line 54 produce a compile-time error. But updated version produce errors too. Is this an SDK issue?
```
unexpected analysis errors in type_inference_A33_t02.dart:
- Line 27, column 26: COMPILE_TIME_ERROR.PATTERN_TYPE_MISMATCH_IN_IRREFUTABLE_CONTEXT
  The matched value of type 'T Function<T>(T)' isn't assignable to the required type 'int Function(int)'.

- Line 33, column 26: COMPILE_TIME_ERROR.PATTERN_TYPE_MISMATCH_IN_IRREFUTABLE_CONTEXT
  The matched value of type 'T Function<T>(T)' isn't assignable to the required type 'int Function(int)'.

- Line 34, column 7: COMPILE_TIME_ERROR.PATTERN_TYPE_MISMATCH_IN_IRREFUTABLE_CONTEXT
  The matched value of type 'T Function<T>(T)' isn't assignable to the required type 'String Function(String)'.

- Line 45, column 44: COMPILE_TIME_ERROR.PATTERN_TYPE_MISMATCH_IN_IRREFUTABLE_CONTEXT
  The matched value of type 'T Function<T>(T)' isn't assignable to the required type 'int Function(int)'.
```

sgrekhov:
Updated, but there are still compile-time errors

LanguageFeatures/Patterns/type_inference_A33_t02.dart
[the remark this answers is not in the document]
eernstg: Yes, the coercion should take place when line 45 has `<String, T Function<T>(T)>{"key1": int Function(int) f2}`. However, line 52 should be able to work in the previous version, using the type `int Function(int)` for `x` (otherwise we won't have the coercion at all, and it should work to do it like that).

In both cases, it looks like the CFE just does not insert coercions, so it makes no difference whether the type difference occurs in the correct place or a different place. ;-)

However, I'm arguing in https://github.com/dart-lang/language/issues/2857#issuecomment-1458854384 that we should drop these "hidden coercions" entirely because they are a rather bad fit for various generalizations, in particular for one generalization which has been mentioned as desirable by members of the language team several times: Formal parameters expressed via patterns.

So we'll see how it goes, but for now we should stick to the current specification, which means that we should get the coercions as mentioned above.

eernstg (commented):
Just one thing remaining: I think line 52 in LanguageFeatures/Patterns/type_inference_A33_t02.dart needs to be changed back to the previous version, such that the test will actually have a coercion (and the error which is reported should not be reported, so the failure is OK).

sgrekhov:
Thank you. Reverted

eernstg (approved):
LGTM

PR #1872, opened 2023-02-21
#1401. Pattern-for-in statement and element tests

sgrekhov:
Please note the following.
1. Tests for for-in-element are written for `List` only. Please review them first and I'll then add the similar tests for `Map` and `Set`
2. Some tests check both static and dynamic semantic. I believe it's Ok if type check is part of the runtime semantics and it uses inferred types. 
3. We also need tests for async for-in loop. I'll add them in a separate PR. Again, let's review this one first

### DIFF
--- a/LanguageFeatures/Patterns/execution_pattern_for_in_element_A01_t01.dart
+++ b/LanguageFeatures/Patterns/execution_pattern_for_in_element_A01_t01.dart
@@ -1,0 +1,117 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion
+/// A collection element of the form:
+///
+/// for (<keyword> <pattern> in <expression>) <element>
+///
+/// Where <keyword> is var or final is treated like so:
+/// 1. Let I be the static type of <expression>, inferred using context type
+///   schema Iterable<P> where P is the context type schema of <pattern>.
+/// 2. Calculate the element type of I:
+///     i. If I implements Iterable<T> for some T then E is T.
+///     ii. Else if I is dynamic then E is dynamic.
+///     iii. Else it is a compile-time error.
+/// 3. Type check <pattern> with matched value type E.
+/// 4. If there are no compile-time errors, then execution proceeds as the
+///   following code, where id1 and id2 are fresh identifiers and append() is
+///   an operation to add an element to the surrounding collection being built:
+/// ```dart
+/// I id1 = <expression>;
+/// Iterator<E> id2 = id1.iterator;
+/// while (id2.moveNext()) {
+///   <keyword> <pattern> = id2.current;
+///   append(<element>);
+/// }
+/// ```
+/// @description Checks that if `I` implements `Iterable<T>` for some `T` then
+/// `E` is `T`. Test `T` inferred from `<pattern>` to `<expression>` and from
+/// `<expression>` to `<pattern>`
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,records
+
+import "../../Utils/expect.dart";
+import "../../Utils/static_type_helper.dart";
+import "patterns_lib.dart";
+
+main() {
+  var l1 = [
+    0,
+    for (var (num v1) in [1, 2, 3]) v1,
+    4
+  ];
+  Expect.listEquals([0, 1, 2, 3, 4], l1);
+  l1.expectStaticType<Exactly<List<num>>>();
+
+  var l2 = [
+    0,
+    for (final (v2) in <num>[1, 2, 3]) v2,
+    4
+  ];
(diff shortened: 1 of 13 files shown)


eernstg:
OK, fine! Taking a look now.

LanguageFeatures/Patterns/execution_pattern_for_in_statement_A01_t02.dart
```diff
@@ -0,0 +1,113 @@
...
+
+  log = "";
+  for (var Square<Centimeter>(area: v11) in [Square(1)] as dynamic) {
+    v11.expectStaticType<Exactly<Unit<Centimeter>>>();
+    log += "$v11;";
+  }
+  Expect.equals("1;", log);
+
+  log = "";
+  for (final Square(area: v12) in
```
eernstg, line 107: This one is interesting!

So what do we do when an objectPattern receives a matched value type of `dynamic`?

> Apply downwards inference from `M` to infer type arguments for `X` if needed.

In this case `M` is `dynamic`, so we would presumably use instantiation to bound in order to find the type argument of `Square`. OK, line 109 tests for `Unit<MetricUnits>`.

OK, great!

LanguageFeatures/Patterns/execution_pattern_for_in_statement_A02_t01.dart
```diff
@@ -0,0 +1,66 @@
...
+//                     ^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  for (final <int>[v2] in <List<num>>[[1], [2], [3]]) {}
+//                        ^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  for (var <String, int>{"k1": v3} in <String, num>[{"k1": 1}]) {}
```
eernstg, line 47: Probably
```suggestion
  for (var <String, int>{"k1": v3} in <Map<String, num>>[{"k1": 1}]) {}
```

LanguageFeatures/Patterns/execution_pattern_for_in_statement_A03_t01.dart
```diff
@@ -0,0 +1,78 @@
...
+///   following code, where id1 and id2 are fresh identifiers:
+/// ```dart
+/// var id1 = <expression>;
+/// var id2 = id1.iterator;
+/// while (id2.moveNext()) {
+///   <keyword> <pattern> = id2.current;
+///   { <statement> }
+/// }
+/// ```
+/// @description Checks that it is a compile-time error if final variable
```
eernstg, line 28:
```suggestion
/// @description Checks that it is a compile-time error if a final variable
```

eernstg (commented):
Looks good!

A couple of tests look like they could be considerably more tight (especially, when typing information is transferred from the pattern to the initializing expression or vice versa, we could test the outcome more directly at the location where it happens).

Also some tests look like they aren't testing very much. This might be somewhat unfair ;-), we _do_ need to test the actual behavior, that is `Expect.listEquals(...)`. On the other hand, there are a number of test cases whose special property is a different distribution of explicit typing information, and in some cases they don't seem to bring anything new.

Anyway, take a look at the comments and complain if I'm being unreasonable. ;-)

LanguageFeatures/Patterns/execution_pattern_for_in_element_A01_t01.dart
```diff
@@ -0,0 +1,117 @@
...
+    0,
+    for (final (v2) in <num>[1, 2, 3]) v2,
+    4
+  ];
+  Expect.listEquals([0, 1, 2, 3, 4], l2);
+  l2.expectStaticType<Exactly<List<num>>>();
+
+  var l3 = [
+    0,
+    for (var <num>[v3] in [[1], [2], [3]]) v3,
```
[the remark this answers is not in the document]
sgrekhov: Again` Storing.stored` has type `List<dynamic>` here and for all cases below.

sgrekhov (commented):
I'm updating these tests now, question arised

LanguageFeatures/Patterns/execution_pattern_for_in_element_A01_t01.dart
```diff
@@ -0,0 +1,117 @@
...
+import "patterns_lib.dart";
+
+main() {
+  var l1 = [
+    0,
+    for (var (num v1) in [1, 2, 3]) v1,
+    4
+  ];
+  Expect.listEquals([0, 1, 2, 3, 4], l1);
+  l1.expectStaticType<Exactly<List<num>>>();
```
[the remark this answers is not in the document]
eernstg: Yes, that would be a bug: When it's `List<dynamic>` I suspect that the context type is `_`, but it should have been `Iterable<num>`, because the type schema of the pattern is `num`.

eernstg:
This is the PR I was referring to when I mentioned that there was one PR where there would be a need to do something more substantial.

I tried to answer the questions I could spot. Please give me a heads up when it is ready for a full review.

sgrekhov:
Updated. Please review

eernstg (approved):
LGTM

PR #1880, opened 2023-02-22
Fixes #1876. Don't expect type promotion in shared_case_scope_A02_t03

### LINKED ISSUE #1876: co19/LanguageFeatures/Patterns/shared_case_scope_A02_t03
See https://github.com/dart-lang/sdk/issues/51464 for discussion.

```dart
String test1(num? x) {
  switch (x) {
    case final a when a is int:
    case final a when a is num:
      a.expectStaticType<Exactly<num>>();
      print(a);
      return "match";
    default:
      return "default";
  }
}
```
This code does not give `num a`, as it is specified right now, it gives `num?`.

```
FAILED: dart2analyzer-none release_arm64 co19/LanguageFeatures/Patterns/shared_case_scope_A02_t03
Expected: Pass
Actual: CompileTimeError

--- Command "dart2analyzer" (took 06.000504s):
DART_CONFIGURATION=ReleaseARM64 sdk/bin/dart --packages=.dart_tool/package_config.json pkg/analyzer_cli/bin/analyzer.dart --use-analysis-driver-memory-byte-store --dart-sdk=/Users/scheglov/Source/Dart/sdk.git/sdk/sdk -Dtest_runner.configuration=custom-configuration-1 --enable-experiment=patterns --ignore-unrecognized-flags --packages=/Users/scheglov/Source/Dart/sdk.git/sdk/.dart_tool/package_config.json --format=json /Users/scheglov/Source/Dart/sdk.git/sdk/tests/co19/src/LanguageFeatures/Patterns/shared_case_scope_A02_t03.dart

unexpected analysis errors in shared_case_scope_A02_t03.dart:
- Line 47, column 26: COMPILE_TIME_ERROR.TYPE_ARGUMENT_NOT_MATCHING_BOUNDS
  'num Function(num)' doesn't conform to the bound 'num? Function(num?)' of the type parameter 'R'.

- Line 59, column 26: COMPILE_TIME_ERROR.TYPE_ARGUMENT_NOT_MATCHING_BOUNDS
  'num Function(num)' doesn't conform to the bound 'num? Function(num?)' of the type parameter 'R'.

--- Re-run this test:
python3 tools/test.py -m release -c dart2analyzer -a arm64 co19/LanguageFeatures/Patterns/shared_case_scope_A02_t03
```

sgrekhov:


LanguageFeatures/Patterns/shared_case_scope_A02_t03.dart
```diff
@@ -32,7 +32,7 @@
 ///   the body resolves to a variable in s that isn't shared.
 ///
 /// @description Checks that shared variables can be used in a shared case scope
-/// body. Test the case when the corresponding variables in vs are promoted
```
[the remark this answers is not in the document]
stereotype441: @eernstg said:
> I guess the rationale is that we won't promote to a type that isn't a 'type of interest', and an approach where we use the least upper bound of several types could easily yield a type that nobody ever mentioned in the code so far.

This matches my recollection as well.  And I agree with your assessment of what the "intersection" operation does.

> The issue [dart-lang/language#2856](https://github.com/dart-lang/language/issues/2856) seems to propose that we find an intersection which is "better" (that is, it allows the given variable to get a more special type). I've commented on the issue in order to suggest a way this could be done. In any case, that's not a feature that we have at this time.
> 
> So I'm suggesting that we adjust this test such that it only tests cases which are expected to remain valid even in the case where some proposal like [dart-lang/language#2856](https://github.com/dart-lang/language/issues/2856) is adopted.

I have mixed feelings about this.  On the one hand, it's nice for tests to be orthogonal and future-proof, so that when we change the behaviour of the language, we have fewer tests to update.  So yeah, assuming we decide to go ahead with https://github.com/dart-lang/language/issues/2856, it would be convenient if these tests would still work.

On the other hand, it's also sometimes helpful to have tests that *haven't* been designed with a lot of orthogonality in mind, because (a) it makes them more likely to exercise corner cases we didn't think of during the design, and (b) it gives us a better chance of noticiing if we make an accidental change to the implementation's behaviour.

We have a bunch of tests in the SDK repo under `tests/language`, and traditionally they tend to be designed in a fairly orthogonal and future-proof way, because they're often written by same people who are developing the product.  Whereas traditionally the co19 tests tend to be a lot less orthogonal and cover more weird corner cases, precisely because they're *not* written by the same team that implements the language.  So my personal opinion is that @sgrekhov should go ahead and follow his own instincts for how to structure the tests, and not worry too much about planning for the future, so that we keep those advantages.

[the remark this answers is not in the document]
eernstg: Thanks, @stereotype441! The difference between tests written by implementers or language designers and tests written by others is certainly something I've noted and appreciated as well. Another relevant distinction is between tests that are written based on a specification vs. tests that are written based on a broader preexisting amount of knowledge that includes detailed knowledge about the implementation. It's good that we have both.

About the intersection operation: In terms of the concrete example, I'll conclude that we _should_ get a promotion to `num` in the case where a variable has the declared type `num?` and is promoted as to `num` in every branch (it can be further promoted, say, to `int`, in one or more branches, but it _must_ be promoted to `num` at some point in every branch, otherwise `num` won't be available as the result of the shared promotion).

LanguageFeatures/Patterns/shared_case_scope_A02_t03.dart
```diff
@@ -44,7 +44,7 @@ String test1(num? x) {
   switch (x) {
     case final a when a is int:
```
eernstg, line 45: We can make sure that `a` is promoted to `num` in every branch (and may or may not be promoted further) by including an `a is num`. So:
```suggestion
    case final a when a is num && a is int:
```
The promotion chains would then be `[int, num, num?]` and `[num, num?]`, and the intersection is `[num, num?]` and hence `a` has type `num`.

LanguageFeatures/Patterns/shared_case_scope_A02_t03.dart
```diff
@@ -56,7 +56,7 @@ String test2(num? x) {
   switch (x) {
     case var a when a is int:
     case var a when a is num:
-      a.expectStaticType<Exactly<num>>();
+      a.expectStaticType<Exactly<num?>>();
```
eernstg, line 59: OK, this case will then test that we do _not_ get a promotion when we have promotion chains like `[int, num?]` and `[num, num?]`, because `[int, num?] ^ [num, num?] == [num?]`.

LanguageFeatures/Patterns/shared_case_scope_A02_t03.dart
```diff
@@ -44,7 +44,7 @@ String test1(num? x) {
   switch (x) {
     case final a when a is int:
     case final a when a is num:
-      a.expectStaticType<Exactly<num>>();
+      a.expectStaticType<Exactly<num?>>();
```
eernstg, line 47:
```suggestion
      a.expectStaticType<Exactly<num>>();
```

eernstg (commented):
A few extra comments. It has now been clarified what 'intersection' means, so we can adjust the tests such that we get (or do not get) the promotion.

sgrekhov:
Thank you for clarification. Updated. Please review

sgrekhov:
Description changed

eernstg (approved):
Thanks, LGTM!

Issue #1877, opened 2023-02-21
co19/LanguageFeatures/Patterns/type_inference_A28_t05


scheglov:
https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#type-checking-and-pattern-required-type says "These patterns are prohibited in an irrefutable context if the matched value isn't assignable to that type."

However here we are not in an irrefutable context, so no error is required.
```dart
  switch([]) {
    case <String>[int v]:
//                    ^
// [analyzer] unspecified
// [cfe] unspecified
    default:
  }
```

scheglov:
Similarly `co19/LanguageFeatures/Patterns/type_inference_A29_t04`

scheglov:
```
FAILED: dart2analyzer-none release_arm64 co19/LanguageFeatures/Patterns/type_inference_A28_t05
Expected: Pass
Actual: MissingCompileTimeError

--- Command "dart2analyzer" (took 28ms):
DART_CONFIGURATION=ReleaseARM64 sdk/bin/dart --packages=.dart_tool/package_config.json pkg/analyzer_cli/bin/analyzer.dart --use-analysis-driver-memory-byte-store --dart-sdk=/Users/scheglov/Source/Dart/sdk.git/sdk/sdk -Dtest_runner.configuration=custom-configuration-1 --enable-experiment=patterns --ignore-unrecognized-flags --packages=/Users/scheglov/Source/Dart/sdk.git/sdk/.dart_tool/package_config.json --format=json /Users/scheglov/Source/Dart/sdk.git/sdk/tests/co19/src/LanguageFeatures/Patterns/type_inference_A28_t05.dart

static error failures:
- Missing expected unspecified error at line 46, column 23, length 1.

- Missing expected unspecified error at line 52, column 23, length 1.

- Missing expected unspecified error at line 57, column 18, length 1.

--- Re-run this test:
python3 tools/test.py -m release -c dart2analyzer -a arm64 co19/LanguageFeatures/Patterns/type_inference_A28_t05
[00:09 |  13% | +   59 | -    1]

FAILED: dart2analyzer-none release_arm64 co19/LanguageFeatures/Patterns/type_inference_A29_t04
Expected: Pass
Actual: MissingCompileTimeError

--- Command "dart2analyzer" (took 25ms):
[...]

PR #1881, opened 2023-02-22
#1877. Tiny typo fixed

sgrekhov:


LanguageFeatures/Patterns/type_inference_A29_t04.dart
```diff
@@ -52,7 +52,7 @@ String test2() {
 
 String test3() {
   if ({} as dynamic case <String, int>{"key1": String a}) {
```
eernstg, line 54: Wouldn't this be yet another case where there is no compile-time error? The context is refutable, and we're type checking `String a` with a matched value type of `int`, but that's OK in a refutable context.

sgrekhov:
Yes, this test should be rewritten. Closing this

PR #1886, opened 2023-02-22
Fixes #1877. No compile-time error if type-test is failed in a refutable context

### LINKED ISSUE #1877: co19/LanguageFeatures/Patterns/type_inference_A28_t05

sgrekhov:


eernstg (approved):
LGTM

PR #1888, opened 2023-02-23
#1401. Test for matching logical and relational patterns

sgrekhov:


LanguageFeatures/Patterns/matching_relational_A02_t02.dart
```diff
@@ -0,0 +1,100 @@
...
+///   a. Let r be the result of v == c.
+///   b. The pattern matches if r is false and fails otherwise. This takes into
+///     account the built-in semantics that null is only equal to null.
+/// iv. Else the operator is a comparison operator op:
+///   a. Let r be the result of calling op on v with argument c.
+///   b. If r is not a bool then throw a runtime error. This can happen if the
+///     operator on v's type returns dynamic.
+///   c. The pattern matches if r is true and fails otherwise.
+///
+/// @description Check that in case `c != null` only `v == null` doesn't match.
```
eernstg, line 35: We can't actually verify that _every_ other object than null will be a non-match, but we can try it out with some specific objects. ;-)

LanguageFeatures/Patterns/matching_relational_A02_t02.dart
```diff
@@ -0,0 +1,100 @@
...
+  String log = "";
+  C(this.value);
+
+  @override
+  bool operator ==(Object? other) {
+    log += "C($value)==$other;";
+    if (other is int) {
+      return this.value == other;
+    }
+    return this == other;
```
eernstg, line 53: Infinite loop?

LanguageFeatures/Patterns/matching_relational_A02_t02.dart
```diff
@@ -0,0 +1,100 @@
...
+    != null => "null",
+    _ => "no match"
+  };
+}
+
+main() {
+  C c1 = C(1);
+  C c2 = C(null);
+  Expect.equals("null", test1(c1));
+  Expect.equals("", c1.log);
```
eernstg, line 84: That's cool! We can actually watch the evaluation of the equality test and see that it doesn't give rise to an invocation of `C.==` at all.

LanguageFeatures/Patterns/matching_relational_A03_t01.dart
```diff
@@ -0,0 +1,114 @@
...
+///   a. Let r be the result of v == c.
+///   b. The pattern matches if r is false and fails otherwise. This takes into
+///     account the built-in semantics that null is only equal to null.
+/// iv. Else the operator is a comparison operator op:
+///   a. Let r be the result of calling op on v with argument c.
+///   b. If r is not a bool then throw a runtime error. This can happen if the
+///     operator on v's type returns dynamic.
+///   c. The pattern matches if r is true and fails otherwise.
+///
+/// @description Check that it is a run-time error if result of calling `op` on
```
eernstg, line 35:
```suggestion
/// @description Check that it is a run-time error if the result of calling `op` on
```

eernstg (commented):
Looks good! I think there's an infinite loop which doesn't actually run in these tests because all test cases avoid running that particular line of code. But it makes sense to fix it, and the fix seems to be very simple.

sgrekhov:
Fixed. An infinite loop, `Object?` as argument of  `==` operator and a grammar. Please rereview

eernstg (approved):
LGTM

PR #1894, opened 2023-02-24
#1401. Matching tests added

sgrekhov:


LanguageFeatures/Patterns/matching_cast_A01_t01.dart
```diff
@@ -0,0 +1,74 @@
...
+  switch (o) {
+    <num>[var v] as List<int> => "match",
+    _ => "no match"
+  };
+
+main() {
+  Expect.throws(() {
+    var (<num>[v] as List<int>) = [3.14];
+  });
+  var (<num>[v] as List<int>) = [42];
```
eernstg, line 59: I think it would be helpful to have an explicit type argument on the list literal in the initializing expression, because this simplifies the reasoning about what we're testing: Then we know immediately that the initializing expression evaluates to a `List<int>`, and then we know that we're testing the step where a `List<int>` is checked for being a `List<int>` at run time.

It's the usual consideration about testing a specific step in a focused and precise manner (and in particular: not testing type schema computation or initializing expression inference).

Also, a reader would need to know the rules quite precisely in order to avoid some misunderstandings, e.g., "the pattern yields the type `List<num>`, so we probably infer the initializing expression as `List<num>`". This would confuse the situation because we might then expect the cast to fail.

The previous case (line 57) would actually be clarified by having a type argument on the initializing expression as well (but that case is a bit less confusing because we couldn't possibly end up with an inferred type argument on the initializing expression that would make the cast succeed).

LanguageFeatures/Patterns/matching_cast_A01_t01.dart
```diff
@@ -0,0 +1,74 @@
...
+  Expect.equals(42, v);
+
+  Expect.throws(() {
+    test1([3.14]);
```
eernstg, line 63: Here we'd need to take an even closer look in order to know whether this means `test1(<num>[3.14])` or `test1(<T>[3.14])` for some other `T` (`Object`, `Object?`, `dynamic`, ...).

So it's still helpful to have an explicit type argument on the list literal.

Similarly for the remaining cases.

eernstg (commented):
First, I'm suggesting (as usual ;-) that we add some explicit type arguments, in order to ensure that we're firmly stating the properties that define the step that we're testing (rather than stating some properties that involve several other steps which are not being tested).

But there was another issue which came up, and that issue should probably be split off into a separate PR:

In these tests, but also in a bunch of tests that have already been landed, there is a class that overrides operator `==`, but not `hashCode`.

We would need to be really, really careful that those objects are _never_ added to a `Set`, or used as a key in a `Map`.

I think it would be a very good idea to search for all those overriding operator `==` declarations, and ensure that there is a corresponding `hashCode` getter.

In general, the dangerous situation is when two objects compare equal according to `==`, but they have different `hashCode` values: It would then be impossible to find the object in situations where it is assumed that it is only necessary to look among objects with the same hash code (e.g., in a hash map).

sgrekhov:
Added type arguments. It's also a good poing regarding the `hashCode`. I'll add it in a separate pull request

LanguageFeatures/Patterns/matching_variable_A01_t01.dart
```diff
@@ -0,0 +1,102 @@
...
+  Expect.equals("match-1", test3(1 as dynamic));
+  Expect.equals("match-2", test3("1"));
+  Expect.equals("match-2", test3("1" as dynamic));
+  Expect.equals("match-1", test3(answer));
+  Expect.equals("match-3", test3(null));
+  Expect.equals("match-3", test3(3.14));
+  Expect.equals("no match", test3(true));
+
+  Expect.throws(() {
+    var [int v] = ["42"] as dynamic;
```
eernstg, line 94: This one is slightly dangerous (in that it tests things which are outside the topic): It might just be the case that the test succeeds for the wrong reason if the list pattern gets a required type which is not a supertype of the actual type argument of the initializing expression value.

For instance, if we actually have the following after phase 1/2/3 then we would have a dynamic error which is caused by the list pattern requiring an `int` list, not by the variable pattern requiring an `int`:

```dart
    var <int>[int v] = ["42"] as dynamic;
```

Also, as usual, we already have other tests where it is tested that the list pattern gets the correct actual type argument. So the following would be aimed more precisely at the topic:

```suggestion
    var <dynamic>[int v] = ["42"] as dynamic;
```

Similarly for line 97 and 100.

eernstg (commented):
Just one tiny thing.

LanguageFeatures/Patterns/matching_variable_A01_t01.dart
sgrekhov: Changed. Please rereview

eernstg: Thx!

eernstg (approved):
LGTM

PR #1906, opened 2023-03-06
Fixes #1901. Add `void` to prevent wrong failure

### LINKED ISSUE #1901: src/Language/Classes/definition_t15.dart is failing with records enabled
It correctly reports the syntax error on the invalid `()` class body. But it also reports an error about `main()` returning null since the `()` is parsed as the return type of `main()`. A simple fix is:

```
class A ()
//       ^
// [analyzer] unspecified
// [cfe] unspecified

void main() { // <-- void
  try {
    new A();
  } catch (e) {}
}
```

sgrekhov:


eernstg (approved):
LGTM

PR #1907, opened 2023-03-07
Fixes #1899. Language version 2.19 added to legacy mixins tests

### LINKED ISSUE #1899: Set language version of tests mixing in classes to 2.19
Dart 3.0 makes it an error to mixin a class that isn't marked `mixin class`. This breaks a lot of existing Dart SDK tests that were doing that, often for unrelated reasons:

* Many of these tests are older than mixin declaration syntax so mixing in a class was the only way to test any mixin functionality.

* Some of them are deliberately mixing in classes, and those classes should just be marked `mixin`.

* Some of them are mixing in classes from the core libraries that are no longer allowed to be mixins.

* There are probably other reasons I'm not thinking of.

Eventually, we'll want to figure out why each of these tests is mixing in a class and fix it appropriately. But, for now, in order to be able to enable the "class-modifiers" flag, we want to just get the tests passing and continuing to behave the way they do today.

The easiest way to fix most of those tests is to simply pin them to language version 2.19. If the class that the test is mixing in is defined in the same library, then this will preserve the current behavior. It's only an error to mixin a non-`mixin` class if the class is declared in a Dart 3.0 library. This probably won't fix *all* test failures since some tests may be mixing in classes from other libraries, but it should fix most.

I landed [a fix](https://dart-review.googlesource.com/c/sdk/+/285923). for the language/ and lib/ tests today. Can you do a similar fix for the co19 tests?

Here is tweaked version of the very hacky script I wrote to add the `@dart=2.19` comments to all of the SDK tests. It includes a list of all of the co19 tests that were failing because of mixin errors as of a couple of days ago:

https://gist.github.com/munificent/b81adc0eaec0db34be713755982cc72c

You may need to tweak it a bit to get it to work, but that should get you started.

Thank you, and let me know if there's anything I can do to help.

cc @jakemac53

sgrekhov:


eernstg (approved):
As far as I can see, the only change which is not adding the block `TODO ... @dart=2.19` is in generator.dart. I think it's a good approach to put the TODO in all those tests for now. We can then settle the further adjustments later on.

LGTM!

PR #1910, opened 2023-03-08
Fixes #1909. Update tests not to fail with patterns feature enabled

### LINKED ISSUE #1909: Couple more test changes related to switches and primitive equality
I found a few more tests that need to be updated because of the changes to switch statements and primitive equality:

**No longer an error now that `==` is a constant expression is the LHS has primitive equality:**

co19/Language/Expressions/Constants/equals_expression_t03

**Different error for non-exhaustive enum switches:**

co19/LanguageFeatures/nnbd/weak/static_errors_A30_t04
co19/LanguageFeatures/nnbd/weak/static_errors_A30_t07
co19/LanguageFeatures/nnbd/weak/static_errors_A30_t08

The error for a switch statement that doesn't cover all enum cases is different now because it's a more general "non-exhaustive switch" error. Since these are explicitly weak mode tests, the right fix is probably to pin them to `@dart=2.19` and then they should continue to report the old error.

Thanks for the quick turnaround on the other ones. :)

sgrekhov:


Language/Expressions/Constants/equals_expression_t03.dart
```diff
@@ -9,22 +9,21 @@
...
 }
 
 final constList = const [
-//                ^
-// [cfe] unspecified
-   const A() == "hello",
-// ^
-// [analyzer] unspecified
+  const A() == "hello",
+  const A() != "world"
```
eernstg, line 26: To fit the description, we might as well have a couple of cases like `1.5 == A()` and `B() == null`, where `B` is a class that does not have primitive equality (because _any_ constant is OK when the right operand is `null`).

Language/Expressions/Constants/equals_expression_t03.dart
```diff
@@ -9,22 +9,21 @@
 /// • An expression of one of the forms e1 == e2 or e1 != e2 where e1 and e2 are
 ///   constant expressions that evaluate to a numeric, string or boolean value
 ///   or to null.
-/// @description Checks that it is a compile-time error when an expression of
-/// the form e1 == e2 where e1 does not evaluate to a numeric, string or boolean
-/// value, is used to initialize a constant variable.
+/// @description Checks that it is not an time error when an expression of
+/// the form e1 == e2 or e1 != e2 where e1 does not evaluate to a numeric,
+/// string or boolean value, is used to initialize a constant variable.
```
eernstg, line 14: I think we need to express this in a slightly more direct way (the negations are piling up). The actual rule today is that `e1 == e2` is constant if `e1` and `e2` are constant, `e1` evaluates to an object of type `double` or an object with primitive equality, or `e2` is null (so `e1 == null` is constant for _any_ constant `e1`).
```suggestion
/// @description Checks that it is not a compile-time error when an expression of
/// the form `e1 == e2` or `e1 != e2` is used to initialize a constant variable, and
/// `e1` does not evaluate to a numeric, string or boolean value, as long as the
/// receiver `e1` has primitive equality or is a `double`.
```

I moved the long constraint to the end; English sentences tend to avoid having anything large (like 'where ... value') between the noun phrase and the verb phrase.

eernstg (commented):
Suggesting a couple of extra cases, to fit the test description.

sgrekhov:
Updated the test. I modified `@assert` to match the current spec and changed the description accordingly

eernstg (approved):
LGTM

PR #1921, opened 2023-03-10
Fixes #1917. Use `const(...)` for logical operator in switch

### LINKED ISSUE #1917: co19/Language/Statements/Switch/syntax_t01.dart expectation should be updated
This case https://github.com/dart-lang/co19/blob/07279694bd8d95a8e5aea258624fdfb3feb01f3f/Language/Statements/Switch/syntax_t01.dart#L33, when opted into patterns, will only match `value` against both `true` and `false`, which will always be `false`.

Prior to patterns it would have matched against the expression `true && false` (ie: `false`).

So the expectation here https://github.com/dart-lang/co19/blob/07279694bd8d95a8e5aea258624fdfb3feb01f3f/Language/Statements/Switch/syntax_t01.dart#L85 should be `-1`.

sgrekhov:


eernstg (approved):
LGTM

PR #1922, opened 2023-03-10
Fixes #1919. Don't use Dart native classes for runtime type checks

### LINKED ISSUE #1919: LanguageFeatures/Records/runtime_type_A01_t01.dart should use new classes rather than num/int
Although `int` may not be subclassed by external users of the language, the implementations internally subclass it, causing the test to fail. See https://github.com/dart-lang/language/issues/2895 for discussion as well as an example of how the test can be rewritten.

sgrekhov:


eernstg (approved):
LGTM

PR #1934, opened 2023-03-14
#1401. Matching tests for record and object patterns

sgrekhov:


LanguageFeatures/Patterns/matching_object_A02_t01.dart
```diff
@@ -0,0 +1,84 @@
...
+  var Square(area: v1, size: v2) = o;
+}
+
+main() {
+  Square square = Square(1, logger);
+  Expect.equals("match", test1(square));
+  Expect.equals("Square.area:(1==1);Square.size", log);
+  log = "";
+  Expect.equals("match", test2(square));
+  Expect.equals("Square.sizeSquare.area:(1==1);", log);
```
eernstg, line 77: So `size` doesn't add a semicolon?

LanguageFeatures/Patterns/matching_object_A02_t01.dart
```diff
@@ -0,0 +1,84 @@
...
+  log = "";
+  Expect.equals("match", test3(square));
+  Expect.equals("Square.area:(1==1);Square.size", log);
+  log = "";
+  test4(square);
+  Expect.equals("Square.areaSquare.size", log);
```
eernstg, line 83: No `:(1==1);`, or something like that?

LanguageFeatures/Patterns/matching_object_A02_t02.dart
```diff
@@ -0,0 +1,89 @@
...
+  Expect.equals("Square.area:(1==1);Square.size", log);
+  log = "";
+  Expect.equals("match", test2(square));
+  Expect.equals("Square.sizeSquare.area:(1==1);", log);
+  log = "";
+  Expect.equals("match", test3(square));
+  Expect.equals("Square.area:(1==1);Square.size", log);
+  log = "";
+  test4(square);
+  Expect.equals("Square.areaSquare.size", log);
```
eernstg, line 88: Again surprising that area doesn't include something like `:(1==1);`.

LanguageFeatures/Patterns/matching_record_A01_t01.dart
```diff
@@ -0,0 +1,130 @@
...
+}
+
+void test4(dynamic o) {
+  var (int v1, num _) = o;
+}
+
+main() {
+  Expect.equals("no match", test1(()));
+  Expect.equals("no match", test1((1,)));
+  Expect.equals("no match", test1((1, "2")));
```
eernstg, line 76: Considering the description, 

> if the runtime type of `v` is not a subtype of the
> required type of `p` then the match fails

this is only about the shape test (because the required type of a record pattern is always `Object?` at every field, in some overall shape). In order to actually test the shape check we should probably have a subpattern for each field which is guaranteed to match. An easy choice would be `_` in every field.

We could of course also interpret the description as applicable recursively (so we'd also have subpatterns, and those subpatterns and their matching behavior should also be tested). In that case we could use `T _` where `T` is some appropriate type.

There would presumably be some other test where the description says that we're testing the matching step for each of the fields of the record according to a constant (that is, we're testing their value, not the required type).

(Surely we already have that kind of test, named something like `...constant...', right? In that case we should be able to reduce this one such that it tests required types and match failure.)

LanguageFeatures/Patterns/matching_record_A01_t02.dart
```diff
@@ -0,0 +1,114 @@
...
+String test3(Object o) {
+  return switch (o) {
+      (c1, c2) => "match-1",
+      (c1, name1: c2, c3, name2: c4) => "match-2",
+      _ => "no match"
+    };
+}
+
+main() {
+  Expect.equals("no match", test1((c1, c2, c3)));
```
eernstg, line 94: This is again testing values, not required types.

LanguageFeatures/Patterns/matching_record_A02_t01.dart
```diff
@@ -0,0 +1,122 @@
...
+/// i. If the runtime type of v is not a subtype of the required type of p, then
+///   the match fails.
+/// ii. For each field f in p, in source order:
+///   a. Access the corresponding field in record v as r.
+///   b. Match the subpattern of f against r. If the match fails, the record
+///     match fails.
+/// iii. The match succeeds if all field subpatterns match.
+///
+/// @description Checks that for each field `f` in `p` in a source order the
+/// corresponding field in record `v` is matched against subpattern
```
eernstg, line 27: Sounds like this one is the correct test for doing all those pattern matching cases where we test values in the subpatterns!

LanguageFeatures/Patterns/matching_record_A02_t02.dart
```diff
@@ -0,0 +1,112 @@
...
+/// i. If the runtime type of v is not a subtype of the required type of p, then
+///   the match fails.
+/// ii. For each field f in p, in source order:
+///   a. Access the corresponding field in record v as r.
+///   b. Match the subpattern of f against r. If the match fails, the record
+///     match fails.
+/// iii. The match succeeds if all field subpatterns match.
+///
+/// @description Checks that for each field `f` in `p` in a source order the
+/// corresponding field in record `v` is matched against subpattern
```
eernstg, line 27: OK, again matching values. So there is absolutely no problem in changing LanguageFeatures/Patterns/matching_record_A01_t01.dart and ...2.dart such that they only test required types.

eernstg (commented):
A few comments.

In some cases I'm just asking why the log looks funny (missing semicolon?). I did not check the code which is called, so it's probably fine, it just looks funny.

However, a couple of tests starting with LanguageFeatures/Patterns/matching_record_A01_t01.dart would need an adjustment: They could fail in some cases because there is a value matching failure, but their description says that they are testing the matched value type against the required type of a pattern.

sgrekhov:
Updated logging and records tests. Please review

LanguageFeatures/Patterns/matching_record_A01_t01.dart
```diff
@@ -83,7 +83,7 @@ main() {
   Expect.equals("match-2", test1((3.14, name: "pi") as (num, {Object name})));
   Expect.equals("match-3", test1((value: 42, name: "answer")));
   Expect.equals("match-3",
-      test1((value: 42, name: "answer") as ({int value, Object name})));
+  test1((value: 42, name: "answer") as ({int value, Object name})));
```
eernstg: Oops, I'm afraid the indentation was actually better before (lines 86, 100, 114).

LanguageFeatures/Patterns/matching_record_A01_t01.dart
```diff
@@ -33,42 +33,42 @@ import "../../Utils/expect.dart";
...
+    (_, name: String _) => "match-2",
+    (value: int _, name: String _) => "match-3",
+    _ => "no match"
+  };
 }
 
 void test4(dynamic o) {
   var (int v1, num _) = o;
-}
+  }
```
eernstg: Indented by accident?

LanguageFeatures/Patterns/matching_object_A02_t01.dart
eernstg: It's considerably more readable now, thanks!

LanguageFeatures/Patterns/matching_record_A01_t01.dart
eernstg: Thanks again, looks good now!

eernstg (approved):
LGTM!

I fixed a couple of indentation issues in an extra commit (to avoid one more iteration), and I'll land it now.

LanguageFeatures/Patterns/matching_record_A01_t01.dart
eernstg: I fixed these.

LanguageFeatures/Patterns/matching_record_A01_t01.dart
eernstg: I fixed this one, too.

PR #1936, opened 2023-03-14
Fixes #1930. Make runtime type of the collection not assignable to the tested pattern

### LINKED ISSUE #1930: LanguageFeatures/Patterns/execution_pattern_for_in_element_A01_t03.dart
There are three expected errors in this test that I'm not sure should be errors.

https://github.com/dart-lang/co19/blob/11b06deb80243095ffd0dfc338623b646dcb7c36/LanguageFeatures/Patterns/execution_pattern_for_in_element_A01_t03.dart#L42

https://github.com/dart-lang/co19/blob/11b06deb80243095ffd0dfc338623b646dcb7c36/LanguageFeatures/Patterns/execution_pattern_for_in_element_A01_t03.dart#L61

https://github.com/dart-lang/co19/blob/11b06deb80243095ffd0dfc338623b646dcb7c36/LanguageFeatures/Patterns/execution_pattern_for_in_element_A01_t03.dart#L67

Each of these expect there to be a runtime error when destructuring but I think in these cases they all get runtime casts inserted that pass without issue. All the `num`s are in fact `int`s at runtime.

The other expectations in the file do have errors because the difference between the `num`s and `int`s are embedded into the reified generic type arguments so the casts do fail at runtime. For example casting a `List<num>` as a `List<int>`.

@eernstg Could you confirm?

sgrekhov:


eernstg (approved):
LGTM

PR #1943, opened 2023-03-16
Fixes #1939. Missed line added to variable_declaration_A04_t03.dart

### LINKED ISSUE #1939: Missing line in a patterns test
In the test . . . the last `for` statement doesn't follow the same principle as all of the previous `for` statements above it: setting the variable `visited` to `true` at the end of the loop body. It causes the test to fail at runtime.

https://github.com/dart-lang/co19/blob/f6f980bd5e2c26085b0394c66fb3665a91802783/LanguageFeatures/Patterns/variable_declaration_A04_t03.dart#L79-L85

sgrekhov:


eernstg (approved):
LGTM

PR #1948, opened 2023-03-20
Fixes #1946. Duplicated test removed

### LINKED ISSUE #1946: LanguageFeatures/Patterns/matching_record_A01_t02
This test matches `(1, 2)` against pattern `(c1, c2)` where c1 and c2 have a logging `operator==` and doesn't expect anything to be logged:

https://github.com/dart-lang/co19/blob/3f7c6897e55910b3d42e74cb83bc6b441afca16c/LanguageFeatures/Patterns/matching_record_A01_t02.dart#L99-L100

https://github.com/dart-lang/co19/blob/3f7c6897e55910b3d42e74cb83bc6b441afca16c/LanguageFeatures/Patterns/matching_record_A01_t02.dart#L76

https://github.com/dart-lang/co19/blob/3f7c6897e55910b3d42e74cb83bc6b441afca16c/LanguageFeatures/Patterns/matching_record_A01_t02.dart#L41-L42

A matched value has a static type `Object`, so it should be type checked against `(Object?, Object?)` before proceeding with matching field sub-patterns. That type check doesn't reject `(1, 2)`.

https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#type-checking-and-pattern-required-type

This problem also appears in other patterns in `test2` and `test3` functions of this test.

sgrekhov:


LanguageFeatures/Patterns/matching_record_A01_t02.dart
```diff
@@ -63,9 +63,9 @@ const c4 = C(4, logger);
 
 String test1(Object o) {
   switch (o) {
-    case (C c1, C c2):
+    case (c1, c2):
```
eernstg, line 66: The run-time type of the matched object here is `(Null, int)` and the required type of the record pattern is `(Object?, Object?)`: the required type of a record pattern [_only_ tests the shape](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#:~:text=the%20matched%20value%20type.-,The%20required%20type,-of%20p%20is%20a).

The required types of the field patterns is not mentioned explicitly, but we have no reason to assume that it is different from `Object?`. The matching process will check for `theConstant == theMatchedValue`, and there is no need to perform a type check first (actually, the type check could be wrong, because `theConstant` could claim that it is `==` to absolutely anything it wants, and then we shouldn't reject `theMatchedValue` because of its type).

This means that we still won't fail the match based on the required type as stated in the description.

I think the test will have to rely on different shapes, and on non-records, that is, stuff like `test1(true)`, `test1(())`, `test1((notNameAnything: 5,)`.

At the same time, this sounds like a kind of test which is already covered. Do you think this is just a duplicate of some existing tests? I know it's not very convenient to search through all tests, but perhaps a quick grep for 'shape' could reveal something?

eernstg (commented):
I'm afraid this one still doesn't test precisely what's stated in the description.

sgrekhov:
@eernstg https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Patterns/matching_record_A01_t01.dart tests record's shape. I'll delete this test then, if you don't mind

eernstg:
OK, sounds good! It's very useful to keep an eye on potential duplication.

eernstg (approved):
LGTM

Issue #1874, opened 2023-02-21
[Class modifiers] The `Function` class will be final in Dart 3.0


lrhn:
In pre-3.0 code the `Function` class can be implemented, extended and (probably) mixed in, it's just ignored and has absolutely no effect.
In 3.0 code, the `Function` class will have a `final` modifier which prevents all of the above.

The following tests seem to be affected by that (according to a `grep`):
```
Language/Expressions/Method_Invocation/Ordinary_Invocation/accessible_instance_member_t07.dart:class A extends Function {
Language/Expressions/Method_Invocation/Ordinary_Invocation/accessible_instance_member_t08.dart:class A extends Function {
```

If there are more examples (didn't find any `implements`?) or ones that go through indirections, say something like `implements IdAlias<Function>` with `typedef IdAlias<T> = T;`, then they might also need fixing.

sgrekhov:
@mit-mit the log is unavailable for me. Could you, please, provide a link to the gerrit CL?

mit-mit:
Sorry about that. It's here: https://dart-review.googlesource.com/c/sdk/+/284185?tab=checks

And the test results: https://ci.chromium.org/ui/p/dart/builders/try/analyzer-linux-release-try/63852/overview

sgrekhov:
Now I see. Thank you!

sgrekhov:
If you don't mind I'll fix if when CL is landed

sgrekhov:
Fixed by ba5ea0ef34d3f67b169b4b30ed2a0a5f1a354cef and cc94e132caaad644c618d13b77ec2c27e707d263

PR #1960, opened 2023-03-23
#1874. Update LanguageFeatures/nnbd/weak/overriding tests to work with class modifiers

sgrekhov:
This simple change enables a large bunch of tests to work with class modifiers.


eernstg (approved):
LGTM

PR #1961, opened 2023-03-23
Fixes #1937. Add `@dart=2.19` to class implementing `LinkedHashSet`

### LINKED ISSUE #1937: Failures on Add more class modifiers to dart:collection.
There are new test failures on [Add more class modifiers to dart:collection.](https://dart-review.googlesource.com/c/sdk/+/288240).

The tests
```
co19/LanguageFeatures/Set-literals/exact_types_of_literals_A01_t01 MissingCompileTimeError (expected Pass)
```
are failing on configurations
```
analyzer-asserts-linux
cfe-strong-linux
cfe-weak-linux
```

The `LinkedHashSet` class is (attempted to be) made `final`, but this test tries to `implement` the type.

If this lands, the test needs to be updated.

sgrekhov:
I don't see any other ways how to fix this

### DIFF
--- a/LanguageFeatures/Set-literals/exact_types_of_literals_A01_t01.dart
+++ b/LanguageFeatures/Set-literals/exact_types_of_literals_A01_t01.dart
@@ -13,6 +13,8 @@
 /// LinkedHashSet<T>
 /// @author sgrekhov@unipro.ru
 
+// @dart=2.19
+
 import "dart:collection";
 
 class C<T> with SetMixin<T> implements LinkedHashSet<T> {


eernstg:
Makes sense!

I don't know about other classes related to `LinkedHashSet`, but given that it's made final the only possible reopening is that a subtype of `LinkedHashSet` could remove `interface` (that is, it would have the modifier `base`). This would enable something like `class C<T> extends TheReopeningSubtypeOfLinkedHashSet<T> with SetMixin<T>`.

However, reopening is not that nice anyway so I wouldn't expect such a subtype to exist. So it's fine to say that this test just can't become a 3.0 test, it's living in the past (as long as that past is alive ;-).

eernstg (approved):
LGTM

PR #1965, opened 2023-03-23
Fixes #1964. Split matching_map_A01_t01.dart into two tests. The second one is for sound nullability only

### LINKED ISSUE #1964: co19/LanguageFeatures/Patterns/matching_map_A01_t01 fails in unsound mode
It assumes that `int?` cannor be assigned to `int` but it isn't so in unsound mode

sgrekhov:


eernstg (approved):
LGTM.

(There will be a need to adjust these because they use `{...}` as a map pattern, but I assume that's in another PR).

PR #1967, opened 2023-03-24
#1401. If- and for-elements tests for maps and sets added

sgrekhov:


LanguageFeatures/Patterns/execution_pattern_if_case_element_A01_t08.dart
```diff
@@ -0,0 +1,70 @@
...
+  Expect.setEquals({0, 2}, s2);
+
+  log = "";
+  var s3 = {
+    0,
+    if (Square(2, logger) case Square(area: unit4) when guard(false)) 1,
+    2
+  };
+  Expect.equals("Square.area:(4==4);guard(false);", log);
+  Expect.setEquals({0, 2}, s3);
```
eernstg, line 69: Is the `s3` test case an duplicate of the `s2` test case? I can't spot the difference.

LanguageFeatures/Patterns/execution_pattern_if_case_element_A03_t03.dart
```diff
@@ -0,0 +1,42 @@
...
+///     c. Else, evaluate the else element if there is one and yield the result
+///       into the collection.
+///   ii. Else there is no guard clause. Evaluate the then element and yield the
+///     result into the collection.
+/// 4. Else the match failed. Evaluate the else element if there is one and
+///   yield the result into the collection.
+///
+/// @description Check that it is a compile-time error if the value yielded by
+/// pattern-if-case element is not assignable to the collection element type
+/// @author sgrekhov22@gmail.com. Test a set literal
```
eernstg, line 22: I can't see any set literals here?

eernstg (commented):
I think LanguageFeatures/Patterns/execution_pattern_if_case_element_A03_t03.dart needs to be adjusted (looks like it is testing one thing, but the `@description` describes another thing, and the whole test is probably a duplicate of something which already exists).

A handful of other things: I think a particular outcome which is similar to the Fibonacci sequence needs a comment (this affects 5-6 libraries, I think). It's good, though, that we test the case where one key/value pair is _overwritten_ by another one in the same collection literal. We might very well not have covered that otherwise, so please don't change that. ;-)

LanguageFeatures/Patterns/execution_pattern_if_case_element_A03_t03.dart
sgrekhov: Fixed. Thanx!

LanguageFeatures/Patterns/execution_pattern_if_case_element_A01_t08.dart
sgrekhov: Indeed. Deleted a copy

eernstg (approved):
LGTM!

PR #1974, opened 2023-03-29
#1401. Exhaustiveness tests added

sgrekhov:


eernstg (commented):
Just a couple of things to consider.

We'd need more testing about exhaustiveness: This set of tests is basically testing whether exhaustiveness is required in the correct situations, and that's fine, but we would also need some tests exploring whether exhaustiveness exists or not with a given set of cases.

eernstg (approved):
LGTM

sgrekhov:
Tests updated according to the review. Tests in this PR are not all tests regarding the exhaustiveness. They covers only this [part of the spec](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/feature-specification.md#exhaustiveness-and-reachability). I'm working now on the set of the tests that should cover https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/exhaustiveness.md

eernstg:
> They cover only

Yes, thought so! ;-)

> .../exhaustiveness.md

Very good! That document needs a bit of interpretation, but we'll just look into that when it comes up.

PR #1977, opened 2023-03-30
Fixes #1976. Update map pattern tests according to the new matching rules

### LINKED ISSUE #1976: Update co19 tests to reflect the changed rule about map pattern matching
The language team decided that map pattern matching should be 'loose' rather than 'strict', in the sense that there is no support for the rest element in a map pattern (such as the `...` in `{"key1": value1, ...}`), and map pattern matching will allow key/value pairs to exist even though they are not matched (that is, `{"key1": value1}` now works in the same way as `{"key1": value1, ...}` used to work).

The co19 tests need to be updated correspondingly: Map patterns with `...` should be a compile-time error, and existing tests using map patterns with `...` should be updated to not have that, and they should expect the same behavior as they did before this change (with `...` and with the old semantics). Tests about map patterns where `...` did not occur should behave the same if a `when` clause is added where the length is tested. And tests about map patterns where `...` did not occur should no longer use `length` to verify that there are no non-matched key-value pairs (they may still call `length` in order to verify that there are _enough_ key/value pairs to satisfy the given set of field patterns).

sgrekhov:


LanguageFeatures/Patterns/variable_declaration_A01_t01.dart
```diff
@@ -54,7 +54,7 @@ main() {
 // [analyzer] unspecified
 // [cfe] unspecified
 
-  Map<String, num> {"pi": a4, ...} = {"pi": 3.14, "e": 2.71};
+  Map<String, num> {"pi": a4} = {"pi": 3.14, "e": 2.71};
 //                 ^^^^^^^^^^^^^^^
```
eernstg: Presumably the placement of '^^^^' is ignored for 'unspecified' errors, but we might want to delete those extra ^'s because that's the style otherwise.

eernstg (approved):
LGTM

I removed a couple of `^^^`, just so the carets are aligned with the error location. (Approximately, because the location of `^^^` ignored for 'unspecified' errors, but just so we follow a consistent style).

LanguageFeatures/Patterns/variable_declaration_A01_t01.dart
sgrekhov: Thank you!

eernstg (approved):
LGTM.

It would be really nice if we could easily see each construct which is intended to be a syntax error (the spec parser rejects many parts of this PR, and it isn't easy to see that it is all about intentional syntax errors).

I adjusted a couple of `^^^^^` lines.

LanguageFeatures/Patterns/variable_declaration_A01_t01.dart
eernstg: Thanks! ;-)

PR #1987, opened 2023-04-04
#1401. Exhaustiveness algorithm tests

sgrekhov:
@eernstg this PR needs some comments. First, exhaustiveness algorithm described there is complicated and not all of its parts/steps can be tested. So, I tried to test what it is possible to test. Some trivial cases are not tested because they are already covered by `Patterns/exhaustiveness_A*`. 

Tests in this PR can be separated in two groups. Test whose names names started with `lifting_*` are testing [Lifting types and patterns to space unions](https://github.com/dart-lang/language/blob/master/accepted/future-releases/0546-patterns/exhaustiveness.md#lifting-types-and-patterns-to-space-unions) where possible. I was not able to invent tests for some cases. 

Tests with names `exhaustiveness_*` are testing algorithm itself. Again, not all patterns checked, partially to avoid duplication, partially because I didn't find anything that can be added to already existing checks in  `Patterns/exhaustiveness_A*`. Assertion text in this group of tests is not copied from spec, as usual, but written by me to clarify what we are testing here. So, if we need more detailed assertion, please let me know.

Waiting for your review/suggestions

### DIFF
--- a/LanguageFeatures/Patterns/Exhaustiveness/exhaustiveness_map_A01_t01.dart
+++ b/LanguageFeatures/Patterns/Exhaustiveness/exhaustiveness_map_A01_t01.dart
@@ -1,0 +1,43 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Switch statements and expressions with map as a matched type can
+/// never be exhaustive
+///
+/// @description Check that map pattern cannot be exhaustive
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+String test1(Map<bool, bool> m) =>
+  switch (m) {
+//^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    {true: true} => "case-1",
+    {true: false} => "case-2",
+    {false: true} => "case-3",
+    {false: false} => "case-4",
+  };
+
+String test2(Map<bool, bool> m) {
+//     ^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  switch (m) {
+    case {true: true}:
+      return "case-1";
+    case {true: false}:
+      return "case-2";
+    case {false: true}:
+      return "case-3";
+    case {false: false}:
+      return "case-4";
+  }
+}
+
+main() {
+  print(test1({true: false}));
+  print(test2({true: false}));
+}
--- a/LanguageFeatures/Patterns/Exhaustiveness/guard_A01_t02.dart
+++ b/LanguageFeatures/Patterns/Exhaustiveness/guard_A01_t02.dart
@@ -1,0 +1,87 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion To tell if the set of cases in a switch statement or expression
+/// are exhaustive over the matched value type:
+/// 1. Lift the matched value type to a space union value.
+/// 2. Discard any cases that have guards. Since static analysis can't tell when
+///   a guard might evaluate to false, any case with a guard doesn't reliably
+///   match values and so can't help prove exhaustiveness.
+/// 3. Lift the remaining case patterns to a set of space unions cases.
+/// 4. The switch is exhaustive if is-exhaustive with value and cases is true
+///   and false otherwise.
+///
+/// @description Check that any cases with guards are ignored when calculating
+/// an exhaustiveness
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns,class-modifiers
+
+import "exhaustiveness_lib.dart";
+import "../../../Utils/expect.dart";
+
+String testBool1(bool b) {
+  switch (b) {
+    case true:
+      return "true";
+    case false when b == false:
+      return "false1";
+    case false:
+      return "false2";
+  }
+}
+
+String testBool2(bool b) =>
+  switch (b) {
+    true when b == true => "true1",
+    true => "true2",
+    false =>  "false"
+  };
+
+String testEnum1(Suit suit) {
+  switch (suit) {
+    case Suit.club:
+    case Suit.diamond:
+    case Suit.heart when suit == Suit.heart:
+    case Suit.heart:
+    case Suit.spade:
+      return "exhaustive";
+  }
+}
+
+String testEnum2(Suit suit) =>
+  switch (suit) {
+    Suit.club => "club",
+    Suit.diamond => "diamond",
+    Suit.heart when suit == Suit.heart => "heart1",
+    Suit.heart => "heart2",
+    Suit.spade => "spade"
+  };
+
+String testSealed1(Card card) {
+  switch (card) {
+    case Pip _ when true:
+      return "Pip1";
+    case Pip _:
+      return "Pip2";
+    case Face _:
+      return "Face";
+  }
+}
+
+String testSealed2(Card card) =>
+  switch (card) {
+    Pip _ => "Pip",
+    Face _ when true => "Face1",
+    Face _ => "Face2"
+  };
+
+main() {
+  Expect.equals("false1", testBool1(false));
+  Expect.equals("false", testBool2(false));
+  Expect.equals("exhaustive", testEnum1(Suit.club));
+  Expect.equals("heart1", testEnum2(Suit.heart));
+  Expect.equals("Face", testSealed1(Jack(Suit.club)));
+  Expect.equals("Face1", testSealed2(Jack(Suit.club)));
+}
--- a/LanguageFeatures/Patterns/Exhaustiveness/lifting_cast_pattern_A01_t01.dart
+++ b/LanguageFeatures/Patterns/Exhaustiveness/lifting_cast_pattern_A01_t01.dart
@@ -1,0 +1,43 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion The lifted space union for a pattern with matched value type M is
+/// ...
+/// Cast pattern:
+/// The space union spaces for a cast pattern with cast type C is a union of:
+/// - The lifted space union of the cast's subpattern in context C.
+/// - For each space E in the expanded spaces of M:
+///   a. If E is not a subset of C and C is not a subset of M, then the lifted
+///     space union of E.
+///
+/// @description Check a lifted space of a cast pattern in case of not sealed
+/// type
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+int test1(Object obj) {
+//  ^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  switch (obj) {
+    case int(isEven: true) as int:
+      return 1;
+    case int _:
+      return 2;
+  }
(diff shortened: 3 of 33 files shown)


eernstg:
Oh, great! I thought we might not have this kind of test before the cut-off date.

I understand that the algorithm is complicated, and also that not all moving parts are observable, so these difficulties are just as I'd expect. I've had a long meeting with @johnniwinther today to clarify various parts of this feature.

LanguageFeatures/Patterns/Exhaustiveness/exhaustiveness_map_A01_t01.dart
```diff
@@ -0,0 +1,43 @@
...
+
+String test1(Map<bool, bool> m) =>
+  switch (m) {
+//^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    {true: true} => "case-1",
+    {true: false} => "case-2",
+    {false: true} => "case-3",
+    {false: false} => "case-4",
```
eernstg, line 21: OK, this is not an exhaustive list of cases.

But you could say that we didn't even try: An empty map will obviously not match any of these. A map that only has `true` as a key or `false` as a key should be matched, and so should a map that has both.

The problem is that we can't specify the empty map without using a guard, and a guarded case is always ignored when we compute exhaustiveness.

So you might say that it is trivially true that a map pattern cannot be exhaustive, because it will never specify the empty map.

Hmm, what about `Map<bool, bool>(length: <= 0)`? That one doesn't have a guard, so it should be accessible for the exhaustiveness algorithm, but I don't think we can expect the algorithm to take that into account. Still, we could include `Map<bool, bool>(length: <= 0) => "case-0",` just to try it out.

LanguageFeatures/Patterns/Exhaustiveness/exhaustiveness_map_A01_t01.dart
```diff
@@ -0,0 +1,43 @@
...
+    case {true: true}:
+      return "case-1";
+    case {true: false}:
+      return "case-2";
+    case {false: true}:
+      return "case-3";
+    case {false: false}:
+      return "case-4";
+  }
+}
```
eernstg, line 38: Presumably this is a again the "can't return null" problem, a comment would be nice.

LanguageFeatures/Patterns/Exhaustiveness/guard_A01_t02.dart
```diff
@@ -0,0 +1,87 @@
...
+/// 1. Lift the matched value type to a space union value.
+/// 2. Discard any cases that have guards. Since static analysis can't tell when
+///   a guard might evaluate to false, any case with a guard doesn't reliably
+///   match values and so can't help prove exhaustiveness.
+/// 3. Lift the remaining case patterns to a set of space unions cases.
+/// 4. The switch is exhaustive if is-exhaustive with value and cases is true
+///   and false otherwise.
+///
+/// @description Check that any cases with guards are ignored when calculating
+/// an exhaustiveness
```
eernstg, line 16: Perhaps say that exhaustiveness can be recognized based on the cases without guards, even in the case where cases with guards exist as well.

LanguageFeatures/Patterns/Exhaustiveness/lifting_cast_pattern_A01_t01.dart
```diff
@@ -0,0 +1,43 @@
...
+//  ^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  switch (obj) {
+    case int(isEven: true) as int:
+      return 1;
+    case int _:
+      return 2;
+  }
+}
```
eernstg, line 30: Comment on "can't return null" as usual.

LanguageFeatures/Patterns/Exhaustiveness/lifting_cast_pattern_A01_t01.dart
```diff
@@ -0,0 +1,43 @@
...
+/// ...
+/// Cast pattern:
+/// The space union spaces for a cast pattern with cast type C is a union of:
+/// - The lifted space union of the cast's subpattern in context C.
+/// - For each space E in the expanded spaces of M:
+///   a. If E is not a subset of C and C is not a subset of M, then the lifted
+///     space union of E.
+///
+/// @description Check a lifted space of a cast pattern in case of not sealed
+/// type
```
eernstg, line 15: The point would be that these switches _are_ actually exhaustive, but the exhaustiveness algorithm can't see it because it has no space modeling "everything which is not `int`".

Could the description say something like that, in addition to what it says currently?

LanguageFeatures/Patterns/Exhaustiveness/lifting_cast_pattern_A01_t02.dart
```diff
@@ -0,0 +1,58 @@
...
+sealed class A {
+  final int field;
+  A(this.field);
+}
+class B extends A {
+  B(int field) : super(field);
+}
+class C extends A {
+  C(int field) : super(field);
+}
```
eernstg, line 31: I think we should have a `class BC extends B implements C {}` in order to see that nothing breaks if we're calling the test methods with a `BC`.

LanguageFeatures/Patterns/Exhaustiveness/lifting_constant_pattern_A01_t01.dart
```diff
@@ -0,0 +1,41 @@
...
+      return "true";
+    case false:
+      return "false";
+  }
+}
+
+String test2(bool o) => switch (o) {
+      true => "true",
+      false => "false"
+    };
```
eernstg, line 34: Indentation?

LanguageFeatures/Patterns/Exhaustiveness/lifting_constant_pattern_A02_t01.dart
```diff
@@ -0,0 +1,52 @@
...
+// SharedOptions=--enable-experiment=patterns
+
+class C {
+  final bool v;
+  const C(this.v);
+
+  bool operator ==(covariant bool other) => v == other;
+}
+
+const True = C(true);
```
eernstg, line 26: I think it's confusing that these constant variables have capitalized names. Perhaps `_true` and `_false` could be used, such that we don't violate basic expectations about capital letters?

LanguageFeatures/Patterns/Exhaustiveness/lifting_identifier_pattern_A01_t01.dart
```diff
@@ -0,0 +1,40 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion The lifted space union for a pattern with matched value type M is
+/// ...
+/// Variable pattern or identifier pattern: The lifted space union of the static
+/// type of the corresponding variable.
+///
+/// @description Check that a lifted space for an identifier pattern is a union
+/// of the static type of the corresponding variable.
```
eernstg, line 11: I believe that's not true (the spec should be adjusted here): In a matching context, an identifier pattern is a reference to a constant variable, and the space for that pattern is a pattern with this type _and_ a constant constraint denoting that particular constant (somehow, perhaps just by means of the variable itself).

However, we can't really observe the space anyway.

I think the test is OK, but we should note that the description should be different, and the spec should be adjusted.

LanguageFeatures/Patterns/Exhaustiveness/lifting_null_check_pattern_A01_t01.dart
```diff
@@ -0,0 +1,40 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+String test1(Object? o) {
+  switch (o) {
+//^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    case Object _?:
```
eernstg, line 25: Perhaps
```suggestion
    case Object? _?:
```
such that the subpattern does not prevent exhaustiveness, it is only non-exhaustive because of the null-check pattern.

LanguageFeatures/Patterns/Exhaustiveness/lifting_null_check_pattern_A01_t02.dart
```diff
@@ -0,0 +1,45 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=patterns
+
+import "exhaustiveness_lib.dart";
+import "../../../Utils/expect.dart";
+
+String test1(Card? c) {
+  switch (c) {
+    case Pip _?:
```
eernstg, line 25:
```suggestion
    case Pip? _?:
```
again, to ensure that it is the null-check pattern that has this effect, not the subpattern.

LanguageFeatures/Patterns/Exhaustiveness/lifting_null_check_pattern_A01_t02.dart
```diff
@@ -0,0 +1,45 @@
...
+      return "Pipe";
```
eernstg, line 26: Is there a point in using "Pipe" rather than "Pip"?

eernstg (approved):
I'm landing this now. I think it's better to create a new PR making various adjustments based on the review comments, because all those adjustments consist of adding a few new cases or changing something that doesn't change the nature of the test (e.g., changing `@description`).

So we can land this immediately, and then land an adjusted version later on. This allows the tests added here to be taken into account as soon as they possibly can be taken into account, and then it's just fine that there can be a later PR that adds a few more cases etc.

LanguageFeatures/Patterns/Exhaustiveness/exhaustiveness_map_A01_t01.dart
sgrekhov: I think it makes sense to add `Map<bool, bool>(length: <= 0)` as a test that will proof that lifted space union for a relational pattern is an empty space union. Thanx!

LanguageFeatures/Patterns/Exhaustiveness/lifting_cast_pattern_A01_t01.dart
sgrekhov: Waiting for https://github.com/dart-lang/co19/issues/2010#issuecomment-1498779364

LanguageFeatures/Patterns/Exhaustiveness/lifting_cast_pattern_A01_t02.dart
sgrekhov: Added

LanguageFeatures/Patterns/Exhaustiveness/lifting_constant_pattern_A02_t01.dart
sgrekhov: Renamed

LanguageFeatures/Patterns/Exhaustiveness/lifting_null_check_pattern_A01_t01.dart
sgrekhov: Fixed

LanguageFeatures/Patterns/Exhaustiveness/lifting_null_check_pattern_A01_t02.dart
sgrekhov: It was a typo, of course :)

LanguageFeatures/Patterns/Exhaustiveness/lifting_null_check_pattern_A01_t02.dart
sgrekhov: Fixed

sgrekhov (commented):
I'm preparing PR based on these notes. Just submitting my notes first, PR will be ready soon (I'm waiting for answer to https://github.com/dart-lang/co19/issues/2010#issuecomment-1498779364)

PR #1999, opened 2023-04-05
Fixes #1991. Runtime error fixed in matching_map_A02_t03.dart

### LINKED ISSUE #1991: LanguageFeatures/Patterns/matching_map_A02_t03
While matching

https://github.com/dart-lang/co19/blob/4564cc42803a8f2abc65f7f10b32e2ac697c69c7/LanguageFeatures/Patterns/matching_map_A02_t03.dart#L83

against

https://github.com/dart-lang/co19/blob/4564cc42803a8f2abc65f7f10b32e2ac697c69c7/LanguageFeatures/Patterns/matching_map_A02_t03.dart#L95

The `"key2"` is also queried from the map (after querying `"key1"`), so the following expectation is not correct:

https://github.com/dart-lang/co19/blob/4564cc42803a8f2abc65f7f10b32e2ac697c69c7/LanguageFeatures/Patterns/matching_map_A02_t03.dart#L97

sgrekhov:


LanguageFeatures/Patterns/matching_map_A02_t03.dart
```diff
@@ -94,10 +94,10 @@ main() {
   Expect.throws(() {
     var <String, int?>{"key1": v1, "key2": v2} = map;
   });
-  Expect.equals("[key1];containsKey(key1);", map.log);
+  Expect.equals("[key1];containsKey(key1);[key2];containsKey(key2);", map.log);
```
eernstg, line 97: Does it do this? I would expect the invocation of `["key2"]` to yield a non-null value, and we would then _not_ invoke `containsKey("key2")`.

eernstg (commented):
I think this might still need an adjustment.

LanguageFeatures/Patterns/matching_map_A02_t03.dart
sgrekhov: `map` here is
```dart
final map = MyMap<String, int?>({"key1": null, "keyX": 42});
```
So `[key2]` should return `null`. But value type of the `map` is nullable (`int?`) so I'm expecting call of `containsKey(key2)` after `[key2]`

eernstg: Oh, it was "keyX", I missed that. Thanks!

PR #2000, opened 2023-04-05
Fixes #1996. Missed error expectation added

### LINKED ISSUE #1996: Additional unmarked error in test
The test co19/LanguageFeatures/Patterns/switch_expression_A01_t01 has an extra error with no expectation (missing `,` after the first case in the first switch).

sgrekhov:


eernstg (approved):
LGTM.

We could also just add that comma, but this is already a syntax error test, so it's OK.

PR #2002, opened 2023-04-05
Fixes #1995. Use always-exhaustive type for testing spaces in switch statements

### LINKED ISSUE #1995: Some switch case exhaustiveness tests are missing expected errors.
Some tests are assuming that exhaustiveness for switch statements is available for the analysis which checks that functions always return a non-nullable value when required.  Unfortunately, this is not true for the time being, see [here](https://github.com/dart-lang/language/issues/2977).  Therefore, the following tests need to be adjusted either to expect an additional error or to provide some exit path (return or throw) after the switch.

```
co19/LanguageFeatures/Patterns/Exhaustiveness/lifting_null_assert_pattern_A01_t01
co19/LanguageFeatures/Patterns/Exhaustiveness/lifting_null_check_pattern_A01_t01
... // Under triage, more to come
```

cc @eernstg

sgrekhov:


eernstg (approved):
Leaf suggested that an additional error should be expected, or a `return` or `throw` statement should be added. The actual change here is different, so it's worth thinking a little bit about what it does.

I can see that it does eliminate the "return null" error, because the switch statement is now subject to exhaustiveness checking, and it succeeds.

I think we'll get a heads up if and when the specification and the implementations are updated to do something else, cf. https://github.com/dart-lang/language/issues/2977, so it should not be a problem that we are no more testing the situation discussed in #2977.

PR #2013, opened 2023-04-06
Fixes #2006. `Pointer` can have top types as type arguments

### LINKED ISSUE #2006: [ffi] `Pointer`s can have underspecified type arguments
* co19/LibTest/ffi/Pointer/Pointer_A02_t03

This test expects errors on type arguments for `Pointer`, but it's perfectly fine to have `Pointer`s with underspecified arguments. Someone might `cast` later before use.

* https://github.com/dart-lang/sdk/issues/51958

sgrekhov:


LibTest/ffi/Pointer/Pointer_A02_t03.dart
```diff
@@ -15,28 +14,12 @@ import 'package:ffi/ffi.dart';
...
-//        ^^^^^^^
-// [cfe] unspecified
-// [analyzer] unspecified
-
-  Pointer<Never> p5 = Pointer.fromAddress(p.address);
-//        ^^^^^
-// [cfe] unspecified
-// [analyzer] unspecified
+  try {
+    Pointer<dynamic> p1 = Pointer.fromAddress(p.address);
```
dcharkes, line 18: Should not have to be in a try.

`p.address` could just be 0x0 for the purpose of this test.

Note that the point of this test is to test that Pointer behaves exactly like the language spec, any class can have top types in their type arguments.

```dart
class Foo<T extends NativeType> {
  external factory Foo.fromAddress(int ptr);
}

void main() {
  Foo<dynamic> p1 = Foo.fromAddress(0x0);
  Foo<void> p2 = Foo.fromAddress(0x0);
  Foo<Object?> p3 = Foo.fromAddress(0x0);
  Foo<Never> p4 = Foo.fromAddress(0x0);
```

Maybe @eernstg can tell us why factory constructors are different from constructors?
 (My hunch is that dynamic and void are only allowed as a return type but not as an actual type at runtime. But why `Object?` works is a mystery to me.)

```dart
class Foo<T extends NativeType> {
  Foo();
  external factory Foo.fromAddress(int ptr);
}

void main() {
  Foo<dynamic> p1 = Foo.fromAddress(0x0);
  Foo<void> p2 = Foo.fromAddress(0x0);
  Foo<Object?> p3 = Foo.fromAddress(0x0);
  Foo<Never> p4 = Foo.fromAddress(0x0);

  Foo<dynamic>(); // not allowed
  Foo<void>(); // not allowed
  Foo<Object?>(); // not allowed
  Foo<Never>();
```

dcharkes (approved):
lgtm once @eernstg has replied.

LibTest/ffi/Pointer/Pointer_A02_t03.dart
eernstg: As far as I can see, `NativeType` would be considered as any other class for the purpose of checking the correctness of actual type arguments (one declaration is `abstract final class NativeType {}`, which makes it a subclass of `Object`, apart from whatever magic it is subject to).

So if there are no magic exceptions then `Pointer<dynamic> p1 = Pointer.fromAddress(p.address);` should infer as `Pointer<dynamic> p1 = Pointer<NativeType>.fromAddress(p.address);` which would not be an error. The type argument for the instance creation expression is selected by inference based on the constraints, in particular `T <: NativeType` from the declared bound.

The type of `p1` is super-bounded, but that is not an error.

However, passing `dynamic`, `void`, or `Object?` as a type argument to the instance creation itself (using a factory constructor or a generative constructor, that should not make a difference) would be an error, because it violates the bound (and the run-time type of an object cannot be super-bounded, that could violate soundness for the code in the body of the class declarations of that object and its superclasses).

The type `dynamic` and the type `void` are usable as types, just like `Object?` or `int`, so there is nothing special about using them as type arguments.

eernstg (approved):
LGTM, but I won't land it now because there is an ongoing discussion.

LibTest/ffi/Pointer/Pointer_A02_t03.dart
dcharkes: Thanks for the explanation @eernstg !

Happy to land this after the try is removed and 0x0 is used for the address.

sgrekhov:
Changed to use `0x0` address. Please review

dcharkes (approved):

eernstg (approved):
LGTM

PR #2014, opened 2023-04-06
Fixes #2007. Fix UintPtr_A01_t01.dart not to fail on 32-bit platforms

### LINKED ISSUE #2007: [ffi] UintPtr_A01_t01 test expectation
* LibTest/ffi/UintPtr/UintPtr_A01_t01.dart

The test fails on ia32 because it's testing for a signed `intptr_t` rather than the unsigned `uintptr_t`.

Therefore 0x8000000 is a perfectly fine value and will not come back as a negative value.

To test a boundary condition, one should test 0x100000000 and expect 0x0 as answer.

* https://github.com/dart-lang/sdk/issues/51958

sgrekhov:


dcharkes (approved):
Thanks @sgrekhov !

eernstg (approved):

PR #2015, opened 2023-04-06
Fixes #2008. Split pattern_context_A01_t01.dart test to multiple methods

### LINKED ISSUE #2008: Pattern test fails due to cascading parse errors
The test co19/LanguageFeatures/Patterns/pattern_context_A01_t01

tests a large number of illegal patterns in one method.  The CFE is failing this, I believe because it hits a parse error on the first illegal pattern, and then after some number of subsequent errors, stops reporting more cascading syntax errors.  There may be a parser recovery issue here, I'm not sure (cc @johnniwinther @stereotype441 ) but I think for the purposes of this test it might be better to split it into either multiple methods (to allow better recovery, assuming the recovery is good enough) or multiple tests.

cc @eernstg

sgrekhov:
Updated test is not failing on CFE

### DIFF
--- a/LanguageFeatures/Patterns/pattern_context_A01_t01.dart
+++ b/LanguageFeatures/Patterns/pattern_context_A01_t01.dart
@@ -18,44 +18,69 @@
 
 import "patterns_lib.dart";
 
-main() {
+test1() {
   var x1 || x1 = 42;
 //       ^
 // [analyzer] unspecified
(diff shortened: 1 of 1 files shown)


leafpetersen (approved):
LGTM, thanks!

eernstg (approved):
LGTM

PR #2016, opened 2023-04-06
Fixes #2009. Update scope_A01_t01.dart according to the current CFE behavior

### LINKED ISSUE #2009: Test expects multiple errors where CFE only reports one
The test co19/LanguageFeatures/Patterns/scope_A01_t01 expects two errors resulting from a scope collision.  The CFE only reports one, however, it seems to correctly report some error for any combination of those scope collisions, which in the opinion of the team is a reasonable way to report errors here.  See https://github.com/dart-lang/sdk/issues/51969 for discussion.  This test should be updated either to simply reflect the CFE choice of where to report errors, or to split the error expectations into multiple methods each with a single error.  cc @eernstg @johnniwinther

sgrekhov:


leafpetersen (approved):
LGTM, thanks!

eernstg (approved):
LGTM

Issue #1959, opened 2023-03-23
[Class modifiers] Add tests for class modifiers


sgrekhov:
Add tests for [class modifiers](https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md) language feature

PR #2021, opened 2023-04-14
#1959. More basic and mixin restrictions tests

sgrekhov:


LanguageFeatures/Class-modifiers/mixin_restrictions_A04_t04.dart
```diff
@@ -0,0 +1,511 @@
...
+/// - S is a mixin or mixin class declaration (necessarily from a post-feature
+///   library), or
+/// - S is a non-mixin class declaration which has Object as superclass and
+///   declares no generative constructor, and either
+///   - K is a pre-feature library, or
+///   - K is a platform library and L is a pre-feature library.
+///
+/// @description Check that it is a compile-time error if a class declaration
+/// mixes in class not declared as `mixin`. Test the case when mixed in classes
+/// defined in the same library
```
eernstg, line 16: Oops, we don't have those declarations (such as `Class` or `BaseClass`) in this library.

LanguageFeatures/Class-modifiers/mixin_restrictions_A04_t07.dart
```diff
@@ -0,0 +1,77 @@
...
+/// library L has S from library K as a declared mixin, unless:
+/// - S is a mixin or mixin class declaration (necessarily from a post-feature
+///   library), or
+/// - S is a non-mixin class declaration which has Object as superclass and
+///   declares no generative constructor, and either
+///   - K is a pre-feature library, or
+///   - K is a platform library and L is a pre-feature library.
+///
+/// @description Check that it is not an error if a class declaration mixes in
+/// class not declared as `mixin` but defined in a pre-feature library
```
eernstg, line 15:
```suggestion
/// a class not declared as `mixin`, but defined in a pre-feature library
```

LanguageFeatures/Class-modifiers/mixin_restrictions_A04_t08.dart
```diff
@@ -0,0 +1,36 @@
...
+///   - K is a platform library and L is a pre-feature library.
+///
+/// @description Check that it is not an error if a class declaration from
+/// pre-feature library mixes in a class not declared as `mixin` but defined in
+/// a platform library
+/// @author sgrekhov22@gmail.com
+
+// @dart=2.19
+
+class C1 with Sink { // abstract class
```
eernstg, line 21: `Sink` is actually an `abstract interface`, too. Perhaps `with Symbol`?

eernstg (commented):
Looks good! I think there is a 15 line typo in LanguageFeatures/Class-modifiers/mixin_restrictions_A04_t04.dart, and I mentioned one or two regular typos.

But there one change which is slightly more substantial (not much editing, but I think it makes a difference): We could test some `typedef` behaviors in the case where the type alias "leaks" the class, that is, we have something like `typedef C = _C;` such that we get to check that `C` can be used just like a type alias whose expansion is a public class.

LanguageFeatures/Class-modifiers/class_modifiers_lib.dart
```diff
@@ -25,10 +25,19 @@ abstract base mixin class AbstractBaseMixinClass {}
...
-typedef TypedefBaseClass = BaseClass;
+typedef TypedefAbstractClass = AbstractClass;
+typedef TypedefAbstractBaseClass = AbstractBaseClass;
+typedef TypedefAbstractInterfaceClass = AbstractInterfaceClass;
+typedef TypedefAbstractFinalClass = AbstractFinalClass;
+typedef TypedefMixinClass = MixinClass;
+typedef TypedefBaseMixinClass = BaseMixinClass;
+typedef TypedefAbstractBaseMixinClass = AbstractBaseMixinClass;
+typedef TypedefMixin = Mixin;
+typedef TypedefBaseMixin = BaseMixin;
```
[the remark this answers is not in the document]
sgrekhov: Type aliases are not used in the current tests now. Let's add alisases for private classes when we'll addind actual tests that restrictions cannot be subverted by `typedef`.

sgrekhov (commented):
Typos are fixed. As for type aliases for private classes I'd suggest to add them later when we'll add actual tests for typedefs. I need some time to figure out the way how to test them

PR #2029, opened 2023-04-25
#1959. Versioning tests added

sgrekhov:
Please note that `versioning_A03_t*` and `versioning_A04_t*` are failing on the edge SDK

### DIFF
--- a/LanguageFeatures/Class-modifiers/versioning_A03_t02.dart
+++ b/LanguageFeatures/Class-modifiers/versioning_A03_t02.dart
@@ -9,156 +9,419 @@
 /// ignores the base modifier.
 ///
 /// @description Checks that it is a compile-time error if a declaration from a
-/// post-future library is not marked `base`, `final` or `sealed` implements a
-/// platform library class marked `final` and the superinterface chain goes
-/// through a pre-feature library declaration
+/// post-feature library implements a platform library class marked `final` and
+/// the superinterface chain goes through a pre-feature library declaration
 /// @author sgrekhov22@gmail.com
 
 import "versioning_pre_feature_lib.dart";
 
-class ClassImplementsPreFeatureExtendsFinal implements PreFeatureExtendsFinal {}
+class ClassImplementsPreFeatureExtendsFinal implements PreFeatureExtendsFinal {
 //    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 // [analyzer] unspecified
 // [cfe] unspecified
+  dynamic noSuchMethod(Invocation i) {}
+}
 
-class ClassImplementsPreFeatureImplementsFinal implements PreFeatureImplementsFinal {}
+class ClassImplementsPreFeatureImplementsFinal implements PreFeatureImplementsFinal {
 //    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 // [analyzer] unspecified
 // [cfe] unspecified
+  dynamic noSuchMethod(Invocation i) {}
+}
 
-class ClassImplementsPreFeatureWithFinal implements PreFeatureWithFinal {}
+class ClassImplementsPreFeatureWithFinal implements PreFeatureWithFinal {
 //    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 // [analyzer] unspecified
 // [cfe] unspecified
+  dynamic noSuchMethod(Invocation i) {}
+}
 
-interface class InterfaceClassImplementsPreFeatureExtendsFinal implements PreFeatureExtendsFinal {}
+base class BaseClassImplementsPreFeatureExtendsFinal implements PreFeatureExtendsFinal {
+//         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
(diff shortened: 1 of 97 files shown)


LanguageFeatures/Class-modifiers/versioning_A04_t01.dart
```diff
@@ -0,0 +1,125 @@
...
+/// restrictions for code in post-feature libraries.
+///
+/// @description Checks that it is a compile-time error if a declaration from a
+/// post-future library extends a platform library class marked `interface` and
+/// the superinterface chain goes through a pre-feature library declaration
+/// @author sgrekhov22@gmail.com
+
+import "versioning_pre_feature_lib.dart";
+
+class ExtendsPreFeatureImplementsInterface extends PreFeatureImplementsInterface {}
```
eernstg, line 15: I don't think these should be errors.

The class `PreFeatureImplementsInterface` is actually a perfectly good class also in the new language, because `implements InterfaceClass` does not imply any constraints (you don't have to have any particular class modifiers at all on a class that `implements InterfaceClass`).

So the migration of `PreFeatureImplementsInterface` will be a no-op with respect to class modifiers, and there is no need to flag these classes as errors.

eernstg: I guess the point is that restrictions associated with `interface` aren't transitive in the first place.

eernstg (commented):
Some of these tests are obvious. For example, errors expected in LanguageFeatures/Class-modifiers/versioning_A04_t02.dart must be errors, because we can't mix in a non-mixin in new code.

However, several other libraries are quite tricky. For example, LanguageFeatures/Class-modifiers/versioning_A04_t01.dart, where `PreFeatureImplementsInterface` seems to be a perfectly OK class, in legacy code as well as new code, and hence we shouldn't expect any errors from its subtypes. `PreFeatureWithInterface` is really tricky, as explained in a comment in this library.

So please take a look, fix the cases that are clear, and complain about the vagueness of the rest as needed. ;-)

LanguageFeatures/Class-modifiers/versioning_pre_feature_lib.dart
```diff
@@ -0,0 +1,21 @@
...
+
+/// @description Library containing definitions of pre-feature classes which
+/// have modifiers in their superdeclarations
+/// @author sgrekhov22@gmail.com
+
+// @dart=2.19
+
+library versioning_pre_feature_lib;
+
+class PreFeatureExtendsFinal extends Function {} // abstract final
```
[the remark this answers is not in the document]
sgrekhov: Changed to `Runes` and `BigInt` (`Runes` cannot be used as a mixin because has a generative constructor, so I used `BigInt` for it)

sgrekhov (commented):
Seems that https://github.com/dart-lang/language/issues/3019 answered all the questions, so I updated the tests accordingly. Please review

LanguageFeatures/Class-modifiers/versioning_A03_t02.dart
```diff
@@ -0,0 +1,427 @@
...
+}
+
+mixin MixinOnPreFeatureWithFinal on PreFeatureWithFinal {
+//    ^^^^^^^^^^^^^^^^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  dynamic noSuchMethod(Invocation i) {}
+}
+
+base mixin BaseMixinOnPreFeatureExtendsFinal on PreFeatureExtendsFinal {
```
eernstg, line 354: These three cases are tricky, too.

We do have a rule that the `on` type of a mixin cannot be a `final` class of a different library, https://github.com/dart-lang/language/issues/2933, but I can't find that in the feature spec.

We have a similar rule for a `sealed` class in a different library ([here](https://github.com/dart-lang/language/blob/main/accepted/future-releases/class-modifiers/feature-specification.md#:~:text=A%20declaration%20depends%20directly%20on%20a%20sealed%20declaration%20from%20another%20library)).

But none of those rules are applicable here, so I'd expect this one plus the next two to be non-errors.

eernstg (commented):
I think we need to change three `mixin` cases in LanguageFeatures/Class-modifiers/versioning_A03_t02.dart to non-errors.

LanguageFeatures/Class-modifiers/versioning_A03_t02.dart
sgrekhov: Moved these cases to a separate test which is not expecting any errors

eernstg: As far as I can see we still have `MixinOnPreFeatureExtendsFinal`, `MixinOnPreFeatureImplementsFinal`, `MixinOnPreFeatureWithFinal` in LanguageFeatures/Class-modifiers/versioning_A03_t02.dart. Did you move some other declarations?

eernstg (commented):
I think we still have a wrong expectation that `mixin M on NastyLegacy ...` should be an error.

LanguageFeatures/Class-modifiers/versioning_A03_t02.dart
sgrekhov: No problem, I too missed that `on` relation is like an `implements` and restrictions for it are transitive.  Moved these tests to a separate test with appropriate description.

eernstg (approved):
Great, LGTM!

PR #2031, opened 2023-04-28
#1959. Grammar tests added

sgrekhov:


LanguageFeatures/Class-modifiers/grammar_A01_t01.dart
```diff
@@ -0,0 +1,49 @@
...
+///
+/// mixinClassModifiers ::= 'abstract'? 'base'? 'mixin'
+///
+/// mixinDeclaration  ::= 'base'? 'mixin' typeIdentifier typeParameters?
+///                       ('on' typeNotVoidList)? interfaces?
+///                       '{' (metadata classMemberDeclaration)* '}'
+///
+/// @description Check that it is a compile-time error if any of
+/// keywords/built-in identifiers `abstract`, `final`, `interface`, `mixin`, or
+/// `class` is used as an identifier
```
eernstg, line 23: 'keyword' is not specified, that's just a loose word which can be used when we want to give a hint that we are talking about a special word (and it could include `hide` and other words which are actually just plain identifiers, but occur somewhere in the grammar).

`abstract`, `interface`, `mixin`, are built-in identifiers.
`final`and `class` are reserved words.

So `final` and `class` cannot be an identifier at all, but `abstract`/`interface`/`mixin` can be the names of everything except types.

There is no need to test the treatment of reserved words, they have been unchanged for many years, and they have been tested before this feature. Similarly, there is no need to test `abstract` or `mixin`, that's been done earlier.

So we should test `interface` as a built-in identifier, it's a new thing that we are actually using that word in the grammar.

That's a small test, but that's OK. ;-)

LanguageFeatures/Class-modifiers/grammar_A01_t03.dart
```diff
@@ -0,0 +1,39 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=class-modifiers
+
+main() {
+  var base = 1;
+  var sealed = 2;
+  var interface = 3;
+  var abstract = 4;
+  var mixin = 5;
```
eernstg, line 32: Again, to avoid duplication that accumulates over time, I think we should omit the tests dealing with `abstract` and `mixin`.

LanguageFeatures/Class-modifiers/grammar_A02_t02.dart
```diff
@@ -0,0 +1,70 @@
...
+
+main() {
+  print(C1);
+  print(C2);
+  print(C3);
+  print(C4);
+  print(C5);
+  print(C6);
+  print(C7);
+}
```
eernstg, line 70: Cool! The parser could get confused about these situations, e.g., it could crash, and we should test this.

eernstg (commented):
Looks good! There is one issue about the classification of "keywords". And I think we have to be tough on duplication. So I commented on that.

sgrekhov:
Updated. Please rereview

eernstg (approved):
LGTM

PR #2032, opened 2023-05-02
#1401. Async for-in tests for patterns

sgrekhov:


LanguageFeatures/Patterns/execution_pattern_await_for_in_statement_A02_t01.dart
```diff
@@ -0,0 +1,147 @@
...
+/// 4. If there are no compile-time errors, then execution proceeds as the
+///   following code, where id1 and id2 are fresh identifiers:
+/// ```dart
+/// var id1 = <expression>;
+/// var id2 = id1.iterator;
+/// while (id2.moveNext()) {
+///   <keyword> <pattern> = id2.current;
+///   { <statement> }
+/// }
+/// ```
```
eernstg, line 27: I just created https://github.com/dart-lang/language/issues/3035 in order to clarify the situation: Do we or do we not support await-for constructs with a pattern declaration?

If so then there will probably be a new section in the patterns feature specification about await-for-in statements, and another one about await-for-in elements, or the existing ones about the synchronous for-in constructs (without `await`) would be generalized to mention the asynchronous ones.

In any case, the above text does not fit an asynchronous for construct of any kind. So we'd need some kind of TODO entry (in the test, or in some other location, whatever works), saying that the above spec text must be updated to cover the asynchronous constructs when that is possible.

So right now we're going to have a complex relationship between the @assertion, the @description, and the test. For instance, the rewrite to a while loop shown above will not work, and hence it doesn't really make sense to mention `I` in the @description.

I'll try to fit the pieces together anyway, and then we'll have another round of adjustments once the specification is available.

LanguageFeatures/Patterns/execution_pattern_await_for_in_statement_A02_t03.dart
```diff
@@ -0,0 +1,79 @@
...
+
+// SharedOptions=--enable-experiment=patterns,records
+
+import "../../Utils/expect.dart";
+import "patterns_lib.dart";
+
+main() async {
+  Expect.throws(() {
+    await for (var (int v1)
+        in Stream<num>.fromIterable([1, 2, 3.14]) as dynamic) {}
```
eernstg, line 41: Can we verify that the run-time error occurs because the stream is a `Stream<num>` where `num` is not assignable to `int`, and it does _not_ occur when we get the first actual element type mismatch?

For example, the body of the loop could add the elements to a log, and then we'd expect an empty log rather than a log that contains 1 and 2 (and then there was an exception with 3.14).

LanguageFeatures/Patterns/execution_pattern_await_for_in_statement_A02_t03.dart
```diff
@@ -0,0 +1,79 @@
...
+  });
+
+  Expect.throws(() {
+    await for (final <int>[v2] in Stream<List<num>>.fromIterable([
+      [1],
+      [2],
+      [3]
```
eernstg, line 48: Could make it [3.14] and then have a log.

LanguageFeatures/Patterns/execution_pattern_await_for_in_statement_A02_t03.dart
```diff
@@ -0,0 +1,79 @@
...
+  });
+
+  Expect.throws(() {
+    await for (var (n: int v5)
+        in Stream<({num n})>.fromIterable([(n: 2.1)]) as dynamic) {}
+  });
+
+  Expect.throws(() {
+    await for (var Square<Centimeter>(area: v6)
+        in Stream.fromIterable([Circle(1)]) as dynamic) {}
```
eernstg, line 71: Perhaps use an explicit type argument here, as usual, in order to have the simplest and most perspicuous form that yields a stream with the desired type argument, rather than relying on one or more additional mechanisms to choose that type argument, even implicitly.

LanguageFeatures/Patterns/execution_pattern_await_for_in_statement_A02_t03.dart
```diff
@@ -0,0 +1,79 @@
...
+      [2],
+      [3]
+    ]) as dynamic) {}
+  });
+
+  Expect.throws(() {
+    await for (var <String, int>{"k1": v3}
+        in Stream<Map<String, num>>.fromIterable([
+      {"k1": 1}
+    ]) as dynamic) {}
```
eernstg, line 56: OK, it should be fine to do the log thing above, and not do anything here. I don't think it'll test anything new to add a log and >1 elements in the stream in this case and below.

LanguageFeatures/Patterns/execution_pattern_await_for_in_statement_A02_t04.dart
```diff
@@ -0,0 +1,66 @@
...
+/// ```dart
+/// var id1 = <expression>;
+/// var id2 = id1.iterator;
+/// while (id2.moveNext()) {
+///   <keyword> <pattern> = id2.current;
+///   { <statement> }
+/// }
+/// ```
+/// @description Checks that it is a compile-time error if in an async for-in
+/// loop `I` doesn't implement `Stream<T>` and in not `dynamic`
```
eernstg, line 29:
```suggestion
/// loop `I` doesn't implement `Stream<T>` and is not `dynamic`
```

LanguageFeatures/Patterns/execution_pattern_await_for_in_statement_A03_t01.dart
```diff
@@ -0,0 +1,66 @@
...
+/// ```dart
+/// var id1 = <expression>;
+/// var id2 = id1.iterator;
+/// while (id2.moveNext()) {
+///   <keyword> <pattern> = id2.current;
+///   { <statement> }
+/// }
+/// ```
+/// @description Checks that in an async for-in loop it is a compile-time error
+/// if type check ot the `<pattern>` with matched value `E` fails
```
eernstg, line 29:
```suggestion
/// if the type check of the `<pattern>` with matched value `E` fails
```

LanguageFeatures/Patterns/execution_pattern_await_for_in_statement_A05_t01.dart
```diff
@@ -0,0 +1,69 @@
...
+//           ^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  await for ((n: int v5) in Stream.fromIterable([(n: 2)])) {}
+//           ^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  var v6 = 0;
```
eernstg, line 64: OK, so we're testing some cases where the keyword-less pattern declaration could be considered to be a pattern assignment (in which case there would be no error here because `v6` is in scope), and some cases where there is no such variable in scope.

Makes sense!

eernstg (commented):
This gives rise to a bit of pondering!

I get the impression (from the patterns feature specification, and from the open and the closed language repo issues) that `await for` with patterns have never been discussed, and there is no specification (the word `await` occurs just once in the patterns feature specification, with no connection to `for`).

So my first take on this is actually that they do not include support for patterns, and any attempt to use a pattern in an `await for` is a compile-time error.

In particular, there is no section about `await for` statements corresponding to [this section](https://github.com/dart-lang/language/blob/main/accepted/future-releases/0546-patterns/feature-specification.md#pattern-for-in-statement), and similarly nothing corresponding to [the pattern-for-in element section](https://github.com/dart-lang/language/blob/main/accepted/future-releases/0546-patterns/feature-specification.md#pattern-for-in-element).

If this is true then we're done very quickly: Just test that we get some errors for a couple of simple combinations of await for and patterns.

However, the grammar just mentions that the non-terminal `forLoopParts` is extended to allow patterns, and there is no reference to the applicability of this enhancement with await-for-in constructs. So maybe. ;-)

Checking out the current situation, I get the impression that it has been implemented and lots of things do work already.

So let's test it "as if it were specified to work", and then we can come up with alternate actions to take if it is far from working in some ways.

Looks good! A couple of typos mentioned in comments.

sgrekhov:

LanguageFeatures/Patterns/execution_pattern_await_for_in_statement_A02_t03.dart
eernstg: Got a log now, thanks!

LanguageFeatures/Patterns/execution_pattern_await_for_in_statement_A03_t01.dart
eernstg: There was one more typo where 'ot' occured. Fixed that on github.

eernstg (approved):
LGTM!

PR #2038, opened 2023-05-08
Fixes #1912. More tests for CONST_CANONICAL_TYPE added

### LINKED ISSUE #1912: Add more tests for CONST_CANONICAL_TYPE
See https://github.com/dart-lang/language/issues/2872 for the details

### LINKED ISSUE #1951: Avoid usage of fixed port numbers in `dart:io` server tests - or at least use different ones
Our test runner assumes it can run several tests in parallel. That requires tests to be indifferent on other things running, but some tests assume they can obtain global resources (e.g. bind a socket on a specific fixed port). This can lead to failures such as

```
--- Command "vm" (took 02.000546s):
DART_CONFIGURATION=DebugARM64 xcodebuild/DebugARM64/dart --sound-null-safety -Dtest_runner.configuration=vm-mac-debug-arm64 --ignore-unrecognized-flags --packages=/opt/s/w/ir/.dart_tool/package_config.json /opt/s/w/ir/tests/co19/src/LibTest/io/HttpServer/bind_A01_t02.dart

exit code:
255

stdout:
unittest-suite-wait-for-done

stderr:
Unhandled exception:
SocketException: Failed to create server socket (OS Error: Address already in use, errno = 48), address = 127.0.0.1, port = 1234
#0      _NativeSocket.bind (dart:io-patch/socket_patch.dart:1002:7)
<asynchronous suspension>
#1      test (file:///opt/s/w/ir/tests/co19/src/LibTest/io/HttpServer/bind_A01_t02.dart:25:23)
<asynchronous suspension>

--- Re-run this test:
python3 tools/test.py -n vm-mac-debug-arm64 co19/LibTest/io/HttpServer/bind_A01_t02
```

It seems there's several tests that use `1234` as fixed number to bind a port to:

* LibTest/io/HttpServer/bind_A01_t02.dart
* LibTest/io/HttpServer/bind_A05_t01.dart
* LibTest/io/HttpServer/bind_A05_t02.dart

The ladder 2 tests do not need to use a fixed port, they can use a dynamically allocated port. The former test could use a helper function to loop through a number of possible fixed ports and only fail if it cannot bind to any of them.

sgrekhov:


LanguageFeatures/nnbd/const_evaluation_A10_t02.dart
```diff
@@ -0,0 +1,49 @@
...
+  Expect.notEquals(Never, NullableObject);
+  Expect.notEquals(Never, Void);
+  Expect.notEquals(Never, FutureOr<Never>);
+
+  Expect.notEquals(Object, NullableObject);
+  Expect.notEquals(Object, Void);
+
+  Expect.notEquals(Void, NullableObject);
+
+  Expect.notEquals(NullableObject, FutureOr<Null>);
```
eernstg, line 48: We're testing the `==` operator on reified types, and this means that it does matter which operand comes first. So we should have symmetric coverage: If we're testing `A == B` then we should also test `B == A` (in both cases, expecting `false`, but that's handled by `notEquals`, we just need to know that it's all about `==`, and hence symmetry doesn't come for free, and it can have bugs).

This basically means that we won't have shorter and shorter blocks (`Null` with everyone, followed by `dynamic` with everyone except `Null`, etc. becomes `Null` with everyone, `dynamic` with everyone, etc.).

We can't possibly test all combinations of reified types that should be unequal, but we should test the pairs that are "likely to get it wrong".

We could certainly have all combinations of primitive top types plus `Object?`:

- (Object?, dynamic) // Missing
- (Object?, void) // Missing
- (dynamic, Object?) // Line 35
- (dynamic, void) // Line 36
- (void, Object?) // Line 46
- (void, dynamic) // Missing.

It is already tested that `Object` is unequal to the simple top types. We should test that `Null` is unequal to bottom types (`Never`/`X extends Never`).

It doesn't hurt to test that `Never` and `void` are different, but for those (wildly different) types we can rely on existing practice (which would have been broken a gazillion times if those weren't recognized as different). So it's perfectly fine that we just have handful of those wildly different pairs.

Finally, we should have some composite types with those those "near-equal pairs" that I just tried to outline: `List<void> != List<dynamic>`, `void Function(dynamic) != void Function(void)` (the toughest test for inequality is the case where just one component differs, whereas the toughest test for `identical` is the case where everything changes, but only in ways that are eliminated by normalization).

eernstg: OK, when this one has been adjusted to cover some extra cases, LanguageFeatures/nnbd/const_evaluation_A10_t03.dart can be adjusted to test the same type pairs wrapped in `Set<_>`. And there could be yet another test containing the cases about function types.

LanguageFeatures/nnbd/const_evaluation_A11_t01.dart
```diff
@@ -0,0 +1,60 @@
...
+
+  Expect.identical(cNullableInt, cDoubleNullableInt);
+  Expect.identical(cNull, cNullableNever);
+  Expect.identical(cNull, cNullableNull);
+  Expect.identical(cDynamic, cNullableDynamic);
+  Expect.identical(cVoid, cNullableVoid);
+  Expect.identical(cFutureOrVoid, cVoid);
+  Expect.identical(cFutureOrDynamic, cDynamic);
+  Expect.identical(cFutureOrObject, cObject);
+  Expect.identical(cFutureOrNull, cNullableFutureNull);
```
eernstg, line 59: This is fine! In spite of the fact that I've asked for more cases with the reified type identical/nonEquals tests, we don't need more of these. We can trust the general equality on reified types to be used by these tests as well, and that general equality is already being tested in earlier tests.

LanguageFeatures/nnbd/const_evaluation_A11_t01.dart
```diff
@@ -0,0 +1,60 @@
...
+/// arguments shall be considered identical if those generic type arguments
+/// compare equal using the definition of runtime type object equality defined
+/// above. That is, comparison (or canonicalization) of constant instances of
+/// generic classes is performed relative to the normal forms of their generic
+/// type arguments, and ignoring legacy type annotations as described above.
+/// Hence, an instance of C<T0> compares identical to C<T1> if T0 and T1 have
+/// the same normal form (up to the identity of bound variables), and the
+/// objects are otherwise identical.
+///
+/// @description Checks that two constant are identical if they type arguments
```
eernstg, line 23:
```suggestion
/// @description Checks that two constants are identical if their type arguments
```

LanguageFeatures/nnbd/const_evaluation_A11_t02.dart
```diff
@@ -0,0 +1,103 @@
...
+/// arguments shall be considered identical if those generic type arguments
+/// compare equal using the definition of runtime type object equality defined
+/// above. That is, comparison (or canonicalization) of constant instances of
+/// generic classes is performed relative to the normal forms of their generic
+/// type arguments, and ignoring legacy type annotations as described above.
+/// Hence, an instance of C<T0> compares identical to C<T1> if T0 and T1 have
+/// the same normal form (up to the identity of bound variables), and the
+/// objects are otherwise identical.
+///
+/// @description Checks that two constant are not equal if they type arguments
```
eernstg, line 23:
```suggestion
/// @description Checks that two constants are not equal if their type arguments
```

LanguageFeatures/nnbd/const_evaluation_A11_t02.dart
```diff
@@ -0,0 +1,103 @@
...
+  Expect.notEquals(cFutureOrVoid, cFutureOrNull);
+
+  Expect.notEquals(cFutureOrDynamic, cFutureOrObject);
+  Expect.notEquals(cFutureOrDynamic, cFutureOrNullableObject);
+  Expect.notEquals(cFutureOrDynamic, cFutureOrNull);
+
+  Expect.notEquals(cFutureOrObject, cFutureOrNullableObject);
+  Expect.notEquals(cFutureOrObject, cFutureOrNull);
+
+  Expect.notEquals(cFutureOrNullableObject, cFutureOrNull);
```
eernstg, line 102: Again, there's no need to expand this one even though LanguageFeatures/nnbd/const_evaluation_A10_t01.dart may be expanded somewhat.

LanguageFeatures/nnbd/const_evaluation_A11_t03.dart
```diff
@@ -0,0 +1,103 @@
...
+  Expect.notEquals(cFutureOrVoid, cFutureOrNull);
+
+  Expect.notEquals(cFutureOrDynamic, cFutureOrObject);
+  Expect.notEquals(cFutureOrDynamic, cFutureOrNullableObject);
+  Expect.notEquals(cFutureOrDynamic, cFutureOrNull);
+
+  Expect.notEquals(cFutureOrObject, cFutureOrNullableObject);
+  Expect.notEquals(cFutureOrObject, cFutureOrNull);
+
+  Expect.notEquals(cFutureOrNullableObject, cFutureOrNull);
```
eernstg, line 102: Fine again, thanks!

eernstg (commented):
I'm not quite sure: Perhaps some of these tests are actually the tests of `NORM`, perhaps we are already testing `NORM` in some other tests.

If a test like LanguageFeatures/nnbd/const_evaluation_A10_t01.dart is _the_ `NORM` test, then I think it needs to be expanded with some additional test cases.

If we already have some other `NORM` test then we should consider whether LanguageFeatures/nnbd/const_evaluation_A10_t01.dart is needed at all.

LanguageFeatures/nnbd/const_evaluation_A10_t01.dart
```diff
@@ -0,0 +1,41 @@
...
+main() {
+  Expect.identical(NullableInt, DoubleNullableInt);
+  Expect.identical(Null, NullableNever);
+  Expect.identical(Null, NullableNull);
+  Expect.identical(dynamic, NullableDynamic);
+  Expect.identical(Void, NullableVoid);
+  Expect.identical(FutureOr<void>, Void);
+  Expect.identical(FutureOr<dynamic>, dynamic);
+  Expect.identical(FutureOr<Object>, Object);
+  Expect.identical(FutureOr<Null>, NullableFutureNull);
```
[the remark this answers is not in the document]
sgrekhov: Done as discussed via chat

LanguageFeatures/nnbd/const_evaluation_A10_t02.dart
sgrekhov: Added

sgrekhov (commented):
The idea of this PR was that we are not testing `NORM()` here but testing of missing parts of [Const evaluation and canonicalization](https://github.com/dart-lang/language/blob/main/accepted/2.12/nnbd/feature-specification.md#const-evaluation-and-canonicalization). In particular Type literals equality (`const_evaluation_A10*`) and contants instances equality (`const_evaluation_A11*`). But, as we agreed, existing normalization tests also should be checked and, probably, adjusted. But it will be a separate PR

sgrekhov:
Ups! I submitted my changes to the wrong branch. Have to do a cherry-pick

sgrekhov:
Done. Please review

LanguageFeatures/nnbd/const_evaluation_A10_t01.dart
```diff
@@ -0,0 +1,132 @@
...
+      Map<Object?, Object>, Map<FutureOr<Object?>, FutureOr<Object>>);
+  Expect.identical(
+      Map<Future<Never>, Future<Null>?>, Map<FutureOr<Never>, FutureOr<Null>>);
+  Expect.identical(Map<FutureOr<int?>, dynamic>,
+      Map<FutureOr<DoubleNullableInt>, NullableDynamic>);
+  Expect.identical(Map<void, Object?>, Map<NullableVoid, DoubleNullableObject>);
+  Expect.identical(Map<Null, Null>, Map<NullableNever, NullableNull>);
+  Expect.identical(
+      Map<FutureOr<int?>, int?>, Map<FutureOr<int?>?, DoubleNullableInt>);
+  Expect.identical(Map<F1, F2>, Map<F1_NORM, F2_NORM>);
```
eernstg, line 109: Wouldn't the consistent ordering be this one?:
```suggestion
  Expect.identical(Map<F1_NORM, F2_NORM>, Map<F1, F2>);
```

LanguageFeatures/nnbd/const_evaluation_A10_t01.dart
```diff
@@ -0,0 +1,132 @@
...
+  Expect.identical(Map<F3, RecordPositionalVoid>,
+      Map<F3_NORM, RecordPositionalNullableVoid>);
```
eernstg, line 111: Perhaps:
```suggestion
  Expect.identical(Map<F3_NORM, RecordPositionalVoid>,
      Map<F3, RecordPositionalNullableVoid>);
```

You could argue that it yields even better coverage to have normalization that needs to change the first type argument of the first argument and the second type argument of the second argument, rather than changing both type arguments of the second argument.

Nevertheless, I think the coverage is already very good, and it's probably easier to keep track of what's happening when the ordering is the same everywhere.

LanguageFeatures/nnbd/const_evaluation_A10_t01.dart
```diff
@@ -0,0 +1,132 @@
...
+  testXExtendsT1();
+  testXExtendsT2();
+
+  // NORM(R Function<X extends B>(S)) = R1 Function<X extends B1>(S1)
+  // where R1 = NORM(R)
+  // and B1 = NORM(B)
+  // and S1 = NORM(S)
+  Expect.identical(F1, F1_NORM);
+  Expect.identical(F2, F2_NORM);
+  Expect.identical(F3, F3_NORM);
```
eernstg, line 124:
```suggestion
  Expect.identical(F1_NORM, F1);
  Expect.identical(F2_NORM, F2);
  Expect.identical(F3_NORM, F3);
```

LanguageFeatures/nnbd/const_evaluation_A10_t03.dart
```diff
@@ -0,0 +1,71 @@
...
+/// runtime type objects) are considered equal if and only if the runtime type
+/// objects T1 and T2 corresponds to the types S1 and S2 respectively, and the
+/// normal forms NORM(S1) and NORM(S2) are syntactically equal up to equivalence
+/// of bound variables and ignoring * modifiers on types.
+/// ...
+/// Two constant type literals T1 and T2 compare as identical if they are equal
+/// using the definition of runtime type equality specified above.
+///
+/// @description Checks that two constant type literals `T1` and `T2` are not
+/// equals if they are not equal according to the runtime type equality.
```
eernstg, line 15:
```suggestion
/// equal if they are not equal according to the runtime type equality.
```

eernstg (commented):
Very nice, thanks!
I've suggested a couple of typo-level changes (one thing: >80 char lines may be unavoidable when there is an error expectation which is that long, but I think it's worth the trouble to move comments around a little bit in order to avoid long lines when we can do that without breaking the testing behavior).

eernstg (commented):
Very nice, thanks!
I've suggested a couple of typo-level changes (one thing: >80 char lines may be unavoidable when there is an error expectation which is that long, but I think it's worth the trouble to move comments around a little bit in order to avoid long lines when we can do that without breaking the testing behavior).

sgrekhov:
Thanx. Done!

LanguageFeatures/nnbd/const_evaluation_A10_t01.dart
```diff
@@ -0,0 +1,132 @@
...
+  NullableNever n
+});
+
+testXExtendsT1<X extends Never>() {
+  Expect.identical(X, Never); // NORM(X extends T) = Never if T is Never
+}
+
+testXExtendsT2<X extends Y, Y extends Never>() {
+  Expect.identical(X,
+      Never); // NORM(X extends T) = Never if T is a type variable Y and NORM(Y) is Never
```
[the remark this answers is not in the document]
eernstg: Thanks!

eernstg (approved):
LGTM

PR #2040, opened 2023-05-08
Fixes #2039. Add and use asyncThrows method

### LINKED ISSUE #2039: Fix co19/LanguageFeatures/Patterns/execution_pattern_await_for_in_statement_A02_t03
This test fails at runtime because `Expect.throws(...);` doesn't catch exceptions thrown by async functions

sgrekhov:


eernstg (approved):
Aha, you already introduced a general `asyncThrows`. Great, thanks!
LGTM.

Issue #1400, opened 2022-08-19
[Extension type] Co19 tests - add tests


itsjustkevin:

sgrekhov:
@itsjustkevin @eernstg where can I find specification of  this feature?

eernstg:
Hi @sgrekhov, discussions about the language design in this area are ongoing, please delay the work on testing this feature until we have a settled design.

sgrekhov:
https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification-views.md

sgrekhov:
Completed

PR #2048, opened 2023-05-12
#1400. Inline classes syntax tests updated and missing one added

sgrekhov:


LanguageFeatures/Inline-classes/syntax_A02_t07.dart
```diff
@@ -0,0 +1,90 @@
...
+// SharedOptions=--enable-experiment=inline-class
+
+import "../../Utils/expect.dart";
+
+final inline class FIC {
+  final int id ;
+  const FIC(this.id);
+}
+
+base class BC extends FIC {
```
eernstg, line 30: This would be a compile-time error: It is not possible to create a subtype of an inline class which isn't another inline class (so it can't be a class, a mixin, or a mixin class).

eernstg: Expect errors for all the other declarations for the same reason.

LanguageFeatures/Inline-classes/syntax_A02_t07.dart
```diff
@@ -0,0 +1,90 @@
...
+/// be extended, implemented and a mixin on it can be declared in the same
+/// library where it is defined
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+import "../../Utils/expect.dart";
+
+final inline class FIC {
+  final int id ;
```
eernstg, line 26:
```suggestion
  final int id;
```

LanguageFeatures/Inline-classes/syntax_A02_t08.dart
```diff
@@ -0,0 +1,42 @@
...
+// SharedOptions=--enable-experiment=inline-class
+
+import "../../Utils/expect.dart";
+
+final inline class FIC {
+  final int id ;
+  const FIC(this.id);
+}
+
+final inline class FIC2 extends FIC {
```
eernstg, line 29: This is also a compile-time error: An inline class can not extend anything (we've had discussions about introducing that as a feature, but it is not part of the currently accepted proposal).

The point is that an inline class never has a subclass relationship to another inline class: The "superclass" has a single field, but the "subclass" doesn't inherit that field, it has to declare a new one. So the relationship is a subtype relationship, as in `implements`.

This is also a good match for the understanding of inline class methods as similar in nature to extension methods: `inline class B implements A1, .. Ak {}` is able to "inherit" members from `A1`, from `A2` etc. because that's how an extension method `on A1` or `on A2` etc. works. Also, if `A1` has `bool get m => true;` and `A2` has `void m() {}`, there is no error as long as `B` declares its own `m`. There is no requirement that `B.m` should be a correct override of `A1.m` or `A2.m`. Again, just like extension methods.

LanguageFeatures/Inline-classes/syntax_A02_t11.dart
```diff
@@ -0,0 +1,122 @@
...
+
+abstract final inline class AbstractFinalInlineClass {
+//             ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  final int id;
+  AbstractFinalInlineClass(this.id);
+}
+
+mixin inline class MixinInlineClass {
```
eernstg, line 78: Actually, `inline mixin class` should be tested, too: That one is temptingly close to something that could exist, but it should still be an error.

eernstg: Same for the 3 next cases.

eernstg (approved):
Looks good! One rule makes a different for most of at least one library, so we need to change a bunch of expectations to expect a compile-time error:

https://github.com/dart-lang/language/blob/main/accepted/future-releases/inline-classes/feature-specification.md#:~:text=A%20compile%2Dtime%20error%20occurs%20if%20an%20inline%20type%20is%20used%20as%20a%20superinterface

sgrekhov:
Updated. Please review

LanguageFeatures/Inline-classes/syntax_A02_t11.dart
```diff
@@ -0,0 +1,37 @@
...
+
+import "../../Utils/expect.dart";
+
+final inline class FIC {
+  final int id ;
+  const FIC(this.id);
+}
+
+final inline class FIC2 implements FIC {
+  int id;
```
eernstg, line 30:
```suggestion
  final int id;
```

LanguageFeatures/Inline-classes/syntax_A02_t13.dart
```diff
@@ -0,0 +1,285 @@
...
+  AbstractBaseMixinInlineClass3(this.id);
+}
+
+inline abstract base mixin class AbstractBaseMixinInlineClass4 {
+//     ^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  final int id;
+  AbstractBaseMixinInlineClass4(this.id);
+}
```
eernstg, line 253: Yep! This library yields a very nice coverage of potential parser issues (a la "oh, nobody tried _that_ before!" ;-).

eernstg (commented):
Just one location seems to be an unintended compile-time error (because an inline class has a non-final instance variable).

sgrekhov:
Typo fixed

LanguageFeatures/Inline-classes/syntax_A02_t08.dart
```diff
@@ -0,0 +1,119 @@
...
+/// non-inline class) or declare a mixin on a `final inline` class even in the
+/// same library where it is defined
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+final inline class FinalInlineClass {
+  final int x;
+  const FinalInlineClass([this.x = 0]);
+}
```
eernstg, line 26: It actually doesn't matter that the inline class is final: It's an error for any non-inline class/mixin/mixin-class to be a direct subtype of an inline class.

I think we ought to have a version of this test (and a version of the corresponding one using `implements`, and something similar for `on`) where the error arises just because the supertype is an inline class (not `final inline`), testing same-library and other-library.

Do we already have this?

LanguageFeatures/Inline-classes/syntax_A02_t11.dart
eernstg: Thanks!

eernstg (commented):
Looks good!

Just checking one thing: Do we have tests where a compile-time error is reported because an inline class has a subtype which is non-inline? It looks like we don't have it in these tests, but it might be done in some other tests that aren't being updated.

eernstg (approved):
LGTM

PR #2050, opened 2023-05-17
#1400. Static analysis of an inline member invocation tests

sgrekhov:


LanguageFeatures/Inline-classes/static_analysis_member_invocation_A02_t01.dart
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Otherwise, a compile-time error occurs if V does not have a
+/// member named m.
```
eernstg, line 6: Just to make it explicit (because it could be easy to misunderstand this point from the text right here), `V` is an inline type and we're talking about members that `V` "has" (as opposed to "declares"). So when we get to `implements AnotherInlineClass, YetAnother` this includes members in the interface of `AnotherInlineClass` plus `YetAnother`.

LanguageFeatures/Inline-classes/static_analysis_member_invocation_A03_t01.dart
```diff
@@ -0,0 +1,41 @@
...
+  IC1(this.id);
+
+  int get asInt => id as int;
+}
+
+inline class IC2<T extends num> {
+  final T id;
+  IC2(this.id);
+
+  double get asDouble => id as double;
```
eernstg, line 26: OK, that will work, but the following would perhaps be more natural, and it allows the test cases below to be written more freely:
```suggestion
  double get asDouble => id.toDouble();
```

LanguageFeatures/Inline-classes/static_analysis_member_invocation_A03_t01.dart
```diff
@@ -0,0 +1,41 @@
...
+}
+
+main() {
+  IC1 ic1 = IC1(42);
+  ic1.asInt.expectStaticType<Exactly<int>>();
+  ic1.id.expectStaticType<Exactly<num>>();
+
+  IC2 ic2 = IC2<double>(3.14);
```
eernstg, line 34: We could use `IC2<dynamic>` to get the extra benefit that we can see that the value of `T` is actually used in line 36, it isn't just blindly using the bound (tiny extra benefit: it also shows that it is possible to use a super-bounded type).

LanguageFeatures/Inline-classes/static_analysis_member_invocation_A03_t01.dart
```diff
@@ -0,0 +1,41 @@
...
+  ic2.asDouble.expectStaticType<Exactly<double>>();
+  ic2.id.expectStaticType<Exactly<num>>();
+
+  IC2<double> ic3 = IC2<double>(3.14);
```
eernstg, line 38: We might actually omit `<double>` from the initializing expression: Type inference on an inline class is potentially running some pieces of new code in the tools. The rest of the test (lines 39-40) should be unaffected because the type of `ic3` will still be `IC2<double>`, we just want to see if anything crashes when it encounters `IC2(3.14)`.

LanguageFeatures/Inline-classes/static_analysis_member_invocation_A03_t01.dart
eernstg: Ah, of course we can't do that! :smiley:

An inline type can _never_ be super-bounded, because that would be unsound during execution of code in the instance member bodies.

I think that will be tested later on, so drop the super-bounded type here.

Using `IC2<int>` can still reveal that we're getting the type from `T`, not from the bound.

LanguageFeatures/Inline-classes/static_analysis_member_invocation_A04_t01.dart
```diff
@@ -0,0 +1,39 @@
...
+  IC2(this.id);
+
+  T foo<T1, T2 extends T>() => id;
+}
+
+main() {
+  IC1 ic1 = IC1(42);
+  ic1.foo.expectStaticType<Exactly<int Function<T1, T2 extends num>()>>();
+
+  IC2 ic2 = IC2<double>(3.14);
```
eernstg, line 34: This one is interesting!

We're creating a ghost of type `IC2<double>` (there is no representation of `double` anywhere, because the ghost is an object that doesn't exist, so we have no place to store it), then we're using covariance of the type parameters to assign the ghost (actually: the representation object) to the variable `ic2`, and the type of `ic2` is `IC2<num>`.

Hence, `double` is completely ignored, and all member invocations will be told that `T == num`.

I agree that this should work exactly as shown in line 35. Cool!

eernstg: But perhaps we should use `IC2<num>` explicitly as the type of `ic2`, following the usual rule that the test should test a single thing in a very focused manner, rather than testing i2b as well as the handling of `T` in instance member invocations.

eernstg (commented):
Interesting stuff! I suggested a few changes.
One thing we do need to have is the last `Object` member: `noSuchMethod`.

LanguageFeatures/Inline-classes/static_analysis_member_invocation_A04_t01.dart
sgrekhov: Changed to `IC2<num>`

sgrekhov (commented):
Tests updated. Please review once again

LanguageFeatures/Inline-classes/static_analysis_member_invocation_A01_t01.dart
```diff
@@ -0,0 +1,26 @@
...
+  final int id;
+  IC(this.id);
+}
+
+main() {
+  IC ic = IC(42);
+  ic.toString();
+  ic.runtimeType;
+  ic.hashCode;
+  ic == ic;
```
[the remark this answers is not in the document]
eernstg: Thanks!

eernstg (approved):
LGTM, thanks!

PR #2053, opened 2023-05-18
#1400. Dynamic Semantics of an Inline Member Invocation tests

sgrekhov:


LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A01_t01.dart
```diff
@@ -0,0 +1,55 @@
...
+///
+/// Then, if args is omitted and Dm is a getter, execute the body of said
+/// getter in an environment where this and the name of the representation are
+/// bound to o, and the type variables of V are bound to the actual values of
+/// T1, .. Ts. If the body completes returning an object o2 then the invocation
+/// evaluates to o2. If the body throws an object and a stack trace then the
+/// invocation completes throwing the same object and stack trace.
+///
+/// @description Check invocation of an inline class getter in the case when
+/// type arguments are omitted. Test that if the body completes returning an
```
eernstg, line 23: I don't really understand the part about omitting the type arguments.

The text in line 5-20 talk about an invocation of `m` on receiver `e`, possibly with an argument part (`e.m<S1..Sk>(a1, .. am)` or `e.m(a1, .. am)`).

We evaluate `e` and get `o`, and then, if the invocation is `e.m` and `m` is a getter, we evaluate the getter body with the representation variable (`id` below) bound to `o2`, and that's the result.

During the evaluation of the getter body, the type variables of the inline class (`T` below) are bound to the type arguments of the receiver type in the invocation (so we need to know what that is, hence the suggestion that it should be specified explicitly below).

We should test that the body is executed with `T` bound to that value. E.g., `return <T>[];` could do it. `id + 1 as T` does depend on `T`, but it isn't so easy to see that `T` has the correct value.

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A01_t02.dart
```diff
@@ -0,0 +1,54 @@
...
+  T get testMe {
+    if (2 > 1) {
+      Error.throwWithStackTrace("I'm inline", st);
+    }
+    return id;
+  }
+}
+
+main() {
+  IC ic1 = IC(42);
```
eernstg, line 47: It would be useful to have an `IC<num>` case and an `IC<int>` case, and throw something that reveals the value of `T`.

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A02_t01.dart
```diff
@@ -0,0 +1,55 @@
...
+/// Evaluation of this invocation proceeds by evaluating e to an object o.
+/// ...
+/// Otherwise, if args is omitted and Dm is a method, the invocation evaluates
+/// to a closurization of Dm where this and the name of the representation are
+/// bound to o, and the type variables of V are bound to the actual values of
+/// T1, .. Ts. The operator == of the closurization returns true if and only if
+/// the operand is the same object.
+///
+/// @description Check invocation of an inline class method in the case when
+/// type arguments are omitted.
```
eernstg, line 22: In this case it makes sense to omit the type arguments, if `getId` is declared as `T getId<Some, Type, Parameters extends SomeBound>()` and it is called as `e.getId()`.

But the text line 15-19 is talking about an invocation where `args` is omitted, that is, `e.getId`, and the result should be a function object with a suitable type and behavior.

Again, the actual type arguments of the inline class should be specified explicitly, and the behavior of the method should reveal that the body of the method is executed with `T` bound to the value from the static type of the receiver.

For example, `IC<int> ic1_1 ...`, and `Function f1_1 = ic1_1.getId;` and `f1_1() == int` if `getId` returns `T`, or something like that.

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A02_t02.dart
```diff
@@ -0,0 +1,42 @@
...
+///
+/// Evaluation of this invocation proceeds by evaluating e to an object o.
+/// ...
+/// Otherwise, if args is omitted and Dm is a method, the invocation evaluates
+/// to a closurization of Dm where this and the name of the representation are
+/// bound to o, and the type variables of V are bound to the actual values of
+/// T1, .. Ts. The operator == of the closurization returns true if and only if
+/// the operand is the same object.
+///
+/// @description Check that `this` is bound to `o`
```
eernstg, line 21: I'm not sure how the test below could be testing this. I would expect something like `Object test() => this;` and then expect `identical(ic1.test(), ic1)`.

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A02_t03.dart
```diff
@@ -0,0 +1,60 @@
...
+///
+/// Evaluation of this invocation proceeds by evaluating e to an object o.
+/// ...
+/// Otherwise, if args is omitted and Dm is a method, the invocation evaluates
+/// to a closurization of Dm where this and the name of the representation are
+/// bound to o, and the type variables of V are bound to the actual values of
+/// T1, .. Ts. The operator == of the closurization returns true if and only if
+/// the operand is the same object.
+///
+/// @description Check that the operator `==`of the closurization returns true
```
eernstg, line 21:
```suggestion
/// @description Check that the operator `==` of the closurization returns true
```

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A03_t01.dart
```diff
@@ -0,0 +1,56 @@
...
+/// it executes the body of Dm in an environment where this and the name of the
+/// representation are bound to o, the type variables of V are bound to the
+/// actual values of T1, .. Ts, and the formal parameters of m are bound to
+/// args1 in the same way that they would be bound for a normal function call.
+/// If the body completes returning an object o2 then the invocation evaluates
+/// to o2. If the body throws an object and a stack trace then the invocation
+/// completes throwing the same object and stack trace.
+///
+/// @description Check that invocation of an inline class method in the case when
+/// type arguments are specified.
```
eernstg, line 26: I think we should have some value arguments as well, such that we can see that they are passed correctly (that is, `ic1.foo<String, double>('Hello', 3.14)`). It would be good if the behavior of the method reveals the value of the type arguments of the inline type as well as the type arguments of the method and the value arguments of the method.

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A03_t02.dart
```diff
@@ -0,0 +1,46 @@
...
+/// invocation proceeds to evaluate args to an actual argument list args1. Then
+/// it executes the body of Dm in an environment where this and the name of the
+/// representation are bound to o, the type variables of V are bound to the
+/// actual values of T1, .. Ts, and the formal parameters of m are bound to
+/// args1 in the same way that they would be bound for a normal function call.
+/// If the body completes returning an object o2 then the invocation evaluates
+/// to o2. If the body throws an object and a stack trace then the invocation
+/// completes throwing the same object and stack trace.
+///
+/// @description Check that `this` is bound to `o`
```
eernstg, line 25: Again, I can't see how a test about the value of `this` can be written without containing an evaluation of `this`. If `this` is returned then the returned value can be compared to the object which was the receiver of the invocation, they should be identical.

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A03_t03.dart
```diff
@@ -0,0 +1,55 @@
...
+
+StackTrace st = StackTrace.fromString("42");
+
+inline class IC<T extends num> {
+  final T id;
+  IC(this.id);
+
+  X testMe<X extends T>() {
+    if (2 > 1) {
+      Error.throwWithStackTrace("I'm inline", st);
```
eernstg, line 41: We should have value arguments as well as type arguments, and the throw object should reveal their values.

eernstg (commented):
Interesting stuff! I do think there are a few things that aren't being tested, and I commented on that.

sgrekhov:
Thank you. I misread the spec and thought that `args` part is about type arguments of inline class itself. Tests updated, please review once again

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A03_t01.dart
```diff
@@ -0,0 +1,49 @@
...
+/// it executes the body of Dm in an environment where this and the name of the
+/// representation are bound to o, the type variables of V are bound to the
+/// actual values of T1, .. Ts, and the formal parameters of m are bound to
+/// args1 in the same way that they would be bound for a normal function call.
+/// If the body completes returning an object o2 then the invocation evaluates
+/// to o2. If the body throws an object and a stack trace then the invocation
+/// completes throwing the same object and stack trace.
+///
+/// @description Check that invocation of an inline class method in the case
+/// when type arguments are specified.
```
eernstg, line 26: OK, so we're testing an inline class method invocation of the form `ic.m<Some, TyArgs>(some, arguments)`.

We should have `this == id` in the body of the method, and that should be equal to the receiver at the call site (this may be in some other test). `T` should be bound to `num` (`int` for line 46++). `key` should be bound to `"key1"` (this one is being tested already).

So it's again the binding of `T` that we can't observe very directly based on the behavior of the code in the test as written.

In this case we have explicit type arguments, which means that there is less complexity before we get to the value of `T`. I can see that the value of `T`, let's call it `U`, is a type that must satisfy `int <: U`, `num <: U`, if `V extends T` is being checked correctly (in this case it is a relationship which should be verified at compile time).

So we have the same issues as previously: The test doesn't make it obvious that know enough about the actual binding of `T`. I think the easiest way to cover that part is to have an `asTypeMap` method, or anything that uses `T` in the body of the method in a way which is observable at the call site.

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A03_t01.dart
eernstg: Got them, thanks!

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A03_t01.dart
eernstg: By the way, it isn't entirely obvious what we're testing here: The description seems to say "Check that invocation ... [does something]", but the 'does something' part is missing.

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A03_t01.dart
```diff
@@ -0,0 +1,49 @@
...
+
+// SharedOptions=--enable-experiment=inline-class
+
+import "../../Utils/expect.dart";
+
+inline class IC<T> {
+  final T id;
+  IC(this.id);
+
+  Map<K, V> asMap<K, V extends T>(K key) => {key: this.id as V};
```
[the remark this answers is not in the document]
eernstg: Just noticed that 'LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A03_t02.dart' is testing the binding of `this`.

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A03_t02.dart
```diff
@@ -0,0 +1,42 @@
...
+inline class IC<T> {
+  final T id;
+  IC(this.id);
+
+  IC test() => this;
+}
+
+main() {
+  IC<num> ic = IC(42);
+  Expect.equals(ic, ic.test());
```
eernstg, line 41: We should have
```suggestion
  Expect.identical(ic, ic.test());
```
Did I miss something that would contradict that expectation?

Thinking again, here is one reason why one might use `equals`: With an `int`, `identical` is _defined_ in terms of `==`, which means that `equals` is guaranteed to be a faithful replacement for `identical`. However, that could be a reason why we'd want to use `class A {}` and an `A` rather than an `int`, such that `identical` matters.

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A03_t02.dart
eernstg: Got it now, thanks!

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A03_t02.dart
eernstg: Also, we might as well test the binding of `id` as well (unless some other test does it already).

sgrekhov:
Thank you for the detailed (as usual) review. Hopefully I did all of the requested changes. Please take another look

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A02_t01.dart
```diff
@@ -30,12 +31,10 @@ inline class IC<T> {
   final T id;
   IC(this.id);
 
-  Map<K, V> asMap<K, V extends T>(K key) => {key: this.id as V};
+  Map<K, Type> asMap<K, V extends T>(K key) => {key: V};
```
eernstg, line 34: Perhaps include `someOtherKey: T` here as well? The point is that the binding of `T` to a value is the really new thing here. We should test that the value of `K` and `V` are also as expected, but we certainly shouldn't ignore `T`.

(Also, when the description says `T1 .. Ts`, it means `T`, not `K, V`.)

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A02_t02.dart
```diff
@@ -29,10 +29,14 @@ inline class IC<T> {
...
+  void test2() {
+    Expect.identical(this, id);
+  }
 }
 
 main() {
   IC ic = IC(42);
-  Expect.identical(ic, ic.test());
+  Expect.identical(ic, ic.test1());
+  ic.test2();
```
eernstg, line 41: Thanks!

LanguageFeatures/Inline-classes/dynamic_semantics_member_invocation_A03_t01.dart
```diff
@@ -22,8 +22,9 @@
 /// to o2. If the body throws an object and a stack trace then the invocation
 /// completes throwing the same object and stack trace.
 ///
-/// @description Check that invocation of an inline class method in the case
-/// when type arguments are specified.
+/// @description Check that in case of invocation of an inline class method in
+/// with the specified type arguments formal parameters of m are bound to
+/// `args1` in the same way that they would be bound for a normal function call
```
eernstg, line 27: Maybe: ..
```suggestion
/// @description Check that in case of an invocation of an inline class method `m`,
/// the specified type parameters and formal parameters of `m` are bound to
/// `args1` in the same way that they would be bound for a normal function call
```

sgrekhov:
Updated. Please rereview

eernstg (approved):
Thanks, LGTM!

PR #2058, opened 2023-05-23
#1400. Static analysis of inline class tests. Part 1

sgrekhov:


LanguageFeatures/Inline-classes/static_analysis_inline_class_A02_t01.dart
```diff
@@ -0,0 +1,42 @@
...
+
+  X2 foo() => id as X2;
+}
+
+main() {
+  V v1 = V(42);
+  Expect.isTrue(v1 is V<num, Object>);
+
+  V<int, num> v2 = V(42);
+  Expect.isTrue(v2 is V<int, num>);
```
eernstg, line 35: So what does it mean to test that it "can be used as a type"?

Checking that it is accepted at compile-time to have it in `is` and `as` is certainly relevant. (Do we also have a test which is about the dynamic semantics? E.g., we should have `v2 is V<num, Object>` as well, but also `v2 is V<num, Never>`, because the actual behavior is to test that `v2 is num`).

Another part of "can be used as a type" could be that it has the expect subtyping relations (unless that's already covered or planned for a different test). So `v2` should be assignable to a variable of type `V<num, Object>` (we expect covariance for all type parameters, as long as we don't have support for anything else), but not `V<int, Never>`.

eernstg: However, we should have some tests showing that the inline type can occur in other positions: Parameter and return types of function types, type variable bounds in classes and functions, and perhaps I forgot some more places.

I believe there would be a separate test for inline types as superinterfaces of anything-other-than-an-inline-class, expecting errors.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A03_t01.dart
```diff
@@ -0,0 +1,49 @@
...
+
+inline class V<X1 extends num, X2 extends B> {
+  final X1 id;
+  V(this.id);
+
+  X2 foo() => id as X2;
+}
+
+main() {
+  V<int, Object> v1 = V(42);
```
eernstg, line 35: I think a stronger test would be `V<int, Object?>`: This is a correctly super-bounded type, but that's still an error because inline classes just don't tolerate super-bounded types (and they also do not tolerate malbounded types, so we should have those, too, and I can see we do have them in line 40, 45).

LanguageFeatures/Inline-classes/static_analysis_inline_class_A04_t01.dart
```diff
@@ -0,0 +1,29 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+class A {}
+class B extends A {}
+class C extends B {}
+
+inline class V<T extends B> {
+  final T id;
```
eernstg, line 20: That won't work: The type annotation of a final instance variable is a covariant position.

But `final void Function(T) f;` would work.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A04_t01.dart
```diff
@@ -0,0 +1,29 @@
...
+  V(this.id);
+}
+
+main() {
+  V<A> v1 = V<B>(B());
```
eernstg, line 25: This should not be an error (`V` is covariant in its type argument). But it's a better fit in a test about the subtype relationships of inline types, 'being in a non-covariant position of the representation type' is a property of the declaration of `V` (of `V.id`, actually), not a property of any usage of the type in client code.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A05_t02.dart
```diff
@@ -0,0 +1,35 @@
...
+import "../../Utils/static_type_helper.dart";
+
+ class V<T extends V<T>> {
+  final T id;
+  V(this.id);
+}
+
+main() {
+  List<V> l = [];
+  l.expectStaticType<Exactly<List<V<V<dynamic>>>>>();
```
eernstg, line 34: This is tricky because the type `V<V<dynamic>>` should be an error in its own right (it is super-bounded). Same issue: the raw type `V` in the type `List<V>` should also be an error.

So I guess the only non-trivial test we can have for i2b is something like `<X extends List<Y>, Y extends num>` which should make it `<List<num>, num>`.

However, this test could be re-purposed to test that it is an error if the inline type isn't regular bounded.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A07_t01.dart
```diff
@@ -0,0 +1,82 @@
...
+//        ^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  Type get runtimeType => V1;
+//         ^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  dynamic noSuchMethod(Invocation invocation) { return null;}
```
eernstg, line 28: Nit -
```suggestion
  dynamic noSuchMethod(Invocation invocation) { return null; }
```
Just guessing that `dart format` prefers that space. Maybe `=> null;` even, to avoid a 3 line format.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A07_t01.dart
```diff
@@ -0,0 +1,82 @@
...
+// [cfe] unspecified
+
+  @override
+  Type get runtimeType => V2<T>;
+//         ^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  @override
+  dynamic noSuchMethod(Invocation invocation) { return null;}
```
eernstg, line 61: Ditto.

eernstg (commented):
I'm asking for a few things. Perhaps they are already planned for 'Part N' for some N, then please notify me.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A02_t01.dart
sgrekhov: > Do we also have a test which is about the dynamic semantics?

Not yet. [Dynamic semantics](https://github.com/dart-lang/language/blob/main/accepted/future-releases/inline-classes/feature-specification.md#dynamic-semantics-of-inline-classes) is a separate chapter of the spec. Will add such tests as tests for this chapter

> Another part of "can be used as a type" could be that it has the expect subtyping relations (unless that's already covered or planned for a different test).

I'll add these tests later as a Subtyping tests in a separate PR

> v2 is V<num, Never>

I'll add this

> However, we should have some tests showing that the inline type can occur in other positions

I'll add this

LanguageFeatures/Inline-classes/static_analysis_inline_class_A04_t01.dart
sgrekhov: I'll remove this test then and add apropriate tests in a `Subtyping` dir

LanguageFeatures/Inline-classes/static_analysis_inline_class_A03_t01.dart
sgrekhov: `V<int, Object?> v1 = V(42);` is not an error. Issue? See also aseparate email

LanguageFeatures/Inline-classes/static_analysis_inline_class_A05_t02.dart
sgrekhov: Hm... Ok, I'll change it.... But please note, there's also missed `inline` keyword for class `V` in this test but there's no any error. Is this an issue?

LanguageFeatures/Inline-classes/static_analysis_inline_class_A07_t01.dart
sgrekhov: Yes, `=> null` is better. I'll change it

sgrekhov:
Some comments before doing the requested changes

sgrekhov:
Ok, thank you. Tests updated, please review once again

LanguageFeatures/Inline-classes/static_analysis_inline_class_A04_t01.dart
eernstg: OK!

LanguageFeatures/Inline-classes/static_analysis_inline_class_A04_t01.dart
eernstg: (I wrote a comment in the overall review entry: We should not forget about this topic)

LanguageFeatures/Inline-classes/static_analysis_inline_class_A05_t02.dart
eernstg: I think this test has been taken out of the set of tests, or renamed --- I can't find anything like `V<V<dynamic>>` any more.

If the `inline` keyword is missing and we're working with a normal class then there is no problem in having super-bounded types, and hence no problem in having a type like `List<V<V<dynamic>>>`.

sgrekhov: Yes I removed it from here and going to add it later in another PR for `Subtyping` (and expect an error)

sgrekhov:
Seems that I wrongly removed test for type parameter in a non-covariant position. Re-added. This may complicate review, sorry for that. I think it'll be better to review changes from all commits

LanguageFeatures/Inline-classes/static_analysis_inline_class_A03_t02.dart
```diff
@@ -0,0 +1,33 @@
...
+/// ...
+/// A compile-time error occurs if the type V<T1, .. Ts> is not regular-bounded.
+///
+/// @description Checks that it is a compile-time error if the type
+/// `V<T1, .. Ts>` is not regular-bounded.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+inline class V<T extends V<T>> {
```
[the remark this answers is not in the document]
eernstg: Cf. https://github.com/dart-lang/language/issues/3109.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A04_t02.dart
```diff
@@ -0,0 +1,35 @@
...
+///
+///   ... // Other members.
+/// }
+/// ...
+/// When s is zero, V<T1, .. Ts> simply stands for V, a non-generic inline type.
+/// When s is greater than zero, a raw occurrence V is treated like a raw type:
+/// Instantiation to bound is used to obtain the omitted type arguments
+///
+/// @description Checks that it is a compile-time error if the type
+/// `V<T1, .. Ts>` is not regular-bounded.
```
eernstg: I don't think there are any types in this library which aren't regular-bounded. It is testing that i2b is used to compute the omitted type arguments in the declared type of `l`.

eernstg (commented):
I'll just send this review now, and then refresh and look at the latest changes.

By the way: I can't see any references to 'covariant' any more (and I also can't see any non-covariant occurrences of a type variable in the representation type), so I guess the test about "..non-covariant position.." will be in some other set of test libraries, perhaps in a different directory, to be written in the future? Just want to make sure we don't forget about it.

eernstg (commented):
Haha, I don't think I can see what happened since I did that refresh: Github says that there are two commits since the most recent review, but looking at the diff mostly seems to show that some libraries were renamed (everything changed, in several libraries).

Anyway, I'm going home now, I'll take another look tomorrow. ;-)

sgrekhov:
Yes, there were two commits since.  Sorry, I understand that it complicates the review. How can we avoid this? What cames to my mind.
1) Don't rename files. Do it as a separate renaming commit. 
2) I'm afraid that's all that I can suggest for now :)

P.S. I'm going to make a commit soon based on the latest review notes. They are two commits behind but mostly still actual

eernstg:
> Do it as a separate renaming commit.

Actually, a separate renaming PR is probably the most convenient approach. So just stick with the given file names until the whole thing is in a good shape, and then reorganize file names as needed in a separate PR. This means that there is no renaming noise in the contents PR, no matter which/how_many commits github uses to create a diff.

I think it's going to be inconvenient for various tracking features if names of tests are changing over time, but as long as they are new tests there's no history to disrupt. So it's probably best to avoid renamings tests that have already been executed by bots.

But, presumably, a renaming PR can be made and processed immediately after the content PR has been processed, so even if the content PR is rolled into the SDK it's only going to disrupt a single row in the test result database, and even that is probably going to be a rare exception.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A04_t01.dart
```diff
@@ -0,0 +1,24 @@
...
+/// an inline class occurs in a non-covariant position in the representation
+/// type
+/// @author sgrekhov22@gmail.com
+// SharedOptions=--enable-experiment=inline-class
+
+inline class V<T extends num> {
+  final void Function(T) f;
+//                    ^
+// [analyzer] unspecified
+// [cfe] unspecified
```
eernstg, line 18: Yes, thanks!

LanguageFeatures/Inline-classes/static_analysis_inline_class_A03_t02.dart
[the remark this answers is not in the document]
eernstg: Looks like they don't want to make it an error. We'll see. ;-)

LanguageFeatures/Inline-classes/static_analysis_inline_class_A02_t01.dart
```diff
@@ -0,0 +1,54 @@
...
+
+class C<T extends V<num, Object>> {
+  V<int, int> bar<T extends V<num, num>>(T t) => t as V<int, int>;
+}
+
+main() {
+  V v1 = V(42);
+  Expect.isTrue(v1 is V<num, Object>);
+  Expect.isTrue(v1 is V<num, Never>);
+  Expect.isTrue(v1 is V<int, String>);
```
eernstg, line 38: Yes, thanks! Now we're covering the case which would fail with a regular class.

eernstg (approved):
LGTM, landing the PR.

I added a few comments, mostly "That's great!" or "we should have a test about [something]", so there is no need to do more before this one is landed.

But it would be great if you could consider whether those tests already exist, or are already planned, or they should be added to a TODO list.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A04_t01.dart
sgrekhov: @eernstg great! Thank you! We do have a natural location for such tests, but I don't see any tests for the statement below there. In the Dart Language spec we have:

> **10.10 Superinterfaces** 
> ...
> When a generic class C declares a type parameter X, it is a compile-time
> error if X occurs in a non-covariant position in a type which specifies a superinterface of C. _For example, the class can not have List<void Function(X)> in its extends or implements clause._
> 

And no any co19 for it. Looks like that this statement were added relatively recently. Created https://github.com/dart-lang/co19/issues/2066 for this. 

BTW without `inline` keyword
```dart
class V<T extends num> {
  final void Function(T) f;
//                    ^
// [analyzer] unspecified
// [cfe] unspecified
  V(this.f);
}

main() {
  print(V);
}
```
Fails in both analyzer and CFE (no expected error). Is this an issue?

PR #2060, opened 2023-05-23
Fixes #2059. Don't expect an error in a weak mode

### LINKED ISSUE #2059: co19/LanguageFeatures/nnbd/weak/const_evaluation_A06_t01 fails in a weak mode
co19/LanguageFeatures/nnbd/weak/const_evaluation_A06_t01 fails in a weak mode
According to https://github.com/dart-lang/sdk/issues/45067#issuecomment-782602525 `null` is allowed in a non-sound mode. So, there should be no error

sgrekhov:
This test used to test a strong mode. Now in a weak mode it fails. As far as I understand here we have class
```dart
class C<T> {
  final T x;
  const C.t(Object? o, T t) : x = o as T;
}
```
When we call
```dart
const c1 = C.t(null, cLegacyInt);
```
`T` is inferred as `int*`. `int*` became `int` and `x = o as int;` failed. But now in a weak mode is doesn't throw, so we should expect no error

### DIFF
--- a/LanguageFeatures/nnbd/weak/const_evaluation_A06_t01.dart
+++ b/LanguageFeatures/nnbd/weak/const_evaluation_A06_t01.dart
@@ -11,6 +11,7 @@
 
 // Requirements=nnbd-weak
 
+import "../../../Utils/expect.dart";
 import '../const_evaluation_lib.dart';
 import 'const_evaluation_legacy_lib.dart';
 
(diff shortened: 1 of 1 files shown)


LanguageFeatures/nnbd/weak/const_evaluation_A06_t01.dart
```diff
@@ -11,19 +11,15 @@
...
 // Requirements=nnbd-weak
 
+import "../../../Utils/expect.dart";
 import '../const_evaluation_lib.dart';
 import 'const_evaluation_legacy_lib.dart';
 
 const dynamic d = null;
 
 main() {
   const c1 = C.t(null, cLegacyInt);
```
eernstg, line 21: OK, tricky as usual. ;-)

The inferred type argument for `C` needs to be a supertype of `Null` and a supertype of `int*` (the type of `cLegacyInt`), but there are several types satisfying this. So we shouldn't have a compile-time error.

Next, no-sound-null-safety execution will not throw when `o as T` is tested in a situation where `o` evaluates to null. So we should indeed expect that there is no run-time error.

Very good, should work!

eernstg (approved):
LGTM, after a bit of thinking.

PR #2068, opened 2023-05-30
#1400. More static analysis of the inline class tests

sgrekhov:


LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t03.dart
```diff
@@ -0,0 +1,47 @@
...
+
+test2<X extends V2<String?, int?>?>(X x) {
+  Object? o = x;
+  x?.id;
+}
+
+main() {
+  test1<V1>(V1(42));
+  test2<V2<String?, int?>>(V2(42));
+}
```
eernstg, line 47: We'd want to check that `V1` is a subtype of `Object?`: `Object? x = V1(24); // OK.`.

Also, we'd want to check that it is a proper subtype: `V1 v1 = Object() as Object?; // Error.`.

But the test would be stronger if we had

```dart
inline class V1 {
  final Object? id;
  V1(this.id);
}
```

because this would allow an expression of type `V1` to have the value `Object()` and the value null, but an expression with static type `Object?` should still not be assignable to `V1`: `V1 x = null as Object?; // Error.`.

This means that we're really checking that `V1` is a proper subtype of `Object?`, we're not just checking whether we can assign an expression of the representation type to a variable of type `V1` (it would be a bug if that's what is being checked, which is the reason why the test shouldn't succeed when we assign an expression of type `Object?`).

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t05.dart
```diff
@@ -0,0 +1,63 @@
...
+/// variable whose type is an inline type, and then relying on implicit
+/// initialization to null.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+inline class V1 {
+  final int it;
+  V1(this.it);
+}
```
eernstg, line 28: It's a stronger test if we use `final int? it;`, because this would allow a variable of type `V1` to have the value null, but it should _still_ be a compile-time error to use an implicit initialization to null to have that effect.

eernstg (commented):
Looks good, but still gets several comments. ;-)

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t04.dart
```diff
@@ -0,0 +1,61 @@
...
+/// instance variable with name id and type R.
+///
+/// We say that the declared representation type of Inline is R, and the
+/// instantiated representation type corresponding to Inline<T1,.. Ts> is
+/// [T1/X1, .. Ts/Xs]R.
+/// ...
+/// Let V be an inline type of the form Inline<T1, .. Ts>, and let R be the
+/// corresponding instantiated representation type. If R is non-nullable then V
+/// is a proper subtype of Object, and V is non-nullable. Otherwise, V is a
+/// proper subtype of Object?, and V is potentially nullable.
```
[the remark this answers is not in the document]
sgrekhov: Please clarify. The example above assumes that `the type of `x` is potentially nullable.` If `X extends T` is not a good way to obtain instance of a potentially nullable type, then how to obtain a potentially nullable type here?

sgrekhov (commented):
Mostly implemented all of the recommendations but still need clarification for `static_analysis_inline_class_A12_t04.dart`

sgrekhov (commented):
Mostly implemented all of the recommendations but still need clarification for `static_analysis_inline_class_A12_t04.dart`

LanguageFeatures/Inline-classes/static_analysis_inline_class_A09_t01.dart
```diff
@@ -0,0 +1,36 @@
...
+/// @assertion If e is an expression whose static type V is the inline type
+/// Inline<T1, .. Ts> and m is the name of a member that V has, a member access
+/// like e.m(args) is treated as an invocation of the inline member m on the
+/// receiver e according to the inline type Inline and with the actual type
+/// arguments T1, ..., Ts, with the actual argument part args.
+///
+/// @description Checks that a member access `e.m(args)` is treated as an
+/// invocation of the inline member m on the receiver `e` according to the
+/// inline type `Inline` and with the actual type arguments `T1, ..., Ts`, with
+/// the actual argument part `args`.
```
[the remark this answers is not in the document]
eernstg: If I understood the situation correctly then that would indeed make sense, thanks!

LanguageFeatures/Inline-classes/static_analysis_inline_class_A11_t05.dart
```diff
@@ -0,0 +1,50 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+import "../../Utils/expect.dart";
+
+extension E1 on int {
+  String foo() => "E1.foo";
+}
+
```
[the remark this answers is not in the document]
eernstg: Thx!

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t04.dart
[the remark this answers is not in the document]
eernstg: We testing whether the nullability properties of an inline type of the form `V` or `V<T1 .. Tk>` is implemented correctly. So `x` should have such an inline type, no matter what. For example:

```dart
void main() {
  var x = V1(1);
  ...
}
```

Next, let's say that the representation type of `V1` is `int`. This is a non-nullable type, and this means that `V1` should be a subtype of `Object`, but not vice versa. So `Object o = x;` should be accepted, but `V1 v = Object()` should be a compile-time error.

If the representation type of `V1` is `int?`, a nullable type, then `V1` should be a subtype of `Object?`, but not vice versa, and `V1` should not be a subtype of any other type (including: `V1` should not be a subtype of `Object`).

So `Object? oq = v1;` should be OK (well, that's an extremely weak test because that's true for all types, but it might fail for inline types because the compile-time subtype query for inline types is new code in the analyzer and compiler. So we should actually test that).

The other part is that `v1` should be considered potentially non-nullable, and this can be confirmed by assigning null to it: `V1 w = null; // Error.`. That should be an error because, roughly, we _cannot know_ whether or not it's OK for the variable `w` to have the value null, so we have to prevent the assignment.

It's true that `X extends SomeNullableType` ensures that `X` has a type which is potentially nullable, but that doesn't matter: We're testing inline classes, not type variables.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t05.dart
eernstg: Thanks!

eernstg: Thanks!

It looks funny that the error is reported at `print(x)`, but this is of course because the variable could be initialized at several different locations (`var x; if (blah) x = 1; else x = 2;`), which means that it is only known for sure that something is wrong at the point where the variable is evaluated, and it isn't 'definitely assigned'.

I think it would be really helpful to add a few words about this in a comment.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A05_t03.dart
```diff
@@ -0,0 +1,55 @@
...
+
+inline class V2<T extends Object> {
+  final int id;
+  V2(this.id);
+}
+
+inline class V3<T extends Object?> {
+  final int id;
+  V3(this.id);
+}
```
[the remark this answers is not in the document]
eernstg: Looks fine!

Yes, please, it would be nice to have a test that confirms that some of those type arguments are reported as errors (unless that is just a duplicate of existing tests about being regular-bounded).

You can create a type that satisfies the bound: `class B extends A<B> {}`, and then use `V<B>`. That's the standard way to obtain a type that satisfies an F-bound: Just write the F-bound itself as a class declaration (`T extends A<T>` becomes `class B extends A<B> {}`).

LanguageFeatures/Inline-classes/static_analysis_inline_class_A10_t01.dart
```diff
@@ -0,0 +1,32 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Similarly, e.m is treated an invocation of the inline member m on
+/// the receiver e according to the inline type Inline and with the actual type
+/// arguments T1, ..., Ts and no actual argument part.
+///
+/// @description Checks that a member access `e.m` is treated as an invocation
+/// of the inline member `m` on the receiver `e` according to the inline type
+/// `Inline` and with the actual type arguments `T1, ..., Ts` and no actual
+/// argument part
```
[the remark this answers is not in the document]
eernstg: Fine, thanks!

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t01.dart
```diff
@@ -0,0 +1,37 @@
...
+/// proper subtype of Object?, and V is potentially nullable.
+///
+/// @description Checks that if an instantiated representation type `R` is
+/// non-nullable then it is not an error to assign it to `Object`
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+inline class V1 {
+  final int? id;
```
eernstg, line 25: Oh, but `int?` is _not_ non-nullable, so we need to use `int` or some other non-nullable type in order to make it correct to test as in line 35 and 36.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t01.dart
```diff
@@ -0,0 +1,37 @@
...
+  V1(this.id);
+}
+
+inline class V2<T1, T2 extends num?> {
+  final num? id;
```
eernstg, line 30: Same here, need to use a non-nullable type as the representation type.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t02.dart
```diff
@@ -0,0 +1,49 @@
...
+/// proper subtype of Object?, and V is potentially nullable.
+///
+/// @description Checks that if an instantiated representation type `R` is
+/// non-nullable then it is a compile-time error to assign `Object` to `V`
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+inline class V1 {
+  final int? id;
```
eernstg, line 25: The description says that the representation type is non-nullable, so it should be non-nullable.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t02.dart
```diff
@@ -0,0 +1,49 @@
...
+  V1(this.id);
+}
+
+inline class V2<T1, T2 extends num?> {
+  final num? id;
```
eernstg, line 30: I think it would be helpful to use a representation type where the phrase 'instantiated representation type' means something.

That is, the representation type should contain one or more type variables, such that it makes a difference which actual type arguments we're passing.

So, for instance, the representation type could be `T1`.

We should then consider `V2<int, .../*this one doesn't matter, let's say Never*/...>`. The point is that the instantiated representation type is `int`, and that is a non-nullable type.

Then it should still be an error to assign `Object()` to a variable of type `V2<int, ...>`.

There may be a different test doing this, but the other side of that coin is that it _is_ an error to assign an expression of type `V2<int?, ...>` to a variable of type `Object`.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t02.dart
```diff
@@ -0,0 +1,39 @@
...
+/// instantiated representation type corresponding to Inline<T1,.. Ts> is
+/// [T1/X1, .. Ts/Xs]R.
+/// ...
+/// Let V be an inline type of the form Inline<T1, .. Ts>, and let R be the
+/// corresponding instantiated representation type. If R is non-nullable then V
+/// is a proper subtype of Object, and V is non-nullable. Otherwise, V is a
+/// proper subtype of Object?, and V is potentially nullable.
+///
+/// @description Checks that if an instantiated representation type `R` is
+/// non-nullable then it is a proper subtype of `Object`
```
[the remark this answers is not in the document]
eernstg: (Good - added new comment there.)

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t03.dart
eernstg: I still don't really understand what `test1` and `test2` are testing. We need to test that when the instantiated representation type is nullable or potentially nullable (phrased as 'not non-nullable') then the inline type is a proper subtype of `Object?`.

`final Object? id;` works fine: This is a nullable representation type. So we need to check that `V1` is a proper subtype of `Object?`, that is

```
Object? o = V1(null); // OK: Confirms 'subtype'.
V1 v = anyExpressionWhatsoever as Object?; // Error: Confirms 'proper'.
```

In `test1` we have `X` whose type is potentially nullable because of the `?` in the bound. I don't think it's possible to learn anything about the nullability of `V1` by means of code whether the type `X` is used, because `X` as potentially nullable no matter what `V1` is.

`test2` has the same property: No code using `X` will help us learn anything about `V2<String?, int?>`, because `V2<String?, int?>?` is nullable anyway, so `X` is potentially nullable no matter what.

sgrekhov (commented):
Thank you for detailed explanation! I was misunderstanding what `potentially nullable` means here and how it should be tested. Updated. Please review

LanguageFeatures/Inline-classes/static_analysis_inline_class_A05_t03.dart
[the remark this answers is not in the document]
eernstg: Thanks!

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t03.dart
eernstg: Sounds good, thanks!

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t03.dart
```diff
@@ -0,0 +1,38 @@
...
+
+// SharedOptions=--enable-experiment=inline-class
+
+inline class V1 {
+  final Object? id;
+  V1(this.id);
+}
+
+inline class V2<T1, T2 extends num?> {
+  final T1? id;
```
eernstg, line 31: I think it would be useful to remove the `?` here and then use `String?` in line 37.

The point is that we're relying on the nullability of the _instantiated_ representation type. When the representation type itself is just `T1` we really must instantiate the representation type (which then yields `String?`) in order to see that the instantiated representation type is 'not non-nullable' (that is: potentially nullable, which is always true when it is nullable, and `String?` is nullable).

So if we use `T1?` and the implementation of some tool forgets to instantiate then there's a bug, but we won't see it.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A12_t04.dart
```diff
@@ -0,0 +1,55 @@
...
+
+// SharedOptions=--enable-experiment=inline-class
+
+inline class V1 {
+  final int? it;
+  V1(this.it);
+}
+
+inline class V2<T1, T2 extends num?> {
+  final T1? id;
```
eernstg, line 31: The test would be stronger if this were `T1` rather than `T1?`, and then ..

Line 46: Declared type `V2<String?, int>` would have a nullable representation type.
Line 51: The raw `V2` as the declared type would also have a nullable representation type, but it's probably better to use `V2<String?, int>` again, just so we're testing `= anObjectQuestion` and `= null` on the same declared type, following the pattern set by line 36 and 41.

eernstg (commented):
Looks good! I think we should take the opportunity to ensure that a couple of specific potential bugs can be detected: When the assertion/description says 'the instantiated representation type', it would be useful if that type really must be instantiated in order to get the right criterion (such that we'll have a test failure if it isn't being instantiated).

eernstg (approved):
LGTM!

PR #2071, opened 2023-06-01
#1400. Static analysis of inline class tests. Part 3

sgrekhov:


LanguageFeatures/Inline-classes/static_analysis_inline_class_A15_t01.dart
```diff
@@ -0,0 +1,74 @@
...
+
+// SharedOptions=--enable-experiment=inline-class
+
+inline class X implements V {
+  final Y id;
+  X(this.id);
+}
+
+inline class Y implements X {
+  final Y id;
```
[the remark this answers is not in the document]
eernstg: One thing we can do to test this is to have inline classes with representation types such that there is a cycle after erasure (cheating: using primary constructors for brevity):

```dart
inline class V<Y>(Y y);
void f<X extends V<Y>, Y extends X>() {} // Error.
```

The point is that the output after erasure is the following:
```dart
void f<X extends Y, Y extends X>() {}
```

and that kind of cycle in a type parameter declaration list is an error.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A16_t01.dart
```diff
@@ -0,0 +1,35 @@
...
+
+// SharedOptions=--enable-experiment=inline-class
+
+inline class V1 {
+  final int id;
+  V1(this.id);
+  V1.x(this.id);
+}
+
+inline class V2 extends V1 {
```
eernstg, line 20: I think we should avoid `extends` because that's already a syntax error (so perhaps we won't even start doing the type analysis).

So `implements V1` is better.

We can still test the superinitializer as if `V1` had allowed us to call its constructor in that manner.

However, we should also test that it is rejected if we try to invoke a constructor of `Object` (which could be done by having a plain `super()` superinitializer).

LanguageFeatures/Inline-classes/static_analysis_inline_class_A17_t01.dart
```diff
@@ -0,0 +1,33 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+inline class V1 {
+  final int id;
+  V1(this.id);
+}
+
+inline class V2 extends V1 {
```
eernstg, line 19: Again, we can't use `extends`, because then every other error will be a derived error. Using `implements V1` should be fine, it is possible, and `super.whatever` should still be an error.

In this case, unfortunately, we can't use any super parameters in an invocation that would actually work if we were calling a constructor of `Object`. 

So we'll probably just have to live with the fact that we can't test that the existence of the super parameter is an error in itself, because it is always also an error because there is no such corresponding parameter of a superconstructor.

eernstg (commented):
Looks fine, with a few comments.

In particular, the definition of 'representation dependency' is different from what I see in some tests. Also, there are several occurrences of `extends`, and that makes the test a syntax error, which means that we (probably) won't be able to see the error which is the actual topic of the test.

sgrekhov (commented):
Tests updated. In some cases it's hard to predict where to expect an error. I'll update these tests when we'll get an implementation (if needed of course). Please review

sgrekhov (commented):
Tests updated. In some cases it's hard to predict where to expect an error. I'll update these tests when we'll get an implementation (if needed of course). Please review

LanguageFeatures/Inline-classes/static_analysis_inline_class_A14_t01.dart
```diff
@@ -0,0 +1,51 @@
...
+/// It is a compile-time error if an inline class declaration has a
+/// representation dependency on itself.
+///
+/// @description Checks that it is a compile-time error if an inline class has a
+/// representation dependency on itself. Test dependency via implements
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+inline class V1 implements V2 {
```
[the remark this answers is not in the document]
eernstg: Sounds good, thanks!

LanguageFeatures/Inline-classes/syntax_A02_t10.dart
```diff
@@ -20,15 +20,15 @@
 // SharedOptions=--enable-experiment=inline-class
```
eernstg, line 20: Can't comment on line 16, so here we go:

We assume the same restrictions about the subtypes of a `final` inline class as the ones we have for regular classes.

Given that `inline class B extends A with M1 .. Mk ...` isn't a thing (which includes the case where `k == 0` and there are no mixins), we only need to consider `inline class B implements A {...}`.

So we require that `B` is also `final` (because it must be `base`, and there is no other way to be `base` for an inline class than being `final`), but it is allowed to declare such a `B` inline class in the same library as `A`.

Outside that library, `inline class D implements A ...` is a compile-time error, just like it is for non-inline classes.

LanguageFeatures/Inline-classes/syntax_A02_t10.dart
```diff
@@ -20,15 +20,15 @@
...
 
 final inline class FIC {
-  final int id ;
-  const FIC(this.id);
+  final int id = 0;
+  FIC();
 }
 
 final inline class FIC2 extends FIC {
```
eernstg, line 27: First, it isn't an error to have a subtype in the same library just because `FIC` is `final`.

Second, we'd want `FIC2` to be correct, such that we can see that it isn't an error.

```dart
final inline class FIC2 implements FIC {
  final int id;
  FIC2(this.id);
}
```

It is an interesting special case that `FIC` uses an initializing expression in the variable declaration, but `FIC2` is going to force the value of `id` to be whatever `FIC2` wants.

However, we never promised that `FIC` could enforce that the value of `id` is zero, it was always the intention that a subtype like `FIC2` would be able to use its own representation object with the members of `FIC`.

As long as the representation object _types_ are aligned (here: `int <: int`), the static analysis will accept the program, and we don't care that the author of `FIC` might think that it is guaranteed that `FIC.id == 0`.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A14_t02.dart
```diff
@@ -0,0 +1,51 @@
...
+/// It is a compile-time error if an inline class declaration has a
+/// representation dependency on itself.
+///
+/// @description Checks that it is a compile-time error if an inline class has a
+/// representation dependency on itself. Test dependency via extends
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+inline class V1 extends V2 {
```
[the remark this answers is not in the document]
eernstg: Thanks!

LanguageFeatures/Inline-classes/static_analysis_inline_class_A14_t03.dart
```diff
@@ -0,0 +1,51 @@
...
+/// It is a compile-time error if an inline class declaration has a
+/// representation dependency on itself.
+///
+/// @description Checks that it is a compile-time error if an inline class has a
+/// representation dependency on itself. Test dependency via extends/implements
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+inline class V1 extends V2 {
```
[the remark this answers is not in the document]
eernstg: Aha, very good, thanks!

LanguageFeatures/Inline-classes/static_analysis_inline_class_A15_t01.dart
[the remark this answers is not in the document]
eernstg: Looks good!

LanguageFeatures/Inline-classes/static_analysis_inline_class_A16_t01.dart
```diff
@@ -0,0 +1,36 @@
...
+
+inline class V1 {
+  final int id;
+  V1(this.id);
+  V1.x(this.id);
+}
+
+inline class V2 implements V1 {
+  final int id = 0;
+  V2() : super();
```
eernstg, line 22: I think we should change the first `V1` constructor to accept an optional positional parameter. Otherwise we'll have an error no matter what, because it is an error to pass no actual arguments to the constructor named `V1`, and then we might not detect whether or not there is a bug in the check that makes an error to have `super()`, no matter how correct it looks relative to the constructors declared by `V1`.

LanguageFeatures/Inline-classes/static_analysis_inline_class_A16_t01.dart
eernstg: Good! I think we need one more change, though. I added a new comment to line 22.

eernstg (commented):
A couple of threads are still alive.

LanguageFeatures/Inline-classes/syntax_A02_t10.dart
sgrekhov: With the requested changes this test became almost the same as the next test, `syntax_t11.dart`. So I changed `syntax_t11.dart` to have an initializing expression in `FIC` and changed this test to check a missing case

sgrekhov (commented):
Thank you. Updated. Also added a couple of missing cases

LanguageFeatures/Inline-classes/syntax_A02_t10.dart
eernstg: Thanks!

LanguageFeatures/Inline-classes/syntax_A02_t10.dart
eernstg: Sounds good!

PR #2074, opened 2023-06-06
Fixes for socket tests

mraleph:
A bunch of fixes to various networking tests which were depending on a specific ordering/timing of networking events. 

Changes to the networking stack done in https://dart-review.googlesource.com/c/sdk/+/307043 are making these fragile tests to fail on Windows.


mkustermann (approved):

LibTest/io/HttpClientRequest/writeAll_A03_t01.dart
```diff
@@ -32,24 +36,24 @@ var localhost = InternetAddress.loopbackIPv4.address;
 
 test(String method) async {
   asyncStart();
-  Iterable objects = [1, "2", 3.14, new C(), null];
+  Iterable objects = [1, "2", 3.25, new C(), null];
```
sgrekhov, line 39: Just wondering. Why `3.25` is better thab `3.14`? Does `3.14` have any floating point comparison issues (like `3.1400000000001 != 3.14`)?

sgrekhov (commented):
LGTM. Just one small question

LibTest/io/HttpClientRequest/writeAll_A03_t01.dart
mraleph: `3.25` is exactly representable as a double precision floating point value, `3.14` is not. So I just want to choose a value that is guaranteed to roundtrip through string formatting without changing. `(3.14).toString()` _happens_ to be `3.14`, because of how `double.toString()` works, but with `3.25` you don't need to think. 

The actual string representation of `3.25` matters here because we set `contentLength` - so the amount of bytes written should match. 

I will add a comment around `contentLength`

mraleph: Removed magical constant and added a comment about `contentLength`.

PR #2075, opened 2023-06-07
#1400. File renaming commit

sgrekhov:
Just rename files to preserve tests enumeration. One test was deleted, it will be readded in the next PR with another name and assertion. "Composing Inline Classes" contains a statement which exactly fits what the test does


eernstg (approved):
LGTM

PR #2077, opened 2023-06-08
#1400. Dynamic semantics tests complemented

sgrekhov:


eernstg (commented):
Very good! One thing needs to be adjusted: In a number of situations it is expected that two function objects are `notEquals`. However, in some of those situations it isn't guaranteed that this is true, because an implementation should be allowed to reuse the same function object if it is able to behave in the correct manner.

In some cases it could even be a top-level function, if it's supposed to be a tear-off of an inline class instance method that doesn't depend on the value of `this` and doesn't have (or doesn't use) any type parameters of the inline class.

```dart
// Source code:
inline class A<X> {
  void m<Y>(int i) { print('Uses $i and $Y, but not X'); }
}

// Compiles into:
void tornOffM<Y>(int i) { print('Uses $i and $Y, but not X'); }

inline class A<X> {
  void m<Y>(int i) => tornOffM<Y>(i);
}
```

With this, a tear-off of `receiverOfTypeA.m` can be compiled as `tornOffM`, and we shouldn't fail a test because this is a constant expression and all those torn off functions are the same object, even across different receivers.

(The same thing is not allowed for a a non-inline class instance method tear-off, they _must_ compare not-equal in the case where they have different receivers.)

eernstg (commented):
Just one `notEquals` remains, but cannot be assumed.

eernstg (approved):
LGTM

eernstg:
Thanks, got it!

PR #2078, opened 2023-06-09
#1400. Composing inline classes tests renumerated

sgrekhov:


eernstg (approved):
LGTM.

In future file rename PRs, please write a couple of words in the description about what is renamed to what.

PR #2081, opened 2023-06-12
#1400. Dynamic semantics inline class tests

sgrekhov:


sgrekhov:
Please note that this PR has no tests for the following statement

> Consider an inline class declaration DV named Inline with representation name id and representation type R. Invocation of a non-redirecting generative inline class constructor proceeds as follows: A fresh, non-late, final variable v is created. An initializing formal this.id has the side-effect that it initializes v to the actual argument passed to this formal. An initializer list element of the form id = e or this.id = e is evaluated by evaluating e to an object o and binding v to o. During the execution of the constructor body, this and id are bound to the value of v. The value of the instance creation expression that gave rise to this constructor execution is the value of this.

I don't think that the above can be tested somehow. If you know the way, please let me know

LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A01_t01.dart
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion At run time, for a given instance o typed as an inline type V,
+/// there is no reification of V associated with o.
+///
+/// @description Check that at a run-time type an instance of an inline class is
```
eernstg, line 8: Perhaps:
```suggestion
/// @description Check that the run-time type of an instance of an inline class is
```

LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A02_t01.dart
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion The run-time representation of a type argument which is an inline
+/// type V is the corresponding instantiated representation type.
+///
+/// @description Check that at a run time an instance of an inline class is
```
eernstg, line 8: The assertion is actually talking about using an inline type as an actual type argument. So if `V<int>` is an inline type with instantiated representation type `int`, a `List<V<int>>` must be a `List<int>` at run time. So it should have that runtimeType, it should accept `as List<int>` at compile time and it should succeed at run time, and it should then be possible to insert an int, and `(thatList as dynamic).add(1.5)` should fail. This means that we have 2-3 different pieces of evidence that this is a `List<int>` and not anything else (of course, it's actually going to be a `MySecretListClass<int>`, but its type argument won't be `num`).

We could also test that `v as int` is accepted at compile time and succeeds at run time, but that's more like an extra case in LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A01_t01.dart.

LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A03_t01.dart
```diff
@@ -0,0 +1,33 @@
...
+import "../../Utils/expect.dart";
+
+inline class V {
+  final int id;
+  V(this.id);
+}
+
+main() {
+  var v = V(42);
+  Expect.isTrue(v is V);
```
eernstg, line 24: This one is a little bit tricky. Here's what I'm thinking about as a natural way to get good coverage:

What we want to show is that `_ is V` works the same as `_ is int`, no matter what we put into the location where we have `_`. So maybe we could do this:

```dart
var objects = [true, 1, 1.5, #foo, <String>[], {'a': () {}}, StringBuffer('Hello')]; // A bunch of objects.

void main() {
  for (var o in objects) {
    Expect.equals(o is V, o is int);
    Expect.equals(o is! V, o is! int);
  }
  ...
}
```

Perhaps it doesn't make much difference, but it does put more emphasis on "`_ is V` works the same as `_ is int`", whereas only asking about these expressions with `v` allows those two expressions to behave in different ways for any other type than `V`, and we'd want to catch that.

LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A03_t02.dart
```diff
@@ -0,0 +1,34 @@
...
+inline class V {
+  final int id;
+  V(this.id);
+
+  String get foo => "Inline";
+}
+
+main() {
+  var v = V(42);
+  Expect.isTrue((v as int).isEven);
```
eernstg, line 26: Here we'd like to know that `_ as V` works the same as `_ as int` (or, perhaps use a representation type which is `List<String>`, just to get the variation; in that case `_ as V` should work the same as `_ as List<String>`).

"they both throw, or both of them don't throw" is less convenient than "they are equal", but it should be expressible using a try/catch and some boolean variables.

LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A03_t03.dart
```diff
@@ -0,0 +1,33 @@
...
+}
+
+inline class V2 {
+  final String id;
+  V2(this.id);
+}
+
+main() {
+  var v = V1(3.14);
+  Expect.throws(() {v as int;});
```
eernstg, line 29: This test might go away if it is already covered by the previous one (because that one could be modified to have cases where it throws as well as cases where it doesn't).

LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A04_t01.dart
```diff
@@ -0,0 +1,26 @@
...
+import "../../Utils/expect.dart";
+
+inline class V {
+  final int id;
+  V(this.id);
+}
+
+main() {
+  Type t = V;
+  Expect.equals(int, t);
```
eernstg, line 25: Very good! It would be useful to have a generic inline class, too, such that the work done in order to find the instantiated representation type isn't a no-op.

Perhaps something like this:

```dart
inline class V<X> {
  final X Function(int) f;
  V(this.f);
}

typedef typeof<X> = X;

void main() {
  Expect.equals(V<bool>, typeof<bool Function(int)>);
}
```

eernstg (commented):
There are a couple of cases where I think it would be useful to use something like a for statement to ensure that we compare expressions like `_ is V` and `_ is int` for several different kinds of objects. The current approach actually only tests the behavior for a single object of type `V`.

sgrekhov (commented):
Tests updated. Please review once again

LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A03_t01.dart
eernstg: Great, thanks!

LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A03_t02.dart
```diff
@@ -14,21 +14,91 @@
...
+  V1(this.id);
+}
+
+inline class V2<T> {
+  final T id;
+  V2(this.id);
+}
 
-  String get foo => "Inline";
+String test(Object instance) {
```
eernstg, line 27: I think this one is a bit like a transposed matrix: We're testing that `o1 as T` and `o2 as T` will both throw or both succeed when `o1` is the value of an inline typed expression and `o2` is the value of an expression whose type is the instantiated representation type of that inline type, and we're doing that for a bunch of different values of `T`.

However, we should actually test that `o is V` and `o is R` for a bunch of different objects `o`, just like LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A03_t01.dart. (So we'd compare `o as V1` with `o as int`, and we'd compare `o as V2<String>` with `o as String`).

Basically, I think we should just have a copy of LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A03_t01.dart where it says `as` rather than `is`, and then we need to have some try/catch code around it in order to detect that we have "both are throwing" or "both succeed", not "one of each".

Let's try to extract what we can learn from this test: I think we're basically learning that `o1` and `o2` have the same run-time type (because they behave same with `o1 as T` and `o2 as T` for a set of different values of `T`). I think that's already covered by the direct test where we check that the run-time type of the value of an expression with an inline type is the representation type. So it should be fine to just change this test such that it's much more like LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A03_t01.dart.

eernstg (commented):
I think the coverage of `as` needs to be revisited, there's a comment in A03_t02 about why.

sgrekhov:
Thank you for detailed explanation! Updated

PR #2082, opened 2023-06-13
#1400. Dynamic semantics inline class test added

sgrekhov:
Please see https://github.com/dart-lang/co19/pull/2081#issuecomment-1587696658
We do have `identical(this, id)` tests but haven't `identical(x, V(x))`. This PR adds it


eernstg (approved):
LGTM!

Issue #2080, opened 2023-06-12
Add tests for "Fully Implementing an Interface"


sgrekhov:
“10.1 Fully Implementing an Interface”  and "11.2.2 Correct Member Overrides" are not covered by co19 tests.  Need to add the coverage

PR #2088, opened 2023-06-19
#2080. Add correct member overrides tests. Part 1.

sgrekhov:
This is the first part of the tests.The second part will contain tests for setters and for default values of a function parameters


eernstg (commented):
Very nice! For instance, it's good that we have a few class modifiers, but no explosion in the number of combinations. Also, all the overrides make sense.

Just one thing: We should also test the cases where the superinterface relationship is established by an `extends` or an `on` clause. Would it work if you simply copy all libraries and change `implements` to `extends` in classes, and `implements` to `on` in mixins?

eernstg (approved):
Great, LGTM!

PR #2092, opened 2023-06-21
Fixes #2087. Add subtyping tests for tearoffs of methods with covariant arguments

### LINKED ISSUE #2087: Add Subtyping tests for function types with covariant-by-declaration parameters
Add Subtyping tests for function types with covariant-by-declaration parameters

sgrekhov:
Subtyping tests are generated. So it makes sense to review the files outside of `generated` folders. Other tests are generated using those ones

### DIFF
--- a/LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A01.dart
+++ b/LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A01.dart
@@ -1,0 +1,31 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We say that a type T0 is a subtype of a type T1 (written
+/// T0 <: T1) when:
+/// Function Type/Function: T0 is a function type and T1 is Function
+///
+/// @description Check that if type `T0` is a function type with a covariant
+/// arguments and `T1` is a function type with arguments that are supertypes of
+/// `T0` arguments, then `T0` is a subtype of `T1`.
+/// @author sgrekhov22@gmail.com
+
+class A {
+  void foo(num n, [Object o = "A"]) => print("A");
+}
+
+class C implements A {
+  void foo(covariant int n, [covariant String c = "C"]) => print("C");
+}
+
+A a = A();
+A c = C();
+void Function(num n, [Object o]) t0Instance = c.foo;
+void Function(num n, [Object o]) t1Instance = a.foo;
+
+void baz(num n, [Object o = ""]) {}
+const t1Default = baz;
+
+//# @T0 = void Function(num n, [Object o])
+//# @T1 = void Function(num n, [Object o])
--- a/LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A02.dart
+++ b/LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A02.dart
@@ -1,0 +1,31 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We say that a type T0 is a subtype of a type T1 (written
+/// T0 <: T1) when:
+/// Function Type/Function: T0 is a function type and T1 is Function
+///
+/// @description Check that if type `T0` is a function type with a covariant
+/// arguments and `T1` is a function type with arguments that are supertypes of
+/// `T0` arguments, then `T0` is a subtype of `T1`.
+/// @author sgrekhov22@gmail.com
+
+class A {
+  void foo({required num n, Object o = "A"}) => print("A");
+}
+
+class C implements A {
+  void foo({required covariant int n, covariant String o = "C"}) => print("C");
+}
+
+A a = A();
+A c = C();
+void Function({required num n, Object o}) t0Instance = c.foo;
+void Function({required num n, Object o}) t1Instance = a.foo;
+
+void baz({required num n, Object o = ""}) {}
+const t1Default = baz;
+
+//# @T0 = void Function({required num n, Object o})
+//# @T1 = void Function({required num n, Object o})
--- a/LanguageFeatures/Subtyping/dynamic/tests/function_type_covariant_arg_fail_A02.dart
+++ b/LanguageFeatures/Subtyping/dynamic/tests/function_type_covariant_arg_fail_A02.dart
@@ -1,0 +1,35 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We say that a type T0 is a subtype of a type T1 (written
+/// T0 <: T1) when:
+/// Function Type/Function: T0 is a function type and T1 is Function
+///
+/// @description Check that if type `T0` is a function type with a covariant
+/// arguments and `T1` is a function type with arguments that are supertypes of
+/// `T0` arguments, then `T0` is a subtype of `T1`. But runtime invocation of
+/// `T0` with argument of `T1`'s type fails
+/// @author sgrekhov22@gmail.com
+
+import '../../../../Utils/expect.dart';
+
+class A {
+  void foo(num n) {
+    print("A");
+  }
+}
+
+class C implements A {
+  void foo(covariant int n) {
+    print("C");
+  }
+}
+
+main() {
+  A a = C();
+  void Function(num n) f = a.foo;
+  Expect.throws(() {
+    f(3.14);
+  });
+}
--- a/LanguageFeatures/Subtyping/static/test_types/function_type_covariant_arg_fail_A01.dart
+++ b/LanguageFeatures/Subtyping/static/test_types/function_type_covariant_arg_fail_A01.dart
@@ -1,0 +1,32 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We say that a type T0 is a subtype of a type T1 (written
+/// T0 <: T1) when:
+/// Function Type/Function: T0 is a function type and T1 is Function
+///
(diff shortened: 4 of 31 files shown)


eernstg:
This is a long one! Looking at it today. ;-)

eernstg:
OK, looks like most of these are generated. Looking at the following, plus one generated library, and assuming that the rest are generated in a consistent manner from these:

- LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A01.dart
- LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A02.dart
- LanguageFeatures/Subtyping/dynamic/tests/function_type_covariant_arg_fail_A02.dart
- LanguageFeatures/Subtyping/static/test_types/function_type_covariant_arg_fail_A01.dart

LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A01.dart
```diff
@@ -0,0 +1,31 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We say that a type T0 is a subtype of a type T1 (written
+/// T0 <: T1) when:
+/// Function Type/Function: T0 is a function type and T1 is Function
+///
+/// @description Check that if type `T0` is a function type with a covariant
+/// arguments and `T1` is a function type with arguments that are supertypes of
+/// `T0` arguments, then `T0` is a subtype of `T1`.
```
eernstg, line 11: We can't really say that. First, this might seem like a tiny thing, but the specification documents do use the word 'parameter' or the phrase 'formal parameter' to denote the declaration that introduces support for receiving values at the start of a method/setter invocation, and they use 'argument' or 'actual argument' to denote the actual value which is being passed. So we should say '`T0` is a function type with a formal parameter that [something]'.

Next, function types can never have a parameter which is covariant (covariant-by-declaration or covariant-by-class), function types just have formal parameters with types. It's entirely a property of a formal parameter of a method/setter declaration that one or more parameters are covariant(-by-whatever).

This also implies that there is no need to test the treatment of function types with covariant parameter types, because they don't exist.

However, a _member signature_ can have formal parameters which are covariant. When we're checking for correct override relations, we transform the member signature into a function type, and in this case we're transforming the type of each parameter which is covariant-by-declaration into a top type (that's `Object?` today) before we check that the overriding declaration's function type is a subtype of the overridden declaration's function type.

Perhaps this is what these tests do?

In any case, the static type of a tear-off of `foo` in `C` is `void Function(int, [String])`, and the static type of `foo` in `A` is `void Function(num, [Object])`, but in this case we only use the latter, so lines 24 and 25 are OK at compile time.

The run-time type of that same tear-off of `C` is `void Function(Object?, [Object?])`, so ithe initialization of `t0Instance` should also succeed at run time. The initialization of `t1Instance` should also succeed, that's the straightforward case, and so is `t1Default`.

LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A02.dart
```diff
@@ -0,0 +1,31 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We say that a type T0 is a subtype of a type T1 (written
+/// T0 <: T1) when:
+/// Function Type/Function: T0 is a function type and T1 is Function
+///
+/// @description Check that if type `T0` is a function type with a covariant
+/// arguments and `T1` is a function type with arguments that are supertypes of
+/// `T0` arguments, then `T0` is a subtype of `T1`.
```
eernstg, line 11: The comment on LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A01.dart would apply here as well.

LanguageFeatures/Subtyping/dynamic/tests/function_type_covariant_arg_fail_A02.dart
```diff
@@ -0,0 +1,35 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We say that a type T0 is a subtype of a type T1 (written
+/// T0 <: T1) when:
+/// Function Type/Function: T0 is a function type and T1 is Function
+///
+/// @description Check that if type `T0` is a function type with a covariant
+/// arguments and `T1` is a function type with arguments that are supertypes of
+/// `T0` arguments, then `T0` is a subtype of `T1`. But runtime invocation of
+/// `T0` with argument of `T1`'s type fails
```
eernstg, line 12: The comments on LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A01.dart about the terminology would apply here as well.

Otherwise this is fine!

LanguageFeatures/Subtyping/static/test_types/function_type_covariant_arg_fail_A01.dart
```diff
@@ -0,0 +1,32 @@
...
+
+/// @assertion We say that a type T0 is a subtype of a type T1 (written
+/// T0 <: T1) when:
+/// Function Type/Function: T0 is a function type and T1 is Function
+///
+/// @description Check that tearing-off a method with covariant arguments strips
+/// these arguments. Therefore, if `T0` is a function type with covariant
+/// arguments and `T0` is a valid override of function type `T1`, but without
+/// `covariant` keyword `T0` is not a valid override of `T1`, then `T0` is not a
+/// subtype of `T1`
```
eernstg, line 13: The comments on LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A01.dart about the terminology would apply here as well.

Otherwise, this is fine!

eernstg (commented):
I had several comments on the terminology used in the non-generated libraries.

For the generated libraries, I could use a bit of help in order to understand how they are generated (such that I can review the code according to the things that actually vary, rather than going over a large number of lines of code that are somehow "the same thing", just changing a small amount of code in a systematic way).

LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A01.dart
sgrekhov: May be then

> Check that if type `T0` is a function type obtained from a
> member signature with a covariant formal parameter and `T1` is a function
> type with formal parameters that are supertypes of those in `T0`, then `T0`
> is a subtype of `T1`.

?

sgrekhov:
> I had several comments on the terminology used in the non-generated libraries.
> 
> For the generated libraries, I could use a bit of help in order to understand how they are generated (such that I can review the code according to the things that actually vary, rather than going over a large number of lines of code that are somehow "the same thing", just changing a small amount of code in a systematic way).

Subtyping tests can be two types. Static and dynamic. Static tests checks that if type `T0` is subtype of type `T1` then instance of `T0` can be used everywhere where we expect type `T1`. As a local variable, as a function's return value, its formal parameter etc.

Dynamic tests do the same but at runtyme. These tests check that `forgetType(t0Instance)` can be used everywhere where we do expect type `T1`. Here `dynamic forgetType(dynamic d) => d;`

In `test_types` folder we define types `T0` and `T1`, their instances `t0Instance` and `t1Instance` and also constant `t1Default` (used as default value of optioanal formal parameter of type `T1`). Then generator use these types for every test case defined in `test_cases` folder. Test cases check that `t0Instance` can be used as local variable of type `T1`, return value etc.

LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A01.dart
eernstg: Maybe it just won't be a good match to talk about covariant parameters in the context of subtype tests at all? We don't have any types (function types, or indeed any type) with covariant parameter types, and hence it seems wrong to try to make a connection between subtype tests and tests about parameters that are covariant-by-declaration.

I think a natural thing to test would be (1) the static type of a method/setter tear-off ignores `covariant` (inherited as well as declared directly in the statically known type); `expectStaticType` should be able to handle this. (2) The dynamic type of a tear-off of a method/getter that has one or more parameters that are covariant-by-declaration has a top type as the parameter type for such parameters; tests like `e.foo is void Function(Object?, [Object?])` should be able to test this, because an actual run-time type of `void Function(T1, [T2])` would yield false for every pair of types `T1`, `T2` which is not top types. Finally, (3), we should test that a method/setter invocation and an invocation of a torn-off method/setter will actually throw if they have a parameter which is covariant-by-declaration, and the actual argument has a type which is OK for some supertype, but not a subtype of the actual parameter type of the function/method which is being called.

(OK, we can't tear off a setter, so we don't need to worry about the static or dynamic type of a torn off setter, but methods can do it, and the non-tear-off part of (3) can be done with setters as well).

We should be able to use a suitable receiver type resp. function type in order to enable testing of (3), but if there is an impossible case then we can always use a dynamic invocation.

eernstg:
Thanks, that's a very useful explanation!

Just one stray thought: If `t1Default` is the value whose type is more general then it can be used as the default value for an optional parameter of type `T1`, but not for `T0`. Is this used to create a compile-time error in the cases where it is actually used as a default value for a parameter of type `T0`? Otherwise it seems a bit surprising because I'd expect that there would be a need for a default value for a parameter of type `T0` now and then. Anyway, I was just curious.

sgrekhov:
> Thanks, that's a very useful explanation!
> 
> Just one stray thought: If `t1Default` is the value whose type is more general then it can be used as the default value for an optional parameter of type `T1`, but not for `T0`. Is this used to create a compile-time error in the cases where it is actually used as a default value for a parameter of type `T0`? Otherwise it seems a bit surprising because I'd expect that there would be a need for a default value for a parameter of type `T0` now and then. Anyway, I was just curious.

No. We don't need a default value for the type `T0`. We only have test cases like
```dart
namedArgumentsFunc1(@T1 t1, {@T1 t2 = t1Default}) {}
```
Here `@T1` is replaced by the generator by actual type `T1` from the file from `test_types` directory. Then we check that `t0Instance` (or `forgetType(t0Instance)`) can be used as actual arguments of `namedArgumentsFunc1` call. We need a constant `t1Default` only to avoid a compile-time error here. We don't need such constant for the type `T0`

eernstg:
OK, thanks again!

LanguageFeatures/Subtyping/dynamic/test_types/function_type_covariant_arg_A01.dart
sgrekhov: Please see https://github.com/dart-lang/co19/pull/2121

PR #2093, opened 2023-06-22
#2080. Add correct member overrides tests. Part 2. Setters

sgrekhov:


Language/Interfaces/Correct_Member_Overrides/correct_member_overrides_A03_t19.dart
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Let m and m′ be member signatures with the same name id. Then m
+/// is a correct override of m′ iff the following criteria are all satisfied:
+/// ...
+/// • If m and m′ are both methods or both setters: Let F be the function type
+/// of m except that the parameter type is the built-in class Object for each
```
chloestefantsova, line 9: I would expect to see `Object?` instead of `Object` in this line. Could you point me to the part of the spec this quote is taken from?

chloestefantsova: Since it's a comment, it can be resolved in a follow-up.

sgrekhov: Thank you! Spec link https://github.com/dart-lang/language/blob/6d50316bac4e4f1d17367bb5440fe5d2da298ad3/specification/dartLangSpec.tex#L5373

chloestefantsova: Thanks! That looks interesting :)

eernstg: With null safety, that should be `Object?`.

eernstg: Spec updates for null safety are in review (...since september 2021...). It is in the PR dart-lang/language#2605: It was missing, but I just corrected that. So it will be landed along with all the other null safety changes.

Issue #2094, opened 2023-06-22
Update mixin tests to remove // @dart=2.19


sgrekhov:
A lot of mixin tests contain
```
// TODO(https://github.com/dart-lang/sdk/issues/51557): Decide if the mixins
// being applied in this test should be "mixin", "mixin class" or the test
// should be left at 2.19.
// @dart=2.19
```
Now class modifiers are part of the language so we can remove this TODO and the language version simply changing `class -> mixin class`

PR #2108, opened 2023-07-03
#2094. Move and rename mixin tests to follow the current spec structure

sgrekhov:
`co19/Language/Mixins/*` tests are all green on all configurations. So, we agreed with @athomas that it is safe to rename such tests


eernstg:
Moved `syntax_t26.dart` to `syntax_t11.dart`, `syntax_t27.dart` to `syntax_t12.dart`, changes a couple of import URIs from double quote to single quote format. All other changes are shown correctly as renames in the github 'files changed' UI.

eernstg (approved):
LGTM

eernstg (commented):
Oops, one more change needed: deferred_t02.dart and deferred_t03.dart have a mixin application class declaration (the kind that uses `=`) with a class body (like `class C = A with M {}`). We should change that such that the reported compile-time error has the correct fault to complain about.

sgrekhov:
Thank you for noticing! Fixed

eernstg (approved):
Thanks! LGTM

eernstg (commented):
Oops, one more thing: `deferred_t01.dart` was changed, too, and there was no syntax error in that one (that's not a mixin application class). I think that one should just be changed back to have `{}`.

sgrekhov:
Really. Sorry. Reverted

eernstg (approved):
No problem, LGTM!

PR #2121, opened 2023-07-10
#2087. Add tests for parameters covariant-by-declaration

sgrekhov:


Language/Classes/Instance_Methods/covariant_A02_t03.dart
```diff
@@ -0,0 +1,49 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion For each parameter p of m where covariant is present, it is a
+/// compile-time error if there exists a direct or indirect superinterface of C
+/// which has an accessible method signature m′′ with the same name as m, such
+/// that m′′ has a parameter p′′ that corresponds to p, unless the type of p is
+/// a subtype or a supertype of the type of p′′
+///
+/// @description Checks runtime type of a tear-off of a method with a covariant
+/// parameter
```
eernstg, line 12: We're testing the dynamic type check here, which is performed according to the declared parameter type of the implementation that actually runs. Perhaps:

```suggestion
/// @description Test that the actual argument is subject to a dynamic type
/// check with the type of the parameter as declared in the actually running
/// implementation.
```

Language/Classes/Instance_Methods/covariant_A02_t03.dart
```diff
@@ -0,0 +1,49 @@
...
+  void m3({covariant int a = 1}) {}
+  void m4({required covariant int a}) {}
+}
+
+main() {
+  C c = C();
+  dynamic d1 = c.m1;
+  Expect.throws(() {
+    d1(Object());
+  });
```
eernstg, line 36: We should also verify that it does not throw with an `int`.

Language/Functions/Formal_Parameters/Covariant_Parameters/no_instance_t01.dart
```diff
@@ -0,0 +1,72 @@
...
+  static void f4({required covariant int i}) {}
+//                         ^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  static void set s1(covariant int i) {}
+//                   ^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
```
eernstg, line 63: Needs local function declarations as well (could be in `main`), and function literals.

sgrekhov:
Tests updated. Please take another look

eernstg (commented):
Looks good! One little bit still missing:

I was worried when I saw that some tests have `covariant` in more than one location in the superinterface graph. The strongest test occurs when it is in just one place (the `extends` type, the `implements` type, the `on` type, or in the newly declared class itself), so we should cover the different kinds of locations with just one occurrence of `covariant`.

A couple of tests have multiple occurrences (Language/Classes/Instance_Methods/covariant_A01_t05.dart and Language/Classes/Instance_Methods/covariant_A01_t06.dart). It's fine to have that test case, too, but we should make sure that we're covering each location separately.

I think there's no occurrence of `covariant` in an `on` type?

Language/Functions/Formal_Parameters/Covariant_Parameters/no_instance_t01.dart
eernstg: Thanks!

sgrekhov:
Updated. Excessive `covariant` were just a copy/paste effect. The intention was to left only one `covariant` in the superinterface graph. Thank you for noticing. Removed.

As for `mixin M on` I forgot about it. Added appropriate tests.

There will be one renaming PR after the merge to reorder these tests. It's much simpler to manage the tests if the similar tests are following each other

eernstg (approved):
Thanks, LGTM!

eernstg (commented):
Oh, one thing needs to be reconsidered after all:

I'm worried that we won't get the static analysis if there are syntax errors. But `no_instance_t01.dart` has a syntax error: Line 87 declares a setter as a local function, but that's a syntax error. The spec_parser complains about line 112-120 or so as well, but that disappears when the setter has been removed.

Also `h2` needs to use parameter type `int?` or have a default value such that we get an error for the correct reason (everything other than the word `covariant` should be correct).

sgrekhov:
Ups, sorry. Fixed.

Issue #2117, opened 2023-07-06
Complete tests for "17.21.3 Superinvocations"


sgrekhov:
We have no tests covering the case when 

```dart
abstract class A { 
  void m(); 
}

abstract class B extends A { 
  void n() { 
    super.m();  // The method 'm' is always abstract in the supertype. • abstract_super_member_reference
  }
}
```
Possibly there are other missing cases as well

sgrekhov:
Add tests for `external` method in a superclass

PR #2122, opened 2023-07-11
#2117. Update accessible_instance_member_t* tests

sgrekhov:
`accessible_instance_member_t03-05` tests were deleted because they have too few differences from `..._t01`.


eernstg (commented):
LGTM.

I noticed one thing: The `@assertion`s are talking about a lookup that succeeds or fails, and the `@description` (in at least one case) talks about a class that 'has' or 'doesn't have' a member with a specific name. This is slightly dangerous because 'have a member' usually refers to an interface (where abstract members count just as much as concrete ones) and 'lookup' is about an actual implementation lookup (where abstract members are ignored).

Anyway, that was just a random comment. Landing!

eernstg (approved):

sgrekhov:
I'll reopen the issue to fix description texts. We also should add a test for the case when member `m` in a superclass is `external` (has no body but not an abstract)

PR #2133, opened 2023-07-14
Fixes #2127. Add tests for covariant-by-class parameters

### LINKED ISSUE #2127: Add co19 tests checking runtime and static types of class members with of covariant-by-class parameters
For class like
```dart
class C<T> {
  void foo(T t) {}
}
```
we should expect that static type of `C<int>.m` is `void Function(int)` but dynamic type `void Function(Object?)`. At run time a type check is performed to ensure that actual type argument is `int`

sgrekhov:


Language/Classes/Instance_Methods/covariant_A04_t01.dart
```diff
@@ -0,0 +1,26 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion For each parameter p of m where covariant is present, it is a
+/// compile-time error if there exists a direct or indirect superinterface of C
+/// which has an accessible method signature m′′ with the same name as m, such
+/// that m′′ has a parameter p′′ that corresponds to p, unless the type of p is
+/// a subtype or a supertype of the type of p′′
+///
+/// @description Checks that tearing-off a method with a covariant-by-class
+/// parameter strips a `covariant` keyword
```
eernstg, line 12: I'm not sure I understand this description. ;-)

I can't see any occurrences of `covariant` in this test, so there's nothing to strip, and if we're talking about function types then they don't ever contain that keyword, so it doesn't make sense to say that we're stripping it off of a function type.

I guess the test is about the static type of a tear-off of an instance method where a parameter is covariant-by-class?

Language/Expressions/Method_Invocation/Instance_Method_Closurization/covariant_parameter_t01.dart
```diff
@@ -0,0 +1,57 @@
...
+typedef Expected2 = void Function([Object?]);
+typedef Expected3 = void Function({Object? a});
+typedef Expected4 = void Function({required Object? a});
+
+main() {
+  C c = C();
+  Expect.equals(Expected1, c.m1.runtimeType);
+  Expect.equals(Expected2, c.m2.runtimeType);
+  Expect.equals(Expected3, c.m3.runtimeType);
+  Expect.equals(Expected4, c.m4.runtimeType);
```
eernstg, line 55: Does that actually work? The specification says that a function object has a run-time type which is a subtype of the function type that we know about. For instance, the VM is allowed to use different internal generated classes for function literals and for top-level functions (or whatever kind of function it needs to give some special treatment). So we could have a `FunctionLiteralVoidFunctionInt` rather than precisely a `void Function(int)` for a function whose declaration has that function type, and we'd have (1) `FunctionLiteralVoidFunctionInt <: void Function(int)`, and if `F` is a function type such that `F <: void Function(int)` and they are not mutual subtypes then we _do not_ have `FunctionLiteralVoidFunctionInt <: F`. (In other words, the function type which is taken from the declaration that gave rise to the function object is the most precise function type that we can give for that function object.)

So even if it works now, I'd say that it isn't future-safe to rely on equality of `runtimeType`, and we should probably just check that the function object `is That Function(Type)`.

Language/Expressions/Method_Invocation/Instance_Method_Closurization/covariant_parameter_t02.dart
```diff
@@ -0,0 +1,49 @@
...
+typedef Expected2 = void Function([Object?]);
+typedef Expected3 = void Function({Object? a});
+typedef Expected4 = void Function({required Object? a});
+
+main() {
+  C<int> c = C<int>();
+  Expect.equals(Expected1, c.m1.runtimeType);
+  Expect.equals(Expected2, c.m2.runtimeType);
+  Expect.equals(Expected3, c.m3.runtimeType);
+  Expect.equals(Expected4, c.m4.runtimeType);
```
eernstg, line 48: Same consideration, probably needs to use `is`.

Language/Expressions/Method_Invocation/Instance_Method_Closurization/covariant_parameter_t03.dart
```diff
@@ -0,0 +1,60 @@
...
+/// where f is an instance method named m which has type parameter declarations
+/// X1 extends B1, . . . , Xs extends Bs, required parameters p1, . . . , pn,
+/// and optional positional parameters pn+1, . . . , pn+k with defaults
+/// d1, . . . , dk.
+/// ...
+/// For each parameter pj , j ∈ 1..n + k, if pj is covariant then Tj is the
+/// built-in class Object?.
+///
+/// @description Check that it is a run-time error if a run-time type of
+/// covariant-by-class parameter is not a subtype `Object?` of a declared type
```
eernstg, line 26: Perhaps:
```suggestion
/// @description Check that it is a run-time error if the run-time type of the
/// actual argument passed to a covariant-by-class parameter is not a 
/// subtype of the declared type of that parameter
```

eernstg (commented):
Looks good! However, I do think it's a bit too unsafe to test against the `runtimeType`.

Language/Expressions/Method_Invocation/Instance_Method_Closurization/covariant_parameter_t01.dart
sgrekhov: It worked. But Ok, chenged to `is`

eernstg: Thanks!

eernstg (approved):
LGTM

PR #2153, opened 2023-07-27
Fixes #2139. Add generator functions element type tests

### LINKED ISSUE #2139: Test element type of generator functions
We've noticed a soundness issue with the element type of generator functions, as mentioned in
https://github.com/dart-lang/language/pull/3218. 

@sgrekhov, it would be great if you could star that pull request, and write some language tests when it has been resolved!

In particular, we should cover the currently unsound behavior associated with examples given [here](https://github.com/dart-lang/language/issues/3148#issuecomment-1607381741), and we should cover the run-time types of objects returned by generator functions. We'd expect `is Iterable<T>` for a suitable `T`, or `is Stream<T>`. It's a bit less convenient to verify that the iterable/stream has an actual type argument which isn't too special (if we expect an `Iterable<num>` we shouldn't get an `Iterable<int>`), but it should be possible to build on this idea:

```dart
X combine<X>(X x, X _) => x;

void main() {
  List<num> xs = <int>[1];
  print('Combine for int:');
  (xs.reduce as dynamic)(combine<int>); // Type correct at run time. Ignore the type checker.
  print('Still alive. Combine for num:');
  xs.reduce(combine<num>);
  print('Not reached');
}
```

sgrekhov:


Language/Functions/element_type_A03_t01.dart
```diff
@@ -0,0 +1,80 @@
...
+  yield null;
+  yield "42";
+  yield Future<Null>.value(null);
+  yield Future<num>.value(1);
+  yield Future<void>.value(1);
+  yield Future<void>.value(null);
+}
+
+main() {
+  Expect.iterableElementsRuntimeIs<dynamic>(f1());
```
eernstg, line 74: Same issue again about testing that the run-time type implements something (perhaps `iterableElementsRuntimeIs` does this already, but then we should just use that everywhere).

Language/Functions/element_type_A03_t02.dart
```diff
@@ -0,0 +1,92 @@
...
+  yield Future<Null>.value(null);
+  yield Future<num>.value(1);
+  yield Future<void>.value(1);
+  yield Future<void>.value(null);
+}
+
+main() async {
+  dynamic r1 = await f1();
+  var v1 = await r1.toList();
+  Expect.iterableElementsRuntimeIs<dynamic>(v1);
```
eernstg, line 76: Same issue again, run-time type implements..

Utils/expect_common.dart
```diff
@@ -355,6 +355,44 @@ class Expect {
...
+        throw ExpectException("Not $T in collection $it.");
+      }
+    }
+  }
+
+  /// Checks that all elements of this `Iterable` don't have type `T`, but may
+  /// be a subtype of `T`. For example
+  /// `iterableElementsRuntimeIsNot<num>([1, 2, 3])` is Ok, but
+  /// iterableElementsRuntimeIsNot<num>([1, 2, 3 as num]) throws an
+  /// `ExpectException`
```
eernstg, line 381: Same kind of issue here: We can't use the elements to test anything like "this iterable does not implement `Iterable<T>` for a given `T`".

Utils/tests/Expect/iterableElementRuntimeIsNot_A01_t01.dart
```diff
@@ -0,0 +1,47 @@
...
+/// which is not exactly the `T`
+///
+/// @description Checks no exception is thrown when `it` is an `Iterable<T>` and
+/// all elements of it have a type `T` (not a subtype)
+/// @author sgrekhov22@gmail.com
+
+import "dart:async";
+import "../../../Utils/expect.dart";
+
+main() {
```
eernstg, line 16: If `iterableElementsRuntimeIs` and `...Not` are updated then this test might just remain unchanged. However, I suspect that there won't be a `...Not`, because that one should be covered by the other one.

Also, it should probably be called `runtimeTypeImplementsIterableOf<X>()` and invoked as `<num>[].runtimeTypeImplementsIterableOf<T>()`, and it would succeed when `T` is `num` and fail when `T` is `Object` or `int`.

Utils/tests/Expect/iterableElementRuntimeIs_A01_t01.dart
```diff
@@ -0,0 +1,51 @@
...
+///
+/// @description Checks no exception is thrown when `it` is an `Iterable<T>` and
+/// all elements of it have a type `T` (not a subtype)
+/// @author sgrekhov22@gmail.com
+
+import "dart:async";
+import "../../../Utils/expect.dart";
+
+main() {
+  Expect.iterableElementsRuntimeIs<int>([1, 2, 3]);
```
eernstg, line 17: A similar comment as for Utils/tests/Expect/iterableElementRuntimeIsNot_A01_t01.dart would fit here.

Utils/tests/Expect/iterableElementRuntimeIs_A01_t02.dart
```diff
@@ -0,0 +1,30 @@
...
+/// which is not exactly the `T`
+///
+/// @description Checks no exception is thrown when `it` is an `Iterable<T>` and
+/// all elements of it have a type `T` (not a subtype)
+/// @author sgrekhov22@gmail.com
+
+import "dart:async";
+import "../../../Utils/expect.dart";
+
+main() {
```
eernstg, line 16: I think a `runtimeTypeImplementsIterableOf<X>()` should not throw, it should return true or false. In that case this test would just expect false rather than a dynamic error.

eernstg (commented):
Looks good, but I do have a few comments.

There are many tests with a description similar to "Check that it is a compile-time error if an element type of a synchronous generator function `f` is not `U`, where `S` is a union-free type of the declared return type of `f` and `S` implements `Iterable<U>`"

I can't read that description. ;-)

First, there's no such thing as 'an element type' of a generator function, there is one specific type which is 'the element type' of that function. Next 'is not `U`' is far too flexible: There is an infinity of ways in which we can make the requirements true, and still satisfy 'is not `U`'.

So we'd have to say something like this:

"Check that if the declared return type of a synchronous generator function `f` is `T`, and `S` is the union-free type derived from `T`, and `S` is of the form `Iterable<U>`, it is a compile-time error to `yield` an expression whose static type isn't assignable to `U`."

But that's just the same thing as this:

"Check that for a synchronous generator function `f` with element type `U` which is not `dynamic`, it is a compile-time error to `yield` an expression whose static type isn't assignable to `U`."

With that, the 'asynchronous' variant would be achieved simply by replacing a space by 'n a'. ;-)

I think all the descriptions containing 'an element type' needs a similar adjustment, but please check all descriptions (I didn't confirm that they _always_ contain 'an element type').

Next, there are many tests where we need to check that a given object has a run-time type that implements `Iterable<T>` for a specific `T`. Here is a way to perform that check:

```dart
// Testing method.

bool runtimeTypeImplementsIterableOf<X>(Object? o) {
  if (o is! Iterable<X>) return false;
  List<X> list = o.toList(growable: true);
  try {
    list.addAll(<X>[]);
  } on TypeError catch (_) {
    return false;
  }
  return true;
}

// Example usage.

void main() {
  Iterable<dynamic> xs = <num>[1];

  print(xs.runtimeType == List<num>); // 'false', it is some subtype.
  print(runtimeTypeImplementsIterableOf<Object>(xs)); // 'false'.
  print(runtimeTypeImplementsIterableOf<num>(xs)); // 'true'.
  print(runtimeTypeImplementsIterableOf<int>(xs)); // 'false'.
}
```

First, we can't use the runtimeType `R` of the target object `o` for anything: It's going to be some subtype of the statically known type, and it certainly won't be `Iterable` or `List` because they are abstract. Next, we can establish that `R <: Iterable<X>`. Finally, we need to ensure that the type argument of `R` at `Iterable` isn't a more special type (e.g., we wouldn't want to accept an object with run-time type that implements `Iterable<Never>`). We can do that by getting hold of the corresponding `List` (so it won't work if it is an infinite iterable, but I think we can handle that), and then test `addAll`.

I mentioned `toList` and `addAll` in several tests, but it's of course much better to have a standard testing function like the one above.

The current version of `iterableElementsRuntimeIs` might be trying to do a similar thing, but it won't work.

Language/Functions/element_type_A01_t03.dart
```diff
@@ -0,0 +1,30 @@
...
+
+FutureOr<Iterable<int>?> foo() sync* {
+  yield 1;
+  yield 2;
+  yield 3;
+}
+
+main() {
+  FutureOr<Iterable<int>?> o = foo() as dynamic;
+  o as FutureOr<Iterable<int>?>;
```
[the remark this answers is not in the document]
sgrekhov: Changed to
```dart
  o as FutureOr<Iterable<int>>;
  Expect.isRuntimeTypeIterable<int>(o);
```

Language/Functions/element_type_A02_t03.dart
```diff
@@ -0,0 +1,32 @@
...
+FutureOr<Stream<int>?> foo() async* {
+  yield 1;
+  yield 2;
+  yield 3;
+}
+
+main() async {
+  dynamic d = await foo();
+  FutureOr<Stream<int>?> o = d;
+  o as FutureOr<Stream<int>?>;
```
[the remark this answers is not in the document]
sgrekhov: Changed to
```dart
  FutureOr<Stream<int>> o = d;
  isRuntimeTypeStream<int>(d);
```

Language/Functions/element_type_A02_t04.dart
```diff
@@ -0,0 +1,42 @@
...
+FutureOr<Stream<int>?> foo() async* {
+  yield 1;
+  yield 2;
+  yield 3;
+}
+
+main() async {
+  var o = await foo();
+  o.expectStaticType<Exactly<Stream<int>?>>();
+  if (o is Future<Stream<int>?>) {
```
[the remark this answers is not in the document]
sgrekhov: Indeed. Updated

Language/Functions/element_type_A02_t06.dart
```diff
@@ -0,0 +1,57 @@
...
+
+FutureOr<Stream<int>?>? f2() async* {
+  yield 1;
+  yield 2;
+  yield 3;
+}
+
+main() async {
+  var res1 = await f1();
+  var it1 = await res1?.toList();
```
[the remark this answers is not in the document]
sgrekhov: The same as the above

sgrekhov (commented):
Tests updated. I renamed checking functions in `Expect` to `Expect.isRuntimeTypeIterable` and `Expect.isNotRuntimeTypeIterable`. Looks like this naming is more in consistance with the [dart style](https://dart.dev/effective-dart/design#prefer-putting-the-most-descriptive-noun-last). If you have better namig ideas, please let me know

I also preserved throwing exceptions by these functions. If we change it to return `bool` the we'll have to use `Expect.isTrue(Expect.(isRuntimeTypeIterable(...)));`. First, we never write a code this way (`Expect.foo(Expect.baz(...))`). Second, in case of failure we'll get an error `Expect.isTrue(false) fails.`. `Expect.isRuntimeTypeIterable<int>(o);` produces `Not Iterable<int>: _SyncStarIterable<dynamic>` which is much more informative

Language/Functions/element_type_A01_t01.dart
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the union-free type of a type `T` as follows:
```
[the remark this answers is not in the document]
eernstg: Thanks!

Lasse's PR was not intended to change anything about the generator element types, so the potential further changes that I mentioned will not occur after all.

Language/Functions/element_type_A03_t01.dart
```diff
@@ -0,0 +1,80 @@
...
+///
+/// We define the element type of a generator function `f` as follows:
+/// Let S be the union-free type derived from the declared return type of f.
+/// ...
+/// Otherwise, if `f` is a generator (synchronous or asynchronous) and `S` is a
+/// supertype of `Object` then the element type of `f` is dynamic
+///
+/// @description Check that element type of a synchronous generator function `f`
+/// is `dynamic`, if `S` is a union-free type of the declared return type of `f`
+/// and `S` is a supertype of `Object`
```
eernstg, line 18:
```suggestion
/// @description Let `f` be a synchronous generator function whose declared return
/// type derives a union-free type which is a supertype of `Object` (including `Object` itself).
/// Then check that its element type is `dynamic`.
```

It shouldn't actually make a difference whether we're talking about the declared return type or the union-free type derived from the declared return type:

If the declared return type is a supertype of `Object` then it is of the form `Object`, `Object?`, `void`, `dynamic`, or `FutureOr<T>` or `T?` where `T` is a supertype of `Object`. For each of those types it is true that it derives a union-free type which is also a supertype of `Object`.

Conversely, if the declared return type `T` derives a union free type `S` which is a supertype of `Object` then `T` itself is `S` wrapped in some number of `FutureOr<_>` and `_?`, and they are also supertypes of `Object`.

Hence, the set of types that derive a union-free type which is a supertype of `Object` is the same set of types as the ones that are themselves a supertype of `Object`.

Anyway, that's nice to know, but let's just stick to the specified rules and keep "that derives a union-free type".

Language/Functions/element_type_A01_t05.dart
```diff
@@ -0,0 +1,50 @@
...
+///
+/// We define the element type of a generator function `f` as follows:
+/// Let S be the union-free type derived from the declared return type of f. If
+/// f is a synchronous generator and S implements Iterable<U> for some U then
+/// the element type of f is U.
+///
+/// @description Check that runtime type of an element type of a synchronous
+/// generator function `f` is `U`, where `S` is a union-free type of the
+/// declared return type of `f` and `S` implements `Iterable<U>`. Test the
+/// run-time types of objects returned by a synchronous generator function
```
eernstg, line 18: Needs an update.

Language/Functions/element_type_A01_t06.dart
```diff
@@ -0,0 +1,51 @@
...
+///
+/// We define the element type of a generator function `f` as follows:
+/// Let S be the union-free type derived from the declared return type of f. If
+/// f is a synchronous generator and S implements Iterable<U> for some U then
+/// the element type of f is U.
+///
+/// @description Check that runtime type of an element type of a synchronous
+/// generator function `f` is `U`, where `S` is a union-free type of the
+/// declared return type of `f` and `S` implements `Iterable<U>`. Test the
+/// run-time types of objects returned by a synchronous generator function
```
eernstg, line 18: Needs an update.

eernstg (commented):
Still some things to think about. ;-)

I think it's really important to make an explicit distinction between `o is Iterable<T>` for some given `T`, and "the run-time type of `o` _implements_ `Iterable<T>`", because the latter is strictly more demanding than the former. Based on that consideration, I've suggested a couple of renames.

Given that we're using a strong test, there is probably no need for `isNotRuntimeTypeIterable` at all, so I've suggested that it gets deleted.

Finally, a couple of descriptions still use the phrase 'an element type' of a function, or some other wording that is confusing.

sgrekhov:
Tests updated. Testing function is renamed to include `Implements` and `...Not...` version deleted. Please take another look.

Language/Functions/element_type_A01_t05.dart
```diff
@@ -0,0 +1,50 @@
...
+}
+
+Iterable<int>? f2() sync* {
+  yield 1;
+  yield 2;
+  yield 3;
+}
+
+main() {
+  Expect.iterableElementsRuntimeIsNot<num?>(f1());
```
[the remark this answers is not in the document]
eernstg: Thanks!

Language/Functions/element_type_A02_t03.dart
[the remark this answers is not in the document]
eernstg: Thanks!

eernstg (approved):
LGTM!

Issue #2142, opened 2023-07-20
Feature renamed to extension type


eernstg:
The language team just committed to the rename: The 'inline class' feature has been renamed as the 'extension type' feature. I think it would be helpful for everybody to have one or more PRs performing renaming of directories, test names, `@description` sections, etc.

The only exception is the experiment flag: This flag will go away as soon as the feature has been released, so it might as well stay the same.

eernstg:
Some properties of extension types are different from the corresponding properties of inline classes, so we'll need some additional PRs about the expected behavior.

In particular, there is no instance variable declaration, and the representation name and type are specified as `(<type> <identifier>)` in the declaration header (the meaning of which is the same as `final <type> <identifier>;` in the declaration body plus a constructor that does `this.<identifier>` to initialize it).

Another new thing: It is allowed for an extension type to declare that it `implements .... T ....` for some `T` that aren't extension types.

But let's look into that later on, for now we just need to stop expecting an error for things like `extension type E(int i) implements int {}`.

eernstg:
Just confirmed with @johnniwinther: The flag that enables the extension type feature will be `inline-class`, so that particular bit escapes the rename. Other than that, the phrase 'inline class' and terms like 'inline-class' should not occur anywhere.

eernstg:
Perhaps this could be done in two steps, such that the diffs will be informative: First move all affected tests from `/inline-classes/` to `/extension-types/` (I probably got the directory name somewhat wrong, but the idea should be clear), and then change the tests to use the new syntax. Or vice versa, whatever is most convenient.

If we do both things in the same PR then I'm afraid it's going to look like deleting a bunch of libraries and adding a bunch of different libraries, with no connection between the two.

sgrekhov:
Why do we need a connection between obsolete inline classes tests and the new extension types ones? Why not to simply have  new tests first, and then delete the old ones? Otherwise there will be mess with renaming, large diffs etc

eernstg:
I thought the "move  then edit" approach would yield the smallest diffs:

If there is no connection then the diff will be 2 times everything (one for delete old library, one for add new library, no connection), and it would be necessary to start from scratch with every test. It's very easy to skip over the 'delete file' changes, but this is still 1 times everything.

[...]

PR #2165, opened 2023-08-07
#2142. Add more static analysis tests

sgrekhov:


LanguageFeatures/Extension-types/static_analysis_extension_types_A02_t01.dart
```diff
@@ -22,19 +22,19 @@ extension type ET1(int id) {
 // [analyzer] unspecified
 // [cfe] unspecified
 
-  ET1.n2(int id) : this();
+  ET1.n2(int id) : this(id);
```
chloestefantsova, line 25: This case doesn't look like it's a compile-time error. Constructor `ET1.n2` declares a formal parameter, `id`, and passes it as the argument to the constructor that initializes the field with the same name. By analogy, the following Dart code compiles and runs without errors:

```dart
class A {
  int foo;
  A(this.foo);
  A.named(int foo): this(foo);
}

main() {}
```

/cc @eernstg

LanguageFeatures/Extension-types/static_analysis_extension_types_A02_t01.dart
```diff
@@ -22,19 +22,19 @@ extension type ET1(int id) {
...
 //^^^
 // [analyzer] unspecified
 // [cfe] unspecified
 
-  ET1.n3(int id) : this.new();
+  ET1.n3(int id) : this.new(id);
```
chloestefantsova, line 30: Similarly here. This line doesn't look like a compile-time error to me.

chloestefantsova (changes requested):

LanguageFeatures/Extension-types/static_analysis_extension_types_A02_t01.dart
eernstg: I also don't see any compile-time errors here.

LanguageFeatures/Extension-types/static_analysis_extension_types_A02_t01.dart
eernstg: Agreed again.

eernstg (commented):
Agreeing. ;-)

eernstg:
What's the next step with this PR?

sgrekhov:
Thank you!  It's an error itself to declare constructor named `new` in that test. Fixed. Please take another look

eernstg:
I think it's safer to use the terminology which is also used in the language specification: `class C { C(); C.named(); }` declares a constructor named `C` and another constructor named `C.named`. There is no such thing as a nameless or an anonymous constructor, constructors have names of the form `<typeIdentifier> ('.' <identifier>)?`.

If we use that terminology then we can only have a constructor named `new` in a class named `new`, which is a syntax error because `new` is a reserved word. Also, if we're considering a constructor declaration of the form `ET.new(int i)` (which could be a primary constructor or a non-primary constructor) then its name is `ET`, not `ET.new`. So a constructor named `new` is not really a thing.

Also, a primary as well as a non-primary constructor can have this kind of declaration. (For the "primary" constructor it's currently specified as `<typeName> <typeParameters>? <representationDeclaration>`, but the `<representationDeclaration>` does allow `new` as in `ET.new(int i)`.)

So why would we have this error?:

> It's an error itself to declare constructor named `new`

eernstg:
Ah, OK! Thanks!

chloestefantsova (approved):
LGTM. Thanks!

Issue #2169, opened 2023-08-10
Test fails on dart2js


johnniwinther:
There are new test failures on [[cfe] Compute instantiated representation type](https://dart-review.googlesource.com/c/sdk/+/319481).

The test
```
co19/LanguageFeatures/Inline-classes/dynamic_semantics_inline_class_A02_t01 RuntimeError (expected Pass)
```
is failing on configurations
```
dart2js-hostasserts-linux-chrome-unsound
dart2js-hostasserts-linux-d8
dart2js-linux-chrome
```

The test uses the runtime type of a list literal, but I think that is not well-defined, at least it fails a runtime on dart2js because the runtime type is `JSArray` and not `List` as expected. @sigmundch Is this the intended behavior?

The test could be rewritten to use a user defined class to avoid this dependency:
```
inline class V1 {
  final int id;
  V1(this.id);
}

inline class V2<T> {
  final T id;
  V2(this.id);
}

class Class<T> {
  T field;
  Class(this.field);
}

main() {
  List<V1> l1 = [];
  Expect.equals(List<int>, l1.runtimeType);
  (l1 as List<int>).add(42);
  Expect.throws(() { (l1 as dynamic).add(3.14);});

  var l2 = Class<V2<String>>(0);
  Expect.equals(Class<String>, l2.runtimeType);
  (l2 as Class<String>).field = "42";
  Expect.throws(() { (l2 as dynamic).field = 42;});
}
```

rakudrama:
I agree that this is a bad test.
The test would not work on the VM if the Lists were changed to Sets or Maps or pretty much anything else.

[`List` is an `abstract` class](https://github.com/dart-lang/sdk/blob/58f4b3a7a6042b0968d7391bb0df7b2dd9e88904/sdk/lib/core/list.dart#L115), so only by some kind of magic could any instance have exactly that type.

PR #2188, opened 2023-08-11
#2142. Change dynamic semantics inline class tests to be extension type tests

sgrekhov:
This also fixes https://github.com/dart-lang/co19/issues/2169


eernstg (approved):
LGTM

PR #2215, opened 2023-08-17
#2142. Add Subtyping tests for extension types

sgrekhov:
Subtyping tests needs some comments, as usual. It makes sense to review `extension_type_*` files in the folders:
`/Subtyping/test_types/`
`/Subtyping/static/test_types/`
`/Subtyping/dynamic/test_types/`
All other tests are generated.

Tests in `/Subtyping/static/test_types/` test static behaviour. Generated tests will look like
```dart
T0 t = t1; // Expect a compile-time error if static type of t1 is not subtype of T0
```

Tests in `/Subtyping/dynamic/test_types/` test  runtime behaviour. Generated tests will look like
```dart
dynamic forgetType(dynamic d) => d;

T0 t = forgetType(t1); // Expect a run-time error if run-time type of t1 is not subtype of T0
```
Tests in `/Subtyping/test_types/` test both static and runtime behavior. In case of extension types we have to test some cases dynamically or statically only. For example
```dart
  extensuin type V(int id) {}
  int i1 = V(42); // Error at compile time
  int i2 = forgetType(V(42)); // No error at run time
```
So we have to test that extension type `V` is not assignable to `int` in static tests only. In dynamic tests we check that `forgetType(V(...))` is assignable to `int` and vice-versa.

Please review


eernstg:
Having looked at the libraries you suggested, I'm not sure about `implements` relations: Do we have subtype tests confirming that `E1 <: E2`, `E0 <: int`, and not `E2 <: E1` and not `int <: E0` (and similarly for some generic types):

```dart
extension type E2(int i) {}
extension type E1(int i) implements E2 {}
extension type E0(int i) implements int {}
```

sgrekhov:
Added missing cases. Please take another look

eernstg (commented):
I'm afraid there's a lot of trailing whitespace. I suppose it can be `grep`ped where it occurs in non-generated files (or is created by the generator)? Otherwise it's good!

sgrekhov:
Updated. Trailing whitespaces removed (yes, it was generator issue and fix affected all tests). But we need to review `/test_types/` only. Please note, that one more test type added.  `/Subtyping/test_types/extension_type_A08.dart`

eernstg:
> Please note, that one more test type added. `/Subtyping/test_types/extension_type_A08.dart`

OK, thanks! It's very helpful to know which parts are actually changed.

However, I'm confused about the changes overall anyway: It used to say (I think) hundred-and-fifty-something files changed; next time it said 253 files changed (I think); but now it says 2840 files changed.

What's going on that changes so many files?

eernstg:
Ah, it's because almost everything had those trailing spaces, and now they have been regenerated without them. OK!

PR #2218, opened 2023-08-17
Fixes #2216. Add test for extension type abstract members

### LINKED ISSUE #2216: Extension types: An abstract member is a compile-time error
We should have tests to verify that this error is reported. Do we have that already?

sgrekhov:


eernstg (approved):
LGTM

Issue #2217, opened 2023-08-17
Extension types: Updates regarding covariance and `UP`


eernstg:
If and when https://github.com/dart-lang/language/pull/3276 lands, we need additional tests:

If `E` is a generic extension type and `S1 <: T1` .. `Sk <: Tk` then `E<S1 .. Sk>` is a subtype of `E<T1 .. tk>`.

It also specifies that the Dart 1 algorithm for computation of the least upper bound of two types (the one which is used in the last case of the function UP, [here](https://github.com/dart-lang/language/blob/main/resources/type-system/upper-lower-bounds.md#:~:text=UP(C0%3CT0%2C%20...%2C%20Tn%3E%2C%20C1%3CS0%2C%20...%2C%20Sk%3E)%20%3D)), must take `Object?` into account:

```dart
extension type E1(Object? o) {} // Depth 1
extension type E2(Object? o) implements E1 {} // Depth 2
extension type E3a(num n) implements E2, num {} // Depth 3
extension type E3b(num n) implements E2, num {} // Depth 3
```

In this case, with the old rule, `E1` and `E2` would not have a well-defined depth, but `num` had depth 1. `E3a` and `E3b` could have depth 2, or it could be undefined because some superinterface paths do not reach `Object`.

We need to consider the depth from `Object?` rather than `Object`, so all the "old" types are now 1 step deeper (`num` now has depth 2), and every type has a well-defined depth (in particular, `E3a` and `E3b` have depth 3).

For `UP(E3a, E3b)` we get `Object?`, because `E2` and `num` both have depth 2.

eernstg:
(Looks like we have covariance tests already in https://github.com/dart-lang/co19/pull/2215, but that shouldn't hurt. The landed version of dart-lang/language#3276 will _surely_ include covariance.)

sgrekhov:
Fixed by #2217

PR #2223, opened 2023-08-21
#2217. Add tests for calculating of the least upper bound for extension types

sgrekhov:
@eernstg if you believe that we need more tests, please let me know

### DIFF
--- a/LanguageFeatures/Extension-types/upper_bound_A01_t01.dart
+++ b/LanguageFeatures/Extension-types/upper_bound_A01_t01.dart
@@ -35,11 +35,11 @@ main() {
   v2.expectStaticType<Exactly<Object?>>();
 
   var v3 = 2 > 1 ? ET1(Object()) : ET3(Object());
-  v3.expectStaticType<Exactly<ET1<Object?>>>();
+  v3.expectStaticType<Exactly<ET1>>>();
 
   var v4 = 2 > 1 ? ET1(Object()) : ET4(Object());
   v4.expectStaticType<Exactly<Object?>>();
 
   var v5 = 2 > 1 ? ET2(Object()) : ET4(Object());
-  v5.expectStaticType<Exactly<ET2<Object>>>();
+  v5.expectStaticType<Exactly<ET2>>();
 }
--- a/LanguageFeatures/Extension-types/upper_bound_A01_t01.dart
+++ b/LanguageFeatures/Extension-types/upper_bound_A01_t01.dart
@@ -1,0 +1,45 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion The Dart 1 algorithm which is used to compute the standard upper
+/// bound of two distinct interface types will work in the same way as it does
+/// today, albeit on a slightly different superinterface graph:
+///
+/// For the purpose of this algorithm, we consider `Object?` to be a
+/// superinterface of `Null` and `Object`, and the path lengths mentioned in the
+/// algorithm will increase by one (because they start from `Object?` rather
+/// than from `Object`).
+///
+/// @description Check that for two extension types with the representation
+/// types `Object?` and `Object` the least upper bound is `Object?` (and if both
+/// of them have representation type `Object` then the least upper bound is
+/// `Object`)
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+import '../../Utils/static_type_helper.dart';
+
+extension type ET1(Object? o) {}
+extension type ET2(Object o) {}
+
+extension type ET3(Object? o) implements ET1 {}
+extension type ET4(Object o) implements ET2 {}
+
+main() {
+  var v1 = 2 > 1 ? ET1(Object()) : ET2(Object());
+  v1.expectStaticType<Exactly<Object?>>();
+
+  var v2 = 2 > 1 ? ET1(null) : ET2("String");
+  v2.expectStaticType<Exactly<Object?>>();
+
+  var v3 = 2 > 1 ? ET1(Object()) : ET3(Object());
+  v3.expectStaticType<Exactly<Object?>>();
+
+  var v4 = 2 > 1 ? ET1(Object()) : ET4(Object());
+  v4.expectStaticType<Exactly<Object?>>();
+
+  var v5 = 2 > 1 ? ET2(Object()) : ET4(Object());
+  v5.expectStaticType<Exactly<Object>>();
+}
--- a/LanguageFeatures/Extension-types/upper_bound_A01_t01.dart
+++ b/LanguageFeatures/Extension-types/upper_bound_A01_t01.dart
@@ -35,7 +35,7 @@ main() {
   v2.expectStaticType<Exactly<Object?>>();
 
   var v3 = 2 > 1 ? ET1(Object()) : ET3(Object());
-  v3.expectStaticType<Exactly<ET1>>>();
+  v3.expectStaticType<Exactly<ET1>>();
 
   var v4 = 2 > 1 ? ET1(Object()) : ET4(Object());
   v4.expectStaticType<Exactly<Object?>>();
--- a/LanguageFeatures/Extension-types/upper_bound_A01_t01.dart
+++ b/LanguageFeatures/Extension-types/upper_bound_A01_t01.dart
@@ -35,11 +35,11 @@ main() {
   v2.expectStaticType<Exactly<Object?>>();
 
   var v3 = 2 > 1 ? ET1(Object()) : ET3(Object());
-  v3.expectStaticType<Exactly<Object?>>();
+  v3.expectStaticType<Exactly<ET1<Object?>>>();
 
   var v4 = 2 > 1 ? ET1(Object()) : ET4(Object());
   v4.expectStaticType<Exactly<Object?>>();
 
   var v5 = 2 > 1 ? ET2(Object()) : ET4(Object());
-  v5.expectStaticType<Exactly<Object>>();
+  v5.expectStaticType<Exactly<ET2<Object>>>();
 }
--- a/LanguageFeatures/Extension-types/upper_bound_A01_t04.dart
+++ b/LanguageFeatures/Extension-types/upper_bound_A01_t04.dart
@@ -11,8 +11,8 @@
 /// algorithm will increase by one (because they start from `Object?` rather
 /// than from `Object`).
 ///
-/// @description Check that the least upper bound calculation algorithm takes
-/// starts from `Object?`
+/// @description Check that the least upper bound calculation algorithm for
+/// nullable types starts superinterfaces tree from `Object?`
 /// @author sgrekhov22@gmail.com
 
 // SharedOptions=--enable-experiment=inline-class
@@ -30,11 +30,11 @@ main() {
   v1.expectStaticType<Exactly<Object?>>();
 
   var v2 = 2 > 1 ? ET1("String") : ET3("String");
-  v2.expectStaticType<Exactly<Object>>();
+  v2.expectStaticType<Exactly<ET1>>();
 
   var v3 = 2 > 1 ? ET2(2) : ET4(3);
   v3.expectStaticType<Exactly<Object?>>();
 
   var v4 = 2 > 1 ? ET2(2) : ET3("String");
-  v4.expectStaticType<Exactly<Object?>>();
+  v4.expectStaticType<Exactly<ET2>>();
 }
--- a/LanguageFeatures/Extension-types/upper_bound_A01_t04.dart
+++ b/LanguageFeatures/Extension-types/upper_bound_A01_t04.dart
@@ -33,8 +33,8 @@ main() {
   v2.expectStaticType<Exactly<ET1>>();
 
   var v3 = 2 > 1 ? ET2(2) : ET4(3);
-  v3.expectStaticType<Exactly<Object?>>();
+  v3.expectStaticType<Exactly<ET2>>();
 
   var v4 = 2 > 1 ? ET2(2) : ET3("String");
-  v4.expectStaticType<Exactly<ET2>>();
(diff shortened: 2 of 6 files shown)


LanguageFeatures/Extension-types/upper_bound_A01_t05.dart
```diff
@@ -22,8 +22,10 @@ import '../../Utils/static_type_helper.dart';
 extension type ET1<T>(T id) {}
 extension type ET2<T>(T id) {}
 
+typedef NullableNum = num?;
+
 extension type ET3<T extends String>(T id) implements ET1<T>, String {}
-extension type ET4<T extends int>(T id) implements ET2<T>, num? {}
+extension type ET4<T extends int>(T id) implements ET2<T>, NullableNum {}
```
eernstg, line 28: This should still be a compile-time error, so I'm afraid there won't be any LUB tests starting from extension types where we encounter a type of the form `T?` in the superinterface graph. But, of course, it's not a problem that we can't have a test for a case that doesn't exist.

(I suppose we already have some other test showing that it is an error for an extension type declaration to have an `implements ... T ...` where `T` is a type of the form `S?`, with or without aliasing.)

eernstg (commented):
It looks like this test expects no errors in a case where we must have an error.

LanguageFeatures/Extension-types/upper_bound_A01_t05.dart
sgrekhov: No we don't have such tests. I've already made a note to myself to add them

sgrekhov: Fixed

LanguageFeatures/Extension-types/upper_bound_A01_t05.dart
```diff
@@ -0,0 +1,46 @@
...
+  v2.expectStaticType<Exactly<Object?>>();
+
+  var v3 = 2 > 1 ? ET3("String") : ET4(3);
+  v3.expectStaticType<Exactly<Object?>>();
+
+  var v4 = 2 > 1 ? ET1<String>("String") : ET3("String");
+  v4.expectStaticType<Exactly<Object>>();
+
+  var v5 = 2 > 1 ? ET1<String?>("String") : ET3("String");
+  v5.expectStaticType<Exactly<Object?>>();
```
eernstg, line 42: This is again a subtype relationship: `ET3<String> <: ET1<String> <: ET1<String?>`, so the result should be `ET1<String?>`.

LanguageFeatures/Extension-types/upper_bound_A01_t05.dart
```diff
@@ -0,0 +1,46 @@
...
+
+  var v6 = 2 > 1 ? ET2<num>("String") : ET4(4);
+  v6.expectStaticType<Exactly<Object?>>();
```
eernstg, line 45: Same situation here, `ET4<int> <: ET2<int> <: ET2<num>`

LanguageFeatures/Extension-types/upper_bound_A01_t01.dart
```diff
@@ -0,0 +1,45 @@
...
+
+main() {
+  var v1 = 2 > 1 ? ET1(Object()) : ET2(Object());
+  v1.expectStaticType<Exactly<Object?>>();
+
+  var v2 = 2 > 1 ? ET1(null) : ET2("String");
+  v2.expectStaticType<Exactly<Object?>>();
+
+  var v3 = 2 > 1 ? ET1(Object()) : ET3(Object());
+  v3.expectStaticType<Exactly<Object?>>();
```
eernstg, line 38: This should be `ET1` because `ET3 <: ET1`, and `UP(S, T) == T` when `S <: T`.

LanguageFeatures/Extension-types/upper_bound_A01_t01.dart
```diff
@@ -0,0 +1,45 @@
...
+
+  var v4 = 2 > 1 ? ET1(Object()) : ET4(Object());
+  v4.expectStaticType<Exactly<Object?>>();
+
+  var v5 = 2 > 1 ? ET2(Object()) : ET4(Object());
```
eernstg, line 43: Should be `ET2`, because of the subtype relationship again.

LanguageFeatures/Extension-types/upper_bound_A01_t04.dart
```diff
@@ -0,0 +1,40 @@
...
+
+main() {
+  var v1 = 2 > 1 ? ET1("String") : ET2(1);
+  v1.expectStaticType<Exactly<Object?>>();
+
+  var v2 = 2 > 1 ? ET1("String") : ET3("String");
+  v2.expectStaticType<Exactly<Object>>();
+
+  var v3 = 2 > 1 ? ET2(2) : ET4(3);
+  v3.expectStaticType<Exactly<Object?>>();
```
eernstg, line 36: This would be `ET2` because `ET4 <: ET2`.

eernstg (commented):
There are a number of cases where the `UP` type function yields a result which is different from that which seems to be expected by the test, mainly because `UP(S, T) == T` when `S <: T`.

We should also have some tests where the compute `UP(E, T)` where `E` is an extension type and `T` is a non-extension type, and vice versa.

LanguageFeatures/Extension-types/upper_bound_A01_t05.dart
```diff
@@ -0,0 +1,46 @@
...
+
+main() {
+  var v1 = 2 > 1 ? ET1<String>("String") : ET2<num>(1);
+  v1.expectStaticType<Exactly<Object>>();
+
+  var v2 = 2 > 1 ? ET1<String>("String") : ET2<String?>("String");
+  v2.expectStaticType<Exactly<Object?>>();
+
+  var v3 = 2 > 1 ? ET3("String") : ET4(3);
+  v3.expectStaticType<Exactly<Object?>>();
```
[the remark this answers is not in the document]
sgrekhov: Added `upper_bound_A01_t06.dart`

sgrekhov (commented):
Tests updated. Please take another look

LanguageFeatures/Extension-types/upper_bound_A01_t06.dart
```diff
@@ -0,0 +1,47 @@
...
+
+main() {
+  var v1 = 2 > 1 ? ET2("ET2") : ET3("ET3");
+  v1.expectStaticType<Exactly<String?>>();
+
+  var v2 = 2 > 1 ? "String" : ET2("ET2");
+  v2.expectStaticType<Exactly<String>>();
+
+  var v3 = 2 > 1 ? "String" : ET3("ET3");
+  v3.expectStaticType<Exactly<String?>>();
```
eernstg, line 34: Please double check this one, it looks like it should be `String`, not `String?`.

LanguageFeatures/Extension-types/upper_bound_A01_t06.dart
```diff
@@ -0,0 +1,47 @@
...
+
+  var v4 = 2 > 1 ? "String" : ET1<String>("ET1");
+  v4.expectStaticType<Exactly<String>>();
```
eernstg, line 37: I think this should be `Object`, `ET1<String>` does not have any subtype links to `String`.

LanguageFeatures/Extension-types/upper_bound_A01_t06.dart
```diff
@@ -0,0 +1,47 @@
...
+
+  var v5 = 2 > 1 ? "String" : ET1<String?>("ET1");
+  v5.expectStaticType<Exactly<String?>>();
```
eernstg, line 40: `Object?` this time, `ET1<String?>` does not have any subtype links to `String?` (or `String`, for that matter), and the rep. type of `ET1<String?>` is nullable.

LanguageFeatures/Extension-types/upper_bound_A01_t04.dart
eernstg: I'm not sure how this was fixed, it looks like no changes (if I remember correctly)?

LanguageFeatures/Extension-types/upper_bound_A01_t05.dart
eernstg: Right, thanks! `ET4<int> <: ET2<int?>` so we do _not_ have the case where one operand is a subtype of the other. So we must run the Dart 1 algorithm.

```mermaid
graph BT;
  ET4["ET4&lt;int>"] --> ET2["ET2&lt;int?>"]
  ET4 --> Object
  ET2 --> ObjectQ["Object?"]
  Object --> ObjectQ
  ET4 --> num
  num --> ComparableN["Comparable&lt;num>"]
  ComparableN --> Object
  ET2n["ET2&lt;num>"] --> Object
```

The only shared supertypes of the two operands is `Object` and `Object?`, so we get `Object`.

LanguageFeatures/Extension-types/upper_bound_A01_t05.dart
```diff
@@ -0,0 +1,46 @@
...
+
+  var v4 = 2 > 1 ? ET1<String>("String") : ET3("String");
+  v4.expectStaticType<Exactly<Object>>();
+
+  var v5 = 2 > 1 ? ET1<String?>("String") : ET3("String");
+  v5.expectStaticType<Exactly<ET1<String?>>>();
+
+  var v6 = 2 > 1 ? ET2<num>(4) : ET4(5);
+  v6.expectStaticType<Exactly<ET2<num?>>>();
```
eernstg, line 45: (Github doesn't want to show the comment on this line, but it should be visible in the 'conversation' tab as 'obsolete'.) I think it should be `Object`.

eernstg (commented):
A few more comments.

LanguageFeatures/Extension-types/upper_bound_A01_t05.dart
```diff
@@ -0,0 +1,46 @@
...
+  v1.expectStaticType<Exactly<Object>>();
+
+  var v2 = 2 > 1 ? ET1<String>("String") : ET2<String?>("String");
+  v2.expectStaticType<Exactly<Object?>>();
+
+  var v3 = 2 > 1 ? ET3("String") : ET4(3);
+  v3.expectStaticType<Exactly<Object?>>();
+
+  var v4 = 2 > 1 ? ET1<String>("String") : ET3("String");
+  v4.expectStaticType<Exactly<Object>>();
```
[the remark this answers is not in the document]
sgrekhov: Updated

LanguageFeatures/Extension-types/upper_bound_A01_t06.dart
sgrekhov: Seems so. First I thought that implementing `ET1<String?>` makes LUB to be `String?`.

LanguageFeatures/Extension-types/upper_bound_A01_t06.dart
sgrekhov: Updated

sgrekhov (commented):
Updated

LanguageFeatures/Extension-types/upper_bound_A01_t05.dart
eernstg: I still think it should be object: The layer sets are `{Object?}` (depth 0) and `{Object}` (depth 1), and none of these sets have >1 elements, so they are both eligible; then we take `Object` because it has the greatest depth. How did you arrive at `Object?`?

LanguageFeatures/Extension-types/upper_bound_A01_t04.dart
eernstg: Thanks!

LanguageFeatures/Extension-types/upper_bound_A01_t06.dart
```diff
@@ -0,0 +1,47 @@
...
+import '../../Utils/static_type_helper.dart';
+
+extension type ET1<T>(T id) {}
+
+extension type ET2<T extends String>(T id) implements ET1<T>, String {}
+extension type ET3<T extends String>(T id) implements ET1<T?>, String {}
+
+main() {
+  var v1 = 2 > 1 ? ET2("ET2") : ET3("ET3");
+  v1.expectStaticType<Exactly<String?>>();
```
eernstg, line 28: Can't be `String?` -- shared superinterfaces are `String`, `Comparable<String>`, `Object`, `Object?`, and `String` wins.

eernstg (commented):
A couple of comments.

sgrekhov:
Thank you. Updated. But if this [error](https://github.com/dart-lang/co19/pull/2223/files#r1302640851) is valid? There are more similar error in these tests

LanguageFeatures/Extension-types/upper_bound_A01_t05.dart
```diff
@@ -0,0 +1,46 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+import '../../Utils/static_type_helper.dart';
+
+extension type ET1<T>(T id) {}
+extension type ET2<T>(T id) {}
+
+extension type ET3<T extends String>(T id) implements ET1<T>, String {}
```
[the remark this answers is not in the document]
eernstg: Immediately sounds wrong. But let's think about it. ;-)

There is nothing wrong in being a subtype of `Object` and also a subtype of `Object?` (almost everything is), and those two types do not satisfy the requirement that usually justifies words like "the type arguments are different" (there are no type arguments at all).

It sounds like the implementation of the analyzer is running code where it is assumed that we have both `C<T1>` and `C<T2>` as superinterfaces of a given interface type, where `C` is some generic type and `T1` and `T2` are not subtypes of each other, even though we don't have that situation at all here. But somehow finding `Object?` as well as `Object` in the superinterface graph causes the confusion.

So it is not a valid error.

LanguageFeatures/Extension-types/upper_bound_A01_t05.dart
eernstg: (I think this thread has been superseded by other threads, please comment again if I'm wrong.)

PR #2227, opened 2023-08-23
#1400. Extension types subtyping tests. Resolve name conflict, add missing tests

sgrekhov:
There were two tests with the same name.
```
LanguageFeatures/Subtyping/test_types/extension_type_A08.dart
LanguageFeatures/Subtyping/dynamic/test_types/extension_type_A08.dart
```
Generator resolves this conflict by adding `_1_` suffix to the filename. But these suffixed tests were forgotten to add to git. This PR renames tests to avoid name conflict and adds missing generated tests


eernstg (approved):
LGTM

PR #2229, opened 2023-08-23
Fixes #2228. Fix syntax error in static_analysis_extension_types_A10_t08.dart

### LINKED ISSUE #2228: Syntax error in extension type test static_analysis_extension_types_A10_t08
I noted a syntax error in `Extension-types/*.dart`: Line 20 of static_analysis_extension_types_A10_t08.dart uses `(,)` as a type, which is not derivable from the grammar. By the way, it is also explicitly mentioned as being an error [here](https://github.com/dart-lang/language/blob/main/accepted/3.0/records/feature-specification.md#:~:text=The%20expression%20(%2C-,),-isn%27t%20allowed.). So we may get the expected error in this test for the wrong reason.

sgrekhov:


eernstg (approved):
LGTM

eernstg (approved):
LGTM

Issue #2119, opened 2023-07-07
co19 minor improvements


sgrekhov:
Placeholder issue for minor co19 improvements (code formatting, assertion/description wording fixes, adding issue numbers etc)

PR #2238, opened 2023-08-25
#2119. Code format and description update in operator tests

sgrekhov:
Just code-style and code formatting changes. Moved them to this separate PR to not make a noise in PR with the actual changes

### DIFF
--- a/Language/Classes/Instance_Methods/Operators/optional_parameter_t05.dart
+++ b/Language/Classes/Instance_Methods/Operators/optional_parameter_t05.dart
@@ -17,8 +17,5 @@ class C {
 }
 
 main() {
-  new C()[];
-//        ^
-// [analyzer] unspecified
-// [cfe] unspecified
+  new C()[1];
 }
--- a/Language/Classes/Instance_Methods/Operators/optional_parameter_t05.dart
+++ b/Language/Classes/Instance_Methods/Operators/optional_parameter_t05.dart
@@ -4,10 +4,10 @@
 
 /// @assertion It is a compile-time error to declare an optional parameter in an 
 /// operator.
+///
 /// @description Checks that a compile-time error is produced if a user-defined
 /// operator [] specifies one optional named parameter.
 /// @author ilya
-
 
 class C {
   operator []({var o}) {}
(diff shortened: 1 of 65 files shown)


eernstg (approved):
LGTM

Language/Classes/Instance_Methods/Operators/optional_parameter_t05.dart
```diff
@@ -4,11 +4,11 @@
 
 /// @assertion It is a compile-time error to declare an optional parameter in an 
 /// operator.
+///
```
eernstg, line 7: Github won't let me comment on line 20, but it has a syntax error: Operator `[]` can be invoked using `e1[e2]` when `e1` is derived from `<primary>` and `e2` from `<expression>`. But there's no `e2` here. I guess that's because the parameter is optional.

The syntax doesn't allow for `new C()[o: 1]`, either.

I guess the most reasonable test would have the declaration (line 13, which is the place where the description indicates that we're testing that the error is actually reported), and then it should use the declarations in the test, but it doesn't have to be a (secondary) syntax error in main.

So `new C()[1];` might be fine, because it avoids the secondary error (we don't expect the _parser_ to be able to understand that `operator []({var o})` should be callable as `C()[]`, it just detects the syntax error and reports it. So that doesn't say anything about the treatment of the operator declaration during compilation.

It is certainly still possible that the static analysis leaves `operator []` in a state where it expects a named parameter, but `C()[o: 1]` would also just be a syntax error.

So it's probably better to simply have `C()[1]` in main in order to make sure that the declaration is used, and then treat this as a secondary error (and write the expectation to expect whatever errors the tools are reporting, if any).

Language/Classes/Instance_Methods/Operators/optional_parameter_t07.dart
```diff
@@ -4,11 +4,11 @@
 
 /// @assertion It is a compile-time error to declare an optional parameter in an 
 /// operator.
+///
```
eernstg, line 7: In line 21, please make adjustments similar to the ones in optional_parameter_t05.

eernstg (commented):
All fine, except for a couple of locations with a syntax error that does not seem to contribute to the purpose of the test.

sgrekhov:
That's reasonable. Thank you. Updated

eernstg (approved):
Thanks, LGTM!

PR #2239, opened 2023-08-25
Fixes #2235. Add missing operator tests

### LINKED ISSUE #2235: Update tests for operators
Existing sests for operators don't fully cover "10.2.1 Operators" chapter of the [specification](https://spec.dart.dev/DartLangSpecDraft.pdf)

sgrekhov:


Language/Classes/Instance_Methods/Operators/syntax_t06.dart
```diff
@@ -0,0 +1,34 @@
...
+/// | ⟨bitwiseOperator⟩
+///
+/// @description Checks that it is a compile-time error if an enum declares an
+/// abstract user-defined operator
+/// @author sgrekhov22@gmail.com
+
+enum E {
+  e1, e2;
+
+  int operator +(int other);
```
eernstg, line 26: This is not actually a syntax error, because an `enum` declaration just contains `<classMemberDeclaration>`s, and they can derive abstract members of all kinds.

However, I don't think we need to consider renaming the test --- we don't actually rely on the names of the form `*syntax*.dart` to indicate that a specific test is a syntax error test.

eernstg (approved):
LGTM.

I added a comment about the naming of one test, but I don't think there's a need to change anything after all.

Issue #2145, opened 2023-07-20
Update Variables and Local Variable declaration according to spec


sgrekhov:
Variables and Local Variable declaration sections of the [specification](https://spec.dart.dev/DartLangSpecDraft.pdf) were updated. See https://github.com/dart-lang/language/commit/af27cfbcc99c4af22c6e71bb84cdb5addbb826a0. Need to update co19 tests accordingly

eernstg:
Right, these updates are quite substantial. The changes are the null safety related, so there should not be anything new for tests that are already targeting the language at a version where null safety has been introduced.

However, it would certainly be valuable to get an extra pair of eyes on the specification and the actual tool behavior, such that we can detect ways in which they disagree, if any. ;-)

PR #2240, opened 2023-08-29
#2145. Update variables tests according to the current spec. Part 2

sgrekhov:


eernstg:
I noted a bunch of changes where a line ends in `\x0D` (that is, carriage return, ASCII 13, `\r`). Presumably this is caused by editing a non-Windows library on a Windows host? One example with many of these is 'Language/Classes/Instance_Methods/Operators/syntax_t01.dart'.

I'm not sure about how to think about this. We usually remove trailing whitespace, and `\r` would certainly count as whitespace, and it occurs just before the newline character (`\0A`, ASCII 10, `\n`), so from that point of view it should simply be removed, no need to think about that for a long time.

On the other hand, it's testing the ability of tools to tolerate non-standard combinations of line-ending character sequences.

On the third hand, if any, we don't have to test that particular part of the parser in hundreds of tests.

Here is some data from a fresh checkout of the co19 repository, in branch co19-2145-2:

```zsh
> file **/*(.) | wc -l
22311
> file **/*(.) | grep 'line terminators' | wc -l
1316
> grep -l ' $' **/*(.) | wc -l
992
```

This means that `file` found 22311 files in the 'co19' directory and subdirectories, 1316 of them have some non-standard line terminators (I think they are all CRLF, that is `\r\n`, but each of those libraries may also contain lots of plain `\n` line endings), and 992 files have trailing spaces (just spaces, haven't search for trailing whitespace of other kinds).

It looks like some of the funny cases (like `\x0D` in the middle of a UN*X-style file where every other line break is just an `\n`) are introduced by this PR, but surely many of them are older. 

We could use a separate PR to fix up the line breaks: Every library in co19 should use `\n` line endings (except for a very small number of tests specifically about line endings, if we have any of those).

With respect to trailing whitespace, we could also have a separate PR to remove it.

The only thing which is inconsistent with normal procedures is if this PR is landed and it _introduces_ trailing whitespace in any number of lines.

Do you have an easy way to find the newly introduced whitespace, and eliminate it? I guess something like this could work to at least find trailing whitespace in modified files:

```bash
> grep -n '\s$'  $(git diff --name-only master..) | less
```

sgrekhov:
Cool! How did you noticed it?
Seems that all the files affected by this PR are ok (have `LF` line endings). Created #2242 for the line endings. At the first glance the problem was introduved by https://github.com/dart-lang/co19/commit/bd1a423957e5e46c0262b1a382f77e58265ebcd1 in May 2021

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A05_t01.dart
```diff
@@ -0,0 +1,39 @@
...
+
+class C {
+  static late var v1;
+  late var v2;
+  var v3;
+  static var v4;
+}
+
+main() {
+  x1 = -1;
```
eernstg, line 25: Let's just use `Expect.throws(() => x1.whatever)` as usual.

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A06_t02.dart
```diff
@@ -0,0 +1,35 @@
...
+
+/// @assertion Case ⟨Setter: Late-final variable with declared type ⟩. A
+/// variable declaration of the form static? late final T id; implicitly induces
+/// a setter with the header void set id(T x). If this setter is executed in a
+/// situation where the variable id has not been bound, it will bind id to the
+/// object that x is bound to. If this setter is executed in a situation where
+/// the variable id has been bound to an object, a dynamic error occurs
+///
+/// @description Checks that it is a dynamic error if this setter is executed in
+/// a situation where the variable has been bound to an object
```
eernstg, line 13: Perhaps?:
```suggestion
/// @description Checks that it is a dynamic error if this getter is executed in
/// a situation where the variable has not been bound to an object
```

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A07_t01.dart
```diff
@@ -0,0 +1,39 @@
...
+/// @assertion Case ⟨Setter: Late-final variable with no declared type, no
+/// initialization⟩. A variable declaration of the form static? late final id;
+/// implicitly induces a setter with the header void set id(dynamic x). An
+/// execution of said setter in a situation where the variable id has not been
+/// bound will bind id to the object that the argument x is bound to. An
+/// execution of the setter in a situation where the variable id has been bound
+/// to an object will incur a dynamic error
+///
+/// @description Checks that the static type of an implicit setter of a late
+/// final variable with no declared type and no initialization is `dynamic`
```
eernstg, line 14: How would we be able to test that? If you want to distinguish a `void Function(dynamic)` from a `void Function(Object?)` then you can use a tear-off, `runtimeType`, and test equality with `typedef F1 = void Function(dynamic);` respectively `typedef F2 = void Function(Object?);` (those function types must be non-equal because type equality makes a distinction between `dynamic` and `Object?`). But you can't tear off a setter...

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A07_t01.dart
```diff
@@ -0,0 +1,39 @@
...
+
+class C {
+  static late final v1;
+  late final v2;
+}
+
+main() {
+  x1 = 1;
+  try {
+    x1.whatever;
```
eernstg, line 27: Also, this doesn't test the signature of the setter at all.

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A07_t02.dart
```diff
@@ -0,0 +1,36 @@
...
+late final x1;
+
+class C {
+  static late final v1;
+  late final v2;
+}
+
+main() {
+  Expect.throws(() {
+    x1;
```
eernstg, line 28: This should be `x1 = 1;` folllowed by `x1 = 2;`, otherwise we don't get to a point where the setter is executed when the variable has already been bound to an object.

eernstg (commented):
Looks good! A couple of tests do look funny, please take another look.

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A03_t01.dart
```diff
@@ -0,0 +1,48 @@
...
+
+/// @assertion Case ⟨Setter: Mutable variable with declared type ⟩. A variable
+/// declaration of one of the forms
+/// • static? late? T id;
+/// • static? late? T id = e;
+/// implicitly induces a setter with the header void set id(T x), whose
+/// execution sets the value of id to the incoming argument x.
+///
+/// @description Checks the static type of an implicit getter of a variable with
+/// no declared type
```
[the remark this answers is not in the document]
sgrekhov: This description was copied from the previous test and forgotten to update. Updated

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A04_t01.dart
```diff
@@ -0,0 +1,54 @@
...
+
+class C {
+  static late var v1 = 1;
+  late var v2 = 2;
+  var v3 = 3;
+  static var v4 = 4;
+}
+
+main() {
+  x1 = -1;
```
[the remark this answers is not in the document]
sgrekhov: Description again. Changed to `setter`. We can't use `expectStaticType` for setters

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A05_t01.dart
sgrekhov: Again. Changed description to `setter`. `Expect.throws(() => x1.whatever)` will test getter, not setter

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A06_t02.dart
sgrekhov: Ups, my fault. Thank you! Test updated to fit the description (and the assertion)

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A07_t01.dart
sgrekhov: Deleted this test

sgrekhov (commented):
Thank you for the detailed review! Tests updated. Some description were changed to fit assertions and the tests. Also setters now actually tests setters. One tests deleted. Please review once again

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A04_t01.dart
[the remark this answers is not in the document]
eernstg: OK, same comment as in Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A03_t01.dart.

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A07_t01.dart
eernstg: OK.

eernstg (commented):
Looks good!

I think we should take the opportunity to tighten the tests concerning the type of the formal parameter of an implicitly induced setter, and I mentioned a couple of things we could do.

Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A02_t01.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2011, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A non-local variable introduces a setter iff it does not have the
+/// modifier final or the modifier const, or it is late and final, but does not
+/// have an initializing expression.
```
eernstg, line 7: This assertion doesn't fit the test: The description and the test is about a local variable, but the assertion is about all those kinds of variables that aren't local.

It's a somewhat inconvenient situation: I don't know exactly how the assertions are organized, but if a test needs to have its assertion updated then I wouldn't be surprised if it would then also need to be renamed and put into a different directory.

We do want to have the test which is actually written here. There is of course another case where the variable is final, isn't initialized in the declaration, and is evaluated at a point where it isn't guaranteed to have been initialized, or it is assigned at a time where it is definitely already initialized, but I think we have those elsewhere already.

Maybe this test is essentially a duplicate, and it should simply be deleted?

Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A02_t03.dart
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2011, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A non-local variable introduces a setter iff it does not have the
+/// modifier final or the modifier const, or it is late and final, but does not
+/// have an initializing expression.
```
eernstg, line 7: Similar to Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A02_t01.dart.

Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A03_t02.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2011, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A non-local variable introduces a setter iff it does not have the
+/// modifier final or the modifier const, or it is late and final, but does not
+/// have an initializing expression.
```
eernstg, line 7: Similar to Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A03_t01.dart.

Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A03_t03.dart
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2011, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A non-local variable introduces a setter iff it does not have the
+/// modifier final or the modifier const, or it is late and final, but does not
+/// have an initializing expression.
```
eernstg, line 7: Similar to Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A03_t01.dart.

Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A03_t04.dart
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A non-local variable introduces a setter iff it does not have the
+/// modifier final or the modifier const, or it is late and final, but does not
+/// have an initializing expression.
```
eernstg, line 7: Similar to Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A03_t01.dart.

eernstg (commented):
Looks very good! (it's a nice improvement over the old version, apart from the fact that there was a need to cover null safety).

There was one expectation that I couldn't understand (Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A06_t01.dart).

Finally, there are a handful of tests starting with Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A02_t01.dart where the assertion says something which isn't on topic for the given test and the given description. If we're lucky then they are just duplicating a test which exists elsewhere already (not textually duplicating another test, but testing something which is already being tested by another test). It is also possible that the 'assertion' should be updated.

In any case, we shouldn't have this kind of discrepancy between the assertion and the description and the test, so we need to consider the situation whenever we see something like that.

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A06_t01.dart
```diff
@@ -0,0 +1,40 @@
...
+//     ^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  C.v1 = n;
+//       ^
+// [analyzer] unspecified
+// [cfe] unspecified
+  C c = C();
+  c.v2 = 2;
```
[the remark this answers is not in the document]
sgrekhov: Typo, of course. Changed to `c.v2 = n;`. Thanx!

Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A02_t01.dart
sgrekhov: Updated assertion and moved this and similar tests to `Srtatements/Local Variable Declaration` directory

sgrekhov (commented):
Typo fixed, assertion and localion of final local variable tests updated

Language/Variables/Implicitly_Induced_Getters_and_Setters/definition_A06_t01.dart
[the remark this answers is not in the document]
eernstg: Thanks!

Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A02_t01.dart
eernstg: Sounds good!

eernstg (approved):
LGTM!

PR #2254, opened 2023-09-01
#2145. Add more local variable declaration tests

sgrekhov:


Language/Statements/Local_Variable_Declaration/declared_type_A03_t01.dart
```diff
@@ -0,0 +1,65 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion The declared type of a local variable with a declaration of one
+/// of the forms late? var v = e; late? final v = e; const v = e; is determined
+/// as follows:
+/// ...
+/// • If the static type of e is of the form X & T where X is a type variable,
+/// the declared type of v is X. In this case v is immediately promoted to X & T
+///
+/// @description Checks that static type of a variable declared by the
```
eernstg, line 12:
```suggestion
/// @description Checks that the static type of a variable declared by the
```

Language/Statements/Local_Variable_Declaration/declared_type_A03_t01.dart
```diff
@@ -0,0 +1,65 @@
...
+    v.isEven;
+    v.expectStaticType<Exactly<T>>();
+  }
+  late var v = t;
+  v.expectStaticType<Exactly<T>>();
+}
+
+test3<T>(T t) {
+  if (t is int) {
+    final v = t;
```
eernstg, line 42: This is an interesting case. 

Double checking with the NNBD feature specification (the updates in the language specification shouldn't disagree with the NNBD feature specification except where something has been changed intentionally, which is basically small bug fixes --- and I don't think we had any of those here):

In the NNBD feature specification, it is not stated explicitly that the promotion to `X & T` is not applied to a `final` local variable when the initializing expression has type `X & T`, but it could be considered to be implied by the reference to `final` a few lines earlier.

However, the implementations _do_ actually promote `final v = e;` in this way, so it would be a breaking change to say that there is no promotion in this case.

It's a weird corner case ("why do we promote `final x = e;` from `X` to `X & int` when `e` has static type `X & int`, but we do _not_ promote `final int? x = e;` to `int` when `e` has static type `int`?"), but I don't see a clear discrepancy, and the promotion is already implemented. (Also, I don't see an overwhelming need to broaden the special exception for `final` variables.)

So it's fine, and we _should_ expect `v` to be promoted to `T & int` here! ;-)

So we still have the same considerations about testing `T & int`.

Language/Statements/Local_Variable_Declaration/declared_type_A04_t01.dart
```diff
@@ -0,0 +1,43 @@
...
+import '../../../Utils/static_type_helper.dart';
+
+test4<T>(T t) {
+  if (t is int) {
+    late final v = t;
+    v.isEven;
+    v.expectStaticType<Exactly<T>>();
+  }
+  late final v = t;
+  v.expectStaticType<Exactly<T>>();
```
eernstg, line 25: Same considerations as usual around testing `T & int`.

Language/Statements/Local_Variable_Declaration/reference_error_A01_t02.dart
```diff
@@ -0,0 +1,41 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if a local variable is referenced at a
+/// source code location that is before the end of its initializing expression,
+/// if any, and otherwise before the declaring occurrence of the identifier
+/// which names the variable.
+///
+/// @description Checks that a variable is introduced into the scope after a
+/// variable declaration statement is evaluated, and the name of this variable
```
eernstg, line 11: That isn't quite accurate: The local declaration introduces the name into the enclosing scope from the beginning of that scope:

```dart
var x = 1;

void main() {
  print(x); // Error (`x` is the local variable below, not the top-level variable).
  var x = true;
}
```

But it is true that we can't use that name (it's a compile-time error no matter how we use it) before the end of the initializing expression in the declaration of that name.

eernstg (commented):
Looks good! Noted a couple of typos. Also, I think the tests can be a bit stronger when it comes to variables whose static type is an intersection type.

Language/Statements/Local_Variable_Declaration/declared_type_A01_t01.dart
```diff
@@ -0,0 +1,27 @@
...
+/// specified type `T` is `T`
+/// @author sgrekhov22@gmail.com
+
+import '../../../Utils/static_type_helper.dart';
+
+main() {
+  late num v1 = 1;
+  late final num v2 = 2;
+  num v3 = 3;
+  final num? v4 = 4;
```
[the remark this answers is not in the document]
eernstg: Thanks!

sgrekhov:
Tests updated. Please review

Language/Statements/Local_Variable_Declaration/declared_type_A03_t01.dart
```diff
@@ -0,0 +1,65 @@
...
+/// statements `late? var v = e; late? final v = e;` is `X` if the static type
+/// of `e` is `X & T` where `X` is a type variable. Also test that `v` is
+/// promoted to `X & T`
+/// @author sgrekhov22@gmail.com
+
+import '../../../Utils/static_type_helper.dart';
+
+test1<T>(T t) {
+  if (t is int) {
+    var v = t;
```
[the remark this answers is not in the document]
eernstg: Thanks!

eernstg (approved):
LGTM!

Issue #2258, opened 2023-09-04
Add  private fields promotion tests


sgrekhov:
See https://github.com/dart-lang/language/issues/2020#issuecomment-1663069707 for the details

PR #2259, opened 2023-09-05
#2258. Update assertions of private fields promotion tests. Add new ones

sgrekhov:
There going to be two more changes.
- Rename and reorder existing tests according to the new assertion text
- Add new tests

### DIFF
--- a/LanguageFeatures/Private-fields-promotion/promotion_A02_t04.dart
+++ b/LanguageFeatures/Private-fields-promotion/promotion_A02_t04.dart
@@ -12,7 +12,7 @@
 /// - There is no implicit noSuchMethod forwarder with the same name elsewhere
 ///   in the library.
 ///
-/// @description Checks that an instance field is promotable even if it is
+/// @description Checks that an instance field is not promotable if it is
 /// external
 /// @author sgrekhov22@gmail.com
 /// @issue 53426
@@ -23,8 +23,8 @@ class A {
   void foo() {}
 }
 
-class C<T extends A> {
-  external final T? _x;
+class C<T> {
+   external final T _x;
 
   void test() {
     if (_x is A) {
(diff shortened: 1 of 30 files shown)


LanguageFeatures/Private-fields-promotion/promotion_A02_t04.dart
```diff
@@ -0,0 +1,41 @@
...
+/// hold:
+/// - It refers to a field (not a getter)
+/// - The field is private
+/// - The field is final
+/// - There is no getter with the same name elsewhere in the library
+/// - There is no non-final field with the same name elsewhere in the library
+/// - There is no implicit noSuchMethod forwarder with the same name elsewhere
+///   in the library.
+///
+/// @description Checks that an instance field is promotable even if it is
```
eernstg, line 15: Presumably a missing update:
```suggestion
/// @description Checks that an instance field is not promotable if it is
```

LanguageFeatures/Private-fields-promotion/promotion_A02_t04.dart
```diff
@@ -0,0 +1,41 @@
...
+class A {
+  void foo() {}
+}
+
+class C<T extends A> {
+  external final T? _x;
+
+  void test() {
+    if (_x is A) {
+      _x.foo();
```
eernstg, line 31: I think this doesn't test the intended behavior: `_x` has type `T?` which is not a type variable, so there is no support for promoting `T?` to an intersection type, unlike the type `T` which could be promoted to, say `T & A`.

On the other hand, `A` is not a subtype of `T?`, because the value of `T` could be a proper subtype of `A`. So we can't promote `_x` to `A`.

Hence, `_x` shouldn't be promoted, even if it were a local variable. Hence, it doesn't really test the special treatment of promotion of private instance variables because there is no promotion to enable in the first place.

We could stick to `T _x` rather than `T? _x`, and then wait for https://github.com/dart-lang/sdk/issues/50677 to sort out the handling of `T?`.

sgrekhov:
Updated. Please review

eernstg (approved):
LGTM, thanks!

PR #2260, opened 2023-09-05
#2258. Rename and reorder private fields promotion tests

sgrekhov:
Without this renaming and reordering it's hard to figure out what is covered by the tests and what not


LanguageFeatures/Private-fields-promotion/promotion_A04_t01.dart
```diff
@@ -12,33 +12,37 @@
 /// - There is no implicit noSuchMethod forwarder with the same name elsewhere
 ///   in the library.
 ///
-/// @description Checks that an instance field is not promotable if it is not
-/// final
+/// @description Checks that if there are other concrete instance getters with
```
eernstg, line 15: We usually use the terminology 'instance member' (as well as 'instance method', 'instance getter', and so on) with different kinds of declarations. When in doubt, we mention the kind of declaration. So we get descriptions like this:

'a class instance member', which includes declarations in a class or mixin class, as well as declarations in a mixin, because they will be added to classes and then they will be class instance members.

'an extension instance member', which is a declaration in an `extension` that does not have the modifier `static`. 'an extension type instance member' is similar, as are others.

So we'd need to say 'there are other concrete class instance getters' here.

With that in mind there is no contradiction between this rule and the fact that an extension instance member does _not_ have the effect that the final instance variable in `C` is made unpromotable.

eernstg (commented):
Looks good! There is just one location where I noticed that a `@description` contradicts itself somewhat (and I commented on this in order to illustrate why it isn't actually a contradiction).

sgrekhov:
Thank you. Fixed

PR #2262, opened 2023-09-06
#1400. Add promotion of a representation variable test

sgrekhov:
Test for promotion of a private representation variable will be added to `LanguageFeatures/Private-fields-promotion`

### DIFF
--- a/LanguageFeatures/Extension-types/promotion_A01_t01.dart
+++ b/LanguageFeatures/Extension-types/promotion_A01_t01.dart
@@ -1,0 +1,39 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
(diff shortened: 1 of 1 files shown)


LanguageFeatures/Extension-types/promotion_A01_t01.dart
```diff
@@ -0,0 +1,39 @@
...
+    if (id is int) {
+      id.isOdd;
+    }
+  }
+}
+
+main() {
+  ET1 et1 = ET1(1);
+  if (et1.id != null) {
+    et1.id.isEven;
```
eernstg, line 32: OK. We'd need to ensure that `et1` is a stable path (and I'm not quite sure how that is checked, but it shouldn't be worse than checking that a path is stable when a private final instance variable is being accessed).

eernstg (approved):
LGTM!

I just added a remark about the ability to recognize a stable path (`x.y.z[something]` and so on). I don't know how general the handling of stable paths is, but if it works to do the things that are done in this test then it's fine! ;-)

PR #2263, opened 2023-09-07
#2258. Update private fields promotion tests according to the new spec. Part 1

sgrekhov:


LanguageFeatures/Private-fields-promotion/promotion_A02_t01.dart
```diff
@@ -12,7 +12,7 @@
 /// - There is no implicit noSuchMethod forwarder with the same name elsewhere
 ///   in the library.
 ///
-/// @description Checks that an instance field is promotable even if it is late
+/// @description Checks that an instance field is not promotable if it is late
```
stereotype441, line 15: Nit: late final instance fields _are_ promotable, but they can't be promoted by _writing_ to them. They can only be promoted by reading from them and checking the type of the value that was read (just like final instance fields that aren't late).

So for example, this works fine:
```dart
class C {
  late final int? _i;
}
main() {
  C c = C();
  c._i = 1;
  if (c._i != null) {
    c._i.isOdd; // OK
  }
}
```

LanguageFeatures/Private-fields-promotion/promotion_A03_t02.dart
```diff
@@ -13,8 +13,8 @@
 ///   in the library.
 ///
 /// @description Checks that an instance field is promotable if all of the
-/// conditions above are met. Test the case when library contains not a final
-/// global getter with the same name
+/// conditions above are met. Test the case when library contains not a global
```
stereotype441, line 16: Nit: remove the word "not"

LanguageFeatures/Private-fields-promotion/promotion_A04_t01.dart
```diff
@@ -43,6 +43,10 @@ extension on C {
 }
 
 main() {
-  A();
-  C(43).testC();
+  C<num?> c = C(43);
+  if (c._x is int) {
+    c._x;
```
stereotype441, line 48: Did you mean `c._x.isOdd`?

stereotype441 (approved):
LGTM assuming my comments are addressed

sgrekhov (commented):
Thank you! Tests updated

stereotype441:
> Thank you! Tests updated

Fixes lgtm!

LanguageFeatures/Private-fields-promotion/not_promotable_A04_t04.dart
```diff
@@ -37,5 +37,12 @@ abstract class C<T> {
 }
```
eernstg, line 37: Github won't allow me to comment on line 15-16, but we should note that it's about an abstract getter, not just about an abstract class (where the getter could be declared, and it could still be concrete).

By the way, the abstract getter could also be implicitly induced by an abstract instance variable declaration, haven't checked whether that's covered.

LanguageFeatures/Private-fields-promotion/promotion_A02_t02.dart
```diff
@@ -22,15 +22,21 @@ class A {
...
 
-class C<T extends A> {
+class C<T extends A?> {
   late final T? _x;
 
   void test(T t) {
     _x = t;
-    _x.foo();
+    if (_x != null) {
+      _x.foo();
```
eernstg, line 31: This one is questionable: `_x` had the type `T?` which is not a type variable, which means that we can't promote it to an intersection type. Also, `Object` is not a subtype of `T?`, so we can't just promote it to `Object`. So we should not get a promotion here.

Anyway, we probably do get that promotion, because both the analyzer and the CFE treat `T?` as a representation of `T` where the nullability is "nullable", not as a type operator which is applied to `T`.

@stereotype441, what's going on with the issues linked from https://github.com/dart-lang/language/pull/2427?

LanguageFeatures/Private-fields-promotion/promotion_A02_t02.dart
```diff
@@ -22,15 +22,21 @@ class A {
   void foo() {}
 }
 
-class C<T extends A> {
+class C<T extends A?> {
   late final T? _x;
 
   void test(T t) {
     _x = t;
```
eernstg, line 29: NB: `_x` is late, so it's not promoted by the assignment.

eernstg (commented):
Looks good!

I think it was somewhat hard to track down the place where it was concluded that a late final isn't promoted by writing, so I'm suggesting that we keep the reference to that particular github comment in the relevant test for now. I think we'll have a feature spec later on, but right now it would be nice if we can find that comment again. ;-)

Just one technical issue: There is a test where it is expected that `T?` is promoted to `T & Object`, and I think that shouldn't happen (according to the specification documents that I can think of), but I also think it actually happens.

I asked for input from Paul for that (hello @stereotype441 ! ;-)

LanguageFeatures/Private-fields-promotion/promotion_A02_t01.dart
eernstg: I think LanguageFeatures/Private-fields-promotion/promotion_A02_t01.dart is a bit out of sync with this consideration: It doesn't actually test that a promotion occurs _based_ on a write operation, it just performs a write and doesn't rely on any promotions at that point. 

The next step in that test is to perform an `is` test, and I think that should give rise to a promotion.

So we should probably not have a `@description` that says "not promotable" and that's it.

Maybe we could use something like 'Checks that an instance field is promotable even if it is late (but it is promoted because of the type test, not because of the assignment)'.

LanguageFeatures/Private-fields-promotion/promotion_A03_t02.dart
eernstg: Closing: This one has been processed.

LanguageFeatures/Private-fields-promotion/promotion_A04_t01.dart
eernstg: Closing: This one has been processed.

LanguageFeatures/Private-fields-promotion/promotion_A02_t02.dart
stereotype441: @eernstg 
I think this is slightly different from what's going on in https://github.com/dart-lang/language/pull/2427. That PR is about the behaviour of UP, which isn't involved here.

What's happening here is that the null check causes flow analysis to invoke the **NonNull** function on the type `T?`. The relevant rules for **NonNull** are:

> **NonNull**(_C_<_T_<sub>1</sub>, ... , _T_<sub>_n_</sub>>) = _C_<_T_<sub>1</sub>, ... , _T_<sub>_n_</sub>>  for class *C* other than Null (including Object).
> **NonNull**(_X_) = X & **NonNull**(B), where B is the bound of X.
> **NonNull**(_T_?) = **NonNull**(_T_)

So, applying those rules, **NonNull**(`T?`) = **NonNull**(`T`) = `T & `**NonNull**(`A?`) = `T & `**NonNull**(`A`) = `T & A`

And it's valid to call `.foo` on the type `T & A`. So I think this test case is just fine.

As to your other question about what's going on with the issues linked from https://github.com/dart-lang/language/pull/2427, I see just one issue, https://github.com/dart-lang/sdk/issues/49691, and it's assigned to @chloestefantsova, so she's probably in the best position to answer that.

LanguageFeatures/Private-fields-promotion/not_promotable_A02_t04.dart
```diff
@@ -0,0 +1,46 @@
...
+
+/// @assertion An instance field is promotable if all the following conditions
+/// hold:
+/// - It refers to a field (not a getter)
+/// - The field is private
+/// - The field is final
+/// - There is no getter with the same name elsewhere in the library
+/// - There is no non-final field with the same name elsewhere in the library
+/// - There is no implicit noSuchMethod forwarder with the same name elsewhere
+///   in the library.
```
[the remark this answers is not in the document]
sgrekhov: Added link to the comment

LanguageFeatures/Private-fields-promotion/not_promotable_A04_t04.dart
sgrekhov: Changed the descriprion to mention an abstract getter and updated `not_promotable_A05_t04.dart` to test an abstract field (I remember that spec doesn't use word `field` and it's better to use `instance variable` but if the feature is called `Private fields promotion` I suppose it's Ok to use word `field`)

LanguageFeatures/Private-fields-promotion/promotion_A02_t02.dart
sgrekhov: No changes here

sgrekhov (commented):

LanguageFeatures/Private-fields-promotion/promotion_A02_t01.dart
eernstg: I think I wrote the previous comment on a web page where your previous comment hadn't been fetched. ;-)
To tests sounds good!

LanguageFeatures/Private-fields-promotion/not_promotable_A02_t04.dart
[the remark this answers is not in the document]
eernstg: Thx!

eernstg (approved):
LGTM

stereotype441:
> Updated. Please take another look

Updates LGTM!

PR #2264, opened 2023-09-08
#1400. Extension type. Add late instance variable test

sgrekhov:
This test fails in CFE. https://github.com/dart-lang/sdk/issues/53467

### DIFF
--- a/LanguageFeatures/Extension-types/syntax_A08_t04.dart
+++ b/LanguageFeatures/Extension-types/syntax_A08_t04.dart
@@ -1,0 +1,69 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A rule for <extensionTypeDeclaration> is added to the grammar,
+/// along with some rules for elements used in extension type declarations:
+///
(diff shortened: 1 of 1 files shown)


eernstg (approved):
LGTM.

I noted that line 37 has the name clash, so we can get a false positive on that one (missing error for having an instance variable at all, but test succeeds because we have two declarations named `id`, and that's enough to get the error we expect).  But that's no real problem because we have the other cases where there is no name clash.

However, we should have testing of the name clashes and the name near-clashes ('Class Member Conflicts'). I created https://github.com/dart-lang/co19/issues/2265 to handle that.

sgrekhov:
> LGTM.
> 
> I noted that line 37 has the name clash, so we can get a false positive on that one (missing error for having an instance variable at all, but test succeeds because we have two declarations named `id`, and that's enough to get the error we expect). But that's no real problem because we have the other cases where there is no name clash.
> 
> However, we should have testing of the name clashes and the name near-clashes ('Class Member Conflicts'). I created #2265 to handle that.

Tha name clash on line 37 here was made by purpose. What if because of some bug implementation will decide that this late variable is the same as the representation variable and won't report an error bacause of it. We have some name clash tests but I'll doublecheck. Thank you

eernstg:
> What if ...

Makes sense, thanks!

PR #2270, opened 2023-09-11
Fixes #2265. Extension types. Add more member conflict tests

### LINKED ISSUE #2265: Do we cover extension type member conflicts and name clashes?
Just noticed a case that we may not cover &mdash; but I did not search very thoroughly, so this issue might be a no-op because we have it already:

We should cover name clashes and name near-clashes with the representation variable of an extension type, because the representation variable is given special treatment in so many ways. Stuff along these lines:

```dart
extension type ET1(int id1) {
  void id() {} // No problems with this one, is just redeclared a few times.
  external int id2;
}

extension type ET2a(int id) {
  void id() {} // Name clash with representation variable.
}

extension type ET2b(int id) {
  set id(_) {} // OK, name is `id=`.
}

extension type ET2c(int id) implements ET1 {
  static int id = 1; // Name clash with representation variable.
}

extension type ET2d(int id) implements ET1 {
  static String get id1 => ''; // Class member conflict with `ET1` representation variable.
  void id2() {} // OK, redeclares.
}
```

sgrekhov:


eernstg (commented):
Looks good! I think a couple of expectations should be changed.

sgrekhov:
Thank you for noticing this! Updated

eernstg (approved):
LGTM

PR #2273, opened 2023-09-12
Fixes #2271. Representation variable is not promotable

### LINKED ISSUE #2271: co19/LanguageFeatures/Extension-types/promotion_A01_t01
```dart
  ET2<num?> et2 = ET2(2);
  if (et2.id != null) {
    et2.id.isEven;
  }
```
Do we expect promotion of `et2` to `ET2<int>` in the declaration?
I think this does not happen, by two reasons:
1. `ET2(2)` is inferred as `ET2<num?>`, not `ET2<int>`.
2. The declaration of `et2` does not promote it to `ET2<int>` anyway, because it is not a type of interest.

sgrekhov:


eernstg:
I'm sorry, wrong thinking on my part! Promotion of the representation variable is actually sound after all. Awaiting further discussion in https://github.com/dart-lang/sdk/issues/53446 (and perhaps there will be a language repository issue for this discussion, but it would then presumably be linked from https://github.com/dart-lang/sdk/issues/53446).

So I'm afraid we will have to abandon this PR.

sgrekhov:
Ok, no problem.

eernstg:
Thanks!

PR #2279, opened 2023-09-14
#2258. Only private field of an extension type is promotable

sgrekhov:
`LanguageFeatures/Extension-types/promotion_A01_t01.dart` deleted because if we change it to test a private field then it became a duplicate of `LanguageFeatures/Private-fields-promotion/promotion_A01_t09.dart`


eernstg (approved):
LGTM

PR #2280, opened 2023-09-15
Fixes #2147. Add access to this tests

### LINKED ISSUE #2147: Add tests for assess to `this`
We have no co19 tests for the following [statement](https://github.com/dart-lang/language/blob/bb001dc06b2909d16221baa5a91c348b0718c712/specification/dartLangSpec.tex#L2826) (10. Classes)

> The current instance (and hence its members) can only be accessed at specific
> locations in a class: We say that a location ℓ has access to this iff ℓ is inside the 
> body of a declaration of an instance member or a generative constructor, or in
> the initializing expression of a late instance variable declaration.

sgrekhov:


eernstg (approved):
LGTM!

PR #2285, opened 2023-09-21
Fixes #2284. Update `rename()` tests according to the changed documentation

### LINKED ISSUE #2284: rename_A02_t03 failing
[rename_A02_t03](https://github.com/dart-lang/co19/blob/00292501dd4d8b0e246130d55a024727e31d62eb/LibTest/io/Link/rename_A02_t03.dart) is failing because it expects that you cannot rename a `Link` to the path of an existing `File`.

That is actually allowed but our documentation was incorrect. I've fixed it in https://github.com/dart-lang/sdk/commit/5b7b8aab40e017d2a13348b36df3a2c2a5c42b71

I could not find any other tests impacted by this but I didn't look very carefully.

sgrekhov:


eernstg (approved):
LGTM.

@brianquinlan, do you have further comments?

eernstg:
Hope you're OK, and will be well soon! Waiting a few days should not be a problem.

LibTest/io/Directory/renameSync_A04_t01.dart
```diff
@@ -0,0 +1,40 @@
...
+import "../../../Utils/expect.dart";
+import "../file_utils.dart";
+
+main() async {
+  await inSandbox(_main);
+}
+
+_main(Directory sandbox) async {
+  Directory srcDir = getTempDirectorySync(parent: sandbox);
+  Directory linkDir = getTempDirectorySync(parent: sandbox);
```
brianquinlan, line 32: Do you need to create a new directory for this? Can't you just create this in the parent?

brianquinlan (approved):

LibTest/io/Directory/renameSync_A04_t01.dart
eernstg: OK, I misunderstood what the test is doing, wrote a comment, and then deleted the comment. Here's a better one. I think.

We're creating a `srcDir` in the sandbox root, call it `src`, and then we're creating a `linkDir` in the sandbox root, call it `lDir`, and then we're creating a link `link` in the sandbox root such that `link` is a link to `lDir`. Then we're renaming `src` to `link`, and that's expected to incur a dynamic error.

The helper method `getTempLinkSync` will, AFAICS, create a symlink in the underlying OS file system, pointing to `target`. Then we rename. The `rename` documentation says that ..

> If `newPath` identifies an existing entity of the same type, that entity is removed first. If `newPath` identifies an existing entity of a different type, the operation fails and an exception is thrown.

So we'd expect the exception because we're renaming a `Directory` to a `Link`. It should not make any difference at all whether the link refers to a directory or to any other kind of file system entity, or even if it is dangling.

So it doesn't hurt to have `linkDir` and letting `link` be a symlink to `linkDir`, but it doesn't actually matter (e.g., `link` could be a link to `not_existing_file_system_entity`, and it should not give rise to any errors to create such a dangling link --- after all, we might create the target later).

Looks like that `linkDir` could indeed be omitted. ;-)

eernstg (commented):
Took another look at `renameSync_A04_t01`.

LibTest/io/Directory/renameSync_A04_t01.dart
sgrekhov: Indeed. Fixed here and in the other places. Thank you!

sgrekhov (commented):
It turns out that if you are going to replace an existing link the can be different results depending on link's target. See https://github.com/dart-lang/sdk/issues/53583 So we need to test all the cases when the replaced link points to a directory, file and to another link. Added these test cases. Please take another look

LibTest/io/Directory/renameSync_A04_t01.dart
sgrekhov: > If newPath identifies an existing entity of the same type, that entity is removed first. If newPath identifies an existing entity of a different type, the operation fails and an exception is thrown.

I think, this statement is removed. https://dart-review.googlesource.com/c/sdk/+/326862/4/sdk/lib/io/file_system_entity.dart

eernstg: @sgrekhov, very good, it sounds like there was something to double check here.

@brianquinlan, as I understand the situation at this point there was a discrepancy between the documented behavior and the actual behavior, and https://dart-review.googlesource.com/c/sdk/+/326862/4/sdk/lib/io/file_system_entity.dart#b279 was the CL where the documentation was corrected to match the actual behavior.

This means that `rename`/`renameSync` would never throw an exception because the `newPath` is a file system entity of a different kind, it would just overwrite the existing entity and succeed. (The operation could surely still throw an exception in the case where the running process doesn't have the required permissions, but that's a property which is shared by all the file system operations.)

But in that case I'd expect that the test should be adjusted such that it does not `Expect.throws` anything.

eernstg:
I'm afraid I'll have to ask for some more input on this:

As far as I can see, methods on `FileSystemEntity` and subtypes thereof may throw a `FileSystemException` in the case where an OS level failure occurred (basically, this is always assumed to be possible, we could always add support for a new platform that has a new kind of OS level error).

This is documented in some cases (like ['directory.dart'](https://dart-review.googlesource.com/c/sdk/+/326862), line 242), and not in others (like https://github.com/dart-lang/sdk/blob/d7b522508e4705f9a5be58b2ff4d3880398eef05/sdk/lib/io/file_system_entity.dart#L279).

Perhaps the point is that the first situation is an error on all platforms (e.g., on gLinux we get "cannot overwrite non-directory '...' with directory '...'", and it might be an error on all other platforms, I don't know for sure). The second situation abstracts over several different kinds of file system entities, and only some of those have such errors (including the one which is mentioned in 'directory.dart').

Next, the pre-change documentation shown [here](https://dart-review.googlesource.com/c/sdk/+/326862/4/sdk/lib/io/file_system_entity.dart#b279) seems to imply that we have (or at least had) some situations where the failure arises in Dart code (not at the OS level). In the given case it was an error for renaming, say, a directory to a name which is already used by a symlink. Presumably that's not an error any more, but there might be other cases where the Dart code throws an exception even though there is no OS level error (and we'll never get far enough to try it). I'm not quite sure how this kind of error would be motivated, but it might be a safety measure because the given operation will be an OS error on some platforms, so we should play safe and make it fail on all platforms.

@brianquinlan, do you agree on this characterization of the kinds of errors that we have in 'dart:io'?

I'm asking because I think it's less than obvious how to test cases like https://github.com/dart-lang/co19/pull/2285/files/3e85921f6eb85c43c14af2d0ee5343752914263c#diff-b9b472b71e013530581d75686bae815ff2a60ac614546dded9531853d328311f, and I'd like to hear a simple story about it if there is one. ;-)

LibTest/io/Directory/renameSync_A04_t01.dart
brianquinlan: @eernstg That's correct. The TL;DR is:
- `File`/`Link` `rename`/`renameSync` can replace other  `File`/`Link` but **not* `Directory`
- `Directory` `rename`/`renameSync` can **not** replace an existing `File`/`Link`/`Directory` **except** on POSIX where it can replace an empty `Directory`

sgrekhov:
`LibTest/io/File/rename(Sync)_A03_t01.dart` tests files renaming to a directory
`LibTest/io/Link/rename(Sync)_A02_t01.dart` test links renaming to a directory

I think it make sense to add a couple of tests for links. To check behavior of link with target directory/file/another likn/not existing entity renaming to a directory. Will do it tomorrow

sgrekhov:
Added new bunch of tests. Now we have a full coverage. File/Directory/Link with target File/Directory/Link/not_existing_entity is renamed and there is existing File/Directory/Link with target File/Directory/Link/not_existing_entity

Sorry that this PR became very big and, most probably, hard for review. I usually trying to avoid ths

eernstg:
Important TL;DR from @brianquinlan about the rules governing the `rename`/`renameSync` operations:

> The TL;DR is:
> - `File`/`Link` `rename`/`renameSync` can replace other  `File`/`Link` but **not* `Directory`
> - `Directory` `rename`/`renameSync` can **not** replace an existing `File`/`Link`/`Directory` **except** on POSIX where it can replace an empty `Directory`

sgrekhov:
I'll add it to the README but, please note https://github.com/dart-lang/sdk/issues/53583 On Windows if there is a `Link` pointing to a `Directory` then you cannot rename `File` to that `Link`

eernstg:
Sounds good!

Ready to land this now, @sgrekhov, @brianquinlan?

sgrekhov:
Added TL;DR to assertion as an informal note. I think this PR can be submitted

eernstg (approved):
LGTM. Some day, someone will be happy that someone included that note! ;-)

I'll land this now.

@brianquinlan, thanks for all the help! I'm landing this PR now. I think we're done, but please create a co19 repo issue if there is still something to do, and then we'll do it in a fresh PR.

Issue #2275, opened 2023-09-12
Add more 'noSuchMethod' tests


sgrekhov:
We only have `Language/Classes/Superinterfaces/no_member_t04.dart` that tests implicit `noSuchMethod` forwarder. Need to improve the coverage and add more tests. We have no co19 tests for the cases described in https://github.com/dart-lang/language/issues/3331

PR #2288, opened 2023-09-28
#2275. More `noSuchMethod` tests added. Part 2

sgrekhov:


Language/Classes/Instance_Methods/Method_noSuchMethod/forwarder_A01_t05.dart
```diff
@@ -0,0 +1,74 @@
...
+}
+
+enum E2 with D2 {
+  e1, e2;
+  String _m();
+  dynamic noSuchMethod(Invocation i) => "E2";
+}
+
+main() {
+  Expect.equals("D1", C1()._m());
```
eernstg, line 68: This kind of test would actually succeed even if an implementation of Dart would use the old approach (it would then detect that there is no `_m` member of the given object `C1()`, and it would invoke `noSuchMethod` using the actual arguments).

So we need to do something which is a bit more circumspect in order to ensure that there _is_ a noSuchMethod forwarder.

One thing we can do is to tear it off: If we had that old semantics and invoke `C1()._m` then it would invoke `noSuchMethod` and we would receive `"C1"`, which would then probably give rise to a dynamic type error because `C1()._m` should yield a function object.

So we'd tear off the method, then test that it is a function object with the expected type (`is String Function()`), and then we'd know that there _is_ a real member named `_m` in the object, and that must be the noSuchMethod forwarder. We could then call the function object and see that it returns `"C1"`, such that we can see that it works as expected.

eernstg: With the old semantics it was necessary to write code in `noSuchMethod` to handle the case where the given `Invocation` is a getter invocation (that's a tear-off operation, so we should return a function object) and when it's a method invocation (that's the case where we'd use the behavior that we have today), so it _was_ possible to get the same behavior.

However, in this case we know what's in the body of the relevant `noSuchMethod`, so if the tear-off works then we also know that there is a noSuchMethod forwarder.

Language/Classes/Instance_Methods/Method_noSuchMethod/forwarder_A03_t01.dart
```diff
@@ -23,53 +23,44 @@
...
+import "no_such_method_lib.dart";
+import "../../../../Utils/expect.dart";
 
-class C1 extends B implements A {
-//    ^^
-// [analyzer] unspecified
-// [cfe] unspecified
-  dynamic noSuchMethod(Invocation i) => "C1";
+class C extends D0 {
+  String _m();
```
eernstg, line 34: I think the reason why we might want to have a private method named `_m`-in-this-library is that it must be a different method than the one named `_m`-in-no_such_method_lib.dart in `D0`, and we can detect this because they behave differently.

But it would still be helpful to have a comment to that effect. Perhaps something like

  // This `_m` is different from `D0._m`: They behave differently.

Language/Classes/Instance_Methods/Method_noSuchMethod/forwarder_A03_t01.dart
```diff
@@ -23,53 +23,44 @@
...
   dynamic noSuchMethod(Invocation i) => "E";
 }
 
 main() {
-  print(C1);
-  print(C2);
-  print(MA);
-  print(E);
+  Expect.equals("C", C()._m());
+  Expect.throws(() {C().test();});
```
eernstg, line 53: I think we ought to test that the `_m`-in-no_such_method_lib.dart method can be torn off, and also that it throws when invoked. So perhaps `test` should tear it off and return the function object? That function object can be called here just fine, and then we can `Expect.throws` it.

Language/Classes/Instance_Methods/Method_noSuchMethod/superclass_A01_t01.dart
```diff
@@ -0,0 +1,59 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if the name m is noSuchMethod
+/// forwarded in a concrete class C, and a superclass of C has an accessible
+/// concrete declaration of m which is not a noSuchMethod forwarder.
+///
+/// @description Checks that it is a compile-time error if the name `m` is
+/// `noSuchMethod` forwarded in a concrete class `C`, and a superclass of `C`
+/// has an accessible concrete declaration of `m` which is not a `noSuchMethod`
+/// forwarder.
```
eernstg, line 12: Perhaps mention that `m` is only noSuchMethod forwarded here if the inherited method has a signature which is not a correct override of the signature in the interface of the class here. Otherwise we'd just have an implemented method which is fine, and that's the one we'd invoke if we call `m()`. Something like ..

```suggestion
/// forwarder. Note that in this situation the inherited `m` must have a
/// signature which is not a correct override of the signature of `m`
/// in this class, otherwise `m` wouldn't be noSuchMethod forwarded.
```

Language/Classes/Instance_Methods/Method_noSuchMethod/superclass_A01_t02.dart
```diff
@@ -0,0 +1,56 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if the name m is noSuchMethod
+/// forwarded in a concrete class C, and a superclass of C has an accessible
+/// concrete declaration of m which is not a noSuchMethod forwarder.
+///
+/// @description Checks that it is a compile-time error if the name `m` is
+/// `noSuchMethod` forwarded in a concrete class `C`, and a superclass of `C`
+/// has an accessible concrete declaration of `m` which is not a `noSuchMethod`
+/// forwarder.
```
eernstg, line 12: An extra sentence would probably be helpful here, too, similarly to Language/Classes/Instance_Methods/Method_noSuchMethod/superclass_A01_t01.dart.

Language/Classes/Instance_Methods/Method_noSuchMethod/superclass_A01_t03.dart
```diff
@@ -0,0 +1,59 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if the name m is noSuchMethod
+/// forwarded in a concrete class C, and a superclass of C has an accessible
+/// concrete declaration of m which is not a noSuchMethod forwarder.
+///
+/// @description Checks that it is a compile-time error if the name `m` is
+/// `noSuchMethod` forwarded in a concrete class `C`, and a superclass of `C`
+/// has an accessible concrete declaration of `m` which is not a `noSuchMethod`
+/// forwarder.
```
eernstg, line 12: Similar extra sentence would be helpful here.

eernstg (commented):
Looks good. I think it would be helpful to add an extra sentence to some comments. More crucially, I think we need to tear off some noSuchMethod forwarders in order to show that they exist: If we just check that the invocation throws then they might not exist and the test could still succeed.

sgrekhov:
Comments updated. As for tear-off we have the following statement in the spec

> A noSuchMethod forwarder is a concrete member of C with the signature taken from the interface of C, and with the same default value for each optional parameter. It can be invoked in an ordinary invocation and in a superinvocation, and when m is a method it can be closurized using a property extraction.

I'm working on the tests for this statement now (it's going to be Part 3 of `noSuchMethod` tests). There I'm going to add the tests that will check that `C().m` will torn-off a forwarder rather then invoke a `noSuchMethod` for a non-existing getter. And then tests static type of the torn-off method. So, will it be ok if we'll add tear-off tests there?

eernstg:
Yep, let's do that. I'll land this one.

eernstg (approved):
LGTM (tear-offs will come later).

PR #2294, opened 2023-10-02
#2275. Add `noSuchMethod` tests. Part 3

sgrekhov:


Language/Classes/Instance_Methods/Method_noSuchMethod/forwarder_A01_t06.dart
```diff
@@ -0,0 +1,105 @@
...
+  Expect.equals("C1", C1().m);
+  Expect.equals("C2", C2().a);
+  Expect.equals("C3", C3().a);
+  Expect.equals("C4", C4().a);
+  Expect.equals("M1", MA1().m);
+  Expect.equals("M2", MA2().a);
+  Expect.equals("M3", MA3().a);
+  Expect.equals("E1", E1.e1.m);
+  Expect.equals("E2", E2.e1.a);
+  Expect.equals("E3", E3.e1.a);
```
eernstg, line 104: These tests will reveal that `noSuchMethod` is invoked, but they won't determine whether `noSuchMethod` is invoked directly or there is a noSuchMethod forwarder.

We could detect that there is a noSuchMethod forwarder if we had another version of this test where `noSuchMethod` would return a value which isn't type correct for the given getter: `noSuchMethod` would be happy to do that, but if there is a noSuchMethod forwarder for the getter then it should catch the case where the returned value has a wrong type.

Language/Classes/Instance_Methods/Method_noSuchMethod/forwarder_A01_t07.dart
```diff
@@ -0,0 +1,150 @@
...
+  log = "";
+  E1.e1.m = "";
+  Expect.equals("E1", log);
+  log = "";
+  E2.e1.a = "";
+  Expect.equals("E2", log);
+  log = "";
+  E3.e1.a = "";
+  Expect.equals("E3", log);
+  log = "";
```
eernstg, line 149: I think it should be possible to make the distinction between the case where we actually have a noSuchMethod forwarder for the setter, and the case where `noSuchMethod` is invoked directly, if we could pass an argument to the setter which is type incorrect (but statically typed as `dynamic`): `noSuchMethod` would be happy to receive that argument, but an actual forwarder should throw.

Language/Classes/Instance_Methods/Method_noSuchMethod/invocation_A01_t01.dart
```diff
@@ -0,0 +1,73 @@
...
+    Expect.equals(42, super.m2(2, s: "2"));
+    Expect.equals(42, this.m1(1, "1"));
+    Expect.equals(42, this.m2(2, s: "2"));
+  }
+}
+
+class C2 with A {
+  test() {
+    Expect.equals(42, this.m1(1, "1"));
+    Expect.equals(42, this.m2(2, s: "2"));
```
eernstg, line 39: The reason why we can't call `super.` in this case would be that the superclass `Object with A` is abstract (a mixin application used as a superclass in a class declaration is always abstract), so there is no noSuchMethod forwarder for `m1` or `m2`.

It would probably be helpful for future readers of this code if there's a comment to that effect.

Also: Did some other test cover the case where `super.m1(...)` should be an error because there is no noSuchMethod forwarder because it's being looked up in a mixin application?

Language/Classes/Instance_Methods/Method_noSuchMethod/invocation_A01_t01.dart
```diff
@@ -0,0 +1,73 @@
...
+  }
+}
+
+class MA = A with M;
+
+enum E with A {
+  e1, e2;
+  test() {
+    Expect.equals(42, this.m1(1, "1"));
+    Expect.equals(42, this.m2(2, s: "2"));
```
eernstg, line 58: A comment would be helpful here, too ("`_Enum with A` is abstract, so `super.` is an error").

eernstg (commented):
Very good! Just one addition would be useful: As far as I can see, we don't catch the case where a getter or setter is noSuchMethod forwarded, but no forwarder has been generated (and `noSuchMethod` is just invoked directly). I believe we would be able to detect that there is a noSuchMethod forwarder by using wrong types and seeing that there is a dynamic type check (so we do get a exception).

Language/Classes/Instance_Methods/Method_noSuchMethod/forwarder_A01_t06.dart
sgrekhov: Added `forwarder_A01_t08.dart` testing it

Language/Classes/Instance_Methods/Method_noSuchMethod/forwarder_A01_t07.dart
sgrekhov: Added `forwarder_A01_t09.dart` testing it

eernstg: Thanks!

Language/Classes/Instance_Methods/Method_noSuchMethod/forwarder_A01_t06.dart
eernstg: Thanks!

Language/Classes/Instance_Methods/Method_noSuchMethod/forwarder_A01_t09.dart
```diff
@@ -0,0 +1,136 @@
...
+  e1,
+  e2;
+
+  dynamic noSuchMethod(Invocation i) {
+    log = "E3";
+  }
+}
+
+main() {
+  Expect.throws(() {C1().m = 1 as dynamic;});
```
eernstg, line 116: Thinking about this one more time, we probably have to use `(C1() as dynamic).m = 1` in order to be safe: It is possible that the compiler will simply generate `C1().m = (1 as dynamic) as String` based on the static knowledge about the interface of `C1`, and then we still won't know whether there's a forwarder or not. But if we do `(C1() as dynamic).m = 1` and get a type error at run time then there _must_ be a forwarder (and it does have a parameter type that works as expected).

sgrekhov:
Makes sense! Updated

eernstg (approved):
LGTM!

Issue #2291, opened 2023-09-28
create_A04_t04 should not expect that the target of the symlink must exist on Windows


brianquinlan:
This is the [relevant code](https://github.com/dart-lang/co19/blob/dcc3cbe37ae5aa0212987d2038d70fba7139a3ed/LibTest/io/Link/create_A04_t04.dart#L46).

I've updated the docs in [this PR](https://dart-review.googlesource.com/c/sdk/+/328661/3/sdk/lib/io/link.dart).

I think that this test can just be deleted.

sgrekhov:
According to the updated docs if we are create Link to a not existing directory then this link should be created and be a file simlink. I'll change the tests accordingly.
If you don't mind I'll do it after the landing of the PR

brianquinlan:
Sure. The PR is documentation-only though.

PR #2299, opened 2023-10-06
#2291. Update `Link.create()` tests according to the documentation

sgrekhov:
These tests written according to the documentation and https://github.com/dart-lang/sdk/issues/53684

This is just the part of the changes. The second part should add tests checking behaviour of links when its target is deleted and another `FileSystemEntity` is created instead. Say delete `File` target and create a `Directory` with the same name instead. 

The third part will update `createSync()` tests

### DIFF
--- a/LibTest/io/Link/create_A01_t01.dart
+++ b/LibTest/io/Link/create_A01_t01.dart
@@ -6,33 +6,36 @@
 ///  String target, {
 ///  bool recursive: false
 ///  })
-/// Creates a symbolic link. Returns a Future<Link> that completes with the link
-/// when it has been created. If the link exists, the future will complete with
-/// an error.
+/// Creates a symbolic link in the file system.
 ///
-/// If recursive is false, the default, the link is created only if all
-/// directories in its path exist. If recursive is true, all non-existing path
-/// components are created. The directories in the path of target are not
-/// affected, unless they are also in path.
+/// The created link will point to the path at `target`, whether that path
+/// exists or not.
 ///
-/// On the Windows platform, this will only work with directories, and the target
-/// directory must exist. The link will be created as a Junction. Only absolute
-/// links will be created, and relative paths to the target will be converted to
-/// absolute paths by joining them with the path of the directory the link is
-/// contained in.
+/// Returns a `Future<Link>` that completes with the link when it has been
+/// created. If the link path already exists, the future will complete with an
+/// error.
 ///
-/// On other platforms, the posix symlink() call is used to make a symbolic link
-/// containing the string target. If target is a relative path, it will be
-/// interpreted relative to the directory containing the link.
-/// @description Checks that this method creates the link
+/// If `recursive` is `false`, the default, the link is created only if all
+/// directories in its path exist. If `recursive` is `true`, all non-existing
+/// parent paths are created first. The directories in the path of target are
+/// not affected, unless they are also in [path].
 ///
-/// @note The test should run with the Administrator priveleges on Windows.
-/// Dart API Spec reads:
+/// On the Windows platform, this call will create a true symbolic link instead
+/// of a junction. The link represents a file or directory and does not change
+/// its type after creation. If `target` exists then the type of the link will
+/// match the type `target`, otherwise a file symlink is created.
+///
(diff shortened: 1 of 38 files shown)


LibTest/io/Link/create_A01_t01.dart
```diff
@@ -50,6 +53,7 @@ _main(Directory sandbox) async {
   await link.create(target.path).then((Link created) {
     Expect.isTrue(created.existsSync());
     Expect.equals(link.path, created.path);
+    Expect.equals(target.path, created.targetSync());
```
eernstg, line 56: Just wondering, is there a danger in mixing sync and async versions of io functions (deadlocks)?

It looks like we have this kind of mixture in some (perhaps lots of) other locations already, but if it was never discussed explicitly then it would be nice to know that it's safe.

LibTest/io/Link/create_A04_t01.dart
```diff
@@ -38,13 +48,12 @@ main() async {
...
-    });
-  }
+  Directory target = getTempDirectorySync(parent: sandbox);
+  Link link = Link(getTempFilePath(parent: sandbox));
+  asyncStart();
+  await link.create(target.path).then((Link created) {
+    Expect.equals(FileSystemEntityType.directory,
+        FileSystemEntity.typeSync(created.path));
+    asyncEnd();
+  });
```
eernstg, line 58: So, just trying to explain the change, this would previously fail because it was executed on Windows in non-privileged mode, and now it's running in a suitable privileged mode (involving "Administrator mode" or a config setting called 'Developer Mode') on Windows?

eernstg: Similarly for some other tests with the same kind of change, and least the next couple of tests.

eernstg (approved):
LGTM.

@brianquinlan, do you have further comments? It should suffice to look at libraries where the diff includes code, and skip the ones that only have comment updates.

I was wondering about one thing: A few libraries have been changed such that code which was previously only executed when `Platform.isWindows` will now be executed on all platforms (as far as I can see). This is probably fine, and this change is probably needed because the test runs will use 'Administrator Mode' on Windows (or a configuration setting called 'Developer Mode'). However, it is slightly puzzling that we can't see any other changes associated with this modifier way of executing the test.

Is this choice of "some-suitably-privileged-mode" specified in some configuration file which is not part of this PR, perhaps outside co19 entirely?

In that case it might be useful to document how the execution mode is chosen (at least for Windows, if that concept just doesn't occur for other platforms), perhaps in those few tests where it matters.

LibTest/io/Link/create_A01_t01.dart
sgrekhov: Tests mainly written usyng sync syntax. But here we are testing async function `create()` so, sync and async code is mixed here. But there were no any failures caused by this. We'll see flaky tests if the code is racy

sgrekhov (commented):
Looks like all Windows bots are run in Administrator mode. I don't know where it is configured

LibTest/io/Link/create_A04_t01.dart
eernstg: OK, thanks!

eernstg:
> all Windows bots are run in Administrator mode

OK, so let's assume that the privileged execution mode it is documented somewhere else (maybe in the location where the administrator mode is actually chosen), and just keep relying on it here. It _is_ already implied by the comments in several tests that it must be executed in a privileged manner.

LibTest/io/Link/create_A01_t01.dart
eernstg: OK, I created https://github.com/dart-lang/sdk/issues/53709 to get some extra input on the sync/async mixture. I think it is safe, but I don't know enough to explain exactly why it would be safe. ;-)

In any case, if we need to do anything then it can be done in another PR.

sgrekhov:
Two more commits were made. Please take another look

eernstg:
@brianquinlan, I'll land this now. Your input is still very welcome, but we'll use a separate PR to handle any issues that you might wish to raise.

PR #2301, opened 2023-10-10
#2291. Update `Link.create()` according to the documentation. Part 2

sgrekhov:
Tests written according to the https://github.com/dart-lang/sdk/issues/53684
On Windows invalid `Link` reports itself as `FileSystemEntityType.link`  (`FileSystemEntityType.notFound` on other platforms). Also, on Windows, Link created as link pointing to a directory cannot point to a file and vice versa


eernstg (approved):
LGTM. Having seen a bunch of these, I'll land this now.

@brianquinlan, if you have any further comments then please notify us and then we can handle them in a subsequent PR.

PR #2308, opened 2023-10-13
Fixes #2307. Don't expect private fields promotion on static fields

### LINKED ISSUE #2307: Two tests in LanguageFeatures/Private-fields-promotion expect to be able to type promote fields of a static value
The tests `LanguageFeatures/Private-fields-promotion/promotion_A01_t07` and `LanguageFeatures/Private-fields-promotion/promotion_A01_t08` both contain code that looks roughly like this:
```dart
enum E {
  e1, e2;

  final int? _x = 42;
}

main() {
  if (E.e1._x != null) {
    E.e1._x.isEven;
  }
}
```

For Dart 3.2, field promotion only considers an expression of the form `target._fieldName` to be stable (and thus promotable) if `_fieldName` represents a promotable field and `target` is one of the following:
- `this` (includes implicit uses of `this`, so bare `_fieldName` is considered promotable inside an instance method of a class)
- `super`
- A local variable that isn't modified or write captured during the time period covered by the potential promotion.
- An expression of the form `target._fieldName` that's considered stable.

We have considered expanding this list to include static constant fields such as `E.e1` (https://github.com/dart-lang/language/issues/3326), but we haven't done so; if we do decide to make this change, we won't make it until a future Dart release.

Accordingly, these two tests need to be updated so that they don't expect expressions like `E.e1._x` to undergo type promotion.

sgrekhov:


eernstg (approved):
LGTM

PR #2309, opened 2023-10-13
Fixes #2306. Fix promoted instance type in not_promotable_A01_t04.dart

### LINKED ISSUE #2306: LanguageFeatures/Private-fields-promotion/not_promotable_A01_t04 has incorrect expectation
`LanguageFeatures/Private-fields-promotion/not_promotable_A01_t04` looks like this:
```dart
class A {
  void foo() {}
}

class C<T> {
   external final T _x;

  void test() {
    if (_x is A) {
      _x.foo();
//       ^^^^^
// [analyzer] unspecified
// [cfe] unspecified
    }
  }
}

main() {
  C c = C();
  if (c._x is A) {
    c._x.foo();
//       ^^^^^
// [analyzer] unspecified
// [cfe] unspecified
  }
  c.test();
}
```

The test is trying to verify that `C._x` does not type promote.

The first expectation (inside `C.test`) works fine, because inside an instance method of `C`, the unpromoted type of `_x` is `T`, and therefore trying to call `_x.foo()` is an effective way of verifying that `_x` wasn't promoted to `A`.

The second expectation (inside `main`) is problematic, because `C c = C();` gets type inferred as `C<dynamic> c = C();`, therefore the unpromoted type of `c._x` is `dynamic`, and so the call to `c._x.foo()` works regardless of whether `c._x` has been type promoted.

One possible fix would be to change the first line of `main` to `var c = C<Object>()`. That would ensure that the unpromoted type of `c._x` is `Object`, and so the call to `c._x.foo()` would be an effective way of verifying that `c._x` hasn't been type promoted.

sgrekhov:


eernstg (approved):
LGTM

PR #2312, opened 2023-10-16
Fixes #2304. Add more `Object` member tests

### LINKED ISSUE #2304: [Extension types] Update tests about Object members
Update co19 tests according to https://github.com/dart-lang/language/pull/3395

sgrekhov:


LanguageFeatures/Extension-types/static_analysis_member_invocation_A01_t04.dart
```diff
@@ -0,0 +1,47 @@
...
+
+extension type ET1(B b) implements A {}
+
+extension type ET2(B b) implements  ET1, B {}
+
+void main() {
+  var e2 = ET2(B());
+  ET1 e1 = e2;
+
+  e2.noSuchMethod(Invocation.method(Symbol("test"), []), true); // OK, signature is `noSuchMethod(Invocation, [bool])`.
```
eernstg, line 41: Looks like >80 chars, so we might as well move the comment up (unless there is something special about tests or co19 that requires the comments on the same line? I can't think of anything...):
```suggestion
  // OK, signature is `noSuchMethod(Invocation, [bool])`.
  e2.noSuchMethod(Invocation.method(Symbol("test"), []), true);
```

Similarly for line 41.

LanguageFeatures/Extension-types/static_analysis_member_invocation_A01_t03.dart
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @override
+  Type get runtimeType {
+    log += "B.runtimeType";
+    return super.runtimeType;
+  }
+
+  @override
+  bool operator ==(Object? other) {
```
eernstg, line 46: We could use a covariant parameter here, in order to verify that the static analysis takes this member signature into account at a call site (just like the `e1`/`e2` examples in other tests), and it doesn't just rely on the signature in `Object`.

eernstg: The variant that has a covariant parameter type would probably need to be in a different library, because we'd (most likely) need to test that statically using some situations that cause a compile-time error (like `myET1 == 42` where the parameter type of `myET1` is `ET1`).

LanguageFeatures/Extension-types/static_analysis_member_invocation_A01_t03.dart
```diff
@@ -0,0 +1,79 @@
...
+import "../../Utils/expect.dart";
+
+String log = "";
+
+class A {
+}
+
+class B implements A {
+  @override
+  int get hashCode {
```
eernstg, line 34: It would be great if we could verify that a call site where the receiver has type `ET2` _will_ use the signature from `B` for `hashCode`, so it would be nice if this signature could be different from the one in `Object`.

One way to do that could be to return `ET3` which is an extension type with `implements int`, and then check that `ET3 _ = myET2.hashCode;` has no compile-time error.

Similarly for `runtimeType`.

The body probably needs to invoke a constructor of `ET3`, because the whole point is that `int` and `Type` shouldn't be assignable to `ET3`.

eernstg (commented):
Looks good!

Just one little twist could be useful: It would be good if we could verify in all cases that the tools use the most specific (combined) member signature for all `Object` members, rather than unconditionally using the signature from `Object`.

LanguageFeatures/Extension-types/static_analysis_member_invocation_A01_t04.dart
sgrekhov: Fixed

sgrekhov (commented):

LanguageFeatures/Extension-types/static_analysis_member_invocation_A01_t05.dart
```diff
@@ -0,0 +1,61 @@
...
+  ET1 e1 = e2;
+
+  int hc1 = e1.hashCode;
+  IntET hc2 = e2.hashCode;
+
+  Type t1 = e1.runtimeType;
+  TypeET t2 = e2.runtimeType;
+
+  bool b1 = e1 == e1;
+  BoolET b2 = e2 == e2;
```
eernstg, line 60: Cool!

LanguageFeatures/Extension-types/static_analysis_member_invocation_A01_t03.dart
eernstg: Thanks!

PR #2315, opened 2023-10-18
#1400. [Extension types] Add more superinterfaces tests

sgrekhov:


LanguageFeatures/Extension-types/superinterfaces_of_extension_type_A03_t02.dart
```diff
@@ -0,0 +1,42 @@
...
+extension type ET2<T extends num>(A<T> _) implements A<num>, A<T> {}
+//                                                             ^
+// [analyzer] unspecified
+// [cfe] unspecified
+extension type ET3<T extends num>(T _) implements ET<T>, ET<num> {}
+
+main() {
+  print(ET1);
+  print(ET2);
+  print(ET3);
```
eernstg, line 41: It takes a considerably more complex implementation to detect whether a whole superinterface graph has this kind of conflict, compared to the case where the two occurrences of `A<...>` are in the same syntactic list. So we should have a couple of test cases where they occur in different syntactic lists (like: both occur in superinterfaces; one in the current declaration and another in a superinterface; and a case where the two types aren't syntactically different: `E3 implements E2a<int>, E2b<String>` and `E2a<X> implements A<X>` and `E2b<X> implements A<X>`). Finally, we should also have some cases where the conflict occurs with an extension type (like a copy of the above, but with `extension type A<X> ...`).

I'm not quite sure those things aren't checked already in some other tests, but it would be great if you could add those cases if they are indeed missing at this time.

LanguageFeatures/Extension-types/superinterfaces_of_extension_type_A08_t01.dart
```diff
@@ -0,0 +1,56 @@
...
+extension type ET4<T extends num>(B<T> _) implements B<T>, B<T> {}
+//                                                         ^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+extension type ET5<T extends num>(T _) implements ET<T>, ET<T> {}
+//                                                       ^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
```
eernstg, line 49: Perhaps add just one case to verify that there is no error if the two occurrences of the same type are not in the same list. E.g., in the current declaration and in a superinterface:

```dart
extension type ET6(A a) implements A {}
extension type ET7(A a) implement A, ET6 {}
```

eernstg (commented):
Looks good! However, I think it would be valuable to add some test cases where the "`A<T1>`, `A<T2>`, `T1 != T2`" error exists because of `A<T1>` and `A<T2>` that occur in different syntactic locations.

eernstg:
If I got the history right about where the little snippets of quoted text came from then this would be an example of 'one in the current declaration and another in a superinterface':

```dart
class A<X> {}
extension type ET1(A<Never> a) implements A<int> {}
extension type ET2(A<Never> a) implements ET1, A<String> {}
```

The point is that it is a conflict that `ET2` implements `A<int>`, and `ET2` implements `A<String>` at the same time, even in the case where the two types of the form `A<...>` do not occur in the same syntactic list.

sgrekhov:
Thank you! Tests added. Please review

LanguageFeatures/Extension-types/superinterfaces_of_extension_type_A03_t02.dart
eernstg: Thanks!

eernstg (approved):
LGTM

PR #2325, opened 2023-10-24
Fixes #2323. Don't test datagram size more than 65503

### LINKED ISSUE #2323: io/RawDatagramSocket/receive_A02_t02.dart should not test maximum packet size
There is no maximum size limit. See https://dart-review.googlesource.com/c/sdk/+/331422

sgrekhov:
@brianquinlan seems that maximum size of 65503 exists on MacOS only. See https://dart-current-results.web.app/#/filter=co19/LibTest/io/RawDatagramSocket/receive_A02_t02. But I didn't find any specification confirming that. So, I preserved the test which confirms that 65503 bytes can be received and deleted the one checking what's going on when UDP size is more than 65503

### DIFF
--- a/LibTest/io/RawDatagramSocket/receive_A02_t01.dart
+++ b/LibTest/io/RawDatagramSocket/receive_A02_t01.dart
@@ -4,8 +4,6 @@
 
 /// @assertion Datagram receive()
 /// Receive a datagram.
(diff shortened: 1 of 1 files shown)


eernstg (approved):
LGTM

PR #2331, opened 2023-10-25
Fixes #2330. Remove obsolete error expectation

### LINKED ISSUE #2330: LanguageFeatures/Extension-types/syntax_A04_t01
The test expects an analyzer diagnostic on the line containing `print(ET);`. This diagnostic was previously being generated as a result of a bug that prevented the analyzer from seeing the declaration of the extension type. That bug been fixed by https://dart-review.googlesource.com/c/sdk/+/331749, so the second diagnostic is no longer being produced (and never ought to have been produced). Please remove the expectation.

sgrekhov:


eernstg (approved):
LGTM

PR #2340, opened 2023-11-01
Fixes #2339. Add tests that extension can't have name `type`

### LINKED ISSUE #2339: Extension can't have name `type`
Add tests checking that extension can't have name `type`. See https://github.com/dart-lang/language/pull/3431

sgrekhov:


eernstg (commented):
Do we already have tests where it is verified that an `extension type` declaration cannot have a name which is an built-in identifier?

(We have some other tests with `class abstract {}` and stuff like that, so we should have one with something like `extension type N(int it) {}` for all built-in identifiers `N`.)

sgrekhov:
Added such tests (and discovered an analyzer's issue). I didn't test the whole list of the reserved words/built-in identifiers because there can be issues when the test expects too much errors (tools may stop reporting errors ay some point). Please take another look

eernstg:
Thanks!

eernstg (commented):
Thanks! There is one issue, though: It should not be an error to have an `extension` or an `extension type` whose name is an `OTHER_IDENTIFIER_NOT_TYPE`.

(Just double checked the grammar in https://github.com/dart-lang/language/pull/3431/files, I don't see how that could be an error. If it is an error after all then I missed something, and I'd need to fix the grammar.)

sgrekhov:
Fixed. Please review

eernstg (approved):
LGTM

PR #2341, opened 2023-11-01
#1400. Add test that extension type and its representation type cannot be in the same constant collection

sgrekhov:


eernstg (approved):
LGTM

sgrekhov:
@parlough thank you for finding this!

PR #2358, opened 2023-11-07
Fixes #2356. Fix return type of `this`

### LINKED ISSUE #2356: Nullable this as non-nullable
The tests

`co19/LanguageFeatures/Extension-types/dynamic_semantics_member_invocation_A02_t02`
`co19/LanguageFeatures/Extension-types/dynamic_semantics_member_invocation_A02_t03`

seem to assume that `this` is non-nullable even when the representation type is not non-nullable:
```
extension type ET1<T>(T id) {
  Object test1() => this;
```
The getter should probably return `Object?` instead.

sgrekhov:


eernstg (approved):
LGTM

Issue #2359, opened 2023-11-07
Some tests require an update due to the recent changes in the nullability rules for extension types


chloestefantsova:
The following tests need to be updated in accordance with https://github.com/dart-lang/language/pull/3434.

* co19/LanguageFeatures/Extension-types/dynamic_semantics_extension_type_A03_t02
* co19/LanguageFeatures/Extension-types/static_analysis_extension_types_A17_t01
* co19/LanguageFeatures/Extension-types/upper_bound_A01_t03
* co19/LanguageFeatures/Extension-types/upper_bound_A01_t05
* co19/LanguageFeatures/Extension-types/upper_bound_A01_t06
* co19/LanguageFeatures/Subtyping/static/generated/extension_type_arguments_binding_A02_t01
* co19/LanguageFeatures/Subtyping/static/generated/extension_type_arguments_binding_A02_t02
* co19/LanguageFeatures/Subtyping/static/generated/extension_type_arguments_binding_A02_t03
* co19/LanguageFeatures/Subtyping/static/generated/extension_type_class_member_A02_t01
* co19/LanguageFeatures/Subtyping/static/generated/extension_type_class_member_A02_t02
* co19/LanguageFeatures/Subtyping/static/generated/extension_type_class_member_A02_t03 
* co19/LanguageFeatures/Subtyping/static/generated/extension_type_global_variable_A02_t01
* co19/LanguageFeatures/Subtyping/static/generated/extension_type_local_variable_A02_t01
* co19/LanguageFeatures/Subtyping/static/generated/extension_type_return_value_A02_t01
* co19/LanguageFeatures/Extension-types/nullability_A02_t01

This was detected while working on the update to the CFE implementing the spec update: https://dart-review.googlesource.com/c/sdk/+/333500.

Issue #2342, opened 2023-11-01
Add tests  for extension types and subtype relations to Object


sgrekhov:
Add tests according to the https://github.com/dart-lang/language/pull/3434

eernstg:
Note that apart from adding new tests, some existing tests updated in https://github.com/dart-lang/co19/pull/2360 need updates in their `@assertion` and/or `@description` sections.

PR #2360, opened 2023-11-08
#2359. Update tests according to the new nullability rules. Part 1

sgrekhov:
This PR contains not all of the tests mentioned in #2359. It contains only the tests that can be updated to not fail with both old and new nullability rules. So, they won't break anything and can be merged safely. 

Other tests will be fixed after the landing of the https://dart-review.googlesource.com/c/sdk/+/333500

### DIFF
--- a/LanguageFeatures/Extension-types/static_analysis_extension_types_A17_t01.dart
+++ b/LanguageFeatures/Extension-types/static_analysis_extension_types_A17_t01.dart
@@ -26,6 +26,6 @@ extension type V1(int id) {}
 extension type V2<T1, T2 extends num?>(T1 id) {}
 
 main() {
-  Object v1 = V1(42);
-  Object v2 = V2<String, int?>("42");
+  Object? v1 = V1(42);
+  Object? v2 = V2<String, int?>("42");
 }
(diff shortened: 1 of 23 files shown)


LanguageFeatures/Extension-types/static_analysis_extension_types_A17_t01.dart
```diff
@@ -26,6 +26,6 @@ extension type V1(int id) {}
 extension type V2<T1, T2 extends num?>(T1 id) {}
```
eernstg, line 26: (Would comment on line 16, but Github won't do that.)

I'm afraid there's a bunch of `@assertion`s that need to be updated here, because the description in the feature specification was updated. 

But, as you mentioned somewhere, this PR will make changes such that the tests will work with the current as well as the upcoming semantics, so maybe that's just according to plan, and these updates will be performed in a later PR?

The corresponding change would be needed in the `@description` as well, but that could of course also be done later.

Please just keep a TODO about this somewhere if it is working as intended that we still have the old `@assertion`s and `@description`s.

eernstg (approved):
LGTM,

landing now, but asking for a TODO somewhere to make sure these tests are revisited in order to update the `@assertion` and `@description` sections later on.

sgrekhov:
https://github.com/dart-lang/co19/issues/2342 is that TODO

PR #2362, opened 2023-11-09
Fixes #2361. Fix flaky tests on web-platforms

### LINKED ISSUE #2361: Flaky tests on the web
The following tests are flaky on our web configurations:
* co19/LibTest/async/Future/Future.delayed_A01_t02 (example failure: https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8765049686744087633/+/u/test_results/ignored_flaky_test_failure_logs)
* co19/LibTest/async/Stream/Stream.periodic_A01_t01 (example failure:  https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8765056430326781265/+/u/test_results/ignored_flaky_test_failure_logs)
* co19/LibTest/async/Timer/Timer_A01_t01 (example failure: https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8765054869684785009/+/u/test_results/ignored_flaky_test_failure_logs)

Turns out this because of the same reason some of Dart's own language tests were failing. Timers can complete before the expected delay on browsers. Currently these tests fail on the first case (0ms), but note that this kind of early timers can happen with bigger delays too. We have seen timers sometimes complete as early as 15ms before the intended time.

To make these tests more reliable, we recommend adding a margin of error. See this change on our end as an example: https://dart-review.googlesource.com/c/sdk/+/333057/4/tests/lib_2/async/periodic_timer2_test.dart

sgrekhov:


eernstg (approved):
LGTM

LibTest/html/Element/onMouseOver_A01_t01.dart
```diff
@@ -4,21 +4,24 @@
...
   var type = 'mouseover';
   var x = document.body;
   if (x != null) {
     asyncStart();
     x.onMouseOver.listen((e) {
       Expect.equals(type, e.type);
-      asyncEnd();
+      if (!fired) {
+        asyncEnd();
+      }
```
sigmundch, line 23: I believe you also need to include `fired = true` here?

LibTest/async/Stream/Stream.periodic_A01_t01.dart
```diff
@@ -6,33 +6,40 @@
...
 /// @author kaigorodov
 
 import "dart:async";
 import "../../../Utils/expect.dart";
 
+// Most browsers can trigger timers too early. Test data shows instances where
+// timers fire even 15ms early. We add a safety margin to prevent flakiness
+// when running this test on affected platforms.
+Duration safetyMargin = const bool.fromEnvironment('dart.library.js')
+    ? Duration(milliseconds: 40)
```
sigmundch, line 21: Note that you may want to also adjust the intervals below for additional coverage. When `periodMs` is 0, 1, 10, and 30, we are no longer validating much about the duration, only that the event was fired.

The case with 100 does validate that at least 60ms have passed. I'm personally OK with that, but if you want additional coverage, we could include another check at 200.

sigmundch (commented):
Thanks @sgrekhov for the quick fixes! One small comment below.

sgrekhov:
@sigmundch thak you for good advices! They are implemented in #2364

PR #2363, opened 2023-11-09
#2350. Add more factory constructor tests. Part 3

sgrekhov:


eernstg (commented):
We should probably take the tests about default values out of this PR, because there is debate about whether that error should be implemented, or the specification should be changed. The tests where we do not have a subtype relationship among the function types of the constructors would be fine to land now.

The tests about the default values can then be uploaded as a different PR, and that one might be landed, or it might need to be updated if the specification gets modified such that there is no error any more.

eernstg (approved):
LGTM

PR #2367, opened 2023-11-13
#2350. Add/update factory constructor tests. Part 4

sgrekhov:


Language/Classes/Constructors/Factories/return_type_A03_t01.dart
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a dynamic type error if a factory returns a non-null object
```
eernstg, line 5: The updated version of this sentence (in [this PR](https://github.com/dart-lang/language/pull/2605)) does not include 'non-null'. The reason is that null never has a type which is a subtype of the actual return type (for all constructors except in the class `Null`, and we don't care exactly how that class is written because we couldn't write an equivalent one in regular Dart code anyway).

So we could have said "It is a dynamic type error if a factory returns null or a non-null object ...", and then we may just as well say "It is a dynamic type error if a factory returns an object ...".

I'm not quite sure how we can best handle this kind of situation (where the current spec differs from dart-lang/language#2605), but it is at least tempting to say that we should check the 2605 specification and see whether it is different from the corresponding part of the "current" specification, such that we don't create/update tests with obsolete (pre-nnbd) expectations. It could also be nice if we can avoid the need to update these `@assertion` paragraphs when 2605 is finally landed (and that's not going to be 2605 as-is, it's going to be several PRs where ).

So how could we proceed such that these things would work well for you?

Language/Classes/Constructors/Factories/return_type_A03_t02.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a dynamic type error if a factory returns a non-null object
+/// whose type is not a subtype of its actual return type.
+///
+/// @description Checks that it is a dynamic type error if a factory returns a
```
eernstg, line 8: The spec always says "the null object" because it's never up for debate which one we're talking about, there is only one null object:
```suggestion
/// @description Checks that it is a dynamic type error if a factory returns the
```

Language/Classes/Constructors/Factories/return_type_A03_t03.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a dynamic type error if a factory returns a non-null object
+/// whose type is not a subtype of its actual return type.
+///
+/// @description Checks that it is a dynamic type error if a factory returns a
```
eernstg, line 8:
```suggestion
/// @description Checks that it is a dynamic type error if a factory returns the
```

eernstg (commented):
Looks good! There is one test, Language/Classes/Constructors/Factories/function_type_t06.dart, where the description refers to a compile-time error which is not the one that the test actually causes.

Apart from that there are couple of typos.

Finally, as the most complex issue, we need to decide on a useful approach to the discrepancy between the current language specification (https://github.com/dart-lang/language/blob/main/specification/dartLangSpec.tex) which is describing a language where null safety has only been introduced in a few chapters, and a future language specification where null safety has been fully integrated (which is probably going to be quite similar to https://github.com/dart-lang/language/pull/2605).

Language/Classes/Constructors/Factories/redirecting_constructor_t01.dart
```diff
@@ -5,10 +5,13 @@
...
-/// redirectingFactoryConstructorSignature:
-///   const? factory identifier (‘.’ identifier)? formalParameterList
-/// ‘=’ typeName (‘.’ identifier)?
-/// ;
+/// ⟨redirectingFactoryConstructorSignature⟩ ::=
+///     const? factory ⟨constructorName⟩ ⟨formalParameterList⟩ ‘=’
+///     ⟨constructorDesignation⟩
+/// ⟨constructorDesignation⟩ ::= ⟨typeIdentifier⟩
+///     | ⟨qualifiedName⟩
+///     | ⟨typeName⟩ ⟨typeArguments⟩ (‘.’ ⟨identifier⟩)?
```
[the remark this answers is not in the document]
sgrekhov: I simply copied this text from `.pdf` version of the spec

Language/Classes/Constructors/Factories/accessible_class_A01_t03.dart
```diff
@@ -0,0 +1,50 @@
...
+/// where R is of one of the forms T<S1 . . . , Sp> or T<S1 . . . , Sp>.id.
+///
+/// It is a compile-time error if T does not denote a class accessible in the
+/// current scope. If T does denote such a class D, it is a compile-time error
+/// if R does not denote a constructor. Otherwise, it is a compile-time error if
+/// R denotes a generative constructor and D is abstract.
+///
+/// @description Checks that a compile error is produced if `T` does not denote
+/// a class accessible in the current scope.
+///
```
eernstg, line 23: Just guessing: These 3 lines should be removed?
```suggestion
```

Language/Classes/Constructors/Factories/parameter_type_t01.dart
```diff
@@ -0,0 +1,52 @@
...
+/// @assertion For the dynamic semantics, assume that k is a redirecting factory
+/// constructor and k′ is the redirectee of k
+///
+/// It is a dynamic type error if an actual argument passed in an invocation of
+/// k is not a subtype of the actual type of the corresponding formal parameter
+/// in the declaration of k.
+///
+/// @description Checks that it is a dynamic type error if an actual argument
+/// type of a redirecting factory constructor call is not a subtype of the
+/// actual type of its corresponding formal parameter.
```
eernstg, line 14: To help future readers, perhaps add a few words stating that this situation can only occur when the static type of the given actual argument is `dynamic`.

Also note that we should check that it is the forwarding constructor parameter declaration that counts: We could have a parameter of type `num` in the forwardee and `int` in the forwarder, and then we should get the dynamic error when we're passing an actual argument whose run-time type is `double` (or some subtype of `double`), even though the constructor which is actually invoked would be perfectly happy about that argument if called directly.

Language/Classes/Constructors/Factories/redirecting_constructor_t01.dart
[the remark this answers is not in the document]
eernstg: ;-)

Language/Classes/Constructors/Factories/return_type_A03_t01.dart
```diff
@@ -0,0 +1,54 @@
...
+import "../../../../Utils/expect.dart";
+
+class A {}
+
+class C extends A {
+  factory C() => "42" as dynamic;
+
+  factory C.f1() => A() as dynamic;
+
+  factory C.f2() => A() as C;
```
eernstg, line 21: You could say that this one doesn't test the situation from the `@description`: We never reach the point where the act of returning occurs, we're just testing that the type cast fails (which would happen anywhere, including this situation which is "one step before returning").

Perhaps just delete this case (plus the one in line 32)?

eernstg (commented):
Looks good! Just a couple of typos and a suggested addition to a comment.

Language/Classes/Constructors/Factories/function_type_t06.dart
```diff
@@ -0,0 +1,48 @@
...
+///     N(T1 x1 . . . , Tn xn, [Tn+1 xn+1=d1, . . . , Tn+k xn+k=dk]) = R;
+/// const? factory
+///     N(T1 x1 . . . , Tn xn, {Tn+1 xn+1=d1, . . . , Tn+k xn+k=dk}) = R;
+/// where R is of one of the forms T<S1 . . . , Sp> or T<S1 . . . , Sp>.id.
+/// ...
+/// The redirectee constructor for this declaration is the constructor k′
+/// denoted by R.
+/// ...
+/// It is a compile error if the function type of k' is not a subtype of the
+/// type of k.
```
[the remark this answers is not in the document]
eernstg: Thanks!

Language/Classes/Constructors/Factories/parameter_type_t01.dart
sgrekhov: Makes sense. Added `parameter_type_t01.dart`

Language/Classes/Constructors/Factories/parameter_type_t01.dart
```diff
@@ -0,0 +1,53 @@
...
+/// constructor and k′ is the redirectee of k
+///
+/// It is a dynamic type error if an actual argument passed in an invocation of
+/// k is not a subtype of the actual type of the corresponding formal parameter
+/// in the declaration of k.
+///
+/// @description Checks that it is a dynamic type error if an actual argument
+/// type of a redirecting factory constructor call is not a subtype of the
+/// actual type of its corresponding formal parameter (that this situation can
+/// only occur when the static type of the given actual argument is `dynamic`).
```
eernstg, line 15: Oh, this is actually not precisely the situation we have below. We'd need to have different parameter types in the redirecting factory constructor and in the redirectee, because the whole point is that we satisfy one of the constraints and violate the other one. Something like this:

```dart
class A {
  factory A(int i) = B;
}

class B implements A {
  B(num n);
}

void main() {
  print('Call B');
  B(1.5 as dynamic); // OK.
  print('Tear off B then call');
  (B.new)(1.5 as dynamic); // OK.
  print('Call A');
  A(1.5 as dynamic); // Throws.
  print('Tear off A then call');
  (A.new)(1.5 as dynamic); // Throws.
}
```

This shows that the formal parameter types of the constructor declared by `A` are taken into account (that is, they aren't just ignored even though the invocation of `A(arg)` is very much the same thing as an invocation of `B(arg)`).

As far as I can see, we never have this distinction between the formal parameter types of the redirecting factory construct and the redirectee in this test.

Language/Classes/Constructors/Factories/parameter_type_t01.dart
eernstg: Thanks! I commented there.

eernstg (commented):
Thanks! The new test needs to make a distinction between the formal parameter types of the redirecting factory constructor and the redirectee, I've commented on that.

(Alternatively, I just haven't understood what 'parameter_type_t01.dart' is doing ;-).

sgrekhov:
`parameter_type_t01` updated, `parameter_type_t02` deleted because the first test does all the work now. Please review

eernstg (approved):
LGTM!

PR #2370, opened 2023-11-14
Fixes #2369. Static type of == operator is always bool

### LINKED ISSUE #2369: Wrong static type on ==
The test `co19/LanguageFeatures/Extension-types/static_analysis_member_invocation_A01_t05` assumes that the static type of `==` can be derived from the return type of the declared `operator ==` method:

```
class B implements A {
  ...
  @override
  BoolET operator ==(Object? other) => BoolET(other == this);
}
...
extension type ET2(B b) implements ET1, B {}

extension type BoolET(bool b) implements bool {}

void main() {
  var e2 = ET2(B());
  ...
  BoolET b2 = e2 == e2;
}
```
Here `e2.==` can be found to have type `BoolET Function(Object?)` as declared in `B`.

If the latest spec is to be believed, the static type of an equality expression is always defined to be `bool`: (§17.27: "The static type of an equality expression is bool.")

cc @eernstg

sgrekhov:


eernstg (approved):
LGTM

Issue #2371, opened 2023-11-14
Add tests for factory constructor generating an enum value


sgrekhov:
After landing of https://github.com/dart-lang/language/pull/3461 add tests like the ones mentioned in https://github.com/dart-lang/language/issues/3460#issue-1987292756

PR #2374, opened 2023-11-15
#2371. Add missing enhanced enums tests, update assertions

sgrekhov:
Tests like ones, mentioned in https://github.com/dart-lang/language/issues/3460 are not added  because https://github.com/dart-lang/language/pull/3461 is not landed yet. They will be added after the landing


eernstg (approved):
LGTM

PR #2376, opened 2023-11-16
#2371. Add tests for factory constructor generating an enum value

sgrekhov:


eernstg (approved):
LGTM!

It's possible that LanguageFeatures/Enhanced-Enum/factory_constructor_A01_t02.dart will cause another error message, because the desugared enum declaration will contain a declaration of a `static const` that has a cyclic dependency. Anyway, that can be handled if and when it occurs.

PR #2385, opened 2023-11-20
#2342. Update nullability tests according to the new rules

sgrekhov:


LanguageFeatures/Extension-types/nullability_A01_t01.dart
```diff
@@ -2,13 +2,11 @@
...
-/// be the corresponding instantiated representation type. If R is non-nullable
-/// then V is a proper subtype of Object, and V is non-nullable. Otherwise, V is
-/// a proper subtype of Object?, and V is potentially nullable.
+/// @assertion An extension type `V` is a proper subtype of `Object?`. It is
+/// potentially non-nullable, unless it implements `Object` or a subtype thereof
 ///
-/// @description Checks that an extension type is not-nullable even if its
-/// representation type is nullable
+/// @description Checks that an extension type is not-nullable even if doesn't
+/// implement subtype of `Object` and its representation type is nullable
```
eernstg, line 9: `extension type E(R it) implements Object {}` is always an error in the case where `R` is nullable, so we shouldn't actually need the added words.

Maybe express the same thing as 'commentary', if that concept exists here? Something like this could be added at the end of this paragraph: 'In this case we know that the extension type doesn't implement `Object` or a subtype thereof, because a nullable representation type would then be a compile-time error'.

LanguageFeatures/Extension-types/nullability_A01_t02.dart
```diff
@@ -2,10 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-/// @assertion Let V be an extension type of the form Name<T1, .. Ts>, and let R
-/// be the corresponding instantiated representation type. If R is non-nullable
-/// then V is a proper subtype of Object, and V is non-nullable. Otherwise, V is
-/// a proper subtype of Object?, and V is potentially nullable.
+/// @assertion An extension type `V` is a proper subtype of `Object?`. It is
+/// potentially non-nullable, unless it implements `Object` or a subtype thereof
```
eernstg, line 6: Leaf mentioned that this wording isn't precise. It should have said "It is potentially non-nullable.If it implements `Object` or a subtype thereof then it is also non-nullable."

This will be fixed soon.

eernstg: I'm not sure which approach is best when it comes to adjusting the '@assertion' (because the feature specification hasn't been fixed yet). WDYT?

LanguageFeatures/Extension-types/static_analysis_extension_types_A17_t03.dart
```diff
@@ -2,31 +2,28 @@
...
-/// ...
-/// Let V be an extension type of the form Name<T1, .. Ts>, and let R be the
-/// corresponding instantiated representation type. If R is non-nullable then V
-/// is a proper subtype of Object, and V is non-nullable. Otherwise, V is a
-/// proper subtype of Object?, and V is potentially nullable.
-///
-/// @description Checks that if an instantiated representation type `R` is
-/// not non-nullable then it is a proper subtype of `Object?` (can be assigned
-/// to `Object?`)
+/// @description Checks that if an extension type doesn't implement subtype of
```
eernstg, line 8: Perhaps:
```suggestion
/// @description Checks that even if an extension type doesn't implement a subtype of
```
The word 'even' is there to indicate that the following isn't implied: "If an extension type `E` does implement a subtype of `Object` then `E` isn't a subtype of `Object?`".

eernstg: There are a couple of other occurrences of 'implement subtype'.

eernstg (commented):
Looks good! There are a couple of typos.

A more pervasive issue is that an `@assertion` includes a sentence from the feature specification which is misleading (it will be fixed, but hasn't been fixed at this time).

I'm not sure what's the best way to proceed with that, maybe just leave it as is and create an issue about updating it when the spec is updated?

LanguageFeatures/Extension-types/nullability_A01_t01.dart
sgrekhov: Added `nullability_A01_t04.dart` testing that it is a compile-time error if an extension type does implement subtype of `Object` and its representation type is nullable.

LanguageFeatures/Extension-types/nullability_A01_t02.dart
sgrekhov: Let's update it later when the spec will be actually updated.

sgrekhov (commented):

LanguageFeatures/Extension-types/nullability_A01_t02.dart
eernstg: OK.

LanguageFeatures/Extension-types/static_analysis_extension_types_A17_t04.dart
```diff
@@ -2,33 +2,27 @@
...
-/// corresponding instantiated representation type. If R is non-nullable then V
-/// is a proper subtype of Object, and V is non-nullable. Otherwise, V is a
-/// proper subtype of Object?, and V is potentially nullable.
-///
-/// @description Checks that if an instantiated representation type `R` is
-/// not non-nullable then an extension type `V` is potentially nullable (it's a
-/// compile-time error to assign `Object?` or `null` to `V`)
+/// @description Checks that if an extension type doesn't implement a subtype of
+/// `Object` then is potentially non-nullable (it's a compile-time error to
+/// assign `Object?` or `null` to it)
```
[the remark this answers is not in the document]
sgrekhov: The idea here is to test that an extension type is a proper subtype of `Object?`, i.e. it can be assigned to `Object?` but `Object?` cannot be assigned to it. And this test checks the latter. Do you believe that we should check here assigning of the `null` only?

LanguageFeatures/Extension-types/static_analysis_extension_types_A17_t04.dart
```diff
@@ -2,33 +2,27 @@
...
+extension type V2<T>(T id) {}
+
+extension type V3(int it) {}
+
+extension type V4<T extends Object>(T id) {}
 
 main() {
   V1 v1_1 = V1(42) as Object?;
-//          ^^^^^^^^^^^^^^^^^
+//                 ^
```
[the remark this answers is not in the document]
sgrekhov: It's just according the current behaviour of the tools. They report an error there
```
static_analysis_extension_types_A17_t04.dart:24:20: Error: A value of type 'Object?' can't be assigned to a variable of type 'V1'.
 - 'Object' is from 'dart:core'.
  V1 v1_1 = V1(42) as Object?;
                   ^
```

LanguageFeatures/Extension-types/nullability_A01_t01.dart
eernstg: OK, thanks!

LanguageFeatures/Extension-types/static_analysis_extension_types_A17_t03.dart
```diff
@@ -5,9 +5,9 @@
 /// @assertion An extension type `V` is a proper subtype of `Object?`. It is
 /// potentially non-nullable, unless it implements `Object` or a subtype thereof
 ///
-/// @description Checks that if an extension type doesn't implement subtype of
-/// `Object` then it is a proper subtype of `Object?` (can be assigned to
-/// `Object?`).
+/// @description Checks that even if an extension type doesn't implement a
+/// subtype of `Object` then it is a proper subtype of `Object?` (can be
+/// assigned to `Object?`).
```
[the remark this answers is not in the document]
eernstg: (I actually don't think any of the other tests mention 'proper subtype', at least in the current version of co19 in the SDK, 460758bf4c298812d47b24f7e5040c75088c3ebf.)

LanguageFeatures/Extension-types/static_analysis_extension_types_A17_t04.dart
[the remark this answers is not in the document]
eernstg: Looking again, it seems likely that the test where 'proper subtype' should actually be tested is LanguageFeatures/Extension-types/static_analysis_extension_types_A17_t03.dart.

So if the tests about assigning `Object?` to an extension typed variable are added over there and deleted from here then I think we'll have a good alignment between the contents of the tests and the descriptions of the tests.

Of course there could be other ways to reorganize that would create a similarly precise alignment, but I do think it would be great if something is done about the fact that t03 says "proper subtype" but doesn't test more than half of it, and this test doesn't say "proper subtype", but it _does_ test the halfpart that was missing from t03. ;-)

sgrekhov:
Moved `  V1 v1_1 = V1(42) as Object?;` tests from `static_analysis_extension_types_A17_t03.dart` to `static_analysis_extension_types_A17_t04.dart`

eernstg (approved):
LGTM

PR #2394, opened 2023-11-23
Merge pre-nnbd into master before deletion

athomas:


athomas:
Outdated fork, will create a new PR.

Issue #2388, opened 2023-11-20
[Extension types] Update tests according to the new setter/method rules


sgrekhov:
Update Extension types tests according to the new rules regarding setter/method conflict. The [change](https://github.com/dart-lang/language/commit/299d20be62cd23fc396d6fa812d6e7d92386ab09)

PR #2397, opened 2023-11-24
#2388. Rename and reorder static_analysis_member_invocation_A06_t* tests

sgrekhov:
Github shows a big diff but in fact it's only renaming and reordering of the tests


eernstg (commented):
LGTM.

Github actually only recognizes one of these operations: Renaming 08 to 04. For instance, it says that 07 was deleted rather than renamed to 06, and it shows a bunch of differences for all the other libraries.

Do you use `git mv` to rename libraries? (.. if not then that could explain why github really doesn't understand the renaming operations). Otherwise it would be very nice to have just a small hint about the renaming operations in the PR description (perhaps something like 'Renaming: 08 -> 04, 07 -> 06'), also in order to help future readers who might want to understand the history of one of these libraries.

eernstg (approved):

sgrekhov:
Ok, will do the next time. I rename files in Intelj IDEA and then do a commit from there, not from a command-line. Sometimes this tool is smart enough to recognise renaming, sometimes not

eernstg:
OK, thanks!

PR #2407, opened 2023-12-01
#2398. Update async tests to avoid false-positive results on web. Language and LanguageFeatures tests

sgrekhov:


Language/Expressions/Await_Expressions/syntax_t03.dart
```diff
@@ -9,9 +9,7 @@
 /// @description Check that it is a compile error if unaryExpression is missing
 /// @author a.semenov@unipro.ru
 
-
 import 'dart:async';
-import '../../../Utils/expect.dart';
 
 f() {
```
mkustermann, line 14: Unused function? (also a few other tests like it)

mkustermann (approved):

Language/Expressions/Await_Expressions/syntax_t03.dart
```diff
@@ -25,6 +23,5 @@ Future test() async {
 }
 
 main() {
-  asyncStart();
-  test().then( (value) => asyncEnd() );
+  test();
```
osa1, line 26: I understand that this test is not run (it has a compile-time error), but wasn't the previous version of `main` more correct?

If we want the simplest possible test that checks the compile-time error then the entire body could be omitted: `main () {}`.

Language/Expressions/Await_Expressions/syntax_t04.dart
```diff
@@ -25,6 +23,5 @@ Stream test() async* {
 }
 
 main() {
-  asyncStart();
-  test().isEmpty.then( (value) => asyncEnd() );
+  test();
```
osa1, line 26: Same as above.

osa1 (approved):
Added a question. Rest LGTM.

Language/Expressions/Await_Expressions/syntax_t03.dart
eernstg: We do have a rule of thumb about `main` in a test with compile-time errors: A language processor (analyzer, compiler, whatever) is allowed to omit parts of the program that are known to have no effect. For instance, it could skip all the processing (including parsing) of the body of a function if that function cannot be called by the given program. This means that a language processor could respond with "OK!" when given a program whose `main` is `void main() {}`, and then we may get missing compile-time errors in all kinds of unreachable declarations.

So in order to allow language processors to perform this kind of optimization, we'd aim to include a minimal amount of code that creates the necessary dependencies. For instance, `print(f)` is going to imply that the function `f` is used, and hence the language processor can't skip it.

We never made this analysis precise, but even in its imprecise form I think it makes sense to preserve the ability to have tools that are skipping unreachable parts of the program.

osa1: Ah, got it. Thanks for the explanation.

Language/Expressions/Await_Expressions/syntax_t03.dart
sgrekhov: Yes. Thank you. Unused code removed

Issue #2004, opened 2023-04-05
[ffi] CFE/analyzer error messages in different locations.


dcharkes:
Some of the diagnostics in Dart FFI get reported in different places for CFE vs analyzer. (On a field/class or on the annotation on that field or class.)

- co19/LibTest/ffi/Array/Array.multi_A01_t02
- co19/LibTest/ffi/Array/Array.multi_A02_t01
- co19/LibTest/ffi/Array/Array.multi_A02_t02
- co19/LibTest/ffi/Array/Array_A02_t01
- co19/LibTest/ffi/Array/Array_A02_t02

Some tests rely on the fix in https://dart-review.googlesource.com/c/sdk/+/293600 which has landed today. (So you need the last build of the VM to get the error messages.)

From the triage in:

* https://github.com/dart-lang/sdk/issues/51958

PR #2411, opened 2023-12-04
#2004. Add deferred libraries tests according to the changed spec

sgrekhov:
Rename again  is shown as delete and add a new file (was shown as a rename on commit).
```
Language/Libraries_and_Scripts/Imports/deferred_import_t01.dart => Language/Libraries_and_Scripts/Imports/semantics_A02_t01.dart
Language/Libraries_and_Scripts/Imports/deferred_import_t02.dart => Language/Libraries_and_Scripts/Imports/semantics_A01_t02.dart
```


Language/Libraries_and_Scripts/Imports/semantics_A02_t03.dart
```diff
@@ -0,0 +1,35 @@
...
+///   evaluated. Closurizing the function also results in a dynamic error.
+/// • For every top level getter g named id in NSimport,i, a corresponding
+///   getter named id with the same signature as g. Calling the getter results
+///   in a dynamic error.
+/// • For every top level setter s named id= in NSimport,i, a corresponding
+///   setter named id= with the same signature as s. Calling the setter results
+///   in a dynamic error that occurs before the actual argument is evaluated.
+///
+/// @description Checks that closurizing a function from a deferred library
+/// before `loadLibrary()` call results in a dynamic error.
```
eernstg, line 25: Perhaps something like the following added words would be helpful for future readers:
```suggestion
/// before the future returned by `loadLibrary()` has been completed
/// results in a dynamic error.
```

Language/Libraries_and_Scripts/Imports/semantics_A03_t01.dart
```diff
@@ -0,0 +1,39 @@
...
+/// NSdeferred then has the following bindings:
+/// ...
+/// • For every class, mixin, enum, and type alias declaration named id in
+///   NSimport,i, a corresponding getter named id with return type Type. Calling
+///   the getter results in a dynamic error.
+///
+/// @description Checks that for every class, mixin, enum, and type alias
+/// declaration in the imported library a corresponding getter with the same
+/// name and return type `Type` is in the current library. Call of this getter
+/// is a dynamic error
```
eernstg, line 20: I think it's a little bit confusing to say that the current library somehow contains those deferred imported entities (both the throwing ones before the `loadLibrary` future has completed, and the real ones that we have after that completion).

The specification says that the entities can be obtained by a look up in a _namespace_ which is associated with the deferred import prefix. This is exactly the same mechanism as the one which is used for an immediate import (they don't switch from "throwing" to "real" entities, but they do rely on a namespace to give access to the imported things).

So the deferred-imported entities are actually no more a part of the current library than all the immediately-imported entities, and we probably don't want to talk about the immediately-imported ones as being part of the current library.

So maybe:

```suggestion
/// @description Checks that for every class, mixin, enum, and type alias
/// declaration in the imported library a corresponding getter with the same
/// name and return type `Type` is present in the namespace denoted by the
/// deferred import prefix. An invocation of this getter is a dynamic error.
```

eernstg (commented):
Looks good! There are just a couple of locations where the description of the situation is slightly confusing.

sgrekhov:
Thank you! Updated

eernstg (approved):
LGTM

PR #2417, opened 2023-12-07
#2119. Run dart formatter on LibTest/async tests

sgrekhov:


LibTest/async/Future/Future.sync_A01_t03.dart
```diff
@@ -19,25 +19,20 @@ import "dart:async";
...
+      Expect.identical(value, e);
     }
-  );
+    asyncEnd();
+  });
 }
 
 main() {
-  asyncStart();
+  asyncStart(9);
```
eernstg, line 35: Yes, that is actually better.

Issue #2415, opened 2023-12-06
Fix `Stream` and `StreamController` test failures


sgrekhov:
`LibTests/async/Stream` and `LibTest/async/StreamController` fails at run time on numerous platforms. Need to check these failures
See https://dart-current-results.web.app/#/filter=co19/LibTest/async. Most probably tests needs to be fixed

PR #2421, opened 2023-12-08
#2415. Change expectations for Stream.asyncMap() according to the current behavior

sgrekhov:
It was decided to change documentation according to the current behaviour. So, `Stream.map()` calls `convert` function once per data event per listener, but  `Stream.asyncMap()` calls it once per data event. It will be documented and the test changed accordingly


eernstg (approved):
LGTM

eernstg (commented):
I LGTM'd this PR just before it was landed. Apparently, I forgot to click the 'Approve' bullet, and (apparently) it cannot be done at this time.

In any case, there is no "missing LGTM" issue in substance, even though it looks like that.

eernstg:
Funny, now we suddenly have the check mark as if the approval has been added.

Issue #2420, opened 2023-12-07
Exhaustiveness uses extension type erasure


eernstg:
@sgrekhov, do we have coverage for the following kind of situation?

```dart
sealed class S {}
class A extends S {}
class B extends S {}

extension type E(B it) {}

void f(S s) {
  var x = switch (s) {
    A() => 1,
    E() => 2,
  };
}
```

In a situation where we wish to determine whether or not a given switch (expression or statement) is exhaustive, we should use the extension type erasure of every extension type that we encounter (as an object pattern in a case, as the type of the matched value, in a cast pattern, whatever).

It is important that we consider the extension type erasure and not just the declared superinterface relationships of the extension types, because the actual exhaustiveness property is determined by the run-time types of the objects that we're working with, and that is again approximated at compile time by the extension type erasure (and not in a useful way by the extension type itself).

I don't remember seeing any tests of this nature, but I might have forgotten. It would be great if you could take a look, and put it on the TODO list if we don't have anything.

sgrekhov:
No, we don't. I'll add them, thank you!

PR #2422, opened 2023-12-11
#2420. Add extension types exhaustiveness tests. Enums

sgrekhov:


LanguageFeatures/Extension-types/exhaustiveness_enum_A01_t01.dart
```diff
@@ -0,0 +1,64 @@
...
+}
+
+extension type ET1<T>(E<T> _) {}
+extension type ET2<T>(E<T> _) implements E<T> {}
+
+String testStatement1(ET1<num> e) {
+  switch (e) {
+    case E.a:
+    case E.c:
+      return "ok";
```
eernstg, line 30: This seems to work, but it would be somewhat esoteric (and this has nothing to do with extension types, the situation exists after erasure).

The reason why this is a very special exception is that we (presumably) rely on the fact that this is a switch over enumerated values. This means that we can iterate over all the elements, and it is _guaranteed_ that we won't encounter any other objects than the ones that are declared in the `enum` declaration itself, and they have a statically known run-time type.

Lots and lots of stuff to know!

OK, I was worried that this would be on the very edge of the capabilities of the exhaustiveness analysis, but it seems to work fine.

In any case, we should probably also have a simpler test where the enumerated type isn't generic, such that it is easy to see that we have an extension type on top of a very straightforward setup (which will be a test that is only about being an extension type, so it's very safe to assume that it will test just that).

eernstg (commented):
LGTM!

I think we should have a set of similar tests where the underlying exhaustiveness of a non-extension type is trivial, because it will be very easy to see with those tests that they are testing exactly the support for a matched value type which is an extension type.

The current tests are also including a tricky case involving enumerated values which are instances of a generic enum, so they are much easier to mess up (say, if they need to be updated somehow in the future).

However, that could just as well be in a different PR, so I'll land this one.

eernstg (approved):

PR #2423, opened 2023-12-12
#2420. Add extension types exhaustiveness tests. Enums, trivial cases

sgrekhov:


LanguageFeatures/Extension-types/exhaustiveness_enum_A01_t04.dart
```diff
@@ -0,0 +1,72 @@
...
+  a, b, c
+}
+
+extension type ET1(E _) {}
+extension type ET2(E _) implements E {}
+
+String testStatement1(ET1 e) {
+  switch (e) {
+    case E.a:
+    case E.a:
```
eernstg, line 27: Presumably this would cause an "unreachable case" warning or something like that. I guess that doesn't cause any trouble with any test run? What's the purpose of having two copies of `E.a`?

eernstg (commented):
Looks good!

I asked one question about LanguageFeatures/Extension-types/exhaustiveness_enum_A01_t04.dart, perhaps it's just a typo.

LanguageFeatures/Extension-types/exhaustiveness_enum_A01_t04.dart
sgrekhov: It was made by purpose. What if this warning will affect exhaustiveness somehow? Let's be sure that not

eernstg:
This is good, I do think it's useful to have some tests which are aimed very directly at the situation which is the topic of the test, "the matched value type is an extension type", and keeping everything else simple. We need complicated cases, too, but if we have the simple ones and they are working then we know that we may need to look at different features (or look at feature interactions) if something complex isn't working. If the simple thing doesn't work then we definitely need to look at the feature in focus (here: a matched value type is an extension type).

Another thing came to mind as well: Do we have exhaustiveness tests using extension types as object patterns? Such tests could also test the use of extension types as object patterns in general (so they should be able to have field patterns, like `case ET(:final someMember?)`). So if we don't have it we should probably have two TODOs.

sgrekhov:
> Do we have exhaustiveness tests using extension types as object patterns? 
Yes we do. I'll add these tests for extension types later in another PR. Right now I'm adding list pattern tests for extension types. Will do a PR soon

LanguageFeatures/Extension-types/exhaustiveness_enum_A01_t04.dart
eernstg: We could also simply remove the duplicate case: We're testing that a matched value type for the purpose of exhaustiveness is treated as its erasure, we aren't testing that switch statements have correct analysis of 'unreachable case' for a bunch of enum values (because we would have that test already, somewhere else).

eernstg:
> PR soon

Sounds good, thanks!

LanguageFeatures/Extension-types/exhaustiveness_enum_A01_t04.dart
sgrekhov: Removed. Please review

eernstg (approved):
LGTM!

PR #2426, opened 2023-12-13
#2420. Add extension types exhaustiveness tests. Maps

sgrekhov:


LanguageFeatures/Extension-types/exhaustiveness_map_A01_t01.dart
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Switch statements and expressions with map as a matched type can
+/// never be exhaustive
```
eernstg, line 6: I couldn't find this claim in the feature specification, and it wouldn't be true, anyway: The matched value type can be of the form `Map<K, V>` for some types `K` and `V`, and we can easily have this and be exhaustive:

```dart
void main() {
  var x = switch (<int, String>{...}) {
    _ => 1,
  };
}
```

If there is a non-extension-type version of this test then a similar assertion/description would presumably need an extra look.

The `@description` is quite different, it talks about a map pattern (the matched value type is on one side of the fence, each pattern is on the other side). So the question is: Can we write a map pattern that successfully matches an arbitrary matched object?

I think you're right that there is no single pattern which will match all maps of any given type: We can't use an ellipsis as a "rest pattern" or anything like that, and we cannot match an empty map using a map pattern.

So what's the conclusion? Probably something like "exhaustiveness with map patterns can only be achieved when there is an exhaustive pattern in addition to any map patterns", like:

```dart
void main() {
  var x = switch (<int, String>{...}) {
    {1: ''} => 1,
    Map() => 2,
  };
}
```

Alternatively, we could try some set of switches whose cases are all map patterns, and see that the result isn't exhaustive. However, that's a very long shot, because we'd need to argue that "we've tried all possibilities!" Otherwise there isn't much that we can conclude based on the result "this one isn't exhaustive; that one neither; or that one; ...".

LanguageFeatures/Extension-types/exhaustiveness_map_A01_t01.dart
```diff
@@ -0,0 +1,78 @@
...
+/// extension type
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+extension type ET1<K, V>(Map<K, V> _) {}
+extension type ET2<K, V>(Map<K, V> _) implements Map<K, V> {}
+
+String test1_1(ET1<bool, bool> m) =>
+  switch (m) { // Empty map is not matched
```
eernstg, line 18: We're also not matching a map with two key/value pairs. There would be 4 of those.

But it doesn't matter much whether they are added or not, because we can't even express the empty map as a map pattern, so we cannot exhaust the conceptual/intuitive set of maps of type `Map<bool, bool>` using map patterns.

However, in this context it could be useful to add an extra test: Start with the same patterns, just so we have something for the exhaustiveness algorithm to work with, and then end with an exhaustive pattern like `Map()`.

This will give us exhaustiveness for the switch expression _if_ the matched value type is treated as its representation type. (If it is treated as the extension type `ET1` then there is no reason to think that `Map()` will exhaust the matched value type). So that would be a quite useful test.

LanguageFeatures/Extension-types/exhaustiveness_map_A01_t01.dart
```diff
@@ -0,0 +1,78 @@
...
+      return "case-1";
+    case {true: false}:
+      return "case-2";
+    case {false: true}:
+      return "case-3";
+    case {false: false}:
+      return "case-4";
+  }
+// There is no return statement here, switch statement is not exhaustive, so an
+// error above occurs because function return type cannot be null
```
eernstg, line 54: Thanks!

eernstg (commented):
There is some amount of confusion in the `@assertion` and `@description`, that should probably be revised (perhaps some older tests have the same?).

In addition to this, I think it would be useful to take the opportunity to test that a switch expression _is_ exhaustive when it exhausts all maps of the given type. This shows that an extension type whose representation type is `Map<K, V>` for some `K` and `V` is actually treated like its representation type during exhaustiveness computations.

sgrekhov:
> However, in this context it could be useful to add an extra test: Start with the same patterns, just so we have something for the exhaustiveness algorithm to work with, and then end with an exhaustive pattern like Map().

A very good idea, thank you! Added this tests and updated descriptions/assertions (including original patterns tests). Please take another look

eernstg (approved):
LGTM!

The whole notion of showing that a particular kind of switch expression 'cannot be exhaustive' is questionable: We're promising that we have tried _all possible ways_ to make it exhaustive, and there is no way that works.

The reality is that any attempt to do this would potentially fail because we've forgotten a particularly smart thing that we might have done. So we've found a bunch of switch expressions that aren't exhaustive (not a hard task ;-), but it would then still be false that this kind of pattern 'cannot be exhaustive'.

We do have delicate properties like this (for example: soundness&mdash;there is no program that will make a variable of type `T` have a value of type `S` unless `S <: T`"), but they are really difficult to get right in an absolute sense, so we just keep trying to improve on them and eliminate failures (in this case: failures to maintain soundness).

I don't think there's a need to play around with the wording any more, but it's probably a good idea to keep in mind that this kind of quantification over all possible something (all programs, all expressions, whatever) is delicate to get right, and we may just choose to avoid promising anything of that kind, unless we specifically need to do that.

PR #2427, opened 2023-12-13
#2350. Add more factory constructors tests

sgrekhov:


Language/Classes/Constructors/Factories/syntax_t06.dart
```diff
@@ -0,0 +1,49 @@
...
+}
+
+enum E {
+  e1(1), e2(2);
+
+  final x;
+
+  const E(this.x);
+//             ^
+// [analyzer] unspecified
```
eernstg, line 36: Why would we expect an error here?

eernstg: I just tried to run this program in DartPad, and I can see the error message. However, it looks like a spurious error message (it says that `x` is initialized at the declaration, but that isn't true), so maybe the analyzer should be adjusted such that it does not emit that error message.

I created https://github.com/dart-lang/sdk/issues/54335 to suggest this change to the analyzer.

I think we should then land this test as it is, and adjust it at some point in the future when the error messages changes, if they do indeed make those changes to the analyzer.

eernstg (approved):
LGTM.

I added a comment, but in the end we should just revisit the relevant test in the future if needed.

PR #2429, opened 2023-12-14
#2139. Fix wrong failure of Language/Functions/element_type_A02_t06

sgrekhov:


Language/Functions/element_type_A02_t06.dart
```diff
@@ -39,6 +39,6 @@ main() async {
   var it1 = await res1?.toList();
   Expect.isRuntimeTypeImplementsIterable<int?>(it1);
 
-  var it2 = (await f2())?.toList();
+  var it2 = await (await f2())?.toList();
```
eernstg, line 42: OK: The inner `await` is fake in the sense that _we_ know there won't be a future there, but the compiler is specified to assume that there could be a future, so we use `await` to obtain a known guarantee that this non-existent future would be awaited. Now we have a `Stream<int>?`.

Very good, now I understand why we need two `await`s here. ;-)

eernstg (approved):
LGTM

PR #2431, opened 2023-12-14
#2420. Add extension types exhaustiveness tests. Variables

sgrekhov:


LanguageFeatures/Extension-types/exhaustiveness_variable_A01_t01.dart
```diff
@@ -0,0 +1,45 @@
...
+
+extension type AET1(A _) {}
+extension type AET2(A _) implements A {}
+extension type BET1(B _) {}
+extension type BET2(B _) implements B {}
+
+test1_1(AET1 a) => switch (a) { B _ => 'B' };
+test1_2(AET2 a) => switch (a) { B _ => 'B' };
+
+test2_1(BET1 b) => switch (b) { B _ => 'B', C _ => 'C' };
```
eernstg, line 32: In this one, the second case would be unreachable, but I guess that doesn't disrupt any test runs (that's just a warning)?

Also, I don't think we can say that this test has anything to do with sealed classes: `B` and `C` are not sealed, and the fact that `A` is sealed doesn't imply any extra guarantees about a case that tests for `B` or `C`. So we're essentially just testing that `switch (e1) { T _ => e2 }` is exhaustive when `T` is the static type of `e1`, with the added twist that one of those types is an extension type and the other one is the corresponding representation type.

However, it does make sense to have this very basic test where we're confirming that a switch expression is exhaustive if it has a single case testing the representation type where the matched value type is the extension type, and vice versa.

This would probably be a new library doing just that, and then `test2_1` could be moved into that new test (and perhaps `C` ends up being unused in this test?).

eernstg (commented):
Looks good! Just one thing: LanguageFeatures/Extension-types/exhaustiveness_variable_A01_t01.dart says that it's about sealed classes, but it contains a test which isn't. However, it's a good idea to have that test anyway, that would just become one more library with a couple of cases of that nature.

sgrekhov:
`test2_1` moved to a separate basic test

eernstg (approved):
LGTM

PR #2438, opened 2023-12-18
Fixes #2428. Add `call` member tests for extension types

### LINKED ISSUE #2428: Implicit `call` method tear-offs with extension typed receivers
The proposal in https://github.com/dart-lang/language/issues/3467 was accepted by the language team. This means that it should be possible to perform `.call` method tear-offs also in the case where the receiver is an expression whose static type is an extension type. In this case it should be possible to tear off both a `call` method which is a non-extension type member (that is, inherited from some "normal" type), and also a `call` method which is an extension type member (from the receiver type or "inherited" from some other extension type).

@sgrekhov, I think no such tests have been written, isn't that true?

sgrekhov:


LanguageFeatures/Extension-types/dynamic_semantics_member_invocation_A03_t07.dart
```diff
@@ -0,0 +1,54 @@
...
+/// argument part (possibly including some actual type arguments) then call it
+/// args. If the invocation omits args, but includes a list of actual type
+/// arguments then call them typeArgs. Assume that V declares the type variables
+/// X1, ..., Xs
+/// ...
+/// Let Dm be the unique declaration named m that V has.
+///
+/// Evaluation of this invocation proceeds by evaluating e to an object o.
+/// ...
+/// Otherwise, the following is known: args is included, and Dm is a method. The
```
eernstg, line 17: This paragraph describes method invocations, not property extractions (also known as tear-offs).

I think this paragraph is more relevant:

> Otherwise, if `args` is omitted and _Dm_ is a method, ...

LanguageFeatures/Extension-types/dynamic_semantics_member_invocation_A03_t07.dart
```diff
@@ -0,0 +1,54 @@
...
+
+extension type ET1(Function it) {
+  String call() => "call from ET1";
+}
+
+extension type ET2(Function it) implements Object {
+  String call() => "call from ET2";
+}
+
+Function get functionGetter1 => ET1(C());
```
eernstg, line 47: Right, we're verifying that `ET1(...)` becomes `ET1(...).call`, which is the topic of this test. We are also verifying that `C()` becomes `C().call` (because otherwise `ET1(C())` would be a compile-time error), which is the standard version of implicit `call` insertion that we've had for years, but we're never using `C.call` in any way.

Perhaps we should just use `ET1(print)` rather than `ET1(C())`, in order to get a simpler and more readable test? Otherwise, we should probably use `C.call` for something, e.g., we could call `it` in the implementation of `ET1.call` and `ET2.call`.

LanguageFeatures/Extension-types/dynamic_semantics_member_invocation_A03_t07.dart
eernstg: Same issue with other tests.

LanguageFeatures/Extension-types/dynamic_semantics_member_invocation_A03_t08.dart
```diff
@@ -0,0 +1,55 @@
...
+main() {
+  ET1 e1 = ET1(C());
+  ET2 e2 = ET2(0);
+
+  Expect.equals("call from C", e1());
+  Expect.equals("call from ET2", e2());
+  Expect.equals("call from C", e1.call());
+  Expect.equals("call from ET2", e2.call());
+  Expect.equals("call from C", (e1.call)());
+  Expect.equals("call from ET2", (e2.call)());
```
eernstg, line 54: Great, thanks!

LanguageFeatures/Extension-types/superinterfaces_of_extension_type_A05_t09.dart
```diff
@@ -0,0 +1,92 @@
...
+/// superinterfaces then the subinterface can resolve the conflict by
+/// redeclaring m.
+///
+/// There is no notion of having a 'correct override relation' here. With
+/// extension types, any member signature can redeclare any other member
+/// signature with the same name, including the case where a method is
+/// redeclared by a getter, or vice versa.
+///
+/// @description Checks that a `call` member can be redeclared as any other
+/// member
```
eernstg, line 36: Very good, I hadn't thought of that!

LanguageFeatures/Extension-types/superinterfaces_of_extension_type_A05_t09.dart
```diff
@@ -0,0 +1,92 @@
...
+class C2 {
+  String get call => "call from C2";
+}
+
+class C3 {
+  void set call(String _) {}
+}
+
+extension type ET1(C1 c) {
+  String call() => "call from ET1";
```
eernstg, line 56: This doesn't actually 'redeclare' `call` because `C1` isn't a superinterface of `ET1`. It would be good if we could add a few words to the description to cover this relationship, but I guess we haven't had a need to talk about it until now:

"An extension type `E` declares a member `m` and the interface of the representation type of `E` has a member named `m` as well."

Note that this could be true also in a case where `E` redeclares `m`, because `E` could have another superinterface `E0` which would also be an extension type, and `E.m` could be a redeclaration of `E0.m`.

It could even be useful to have a test where we verify that this situation doesn't create any confusion about the semantics: `E.m` should be executed, `E0.m` would be shadowed because of the redeclaration, and `R.m` can be called using something like `it.m()` (where `R` is the representation type of `E` and `it` is the representation name).

I added a suggestion about the text in the description.

LanguageFeatures/Extension-types/superinterfaces_of_extension_type_A05_t09.dart
eernstg, line 36: See the comment on line 56:
```suggestion
/// member, and also that an extension type member named `call` can coexist with a 
/// member named `call` in the interface of the representation type.
```

LanguageFeatures/Extension-types/superinterfaces_of_extension_type_A05_t10.dart
```diff
@@ -0,0 +1,86 @@
...
+/// superinterfaces then the subinterface can resolve the conflict by
+/// redeclaring m.
+///
+/// There is no notion of having a 'correct override relation' here. With
+/// extension types, any member signature can redeclare any other member
+/// signature with the same name, including the case where a method is
+/// redeclared by a getter, or vice versa.
+///
+/// @description Checks that a `call` member can be redeclared as any other
+/// member
```
eernstg, line 36:
```suggestion
/// member, and it can coexist with a member named `call` in the interface of the
/// representation type.
```

LanguageFeatures/Extension-types/superinterfaces_of_extension_type_A05_t10.dart
```diff
@@ -0,0 +1,86 @@
...
+class C1 {
+  String call() => "call from C1";
+}
+
+class C2 {
+  String get call => "call from C2";
+}
+
+class C3 {
+  void set call(String _) {}
```
eernstg, line 52: OK, this one is named `call=` rather than `call`, but let's ignore that in the description. Close enough, and we _do_ want to test the 'precludes' case. ;-)

LanguageFeatures/Extension-types/superinterfaces_of_extension_type_A05_t11.dart
```diff
@@ -0,0 +1,106 @@
...
+/// superinterfaces then the subinterface can resolve the conflict by
+/// redeclaring m.
+///
+/// There is no notion of having a 'correct override relation' here. With
+/// extension types, any member signature can redeclare any other member
+/// signature with the same name, including the case where a method is
+/// redeclared by a getter, or vice versa.
+///
+/// @description Checks that a `call` member can be redeclared as any other
+/// member
```
eernstg, line 36: Similar comments a the earlier tests about 'redeclare'.

eernstg (commented):
Looks good! I think the `@assertion` needs to be revised in some of these tests. It describes invocations, not tear-offs. Another thing is that some tests are about a declaration that 'redeclares' `call`: They might benefit from including one more case.

LanguageFeatures/Extension-types/dynamic_semantics_member_invocation_A03_t07.dart
sgrekhov: Tests renamed to `dynamic_semantics_member_invocation_A02_*` because these tests chek this statement

sgrekhov (commented):

eernstg (approved):
LGTM

PR #2442, opened 2023-12-19
Fixes #2419. Add/update tests according to the recent spec changes

### LINKED ISSUE #2419: Add co19 tests for the new specification changes
Dart specification changes are 
- Correct grammar rule for <typeNotVoidNotFunction> to allow import prefix on `Function` https://github.com/dart-lang/language/commit/b44b0f487ae87d28b4f52db00937be57860efcdf
- Correct constant expression spec per language#3468 (language#3495) https://github.com/dart-lang/language/commit/b92280e53da5464ca9486465d5e2b5fc907c1c45

sgrekhov:


eernstg (approved):
LGTM.

However, one thing might need a couple of new tests: The real change for `e1 ~/ e2` is that `e1` and `e2` can have the types `int` or `double` in any combination (previously they had to be `int`).

Did we already test the variants involving `double` operands to `~/` in constant expressions? (in spite of the fact that it was specified to be an error ;-)

If we only have `int` tests then we should have a couple of new tests with doubles and mixtures.

PR #2453, opened 2023-12-25
#2420. Exhaustiveness variable tests. Add switch statement check

sgrekhov:


LanguageFeatures/Extension-types/exhaustiveness_variable_A01_t05.dart
```diff
@@ -0,0 +1,46 @@
...
+extension type AET1(A _) {}
+extension type AET2(A _) implements A {}
+extension type BET1(B _) {}
+extension type BET2(B _) implements B {}
+
+String test1(AET1 a) {
+  switch (a) {
+    case B _: return 'B';
+  }
+}
```
eernstg, line 33: This one is blocked on https://github.com/dart-lang/language/issues/3534: It hasn't yet been decided whether we should use the extension type erasures (as I'd prefer) during flow analysis in order to determine whether or not it is possible to reach the end of the function body in this example.

So we'll have to put this one in a waiting position until https://github.com/dart-lang/language/issues/3534 has been settled.

eernstg (commented):
Blocked on https://github.com/dart-lang/language/issues/3534.

Perhaps split this into one PR that uses return type `String` on a bunch of existing test functions containing switch expressions, and put the new tests into a different PR? We could then land the return-type part, and confine the blockedness to the second PR.

LanguageFeatures/Extension-types/exhaustiveness_variable_A01_t05.dart
sgrekhov: https://github.com/dart-lang/language/issues/3534 has been settled. Now this test should work

sgrekhov (commented):
I believe now this PR is nublocked. Please take another look

LanguageFeatures/Extension-types/exhaustiveness_variable_A01_t05.dart
eernstg: Right!

eernstg (approved):
LGTM!

PR #2454, opened 2023-12-26
#2420. Add object-pattern exhaustiveness tests

sgrekhov:


LanguageFeatures/Extension-types/exhaustiveness_object_A01_t02.dart
```diff
@@ -0,0 +1,78 @@
...
+
+extension type FaceET1<T>(Face<T> _) {}
+extension type FaceET2<T>(Face<T> _) implements Face<T> {}
+
+String test1_1(FaceET1 face) {
+  switch (face) {
+    case Jack(): return 'Jack';
+    case Queen(): return 'Queen';
+    case King(suit: _): return 'King';
+  }
```
eernstg, line 26: This one is blocked on https://github.com/dart-lang/language/issues/3534 and https://github.com/dart-lang/language/issues/3524: If we don't consider `FaceET1` and `FaceET2` as always-exhaustive then line 27 would be considered reachable, and we'd have an error. If we don't consider the extension type erasure of `FaceET1` or `FaceET2` then we can't conclude that the switch statement is exhaustive.

I believe this implies that we can only expect this to compile with no errors if both of those proposals are accepted.

LanguageFeatures/Extension-types/exhaustiveness_object_A01_t03.dart
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Switch statements and expressions with a sealed class as a
+/// matched type are always exhaustive
+///
+/// @description Check that it is no compile-time error if the matched value
+/// type of a switch expression is a sealed class as and the set of cases is an
+/// exhaustive set of object patterns with extension types
```
eernstg, line 10: Yes, this one should work today!

(We _do_ have a decision that object patterns whose type is an extension type should be considered to have the erasure as its type during exhaustiveness computations).

LanguageFeatures/Extension-types/exhaustiveness_object_A01_t04.dart
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Switch statements and expressions with a sealed class as a
+/// matched type are always exhaustive
+///
+/// @description Check that it is no compile-time error if the matched value
+/// type of a switch statement is a sealed class as and the set of cases is an
+/// exhaustive set of object patterns with extension types
```
eernstg, line 10: This one is again blocked.

eernstg (commented):
This one is again blocked (on https://github.com/dart-lang/language/issues/3534 and https://github.com/dart-lang/language/issues/3524, no less).

It should again be possible to split it into two PRs, one which is OK today, and another one which is blocked.

LanguageFeatures/Extension-types/exhaustiveness_object_A01_t02.dart
sgrekhov: @eernstg I believe that `FaceET1` and `FaceET2` are considered as always exhaustive now. So, this shouldn't be blocked any longer. But why we should have an error here? `Face` is a sealed class and `Jack`, `Queen` and `King` is an exhaustive set of descendants. The current implementation doesn't produce an error here

LanguageFeatures/Extension-types/exhaustiveness_object_A01_t04.dart
sgrekhov: Unblocked now? No error in the current implementation

LanguageFeatures/Extension-types/exhaustiveness_object_A01_t02.dart
eernstg: Ah, that's probably because flow analysis trusts a switch statement to be exhaustive when it has been handled by the exhaustivenses algorithm. So with 3524, `FaceET1/2` are always-exhaustive, the exhaustiveness check is satisfied, and the flow analysis doesn't care that it can't see the same thing, it just concludes that we can't reach the end of the function because every branch of this switch will complete returning (that is, it will not complete normally).

So we're good!

LanguageFeatures/Extension-types/exhaustiveness_object_A01_t04.dart
eernstg: Actually, it was probably not blocked after all.

Anyway, the analysis today says: `Face` is always-exhaustive, so we must check exhaustiveness of switch statements (and switch expressions, too, as always). The extension-typed object patterns are checked during exhaustiveness analysis according to their erasure, and we do have all immediate subtypes of `Face` (with some extra tricks around type arguments).

So: Unblocked indeed.

eernstg (approved):
LGTM!

It is a source of confusion that we have exhaustiveness analysis and also flow analysis, and they aren't equally powerful.

PR #2455, opened 2023-12-27
#1400. Add type inference tests for extension types

sgrekhov:
There is no statement from the assertion in the spec but I think it is assumed


eernstg (approved):
LGTM

PR #2457, opened 2023-12-28
Fixes #2440. Add asyncStart/End() to Stream interface tests

### LINKED ISSUE #2440: Missing `asyncStart`/`asyncEnd` in async/StreamController/stream_all_A01_t02
```dart
import "dart:async";
import "../Stream/allTests_A02.lib.dart";

Stream<T> create<T>(Iterable<T> data,
    {bool isError(T element)?, T? defaultValue}) {
  StreamController<T> sc = new StreamController<T>();
  for (T e in data) {
    if (isError != null && isError(e)) {
      sc.addError(e as Object);
    } else {
      sc.add(e);
    }
  }
  new Future(() => sc.close()); // close stream controller later
  return sc.stream;
}

main() {
  test(create);
}
```

Here we need a `asyncEnd` after `sc.clone`, and a corresponding `asyncStart` somewhere.

I'm not sure where would be the best for `asyncStart`. Ideally we want just one `asyncStart` with the right argument. I don't know if it's possible in this test.

sgrekhov:


osa1 (approved):
LGTM, thanks!

PR #2458, opened 2023-12-29
#2420. Add cast-pattern tests for extension types

sgrekhov:


LanguageFeatures/Extension-types/exhaustiveness_cast_A01_t01.dart
```diff
@@ -0,0 +1,100 @@
...
+extension type IntET1(int _) {}
+extension type IntET2(int _) implements int {}
+
+int test1_1(ObjectET1 obj) {
+  switch (obj) {
+    case int(isEven: true) as int:
+      return 1;
+    case int _:
+      return 2;
+  }
```
eernstg, line 37: This one is blocked on https://github.com/dart-lang/language/issues/3534: If I change `ObjectET1` to `Object` then the function has no error (that is, flow analysis then recognizes that it is not possible to reach the end of the function). This implies that (1) the implementation does not use the rule which is recommended in said issue, and (2) if that proposal is implemented then we won't have an error (so it does matter).

LanguageFeatures/Extension-types/exhaustiveness_cast_A01_t01.dart
```diff
@@ -0,0 +1,100 @@
...
+}
+
+int test2_1(Object obj) {
+  switch (obj) {
+    case int(isEven: true) as IntET1:
+      return 1;
+    case int _:
+      return 2;
+  }
+}
```
eernstg, line 56: Same experiment, same outcome: This one is blocked on dart-lang/language#3534.

eernstg (commented):
At least some of these tests are blocked on dart-lang/language#3534.

LanguageFeatures/Extension-types/exhaustiveness_cast_A01_t01.dart
eernstg: Right, `ObjectET1` is not always-exhaustive, so we don't check the switch statement for exhaustiveness and we expect to be able to reach the end of the function, and that's an error. Good!

eernstg (approved):
LGTM

PR #2459, opened 2024-01-02
#2420. Add more dynamic semantics extension type tests

sgrekhov:


LanguageFeatures/Extension-types/dynamic_semantics_extension_type_A06_t01.dart
```diff
@@ -0,0 +1,61 @@
...
+extension type BoolET1(bool _) {}
+extension type BoolET2(bool _) implements bool {}
+extension type const ConstBoolET1(bool _) {}
+extension type const ConstBoolET2(bool _) implements bool {}
+
+main() {
+  var _false = BoolET1(false);
+  var _true = BoolET2(true);
+  const True = ConstBoolET1(true);
+  const False = ConstBoolET2(false);
```
eernstg: Perhaps use standard conventions for naming (we aren't testing non-standard names):

```suggestion
var _false = BoolET1(false);
var _true = BoolET2(true);
const cTrue = ConstBoolET1(true);
const cFalse = ConstBoolET2(false);

main() {
```

This makes `_false` and `_true` standard private names, and it makes `cTrue` and `cFalse` standard camel-case names. I think this is good for the overall readability, and surely we have some other tests verifying that local variables can have "looks just like private" names, and non-types can have capitalized names, even though they shouldn't. ;-)

eernstg: (I had no other comments, so I went ahead and did this.)

eernstg (approved):
LGTM. I adjusted a couple of typo-level things.

PR #2461, opened 2024-01-03
Fixes #2436. Add test for external constructors with initializers

### LINKED ISSUE #2436: Add test for external constructor that initializes fields
Add test like
```dart
class C {
  num x;
  external C(this.x); // Error
}

main() {
  print(C);
}
```

See also https://github.com/dart-lang/language/issues/3516

sgrekhov:


eernstg (approved):
LGTM

eernstg (commented):
LGTM.

We could cover assertions as well, but this is not very important because it's a syntax error already at `:`, and hence it doesn't matter what we put after the `:`, it's still going to be a syntax error.

PR #2467, opened 2024-01-08
#2420. Add constant pattern exhaustiveness tests

sgrekhov:


LanguageFeatures/Extension-types/exhaustiveness_constant_A02_t01.dart
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion If the constant has no primitive equality, then it is not
+/// exhausted
+///
+/// @description Check that if a constant have no primitive equality then it is
+/// not exhausted
```
eernstg, line 9: We might as well stick to the standard phrase which is "has primitive equality" and the negation "does not have primitive equality" (because "has no primitive equality" seems to suggest that it might have one or more of those):
```suggestion
/// @description Check that if a constant does not have primitive equality
/// then it is not exhausted
```

LanguageFeatures/Extension-types/exhaustiveness_constant_A02_t02.dart
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion If the constant has no primitive equality, then it is not
+/// exhausted
+///
+/// @description Check that if a constant have no primitive equality then it is
+/// not exhausted
```
eernstg, line 9: Similar rewrite.

sgrekhov:
Description updated. Please take another look

eernstg (approved):
LGTM

PR #2468, opened 2024-01-08
#2420. Add null-assert pattern exhaustiveness tests

sgrekhov:


eernstg (approved):
LGTM!

(I'm getting used to this: (1) erase the scrutinee type if it is/contains an extension type, (2) check whether it's always-exhaustive, (3) yes, it was, (4) check the switch statement for exhaustiveness, (5) flow analysis just trusts every switch statement which must be exhaustive to be so.)

PR #2469, opened 2024-01-08
#2420. Add another variable pattern exhaustiveness test

sgrekhov:


LanguageFeatures/Extension-types/exhaustiveness_variable_A03_t01.dart
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Exhaustiveness of a variable pattern is determined by the static
+/// type of the corresponding variable.
+///
+/// @description Check that an exhaustiveness of a variable pattern is
```
eernstg, line 8:
```suggestion
/// @description Check that the exhaustiveness of a variable pattern is
```
because the variable has one specific exhaustiveness (property), it can't have several (of which we'd choose one, and say that it's "an exhaustiveness").

We might actually want to talk about "the lifted space union" of the variable pattern, because that's directly using the terminology of the feature specification [exhaustiveness.md](https://github.com/dart-lang/language/blob/main/accepted/3.0/patterns/exhaustiveness.md). This could yield something like this:

```suggestion
/// @description Check that the lifted space union of a variable pattern is
```
It follows implicitly that we're talking about exhaustiveness because that's the only topic where the phrase 'lifted space union' is used.

eernstg (commented):
Looks good, but let's just briefly consider the wording of the description.

sgrekhov:
I was trying to avoid "lifted space union" in the description because it requires detailed knowledge of exhaustiveness specification. But, seems, that it's reasonable to require this knowledge from the one who will deal with exhaustiveness tests :) So, I've updated the description

LanguageFeatures/Extension-types/exhaustiveness_variable_A03_t01.dart
```diff
@@ -5,8 +5,8 @@
 /// @assertion Exhaustiveness of a variable pattern is determined by the static
 /// type of the corresponding variable.
 ///
-/// @description Check that an exhaustiveness of a variable pattern is
-/// determined by the static type of the corresponding variable.
+/// @description Check that a lifted space union of a variable pattern is a
+/// union of the static type of the corresponding variable.
```
eernstg, line 9: Sorry about nit picking one more time, but here's my take on articles:

We use `the X` to indicate a contract with the listener/reader that the set of phenomena described by `X` contains one specific exemplar that we've already agreed on knowing about and focusing on. We use `an X` to indicate that this `X` is new in the discourse.

Much like the distinction between a declaring occurrence of `c` like `var c = new C()` vs. referencing occurrences like `c.foo()`, except that we don't have "the" in Dart, so we need to introduce a named storage location (that is, a variable) in order to be able to refer to a specific phenomenon among the ones that we've introduced.

Natural language just accepts the ambiguity, to some extent: "There were two dogs in the room, a humongous barking mastiff and a sleeping chihuahua. The dog tried to eat me." Ambiguous? Maybe, but natural language users will complain if that's a problem. ;-)

So:

```suggestion
/// @description Check that the lifted space union of a variable pattern is
/// determined by the static type of the corresponding variable.
```

This means that we introduce _a_ variable pattern, then refer to _the_ space which is associated with that pattern (it's uniquely defined when we know the pattern), and finally talk about the value of that space union.

In this particular case I'd suggest saying "is determined by" rather than "is a union of" because we want to emphasize that each static type has a space union (uniquely determined by that type), and that space union is also the space union of the pattern. If we say "is a union of" then we suggest that there are several of those, and we can choose whichever one we want.

eernstg (commented):
OK, nit picking this time. I think we're done next time. ;-)

sgrekhov:
Thank you. Wording updated

eernstg (approved):
LGTM

PR #2470, opened 2024-01-08
#2436. Add expected constant evaluation error for CFE

sgrekhov:
These tests fail in CFE because of the constant evaluation error. I believe it's Ok and updated the tests accordingly


eernstg (approved):
Right, these are derived/secondary errors, and for those we just expect whatever the tools are doing.

PR #2472, opened 2024-01-09
Fixes #2471. Fix timer constructor tests. Add check for microtasks

### LINKED ISSUE #2471: co19/LibTest/async/Timer/Timer.periodic_A02_t01 is flaky
The tests
```
co19/LibTest/async/Timer/Timer.periodic_A02_t01 RuntimeError (expected Pass)
```
are failing on configurations
```
vm-linux-debug-simriscv64
```

The test went from flaky to failing. From the [log](https://dart-ci.appspot.com/log/vm-linux-debug-simriscv64/vm-linux-debug-simriscv64/419/co19/LibTest/async/Timer/Timer.periodic_A02_t01), it seems like a duration is slightly longer than expected. So it will probably revert to flaky again.

sgrekhov:


LibTest/async/Timer/Timer_A01_t01.dart
```diff
@@ -35,16 +33,12 @@ check(int delayms) {
 }
```
eernstg, line 33: Can't comment on line 30, so here we go:

Perhaps it would be better to put `actual` into the message rather than `actual + safetyMargin`: The `actual` value is the time which has actually elapsed, but `actual + safetyMargin` is a value that doesn't really mean anything, it is just larger than `actual` such that it is unlikely that violations of `delay <= actual` will trigger the failure in cases where it's just the browser being imprecise and over-eager.

LibTest/async/Timer/Timer_A02_t02.dart
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion factory Timer(Duration duration, void callback())
+///
+/// A negative duration is treated the same as [Duration.zero]
+///
+/// @description Checks that negative duration is accepted and treated as
+/// [Duration.zero]. Test that timer waits for microtasks queue before the start
```
eernstg: Typo:
```suggestion
/// [Duration.zero]. Test that timer waits for microtask queue before the start
```

eernstg: I'll just fix the typo so we can land this now.

PR #2474, opened 2024-01-10
#2119. Add missing experimental flags

sgrekhov:
Missing! :)


eernstg:
;-D

eernstg (approved):
LGTM

However, I don't quite understand why there is a need for this flag. Isn't every tool already at a version where the flag is on by default?

Issue #2477, opened 2024-01-10
[Extension types] Add more tests for constants and operators


sgrekhov:
See https://github.com/dart-lang/sdk/issues/54545 and https://github.com/dart-lang/sdk/issues/54532

sgrekhov:
Closed as completed

PR #2482, opened 2024-01-12
#2477. Add match operators constants evaluation tests

sgrekhov:


Language/Expressions/Constants/math_operators_A02_t02.dart
```diff
@@ -0,0 +1,53 @@
...
+
+extension type const ExtNum(num _) implements num {}
+
+extension type const ExtInt(int _) implements int {}
+
+extension type const ExtDouble(double _) implements double {}
+
+main() {
+  const c1 = ExtNum(1);
+  Expect.identical(-1, -c1);
```
eernstg, line 43: But here, `-c1` is not in a constant context so we won't get any errors even if it isn't considered constant. Maybe use `const c1 = -ExtNum(1);`?

eernstg: Similar comment on the other cases.

eernstg (commented):
Looks good! Just one thing: I think Language/Expressions/Constants/math_operators_A02_t02.dart needs to be adjusted in order to test a unary minus expression as a constant expression.

sgrekhov:
Indeed. Thank you! Updated

eernstg (approved):
LGTM

PR #2484, opened 2024-01-15
#2477. Add more constant evaluation tests

sgrekhov:
Based on https://github.com/dart-lang/sdk/issues/54491

### DIFF
--- a/LanguageFeatures/Extension-types/static_analysis_extension_types_A20_t07.dart
+++ b/LanguageFeatures/Extension-types/static_analysis_extension_types_A20_t07.dart
@@ -1,0 +1,56 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion The extension type erasure of an extension type V is obtained by
+/// recursively replacing every subterm of V which is an extension type by the
+/// corresponding representation type.
+///
+/// Let X1 extends B1, .. Xs extends Bs be a declaration of the type parameters
+/// of a generic entity (it could be a generic class, extension type, mixin,
+/// typedef, or function). Let BBj be the extension type erasure of
+/// Bj, for j in 1 .. s. It is a compile-time error if
+/// X1 extends BB1, .. Xs extends BBs has any compile-time errors.
+///
+/// @description Check that the erasure of an extension type `V` is obtained by
+/// recursively replacing every subterm of `V` which is an extension type by the
+/// corresponding representation type.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+class Check {
+  const Check(Object? v, {required Object? eq})
+      : assert(v == eq, "Not equal ${(v, eq: eq)}");
+}
+
+extension type const Ext1(Object? value) {}
+extension type const Ext2(Object value) implements Object {}
+
+void main() {
+  const Check([Ext1(1)], eq: <Object?>[1,]);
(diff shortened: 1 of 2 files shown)


LanguageFeatures/Extension-types/static_analysis_extension_types_A20_t07.dart
```diff
@@ -0,0 +1,56 @@
...
+
+void main() {
+  const Check([Ext1(1)], eq: <Object?>[1,]);
+  const Check([Ext2(1)], eq: <Object>[1,]);
+  const Check({Ext1(1)}, eq: <Object?>{1,});
+  const Check({Ext2(1)}, eq: <Object>{1,});
+  const Check({Ext1(1): Ext1(2)}, eq: <Object?, Object?>{1: 2});
+  const Check({Ext2(1): Ext1(2)}, eq: <Object, Object?>{1: 2});
+  const Check({Ext1(1): Ext2(2)}, eq: <Object?, Object>{1: 2});
+  const Check({Ext2(1): Ext2(2)}, eq: <Object, Object>{1: 2});
```
eernstg, line 38: OK, so let me try to understand what we're testing here: We know that the context type for the arguments to `Check` will not cause collection literals to get any particular actual type arguments, which again implies that the collection literals will have type argument `Ext1` respectively `Ext2` based on the given element.

During evaluation (in this case: constant evaluation at compile-time, but that's supposed to work just like run-time evaluation), extension types are erased. So we should get the type argument `Object?` rather than `Ext1`, and `Object` rather than `Ext2`.

The collection literals are evaluated, and the result is canonicalized. Canonicalization will normalize the actual type arguments, but that should not make any difference here.

So we should be able to do `identical` in `Check` (but `==` will also do the job), and we _can_ actually expect the equality/identity to hold.

All good!

The erasure is very simple in this case, so we should probably have another test where the erasure must take more than one step, and the value is obtained from a type argument. For example:

```dart
extension type E1(int it) {}
extension type E2<X>(Map<E1, X> it) {}
```

If we check `[E2<List<E1>>({})]` then it should be a `List<Map<int, List<int>>>`.

LanguageFeatures/Extension-types/static_analysis_extension_types_A20_t07.dart
```diff
@@ -0,0 +1,56 @@
...
+  const Check(((1 as Ext1),), eq: (1,));
+  const Check(((1 as Ext2),), eq: (1,));
+  const Check((Ext1(1) as int,), eq: (1,));
+  const Check((Ext2(1) as int,), eq: (1,));
+  const Check(Ext1((1,)), eq: (1,));
+  const Check(Ext2((1,)), eq: (1,));
+  const Check(Ext1((Ext1(1),)), eq: (1,));
+  const Check(Ext1((Ext2(1),)), eq: (1,));
+  const Check(Ext2((Ext1(1),)), eq: (1,));
+  const Check(Ext2((Ext2(1),)), eq: (1,));
```
eernstg, line 55: All these record tests are more tricky. Records do not have a representation of the types of their components, they only know about their own shape, and then they'll consult each component value for its run-time type.

This means that there is no erasure at any point in this process. We just evaluate the expressions like `Ext1(1)` and get the object `1` of type `int`, and then the record `(1,)` will have the run-time type `(int,)`. The representation type of `Ext1` and `Ext2` could be any supertype of `int`, and it wouldn't make a difference at all.

So I think this isn't actually testing anything about extension type erasures. Have I overlooked something?

eernstg (commented):
Looks good! A couple of comments: We might want to have some cases that are slightly more complex, and perhaps the record tests don't contribute much.

sgrekhov:
Removed record types, added more complicated cases. Please take another look

PR #2492, opened 2024-01-17
Fixes #2487. Add tests for type cast and type `void`

### LINKED ISSUE #2487: Add tests that `as void` is not an expression
> ⟨typeCast⟩ ::= ⟨asOperator⟩ ⟨typeNotVoid⟩
> ⟨asOperator⟩ ::= as
> 

It's a compile-time error if `type` is `void`. Don't forget to check type aliases

sgrekhov:


eernstg (approved):
LGTM

PR #2493, opened 2024-01-17
Fixes #2491. Add test that an extension type cannot implement type `Never`

### LINKED ISSUE #2491: [Extension types] Add tests that extension types cannot implement `Never`
There is no co19 test checking that it is a compile-time error if an extension type implements type `Never`

sgrekhov:
Please note that the statement in the assertion doesn't contain any mentions of type `Never` but I didn't found a better one in the spec. May be the type `Never` was just forgotten there?


eernstg (approved):
LGTM.

One could say that this is just one more test where we determine that an error is reported in the case where the representation type isn't a subtype of an implemented non-extension type. Anyway, it's ok to have this special case even if it is somewhat redundant with other tests where the relation between representation type and implemented types is being tested.

PR #2497, opened 2024-01-18
Fixes #2496. Fix co19 tests that became failing after test runner update

### LINKED ISSUE #2496: Update tests according to the pending change in test runner
Fix the tests that will fail after https://dart-review.googlesource.com/c/sdk/+/345541

sgrekhov:
Please don't merge this PR until landing of https://dart-review.googlesource.com/c/sdk/+/345541


eernstg (approved):
LGTM.

Waiting for https://dart-review.googlesource.com/c/sdk/+/345541.

Language/Libraries_and_Scripts/definition_syntax_t18.dart
```diff
@@ -19,8 +19,9 @@
 /// library begins with the word library (possibly prefaced with any applicable
 /// metadata annotations), followed by a qualified identifier that gives the
 /// name of the library.
```
davidmorgan, line 21: I think some of this is now out of date, corresponding to the change to allow "library;" with no name?

Language/Libraries_and_Scripts/Parts/part_0.dart
```diff
@@ -10,7 +10,4 @@ class Annot {
 }
 
 final foo = "foo";
-//    ^^^
-// [analyzer] unspecified
-// [cfe] unspecified
```
davidmorgan, line 15: Per discussion on

https://github.com/dart-lang/sdk/issues/54661

I think this failure assertion is correct when analyzing `compilation_t01.dart` and incorrect when analyzing `compilation_t04.dart`. We have no way to assert failure in one and not the other, so I think the right fix is to copy `part_0.dart` into a separate version with no failure assertion as in my PR.

I may be wrong, as I commented on the issue there is a lot of complexity here ;)

sgrekhov: Yes, this change seems wrong. Let's wait the conclusion of https://github.com/dart-lang/sdk/issues/54661

sgrekhov:
I prepared a change that fixes all of the new test failures introduced by the change in the test runner. In most cases I simply don't expect analyzer error there. Updated file looks like
```dart
// TODO (sgrekhov) Add expected analyzer error below. For now (Feb, 2024)
// analyzer doesn't analyze parts, this error is not reported by the analyzer
// and this is not an error but an intended behavior.
// See https://github.com/dart-lang/sdk/issues/54661

part of assert;
//      ^^^^^^
// [cfe] unspecified
```
This change will fix the current errors. When the analyzer will be fixed (at least in the test runner) to analyze the parts I'll remove TODO and add error expectations.

I was going to submit this change, but according to the [latest comment](https://github.com/dart-lang/sdk/issues/54661#issuecomment-1930655040)

> For the test runner I expect that it would be natural to report every diagnostic from every reachable file (which would be trivial to implement with what we have today.

If the fix of the test runner is a trivial may be it makes sense to wait for the change first and then fix the tests?

The third option is to allow the test to be failing for now until the test runner will be fixed. Just to add issue number
```dart
// @issue 54661

part of assert;
//      ^^^^^^
// [analyzer] unspecified
// [cfe] unspecified
```

@davidmorgan @eernstg what do you think? Which option is better?

Language/Libraries_and_Scripts/definition_syntax_t18.dart
eernstg: I'm wondering: Do we gain anything by having the `library;` directive in Language/Libraries_and_Scripts/definition_syntax_t18_lib.dart and importing that library into this one? It seems likely to me that we could just avoid having any connection to multiple libraries and reporting errors (or not, as it would be expected today) in a different library than the entry point.

We would then add `library;` to this library (expecting no errors), remove the import, delete Language/Libraries_and_Scripts/definition_syntax_t18_lib.dart, and thus get to a state where this library is a non-problem relative to the test runner's treatment of errors outside the entry point.

Language/Libraries_and_Scripts/Parts/part_0.dart
eernstg: At this point I think we have a plan: Change all tests involving parts such that each part has a uniquely determined owner (in co19, we can't stop anyone "out there" who insists on owning one of these parts). We would then create 4 copies of this part file and use each of them as a part in `compilation_t0n.dart` for `n` in `{1, 2, 3, 4}`. This would eliminate the contradictory expectations.

eernstg:
About the test runner reporting diagnostics from all entities (libraries and parts) that are being analyzed: That sounds like a really nice solution! I'm pushing for that [here](https://github.com/dart-lang/sdk/issues/54661#issuecomment-1931613046).

So let's try to clarify this proposal a bit more, and then the answers to the questions about `part of assert;` etc. might follow automatically.

eernstg (commented):
Added a couple of comments about things that may be resolved at this time.

davidmorgan:
Getting back to this: my PR for the test_runner is about ready, I'll send it out for review. https://dart-review.googlesource.com/c/sdk/+/345541 ... It now adds e2e tests so you can see that expectations for the analyzer are picked up for imported libraries and included parts.

You can see the co19 test failures here https://dart-ci.firebaseapp.com/cl/345541/8 ... a question @eernstg @sgrekhov should I aim to land it with approvals for those failures? I think another option would be to land these co19 test fixes first and roll them into the SDK, as my understand is they should pass with both versions of the test runner.

Thanks!

sgrekhov:
It's our usual practice to fill co19 issue and then to approve the failures. And I fix the failing tests later. So, for me it's ok simply to approve the failures

eernstg:
@davidmorgan wrote:
> land it with approvals for those failures?

Sounds good! Usually, a test failure approval comment will mention an issue where the situation is described and any remaining tasks are tracked. In this case I guess that would be an issue on the analyzer-as-test_runner, requesting that it should report all compile-time errors from the transitive closure of each of the analysis targets. Do we have an issue like that? [This comment](https://github.com/dart-lang/sdk/issues/54661#issuecomment-1934121927) seems to be really close. ;-)

sgrekhov:
co19 issue https://github.com/dart-lang/co19/issues/2496

davidmorgan:
Aaand the test_runner change is in! Indeed I pointed to #2496 in approving the failures.

sgrekhov:
Thank you! Cool! I'll fix the failures

Language/Libraries_and_Scripts/definition_syntax_t18.dart
```diff
@@ -19,8 +19,9 @@
...
-/// @description Checks that it is a compile-time error if the library directive
-/// is missing the library name.
+///
+/// @description Checks that it is not an error if a library name is missing in
+/// the library directive
```
eernstg, line 24:
```suggestion
/// the library directive of the imported library.
```

Language/Libraries_and_Scripts/definition_syntax_t18.dart
eernstg: Looks good now! Should perhaps change the description as mentioned below.

Language/Libraries_and_Scripts/Parts/part_0.dart
eernstg: I think this was done already?

eernstg (approved):
I find it a little difficult to see whether everything is done here. How about we land this, and then we can try to include it in a roll and see if it is working? ;-)

@davidmorgan, @sgrekhov, WDYT?

sgrekhov:
@eernstg I'm going to update this PR tomorrow and send it to review. So, let's wait till tomorrow

eernstg:
Very good, thanks!

sgrekhov:
Updated. All tests that marked as failing here https://dart-ci.firebaseapp.com/cl/345541/10 became Ok. Erik's notes also implemented. Please review

Note that https://dart-review.googlesource.com/c/sdk/+/345541 is still not landed yet

davidmorgan:
Whoops, I thought it was merged, thanks for pointing that out :)

eernstg (approved):
LGTM!

davidmorgan:
Okay, the test_runner change is actually merged now. Thanks again and sorry for the confusion caused last time :)

PR #2498, opened 2024-01-19
#2420. Add more exhaustiveness tests

sgrekhov:
You may say that we already have such tests. But my check showed that we have similar tests for constants only


LanguageFeatures/Extension-types/static_analysis_extension_types_A33_t03.dart
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion An extension type V is always-exhaustive if and only if its
+/// instantiated representation type is always-exhaustive.
+///
+/// @description Check that an extension type is not always-exhaustive if its
+/// representation type is not always-exhaustive but adding an exhaustive case
+/// removes a compile-time error
```
eernstg, line 10: I don't think it's possible to match up a test like this one with the concept of an always-exhaustive type: Such types are taken into account during the static analysis of switch statements (a switch statement must be exhaustive if its scrutinee has a static type which is always-exhaustive), but they simply aren't relevant to switch expressions (they must _always_ be exhaustive, period). We should probably just drop this test.

LanguageFeatures/Extension-types/static_analysis_extension_types_A33_t01.dart
```diff
@@ -0,0 +1,38 @@
...
+    case BoolET(b: false):
+      return "False";
+  }
+}
+
+String test2(BoolET b) =>
+    switch (b) {
+      BoolET(b: true) => "True",
+      BoolET(b: false) => "False"
+    };
```
eernstg, line 31: The notion of `BoolET` being always-exhaustive doesn't matter here: Switch expressions must be exhaustive no matter what they're switching on.

One could perhaps claim that this is a test of the exhaustiveness analysis: It is only possible for for the exhaustiveness analysis to conclude that this switch expression is exhaustive if it uses the extension type erasure to discover that the scrutinee is a `bool` and the cases cover `true` and `false`.

But in that case the description should say something like "testing that the exhaustiveness analysis uses extension type erasure to compute space unions of the matched value type and the patterns". But that might be covered by some other test already? (constant only, perhaps?)

eernstg (commented):
Looks good, but some parts don't really fit the stated topic. In particular, "always-exhaustive" and "switch expression" are not compatible.

LanguageFeatures/Extension-types/static_analysis_extension_types_A33_t01.dart
sgrekhov: Yes, we have such test for constants. Dropped test for switch expression

LanguageFeatures/Extension-types/static_analysis_extension_types_A33_t03.dart
sgrekhov: Ok, dropped

sgrekhov (commented):

eernstg (approved):
LGTM, and thanks! Sorry about being a little bit destructive here. ;-)

PR #2500, opened 2024-01-22
Fixes #2490. Add more await tests for extension types

### LINKED ISSUE #2490: [Extension types] Adjust tests around extension types and await
See recent spec [change](https://github.com/dart-lang/language/commit/c71b62142c6f3d63ed438c12305b6f92485542fe)

sgrekhov:


LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t04.dart
```diff
@@ -0,0 +1,58 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  V3<int>? v3_1 = V3(42);
+  await v3_1;
+//      ^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  V3 v3_2 = null;
```
eernstg, line 53: Probably
```suggestion
  V3? v3_2 = null;
```

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t04.dart
```diff
@@ -0,0 +1,58 @@
...
+  await v3_2;
+//      ^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
```
eernstg, line 58: It would be great if we also had V4, V5, V6 where the representation type has a future type (it could just be `Future<int>`, but also `FutureOr<int>?`), but the extension type still does not have `Future` anywhere as a superinterface.

(The types that don't "admit" to have anything to do with futures and still have a representation object which is a future are probably the most confusing ones, so they shouldn't be allowed to go under the radar).

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t06.dart
```diff
@@ -0,0 +1,37 @@
...
+/// - T is S?, and S is incompatible with await.
+/// - T is X & B, and either:
+/// - B is incompatible with await, or
+/// - B does not derive a future type, and X is incompatible with await.
+/// - T is a type variable with bound S, and S is incompatible with await.
+/// Consider an expression of the form await e. A compile-time error occurs if
+/// the static type of e is incompatible with await.
+///
+/// @description Checks that it is a compile-time error if `await e` occurs, and
+/// the static type of `e` is `X & B`, and `B` is incompatible with await.
```
eernstg, line 17: We don't have any type variables here as far as I can see, so we could never have a type of the form `X & B`. We'd need something like `void foo<X extends ET>(X x)`, and then promote, and then `await x;`.

eernstg (commented):
Looks good! A few tests have issues, though, in particular with intersection types.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t08.dart
```diff
@@ -0,0 +1,34 @@
...
+/// - T is S?, and S is incompatible with await.
+/// - T is X & B, and either:
+/// - B is incompatible with await, or
+/// - B does not derive a future type, and X is incompatible with await.
+/// - T is a type variable with bound S, and S is incompatible with await.
+/// Consider an expression of the form await e. A compile-time error occurs if
+/// the static type of e is incompatible with await.
+///
+/// @description Checks that it is not an error if `await e` occurs, and the
+/// static type of `e` is `X & B` and both `X` and `B` are compatible with await
```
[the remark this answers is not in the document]
sgrekhov: Fixed

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t10.dart
```diff
@@ -0,0 +1,32 @@
...
+/// and `S` is compatible with await.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+extension type ET(Future<num> n) implements Future<Object> {}
+
+test<T1 extends num, T2 extends ET>(T1 t1, T2 t2) async {
+  await t1;
+  await t2;
```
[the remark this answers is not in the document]
sgrekhov: Done

sgrekhov (commented):

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t04.dart
eernstg: Great, thanks!

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t06.dart
```diff
@@ -0,0 +1,55 @@
...
+import 'dart:async';
+
+extension type NumET(num _) {}
+extension type IntET(int _) {}
+
+test1<X extends FutureOr<NumET>>(X x) async {
+  if (x is int) {
+    await x; // No error here, int is compatible with await
+  }
+  if (x is IntET) {
```
eernstg, line 32: Same issue here, there is no promotion of `x`.

eernstg: Similar issues in test2.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t06.dart
```diff
@@ -0,0 +1,55 @@
...
+
+// SharedOptions=--enable-experiment=inline-class
+
+import 'dart:async';
+
+extension type NumET(num _) {}
+extension type IntET(int _) {}
+
+test1<X extends FutureOr<NumET>>(X x) async {
+  if (x is int) {
```
eernstg, line 29: I think this would have to be something like `x is NumET`: We don't get a promotion of `x` to a non-subtype of `FutureOr<NumET>`, and then we aren't testing anything of the form `X & B`.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t07.dart
```diff
@@ -0,0 +1,40 @@
...
+/// - T is X & B, and either:
+/// - B is incompatible with await, or
+/// - B does not derive a future type, and X is incompatible with await.
+/// - T is a type variable with bound S, and S is incompatible with await.
+/// Consider an expression of the form await e. A compile-time error occurs if
+/// the static type of e is incompatible with await.
+///
+/// @description Checks that it is a compile-time error if `await e` occurs, and
+/// the static type of `e` is `X & B` and `B` does not derive a future type, and
+/// `X` is incompatible with await.
```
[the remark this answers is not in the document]
eernstg: Still true, same situation as in LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t06.dart. ;-)

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t08.dart
```diff
@@ -0,0 +1,46 @@
...
+import 'dart:async';
+
+extension type IntET(int _) {}
+extension type ET(Future<int> _) implements Future<int> {}
+
+test<X extends FutureOr<num>>(X x) async {
+  if (x is num) {
+    await x;
+  }
+  if (x is FutureOr<Object>?) {
```
eernstg, line 31: This is not a subtype, no promotion, no intersection type. (So there is no error, because `X` _is_ compatible with await, but we're probably intending to test the handling of intersection types).

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t06.dart
```diff
@@ -0,0 +1,37 @@
...
+/// @issue 54648, 54649
+
+// SharedOptions=--enable-experiment=inline-class
+
+extension type IntET(int i) {}
+extension type ET(Future<int> id) implements Future<int> {}
+
+main() async {
+  ET et = ET(Future<int>.value(2));
+  if (et is int) {
```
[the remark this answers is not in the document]
eernstg: Added a new comment on this library.

eernstg (commented):
Still a couple of issues. I've commented on several cases where the description is about intersection types, but we don't have an expression whose type is an intersection type. In general, we just need to perform the type test on a subtype of the current type of the variable at that point.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t08.dart
sgrekhov: Changed to `FutureOr<int>`

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t07.dart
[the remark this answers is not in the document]
sgrekhov: I'm not 100%  sure but updated. Please review. If the updated test is a valid one now?

sgrekhov (commented):
Updated. Please review, there are some places where I'm not 100% sure

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t06.dart
eernstg: Not yet: The type of `x` is initially `X extends FutureOr<NumET>`. We then test for `int`, but `int` is not a subtype of `FutureOr<NumET>` and hence there's no promotion and no intersection type.

However, `NumET <: FutureOr<NumET>` because `T <: FutureOr<T>` for any `T` (including extension types), so we could have `if (x is NumET) ...` and that would promote the type of `x` to `X & NumET`. Similarly, we could have `if (x is IntET) ...` if we have `IntET <: NumET` (as in the new code).

`NumET` and `IntET` are incompatible with await, but that's OK: If we `await x` then we just determine that `NumET`/`IntET` do not derive a future type, but `X` does, and then `await x` is OK, with type `flatten(X) == flatten(FutureOr<NumET>) == NumET`.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t07.dart
```diff
@@ -19,17 +19,18 @@
...
 
-extension type NumET(num _) {}
-extension type IntET(int _) {}
-extension type FutureET<T>(Future<T> _) {}
+extension type NumET(num _) implements num {}
+extension type IntET(int _) implements NumET {}
 
-test1<X extends Future<NumET>>(X x) async {
+test1<X extends NumET>(X x) async {
   if (x is int) {
```
eernstg, line 28: We still can't promote to `int`, because that isn't a subtype of `NumET`.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t07.dart
```diff
@@ -19,17 +19,18 @@
...
     await x; // No error here, int is compatible with await
   }
+  if (x is FutureOr<IntET>) {
```
eernstg, line 31: Can't promote to `FutureOr<IntET>`.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t07.dart
```diff
@@ -38,10 +39,13 @@ test1<X extends Future<NumET>>(X x) async {
   }
 }
 
-test2<X extends FutureOr<num>>(X x) async {
+test2<X extends num>(X x) async {
   if (x is int) {
```
eernstg, line 43: OK, this works! ;-)

We promote `x` to `X & int`, and `int` is compatible with await, and `flatten(int) == int`.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t07.dart
```diff
@@ -38,10 +39,13 @@ test1<X extends Future<NumET>>(X x) async {
...
     await x; // No error here, int is compatible with await
   }
+  if (x is FutureOr<IntET>) {
```
eernstg, line 46: Can't promote `x` because `FutureOr<IntET>` is not a subtype of `num`.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t08.dart
```diff
@@ -28,7 +28,7 @@ test<X extends FutureOr<num>>(X x) async {
   if (x is num) {
     await x;
   }
-  if (x is FutureOr<Object>?) {
+  if (x is FutureOr<int>?) {
```
eernstg, line 31: Yes, this works! `await x` below should have static type `int`.

eernstg (commented):
This one still has some issues.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t06.dart
sgrekhov: > `NumET` and `IntET` are incompatible with await, but that's OK: If we await x then we just determine that `NumET`/`IntET` do not derive a future type, but `X` does, and then `await x` is OK, with type `flatten(X) == flatten(FutureOr<NumET>) == NumET`.

Sorry, I don't quite understand. Do you want to sat that 
```dart
if (x is IntET) {
  await x;
}
```
is OK? No error?  But, according to the definition

> - T is X & B, and either:
>   - B is incompatible with await, or
In case of `x is IntET`, `IntET` is incompatible with await, criteria met, so `await x` should be a compile-time error.

eernstg: Ah, you're right. We have a slight inconsistency because 'incompatible with await' is more aggressive than _flatten_, so we never get around to try to invoke _flatten_ on the type of `x` because it is already an error, as you mention, according to the 'incompatible' rule to have `await x`.

The underlying question is "do we want to make it an error to have `await e` when the type of `e` is `X & B` where `B` 'is bad' but `X` is 'good'?". The 'incompatible' rule says "yes, don't even bother to look at `X`", but _flatten_ allows us to disregard `B` if it doesn't derive any future types, and then use `X`.

So the question is whether we can actually have this situation at all:

- We do `await x`.
- The static type of `x` is `X & B`.
- `B` is an extension type which is compatible with await.
- `B` does not derive a future type.
- So we use `X` in flatten, and it _does_ derive a future type.

I don't think we can find a type `B` with those properties.

So this basically means that we can't use the 'otherwise' case in the rule about _flatten_ on an intersection type `X & B` where `B` is an extension type.

But then we'd just have to test that we do get the compile-time error when `B` is an extension type that does not implement `Future<T>` for any `T`. Perhaps that's already done in some other test?

Next, we can test that the first case is used in the _flatten_ rule about `X & B` when `X` does not derive a future type, but `B` does (perhaps `X extends Object` and `B == NumFutureET`).

eernstg (commented):
Added a comment.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t07.dart
sgrekhov: Indeed... Removed this

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t06.dart
sgrekhov: > But then we'd just have to test that we do get the compile-time error when `B` is an extension type that does not implement `Future<T>` for any `T`. Perhaps that's already done in some other test?

Below in this test we have `if (x is IntET) {`

> Next, we can test that the first case is used in the flatten rule about `X & B` when `X` does not derive a future type, but `B` does (perhaps `X extends Object` and `B == NumFutureET`).

Added `static_analysis_extension_types_A08_t11.dart` which checks it. In case `X extends Object` `X` is compatible with await, so I used `X extends ObjectET`

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t06.dart
sgrekhov: Fixed

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t07.dart
sgrekhov: Fixed

sgrekhov (commented):
Thank you! Updated once again. Please review

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t07.dart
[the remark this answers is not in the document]
eernstg: OK! Now we get an error because the type of `x` in line 28 is `X & IntET`, and `IntET` is incompatible with await.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t07.dart
```diff
@@ -0,0 +1,50 @@
...
+/// - T is X & B, and either:
+///   - B is incompatible with await, or
+///   - B does not derive a future type, and X is incompatible with await.
+/// - T is a type variable with bound S, and S is incompatible with await.
+/// Consider an expression of the form await e. A compile-time error occurs if
+/// the static type of e is incompatible with await.
+///
+/// @description Checks that it is a compile-time error if `await e` occurs, and
+/// the static type of `e` is `X & B` and `B` does not derive a future type, and
+/// `X` is incompatible with await.
```
eernstg, line 18:
```suggestion
/// the static type of `e` is `X & B`, and `B` is incompatible with await.
```

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t08.dart
```diff
@@ -0,0 +1,46 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+import 'dart:async';
+
+extension type IntET(int _) {}
+extension type ET(Future<int> _) implements Future<int> {}
+
+test<X extends FutureOr<num>>(X x) async {
```
eernstg, line 27: Oh, I missed one thing: The type tested in line 31 is `FutureOr<int>?` which is not a subtype of `FutureOr<num>`. So we still won't get a promotion in line 31 unless we add that `?` on the bound as well, or removed it in line 31.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t06.dart
eernstg: Thanks!

eernstg (commented):
Just one more thing: I think there is still one `?` that prevents promotion.

eernstg (approved):
LGTM!

Issue #2485, opened 2024-01-15
Add more constant evaluation tests


sgrekhov:
There are no co19 tests for constant expressions of the forms like `e as T`, `e is T`, `e is! T` etc. Note https://github.com/dart-lang/sdk/issues/54620

PR #2501, opened 2024-01-23
#2485. Add is T constant evaluation tests

sgrekhov:


Language/Expressions/Constants/as_type_A01_t02.dart
```diff
@@ -73,7 +73,8 @@ main() {
   const fo7 = IntET2(1) as FutureOr<IntET1Alias>;
 
   const f1 = foo as void Function();
-  const f2 = bar<int> as int Function<int>(int);
+  const f2 = as as void Function();
+  const f3 = bar<int> as int Function<int>(int);
```
eernstg, line 77: If this is not intended to be a compile-time error I think we'd do this:
```suggestion
  const f3 = bar<int> as int Function(int);
```

Language/Expressions/Constants/is_type_A02_t02.dart
```diff
@@ -0,0 +1,46 @@
...
+// [cfe] unspecified
+  const Void = void is Type;
+//             ^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  const cIntExtension = IntExtension is Type;
+//                      ^^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  const F = () {} is Function;
```
eernstg, line 32: This is an expression, just not a constant one. Fits in a different test?

eernstg (commented):
Looks good! Just a couple of glitches.

Language/Expressions/Constants/as_type_A01_t02.dart
sgrekhov: Updated

Language/Expressions/Constants/is_type_A02_t02.dart
sgrekhov: Indeed. Moved to appropriate test. Thanx!

sgrekhov (commented):
Thank you! Updated. Please take another look

eernstg (approved):
LGTM

PR #2513, opened 2024-01-30
#2420. Add more exhaustiveness tests

sgrekhov:


LanguageFeatures/Extension-types/exhaustiveness_flow_analysis_A01_t01.dart
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Extension type erasure is not used during flow analysis
+///
+/// @description Check that an extension type erasure is not used during flow
+/// analysis
```
eernstg, line 8: Perhaps:
```suggestion
/// analysis that isn't concerned with run-time type tests
```

eernstg: I don't really understand what we're testing here (e.g., line 17-30).

We could have a variable whose type is an extension type like `ET`, and it could be erased such that it is actually treated as if it had type `int` (which would be a bug). But we don't have that.

We could also have an object pattern like `ET(value: var __)`, but in the given example there is no promotion because `ET` isn't a subtype of `int`. So `i` still has the type `int`, and there is nothing to erase (so we can't really be testing erasure).

The last perspective that I can think of right now is that the flow analysis will (and should) use the extension type erasure when it is detecting that `case ET...` handles all values of type `int`.

However, the switch isn't on an always-exhaustive type, so it isn't required to be exhaustive.

I think we're just going to have the error that `x` isn't definitely assigned in line 27, and, as far as I can see, nothing about the code so far has anything to do with "extension type erasure isn't used for non-run-time-type test flow analysis".

LanguageFeatures/Extension-types/exhaustiveness_variable_A04_t01.dart
```diff
@@ -0,0 +1,36 @@
...
+import '../../Utils/static_type_helper.dart';
+
+extension type ET(int i) {}
+extension type TO(int i) implements ET {}
+
+main() {
+  ET x = ET(1);
+  switch (x) {
+    case int y:
+      x.expectStaticType<Exactly<ET>>;
```
eernstg, line 22: Yep! This is the kind of test I was looking for above: The type of `x` hasn't been erased to `int`.

LanguageFeatures/Extension-types/static_analysis_extension_types_A33_t03.dart
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion An extension type V is always-exhaustive if and only if its
+/// instantiated representation type is always-exhaustive.
+///
+/// @description Check that an extension type is always-exhaustive if its
+/// representation type is always-exhaustive.
```
eernstg, line 9: In this situation, `B` isn't always-exhaustive, which means that `BET` also isn't always-exhaustive, which means that switch statements aren't subject to exhaustiveness analysis.

How would we detect that?

Presumably, we'd have a switch statement whose cases aren't exhaustive, and it should not be an error.

The function could have return type `void`, because we don't want it to be an error if the location after the switch statement is considered reachable (it _must_ be considered reachable, it would be a soundness issue if it is considered unreachable).

So the main point is that we can "measure" the status (is always-exhaustive vs. not) by having that non-exhaustive switch statement, and expecting no error if we expect not-always-exhaustiveness.

LanguageFeatures/Patterns/Exhaustiveness/exhaustiveness_sealed_A03_t01.dart
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Switch statements and expressions with a sealed class as a
+/// matched type are always exhaustive
+///
+/// @description Check that if a matched value of a switch statement is not a
+/// sealed class then flow analysis treats this statement as a not exhaustive
```
eernstg, line 9: I think being 'not a sealed class' is much too broad here. For instance, an enum value is definitely not a sealed class, but covering all values of an enum type can be highly relevant to exhaustiveness.

Again, I think the direct way to check that the compiler/analyzer treats a scrutinee as having a non-exhaustive type is to have a switch statement which isn't exhaustive. That's OK if the scrutinee type is not always-exhaustive, but it is an error if it is always-exhaustive.

LanguageFeatures/Patterns/Exhaustiveness/exhaustiveness_sealed_A03_t01.dart
```diff
@@ -0,0 +1,28 @@
...
+void main() {
+  int x;
+  A a = B();
+  switch (a) {
+    case B():
+      print("x not assigned here");
+    case _ :
+      x = 1;
+  }
+  print(x);
```
eernstg, line 24: I don't think line 24 is a very helpful statement in this test.

The type `A` is not always-exhaustive, so we'd expect the switch statement to be OK even though it isn't exhaustive ... but it _is_ exhaustive because it ends in a catch-all case. So we don't get any information about whether the compiler thought the switch statement was or wasn't required to be exhaustive.

eernstg (commented):
Lots of comments on this one. ;-)

LanguageFeatures/Patterns/Exhaustiveness/exhaustiveness_sealed_A03_t01.dart
sgrekhov: Ups, my bad. Class `B` here shouldn't extends `A`, of course. The main idea of this tests is to check that because `B`  is not a sealed class then it can be a hidden subtype of `A` and `B` somewhere in another library. Therefore in spite of the fact that `case B():` looks unreacable it isn't so and therefore this should be a compile-time error. Updated. Does it makes sense now?

sgrekhov (commented):

LanguageFeatures/Extension-types/exhaustiveness_flow_analysis_A01_t01.dart
```diff
@@ -19,7 +19,7 @@ void main() {
   switch (i) {
     case >0:
       x = 1;
-    case ET1(value: var __):
+    case ET1 _:
```
eernstg, line 22: OK, so now we have a variable pattern rather than an object pattern. This means that we will have a variable (named `_`) whose type is `ET`. It would be a bug if that variable were treated as if it had type `int`. So we could test that be having something like `_.myETMethod()` which would then be an error.

I don't see how the current code would be able to detect that `_` has a wrong type.

eernstg: Similarly for line 35.

LanguageFeatures/Extension-types/exhaustiveness_flow_analysis_A01_t01.dart
eernstg: It's tricky: How do we even detect that flow analysis has a bug which is that it uses the erased type for "anything other than run-time type tests", because "anything other than" is complex in the first place.

But an obvious test would be that a variable pattern like `ET v` will give `v` the type `ET` (and not `int`). This could be tested by using something in the interface of `ET`, or by an assignment to yet another variable like `ET v2 = v;`. 

We could also have a test like `switch (o) { case ET(): o.myETMethod(); ... }` where `o` has a static type that allows promotion to `ET` (e.g., `o` could have the type `Object?`). This would show that promotion of the scrutinee is capable of promoting to an extension type (and it doesn't erase the extension type when using the scrutinee in that case).

LanguageFeatures/Extension-types/static_analysis_extension_types_A33_t03.dart
eernstg: OK. Checking that the issue reported in dart-lang/sdk#54444 has been fixed could be done by checking that no compile-time error is emitted for the following:

```dart
class B {}
extension type BET(B _) {}

String test1(BET b) { // "OK" would mean that the spec change has been implemented.
  switch (b) {
    case B _: return 'B';
  }
}
```

The error used to be emitted because `BET` wasn't erased to `B` during flow analysis, and then flow analysis wouldn't conclude anything useful from knowing that `case B` returns (and then we'd get the "implicitly returns null at end of function" error).

But my understanding was that this series of tests is about the opposite property: We already have other tests about using the erasure to reason about run-time type tests, is this series of tests we want to see that the erasure is not applied for any other reason.

Isn't it true that we already have tests testing that we _do_ get the erasure when reasoning about run-time type tests? (That's basically about switch statements being able or unable to complete normally.)

eernstg (commented):
This is still really tricky. The difficulties arise because some tests claim that they are about "flow analysis does not use erasure...", and that's a really complex property to test.

The simple case that I can think of is that some expressions of type `ET` are erased to `int` _for the purpose of checking control flow based on run-time tests/checks_, and we just need to make sure that the same expressions aren't erased when we perform normal type checking.

sgrekhov (commented):
Partially updated. Github doesn't show me all of the threads on one page therefore I'll comment here the others as well.
1) `exhaustiveness_flow_analysis_A01_t01.dart` updated to check that there is no erasure by calling member of an extension type
2) Yes, we really have tests that check that https://github.com/dart-lang/sdk/issues/54444 is fixed, for example https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Extension-types/exhaustiveness_variable_A01_t08.dart. So I need to rewrite `exhaustiveness_flow_analysis_A02_t01.dart`. But I don't quite understand your suggestion  to use a not exhaustive type and not exhaustive switch statement. Could you, please, provide an example?
3) Added `exhaustiveness_sealed_A03_t02.dart` containing "not possible nut possible" test

Do I forget anything? Please review and advice with p.2.

LanguageFeatures/Extension-types/exhaustiveness_flow_analysis_A01_t01.dart
eernstg: Yes, got it!

LanguageFeatures/Extension-types/exhaustiveness_flow_analysis_A01_t01.dart
```diff
@@ -0,0 +1,30 @@
...
+}
+extension type const ET2(int value) implements int {
+  void confirmET2() {}
+}
+
+void main() {
+  int i = 42;
+  switch (i) {
+    case ET1 v:
+      v.confirmET1(); // This confirms that ET1 is not erased by int
```
eernstg, line 24:
```suggestion
      v.confirmET1(); // This confirms that `ET1` is not erased to int
```
(Tiny thing, but I had another comment, so I could just as well mention this one. ;-)

LanguageFeatures/Extension-types/exhaustiveness_flow_analysis_A01_t01.dart
```diff
@@ -0,0 +1,30 @@
...
+  }
+  switch (i) {
+    case ET2 v:
+    v.confirmET2();
```
eernstg, line 28: I think `dart format` would put it here:
```suggestion
      v.confirmET2();
```

LanguageFeatures/Extension-types/static_analysis_extension_types_A33_t03.dart
eernstg: Thanks!

LanguageFeatures/Extension-types/exhaustiveness_flow_analysis_A02_t01.dart
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Extension type erasure is used by flow analysis for implicit `is`
+/// reachability.
+///
+/// @description Check that an xxtension type erasure is used by flow analysis
```
eernstg, line 8: Typo:
```suggestion
/// @description Check that an extension type erasure is used by flow analysis
```

eernstg: Github _will not show_ the comment starting with 'The main idea of this test was to check that issue described in https://github.com/dart-lang/sdk/issues/54444 is fixed.' I can't really see what the context was, but II think it's fine. ;-)

LanguageFeatures/Extension-types/exhaustiveness_flow_analysis_A01_t01.dart
eernstg: Thanks!

LanguageFeatures/Patterns/Exhaustiveness/exhaustiveness_sealed_A03_t01.dart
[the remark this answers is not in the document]
eernstg: Good, thanks!

eernstg (commented):
Looks really good! Just a couple of typos left.

sgrekhov:
Typos and formatting fixed. I also deleted `LanguageFeatures/Extension-types/exhaustiveness_flow_analysis_A02_t01.dart`. https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Extension-types/exhaustiveness_variable_A01_t08.dart. does the same checks. Please review

eernstg (approved):
LGTM.

Issue #2446, opened 2023-12-20
Tests needs update after change to exhaustiveness


johnniwinther:
Exhaustiveness for cast-patterns has been updated in https://github.com/dart-lang/language/commit/28087025d5da5ee72b94db18c512f83220768896.

With this change some tests needs to be updated:

`co19/LanguageFeatures/Patterns/cast_A02_t02` needs to be updated with new expectations.

`co19/LanguageFeatures/Patterns/Exhaustiveness/lifting_cast_pattern_A02_t01` is no longer valid and needs to be updated to show that the code is no longer expected to compile.

PR #2514, opened 2024-01-31
#2446. Add additional test for cast pattern

sgrekhov:


eernstg (approved):
LGTM.

So we're testing that the lifted space union of the pattern in line 27 covers all of the lifted space union of the scrutinee type `dynamic`. This is true because a case is considered to cover the entire scrutinee type when it is a cast containing a pattern that is guaranteed to match every object that doesn't cause the `as` to throw.

PR #2516, opened 2024-02-01
Fix #2490. Update incompatible with await tests according to the changed rule

### LINKED ISSUE #2490: [Extension types] Adjust tests around extension types and await

sgrekhov:


LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t12.dart
```diff
@@ -0,0 +1,31 @@
...
+/// static type of `e` is `X & B` and `B` does not derive a future type but
+/// compatible with await, and `X` is incompatible with await.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+extension type NumET(num _) implements num {}
+
+test<X extends NumET>(X x) async {
+  if (x is int) {
```
eernstg, line 24: We can't promote from `NumET` to `int` because we don't have `int <: NumET`. However, `await x` will still be an error, because `X` is incompatible with await.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t12.dart
```diff
@@ -0,0 +1,31 @@
...
+/// of the following criteria holds:
+/// - T is an extension type that does not implement Future.
+/// - T is S?, and S is incompatible with await.
+/// - T is X & B, and B is incompatible with await.
+/// - T is a type variable with bound S, and S is incompatible with await.
+/// Consider an expression of the form await e. A compile-time error occurs if
+/// the static type of e is incompatible with await.
+///
+/// @description Checks that it is not an error if `await e` occurs, and the
+/// static type of `e` is `X & B` and `B` does not derive a future type but
```
eernstg, line 15: I think this might not be possible: If `X` is incompatible with await then `X` is probably bounded by an extension type that doesn't implement `Future`, and then it probably isn't possible for `B` to be a type that (1) does not derive a future type, and still (2) is compatible with await. In particular, I think `B` _must_ be an extension type (unless it's a bottom type, but if we promote to a bottom type then the code is unreachable), and an extension type is not compatible with await unless it implements `Future`, and in that case it also derives a future type.

eernstg (commented):
Again some tricky corners to consider. ;-)

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t12.dart
```diff
@@ -0,0 +1,31 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=inline-class
+
+extension type NumET(num _) implements num {}
+
+test<X extends NumET>(X x) async {
+  if (x is int) {
+    await x;
+  }
```
eernstg, line 26: We could probably do something like this:

```dart
extension type E1(Object _) {} // Incompatible with await.
extension type E2(Future<int> _) implements E1, Future<int> {} // Compatible.

Future<void> foo<X extends E1>(X x) async {
  E1 y = x;
  // await y; // Error, reported today.
  await x; // Error, `E` is not compatible with await. Not reported.
  if (x is E2) { // `x` has type `X & E2`.
    await x; // OK, result has type `int`.
  }
}

void main() async {
  await foo(E1(Future.value(1)));
}
```

(By the way, I just tried a completely fresh CFE, it _does_ report an error for the line marked 'Not reported'. Moving fast. ;-)

eernstg (commented):
One more comment.

LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t12.dart
sgrekhov: Ah, yes. Indeed

sgrekhov:
I removed `static_analysis_extension_types_A08_t12.dar`. Indeed we can't do this. So, only update of asserts left.
```dart
extension type E1(Object _) {} // Incompatible with await.
extension type E2(Future<int> _) implements E1, Future<int> {} // Compatible.

Future<void> foo<X extends E1>(X x) async {
  E1 y = x;
  // await y; // Error, reported today.  <- this checked in `static_analysis_extension_types_A08_t01.dart`
  await x; // Error, `E` is not compatible with await. Not reported. <- checked in `static_analysis_extension_types_A08_t07.dart`
  if (x is E2) { // `x` has type `X & E2`.
    await x; // OK, result has type `int`. <- checked in `static_analysis_extension_types_A08_t11.dart`
  }
}

void main() async {
  await foo(E1(Future.value(1)));
}
```
So, I think, there is no need to add the test above. Please take another look

eernstg:
OK, updated asserts is a safe bet: Landing this.

However, I'm still wondering: Do we have a test that covers the case where we can a `await e` where `e` has type `X & B` where `B` is an extension type which is compatible with await, but `X` isn't (because the bound of `X` is an extension type that is not compatible)?

```dart
// Like this.
extension type E1(Object _) {} // Incompatible with await.
extension type E2(Future<int> _) implements E1, Future<int> {} // Compatible.
```

This is a tricky case because an algorithm that searches for an extension type that doesn't implement `Future` could search and find `E1` and then conclude that `X & E2` is incompatible with await, even though we're supposed to use the check on `E2` directly (and ignore that we have a "bad" type in the superinterface graph).

eernstg (approved):
LGTM

sgrekhov:
> OK, updated asserts is a safe bet: Landing this.
> 
> However, I'm still wondering: Do we have a test that covers the case where we can a `await e` where `e` has type `X & B` where `B` is an extension type which is compatible with await, but `X` isn't (because the bound of `X` is an extension type that is not compatible)?
> 
> ```dart
> // Like this.
> extension type E1(Object _) {} // Incompatible with await.
> extension type E2(Future<int> _) implements E1, Future<int> {} // Compatible.
> ```
> 
> This is a tricky case because an algorithm that searches for an extension type that doesn't implement `Future` could search and find `E1` and then conclude that `X & E2` is incompatible with await, even though we're supposed to use the check on `E2` directly (and ignore that we have a "bad" type in the superinterface graph).

Please see https://github.com/dart-lang/co19/blob/5125c01a89fb018a335c74e900ca82e98d58c7c4/LanguageFeatures/Extension-types/static_analysis_extension_types_A08_t11.dart#L19-L37

eernstg:
Great, thanks!

PR #2520, opened 2024-02-05
Fix #2403. Add test for the new UP algorithm

### LINKED ISSUE #2403: Add tests for the new `UP` algorithm
Add tests according to the https://github.com/dart-lang/sdk/issues/54002

sgrekhov:


TypeSystem/Upper-lower-bounds/up_A01_t01.dart
```diff
@@ -0,0 +1,42 @@
...
+
+int? foo() => DateTime.now().millisecondsSinceEpoch.isEven? null : 1;
+
+void f<X extends num>(bool b, X x) {
+  if (x is int) { // `x` promoted to `X & int`.
+    var y1 = b ? x : null;
+    y1.isEven; // Error. Type of `y1` is `int?`
+//     ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
```
eernstg, line 30: The old (malformed) type was `(X & int)?`, so we should actually also test that the type is `int?` (not S?` for any other `S`). We could do that using `expectStaticType`.

`expectStaticType` wouldn't be quite safe, though, because it gives rise to erasure of intersection types (we can't pass `X & int` as a type argument to `expectStaticType`, explicitly or implicitly, so there's no way we can use `expectStaticType` to see the difference between `int?` and `(X & int)?)`. Actually, the reified type corresponding to `(X & int)?` isn't well-defined, because that type shouldn't exist in the first place.

In any case, we should also have something like this: `X? z = x;` and expect an error for that, to verify that the type of `x` is just `int?`, not `(X & int)?`.

TypeSystem/Upper-lower-bounds/up_A01_t01.dart
```diff
@@ -0,0 +1,42 @@
...
+
+    var y2 = b ? foo() : x;
+    y2.isEven;
+//     ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
```
eernstg, line 36: Here we have `UP(int?, X & int) == UP(int?, int) == UP(int, int)? == int?`, which can be tested using `expectStaticType`.

The old type was `UP(int?, X & int) == UP(int, X & int)? == UP(int, int)? == int?`. This means that we get the same result for a different reason , and it makes a lot of sense to check that we still get `int?`. `expectStaticType` could do that, `X? w =x;` would confirm that the result isn't of the form `X & ...`.

TypeSystem/Upper-lower-bounds/up_A01_t02.dart
```diff
@@ -0,0 +1,38 @@
...
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/expect.dart';
+
+int? foo() => DateTime.now().millisecondsSinceEpoch.isEven? null : 1;
+
+void f<X extends num>(X x) {
+  if (x is int) { // `x` promoted to `X & int`.
+    var y1 = 2 > 1 ? x : null;
+    Expect.isTrue(y1?.isEven);
```
eernstg, line 29: `expectStaticType<int?>` should succeed. We can't use `X? z = x;` because we don't want an error here, but the other test covers that property because the type of `x` and the promotion are identical.

Similarly for the case below.

eernstg (commented):
Looks good! I think we need to add a couple of test cases in order to recognize that the new type isn't of the form `(X & S)?`, but otherwise it's all good.

sgrekhov:

TypeSystem/Upper-lower-bounds/up_A01_t01.dart
eernstg: Thanks!

eernstg (commented):
LGTM!

(I added one `expectStaticType` invocation.)

eernstg (approved):

Issue #2522, opened 2024-02-06
Move type-related tests to `TypeSystem` folder


sgrekhov:
Move `LanguageFeatures/Subtyping`, `LanguageFeatures/nnbd/flow-analysis`, `LanguageFeatures/nnbd/type-normalization` to  `TypeSystem` folder.

PR #2526, opened 2024-02-07
#2522. Move type-normalization weak tests to TypeSystem dir

sgrekhov:
I checked `*.status` files there is no mentions of the old location of these weak tests. So, I believe it's safe to move them


eernstg (approved):
LGTM

PR #2528, opened 2024-02-08
Fix #2515. Remove strong-mode tests checking interaction with legacy

### LINKED ISSUE #2515: Fix LanguageFeatures/nnbd tests failures
Some of nnbd tests are failing. See https://dart-current-results.web.app/#/filter=co19/LanguageFeatures/nnbd
Need to triage these failures and fix them if needed

sgrekhov:
According to the https://github.com/dart-lang/sdk/issues/54805#issuecomment-1924195575 strong-mode tests checking interaction with legacy code removed. Deleted tests were failing on all platforms


eernstg (approved):
LGTM.

Just wondering: It looks like we still have libraries like exports_A02_opted_out_lib.dart that contain @dart=2.6, and they will cause errors if they are exported (and today: also if they are imported, we just can't have a 2.6 library in a Dart program at all at this time).

So the purpose of LanguageFeatures/nnbd/weak/exports_A02_t01.dart used to be "test that it is an error to export legacy declarations from an opted-in library", but the purpose _today_ is to get a compile-time error because we have a @dart=2.6 library at all. Right?

sgrekhov:
Today the purpose is the same but for weak mode only. These tests are simply not executed in a strong mode (and this is the only mode for the analyzer)

Issue #2529, opened 2024-02-08
Remove LibTest containg compile-time errors


sgrekhov:
API tests shoildn't check for compile-time errors

PR #2530, opened 2024-02-08
#2529. Remove API tests with compile-time errors from core and html folders

sgrekhov:
These tests doesn't check anything but the trivial fact that `null` cannot be assigned to a non-null variable. There tests were inherited from pre-nnbd times when they made sense. At that times they checked something like "returns 0 if value of this parameter is `null`". Then, during NNBD migration, an expected compile-time errors were added to this tests. But it's better simply delete them


eernstg (approved):
LGTM

In general, it would be great if we can find and delete tests that are redundant or useless, as a continuous maintenance task. Not that I think it's easy. ;-)

PR #2532, opened 2024-02-09
#2529. Fix and enable Iterable/forEach_A03_t01

sgrekhov:


eernstg (approved):
Fun stuff! The `forEach` parameter has type `void Function(E)` and we're passing a function literal of the form `(...) {... return predicate(e); }`, and the only reason why we don't get a compile-time error is that `predicate` has the return type `dynamic`, and it is OK to return `dynamic` to `void`. ;-)

Anyway, LGTM.

PR #2538, opened 2024-02-12
Fix #2485. Update`as` and `is` expressions tests with a function type

### LINKED ISSUE #2485: Add more constant evaluation tests

sgrekhov:


LanguageFeatures/Constructor-tear-offs/ambiguities_A36_t04.dart
```diff
@@ -0,0 +1,69 @@
...
+/// operator), but in the service of keeping our future options open, we choose 
+/// a design that does not rely on those restrictions. For example we omit most 
+/// infix operators from being "continuation tokens", even though they currently 
+/// cannot start a new expression, and therefore cannot follow a > infix 
+/// operator. This leaves us open to allowing some of those operators as prefix 
+/// operators in the future, like we currently allow the - operator.
+///
+/// @description Checks that any other token following the ambiguous > will make
+/// the prior tokens be parsed as comma separated < and > operator invocations.
+/// Test `as`, `as!` and `is` tokens
```
eernstg, line 49: We can't promise that the prior tokens are parsed as anything specific (the parse might very well just fail entirely), we can only promise that they will _not_ be parsed as a `<postfixExpression>` of the form `<primary> <selector>* <typeArguments>`.

In each specific case we may then determine that some other parse will succeed, or the term is a syntax error.

Language/Expressions/Constants/as_type_A01_t02.dart
```diff
@@ -74,7 +73,6 @@ main() {
 
   const f1 = foo as void Function();
   const f2 = as as void Function();
-  const f3 = bar<int> as int Function(int);
```
eernstg, line 76: Let's keep everything about `bar`: The declaration, this const, and the test in line 91.

There is nothing wrong with the test itself, we just can't express it in the given syntactic form. So we'll have to use this in order to make line 77 work, and the rest should then work just fine:

```
const f3 = (bar<int>) as int Function(int);
```

The term `bar<int>` in line 91 should be fine because it is followed by `,`.

eernstg (commented):
We might as well preserve the test cases and just fix the syntax error (I don't think there's a reason to drop the test itself, we just have to write it in a slightly different way).

Language/Expressions/Constants/as_type_A01_t02.dart
sgrekhov: Thank you. Updated

LanguageFeatures/Constructor-tear-offs/ambiguities_A36_t04.dart
sgrekhov: Description updated and more cases added

sgrekhov (commented):

eernstg (approved):
LGTM

PR #2545, opened 2024-02-14
#1400. Add cascade invocation test and more constants tests

sgrekhov:


LanguageFeatures/Extension-types/syntax_A08_t05.dart
```diff
@@ -0,0 +1,42 @@
...
+///
+/// <extensionTypeMemberDeclaration> ::= <classMemberDefinition>
+/// ...
+/// Some errors can be detected immediately from the syntax:
+///
+/// A compile-time error occurs if the extension type declaration declares any
+/// instance variables, unless they are external
+///
+/// @description Checks that it is a compile-time error if an extension type
+/// declaration declares a non-static constant.
```
eernstg, line 28: Good catch! An instance constant doesn't exist, but we should check that it isn't magically brought into existence with an extension type.

eernstg (approved):
LGTM

PR #2547, opened 2024-02-16
Fix #2546. Add constant context tests

### LINKED ISSUE #2546: Add constant context tests
Add tests for "17.3.2 Constant Contexts"

sgrekhov:


Language/Expressions/Constants/Constant_Contexts/constant_context_A01_t01.dart
```diff
@@ -0,0 +1,29 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Let e be an expression; e occurs in a constant context iff one of
+/// the following applies:
+/// • e is an element of a list or set literal whose first token is const, or e
+///   is a key or a value of an entry of a map literal whose first token is
+///   const.
+///
+/// @description Checks that an expression `e` is a constant expression if `e`
+/// is an element of a list whose first token is `const`. In this case `const`
```
eernstg, line 12: I think we should say 'a list literal', because that's a syntactic construct, a plain set could be the result of computations with basically _any_ syntax
```suggestion
/// is an element of a list literal whose first token is `const`. In this case the `const`
```
Same thing for 'map' in the next test.

Language/Expressions/Constants/Constant_Contexts/constant_context_A01_t01.dart
```diff
@@ -0,0 +1,29 @@
...
+  final val;
+  const C(this.val);
+}
+
+main() {
+  var c1 = const [C(1)];
+  var c2 = const {C(2)};
+
+  Expect.identical(const C(1), c1[0]);
+  Expect.identical(const C(2), c2.lookup(const C(2)));
```
eernstg, line 28: Perhaps `c2.first` would be clearer (this means that the reader doesn't have to check the documentation about `lookup` to know that it will definitely return the element from the set, not the argument ;-).

Language/Expressions/Constants/Constant_Contexts/constant_context_A01_t02.dart
```diff
@@ -0,0 +1,28 @@
...
+class C {
+  final val;
+  const C(this.val);
+}
+
+main() {
+  var c = const {C(1): C(2)};
+
+  Expect.identical(const C(2), c[const C(1)]);
+  Expect.identical(const C(1), c.keys.toList()[0]);
```
eernstg, line 27: Again, it's tempting to suggest `c.keys.first`.

Language/Expressions/Constants/Constant_Contexts/constant_context_A02_t01.dart
```diff
@@ -0,0 +1,27 @@
...
+/// @assertion Let e be an expression; e occurs in a constant context iff one of
+/// the following applies:
+/// ...
+/// • e occurs as @e in a construct derived from ⟨metadata⟩.
+///
+/// @description Checks that an expression `e` is a constant expression if `e`
+/// occurs as `@e` in a construct derived from ⟨metadata⟩
+/// @author sgrekhov22@gmail.com
+
+import 'dart:mirrors';
```
eernstg, line 14: Is there a way to say that mirrors are one of the `Requirements` such that this test will only be executed with configurations that have mirrors?

Language/Expressions/Constants/Constant_Contexts/constant_context_A03_t01.dart
```diff
@@ -0,0 +1,25 @@
...
+import '../../../../Utils/expect.dart';
+
+class C {
+  final val;
+  const C(this.val);
+}
+
+main() {
+  const c = const C(1);
+  Expect.identical(c, const C(1));
```
eernstg, line 24: So the question is whether `1` is treated as a constant expression (that is: whether or not it's canonicalized). But values of the type `int` are _always_ canonicalized, there is no way to allocate a new instance of `int` with the value 7, we can just obtain the unique one that everybody else will also obtain if they need 7.

So the argument to the constructor `C` should probably be `D()` where `D` is some other class that allows for constant as well as non-constant creation (or it could be `[]`, just so we don't have to create an extra class).

Language/Expressions/Constants/Constant_Contexts/constant_context_A05_t02.dart
```diff
@@ -0,0 +1,36 @@
...
+/// `e0` is a function literal
+/// @author sgrekhov22@gmail.com
+
+class C {
+  final val;
+  const C(this.val);
+}
+
+main() {
+  const c1 = (){C(1);};
```
eernstg, line 22: This is actually the initializing expression, not a subexpression of any other expression.

Perhaps?:
```suggestion
  const c1 = C((){});
```
`c2` needs a similar adjustment, but `c3` is fine (except that perhaps the `^^^` should be adjusted).

Language/Expressions/Constants/Constant_Contexts/constant_context_A06_t01.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A constant context is introduced in situations where an
+/// expression is required to be constant. This is used to allow the const
+/// modifier to be omitted in cases where it does not contribute any new
+/// information.
+///
+/// @description Checks that an initializers list is not a constant context
```
eernstg, line 10: Typo:
```suggestion
/// @description Checks that an initializer list is not a constant context
```

eernstg (commented):
Looks good! Suggested a couple of adjustments.

Language/Expressions/Constants/Constant_Contexts/constant_context_A01_t01.dart
sgrekhov: Indeed. Thank you!

Language/Expressions/Constants/Constant_Contexts/constant_context_A02_t01.dart
sgrekhov: No, there isn't. There tests are turned off in  `.status` files

Language/Expressions/Constants/Constant_Contexts/constant_context_A03_t01.dart
sgrekhov: Changed to `[]`

Language/Expressions/Constants/Constant_Contexts/constant_context_A05_t02.dart
sgrekhov: Changed to `C((){C(1);})`. According to the description `C(1)` must be a subexpression of a function literal, therefore I added it to the body

sgrekhov (commented):
Thank you. Updated, please take another look

Language/Expressions/Constants/Constant_Contexts/constant_context_A02_t01.dart
eernstg: OK, thanks!

eernstg (commented):
Looks good!

Just one thing&mdash;there was a long comment. It still says 'pending', so it was probably never sent, but I hope it will be sent now.

It's mostly about me being confused. ;-)

The point is that we have a rule that says "`e` is in a constant context if blah-blah", and then we have a test that says something like "assume `e` and not-blah-blah", and the consequence would be that `e` is not in a constant context. However, the test description seems to assume that it is an error, but there are of course lots of situations where an expression isn't in a constant context, and it's not an error.

On top of that, it _is_ an error for a function literal to be a subexpression of a constant expression, so we do get an error as claimed.

The whole thing is just a little unclear.

So please take a look at the comment 'Oh, I misread that!' and see if it makes sense to you. ;-)

sgrekhov:
I was in doubt if we need this text because a function literal in a constant context is an error itself. But eventually, I decided to add it because I was thinking that if we have this statement in the spec, that means there are some reasons to add it. No problem; I just deleted this test. Thank you.

eernstg (approved):
LGTM!

Language/Expressions/Constants/Constant_Contexts/constant_context_A02_t01.dart
a-siva, line 14: 'dart:mirrors' will not work in the VM AOT mode, this test needs to be skipped for AOT builds.

Also 'dart:mirrors' is an unsupported feature, not sure why we need to add new tests to test it's funcitonality

sgrekhov: @a-siva thank you! https://dart-review.googlesource.com/c/sdk/+/353522

Issue #2548, opened 2024-02-19
Add/ update c019 tests for "Statically checked declaration-site variance" feature


sgrekhov:
Specification https://github.com/dart-lang/language/issues/524

PR #2554, opened 2024-02-22
#2548. Add variance modifier declaration tests

sgrekhov:


TypeSystem/type-variance/declaration_A01_t05.dart
```diff
@@ -0,0 +1,34 @@
...
+/// for a type parameter declared by a static extension, a generic function
+/// type, a generic function or method, or a type alias.
+///
+/// @description Check that it is a compile-time error if a variance modifier is
+/// specified for a type parameter declared by an enum
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=variance
+
+enum  E1<in T> {e;}
```
eernstg, line 15: I do not think it should be an error to use statically checked variance here. Where does this rule come from? Perhaps the feature specification should be updated, if it says something that has this error as a consequence (it probably wasn't updated when generalized enums were added).

TypeSystem/type-variance/declaration_A01_t06.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if a variance modifier is specified
+/// for a type parameter declared by a static extension, a generic function
+/// type, a generic function or method, or a type alias.
+///
+/// @description Check that it is a compile-time error if a variance modifier is
+/// specified for a type parameter declared by a type alias
```
eernstg, line 10: Could this be a near-duplicate of TypeSystem/type-variance/declaration_A01_t02.dart?

eernstg (commented):

TypeSystem/type-variance/declaration_A01_t05.dart
sgrekhov: There is no such statement in the spec, I just added it as an experiment and then forgot to delete. Sorry.

TypeSystem/type-variance/declaration_A01_t06.dart
sgrekhov: It could. But let's have this test (at least we are checking non-function types here unlike `declaration_A01_t02.dart`)?

sgrekhov (commented):
Removed the accidental test. Please take another look

TypeSystem/type-variance/declaration_A01_t05.dart
eernstg: OK, thanks!

TypeSystem/type-variance/declaration_A01_t06.dart
eernstg: Good point, let's keep it!

eernstg (approved):
LGTM

eernstg (approved):
LGTM

PR #2555, opened 2024-02-23
#2548. Add tests for out parameter in a contrvariant position

sgrekhov:
Before switching to macroses just want to submit the current work


eernstg (commented):
Looks good!

One typo occurs in many places: 'contrvariant' should be 'contravariant'. Also, 'occurs in the contrvariant position' would probably be more reader-friendly if changed to 'occurs in a contravariant position' such that we don't imply that there is exactly one contravariant position in any given member signature, or, if there are several, that we've already agreed on which one it is.

sgrekhov:
Typo with article fixed. Please clarify where exactly " 'contrvariant' should be 'contravariant'"? I don't see such places

sgrekhov:
Found it, sorry

sgrekhov:
Fixed. Please take another look

eernstg (approved):
LGTM, thanks!

Issue #2559, opened 2024-03-05
Add co19 tests for augmentation libraries language feature


sgrekhov:
Spec https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md

sgrekhov:
Waiting for https://github.com/dart-lang/language/issues/4009 and https://github.com/dart-lang/language/issues/4061

PR #2560, opened 2024-03-05
#2559. Add tests checking that `augment` is a built-in identifier

sgrekhov:
According to the Augmentation libraries [specification](https://github.com/dart-lang/language/blob/main/working/augmentation-libraries/feature-specification.md)

> There is a new built-in identifier, augment, which is used to syntactically mark a declaration as an augmentation of an existing one.

### DIFF
--- a/Language/Expressions/Identifier_Reference/built_in_identifier_A26_t02.dart
+++ b/Language/Expressions/Identifier_Reference/built_in_identifier_A26_t02.dart
@@ -41,10 +41,28 @@ extension Ext<augment> on List {}
 // [analyzer] unspecified
 // [cfe] unspecified
 
+typedef int F1<augment>();
+//             ^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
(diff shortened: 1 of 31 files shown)


Language/Expressions/Identifier_Reference/built_in_identifier_A26_t02.dart
```diff
@@ -0,0 +1,50 @@
...
+void foo<augment>() {}
+//       ^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+extension Ext<augment> on List {}
+//            ^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
```
eernstg, line 43: We could also have a type variable declared by a type alias or by a function type:
```dart
typedef int F1<augment>();
typedef F2<augment extends Comparable<augment>> = int Function();

const void Function<augment>()? c = null;
```

sgrekhov:
Added these test cases to the other buil-in identifiers tests

eernstg (approved):
LGTM!

PR #2561, opened 2024-03-06
#2559. Add tests for augment libraries visibility

sgrekhov:


LanguageFeatures/Augmentation-libraries/defining_augmentation_A02_t02.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Augmentations do not share an import scope with the main library
+/// or each other. The libraries one augmentation imports are visible only to
+/// that file.
+///
+/// @description Checks that imports in an augmentation library are not visible
+/// in a main library even if an augmentation library exports it
```
eernstg, line 10: Do we have a test where an augmentation library 'aug_lib.dart' exports 'lib.dart' such that the augmented library 'auged_lib.dart' must implicitly export 'lib.dart', and then some other library 'main.dart' imports `auged_lib.dart' and must now have access to 'lib.dart' declarations?

The point is that we'd like to confirm that 'lib.dart' is _actually_ exported from 'auged_lib.dart', and not just ignored entirely.

eernstg (commented):
Looks good!

I would hope that we could use the augmented ("owner") library as the entry point in every case, because I'm arguing that it should just be an error to execute an augmentation library (https://github.com/dart-lang/language/issues/3642).

(That's a messy feature, anyway.)

Doesn't the test runner support detecting errors that occur in some other file than the library which is specified as "the library to execute" (i.e., the entry point)? Just checking, https://dart-review.googlesource.com/c/sdk/+/345541 has been merged. I would think that this makes it possible to use the augmented library as "the file to execute" every time.

LanguageFeatures/Augmentation-libraries/defining_augmentation_A01_t02.dart
```diff
@@ -0,0 +1,58 @@
...
+/// to all of the others, including private ones.
+///
+/// @description Checks that declarations in a main library are visible in an
+/// augmentation library
+/// @author sgrekhov22@gmail.com
+/// @issue 55103
+
+// SharedOptions=--enable-experiment=macros
+
+library augment 'defining_augmentation_A01_t02_lib.dart';
```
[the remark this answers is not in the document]
sgrekhov: Updated

LanguageFeatures/Augmentation-libraries/defining_augmentation_A02_t02.dart
sgrekhov: Added `defining_augmentation_A02_t06*` test

sgrekhov:
> Doesn't the test runner support detecting errors that occur in some other file than the library which is specified as "the library to execute" (i.e., the entry point)? Just checking, https://dart-review.googlesource.com/c/sdk/+/345541 has been merged. I would think that this makes it possible to use the augmented library as "the file to execute" every time.

Test runner doesn't support expecting errors in augmentation now. Therefore, some tests from this PR fails in spite of that augmentation has expected error it anyway reported as unexpected. Do we have any plans for supporting expected errors in augmentations?

sgrekhov:
Ok, now test runner does support expected errors in augmented libraries. This change is ready for review now

LanguageFeatures/Augmentation-libraries/defining_augmentation_A02_t04_lib1.dart
```diff
@@ -0,0 +1,17 @@
...
+///
+/// @description Checks that imports in an augmentation library are not visible
+/// in another augmentation library
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+library augment 'defining_augmentation_A02_t04.dart';
+
+import 'augmentation_libraries_lib.dart';
```
eernstg, line 17: Perhaps this was intended?:
```suggestion
import 'augmentation_libraries_lib.dart';
export 'augmentation_libraries_lib.dart';
```

eernstg: Perhaps use `AL` in this library, just to make sure that the import really occurs (importing into an augmentation library may be new code in the tools, so maybe they could do nothing for an import which isn't used).

LanguageFeatures/Augmentation-libraries/defining_augmentation_A02_t04_lib2.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Augmentations do not share an import scope with the main library
+/// or each other. The libraries one augmentation imports are visible only to
+/// that file.
+///
+/// @description Checks that imports in an augmentation library are not visible
+/// in another augmentation library even it exports it
```
eernstg, line 10:
```suggestion
/// in another augmentation library even if it exports it
```

LanguageFeatures/Augmentation-libraries/defining_augmentation_A02_t04.dart
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Augmentations do not share an import scope with the main library
+/// or each other. The libraries one augmentation imports are visible only to
+/// that file.
+///
+/// @description Checks that imports in an augmentation library are not visible
+/// in another augmentation library even it exports it
```
eernstg, line 10:
```suggestion
/// in another augmentation library even if it exports it
```

eernstg (commented):

One thing came up: I think it would be useful to use an imported namespace in one case, because there could be a kind of bug where unused import namespaces are sort of ignored, and then we could get a successful test run for the wrong reason. ;-)

sgrekhov:
Thank you! Updated. Please take another look

sgrekhov:
PR updated, `library augment...` replaced by `augment library...`. Please review

LanguageFeatures/Augmentation-libraries/defining_augmentation_A02_t04.dart
eernstg: Looks like this typo is still there.

LanguageFeatures/Augmentation-libraries/defining_augmentation_A02_t04_lib2.dart
eernstg: Looks like this typo is still there.

LanguageFeatures/Augmentation-libraries/defining_augmentation_A02_t04_lib1.dart
eernstg: I looked at LanguageFeatures/Augmentation-libraries/defining_augmentation_A02_t04.dart and concluded from the `@description` that there should be an export somewhere, and this was the only location where it would fit, as far as I could see.

On the other hand, the `@description` in this file does not mention any exports. Maybe the description should be adjusted in one or the other file?

eernstg (commented):
Looks good! I did notice that a couple of review comments were still lingering, so I looked at them again. Looks like they're still relevant, so it would probably be useful to take another look, perhaps I missed something.

sgrekhov:
Thank you. Updated

eernstg (approved):
LGTM

PR #2562, opened 2024-03-06
#2559. Add tests for augment libraries errors

sgrekhov:


sgrekhov:
This PR has the same issues as https://github.com/dart-lang/co19/pull/2561. So, converted it to a draft. Will bring it back to review after fixing the issues

sgrekhov:
Ready for review now

LanguageFeatures/Augmentation-libraries/defining_augmentation_A04_t01.dart
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if:
+/// ...
+/// - An augmentation library contains any part directives.
+///
+/// @description Checks that it is a compile-time error if an augmentation
+/// library contains any part directives.
```
eernstg, line 10: OK, this might change (Lasse wants to enable it, and I'd support that), but it is clearly specified to be an error at this time, so we'll go with that and then change the test if the rules are changed.

eernstg (approved):
LGTM

eernstg:
One little thing about the terminology: The entry point _is_ a library, and it needs to be a 'script' (in the sense that the language specification uses), that is, it must export a top-level function called `main` that accepts zero, one, or two positional arguments.

So it's not a problem to have a `main` function in an augmentation library, that function is going to be exported from the main library anyway. The only problem that came up was that the library with a name of the form `*_t0[1-9].dart' was an augmentation library, and it is likely to become an error to execute an augmentation library.

PR #2563, opened 2024-03-07
Fixes #2558. Remove `inline-class` experimental flag

### LINKED ISSUE #2558: Remove `inline-class` experimantal flag from co19 tests
Remove `inline-class` experimantal flag from co19 tests

sgrekhov:


eernstg (approved):
LGTM

PR #2564, opened 2024-03-08
#2559. Add applying augmentation library tests

sgrekhov:


LanguageFeatures/Augmentation-libraries/applying_augmentation_library_A04_t01.dart
```diff
@@ -0,0 +1,23 @@
...
+/// ...
+/// - The main library and its augmentations do not all have the same language
+///   version. There is only one user-visible library at the end, and it should
+///   have a consistent version across its entire surface area.
+///
+/// @description Check that it is a compile-time error if an augmentation
+/// library has version lower than main.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
```
eernstg, line 15: Is it even possible to use different language versions, are there several versions where augmentation is supported?

If we do have enough usable language versions then I think it would be safer to make the language version explicit in all participating libraries.

eernstg: Similarly for applying_augmentation_library_A04.

eernstg (commented):
Looks good! I just have one question: Can we make the language versions explicit in those few tests that are about the language version? It seems slightly dangerous to rely on an implicit language version.

sgrekhov:
Specified language version `3.4`. Now the tests check versions `3.3` vs. `3.4`. Please take another look

PR #2565, opened 2024-03-11
#2559. Add augmenting declarations tests

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_declarations_A01_t01.dart
```diff
@@ -0,0 +1,22 @@
...
+/// - An augmenting declaration has no corresponding original declaration to
+///   apply to.
+///
+/// @description Checks that it is a compile-time error if an augmenting
+/// declaration has no corresponding original declaration to apply to
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment class C {}
```
eernstg, line 15: If the language team (1) accepts the proposal that I hinted at in https://github.com/dart-lang/language/issues/3647, or (2) tells me that we already have such a rule then this would be an error simply because it isn't in an augmentation library.

So we ought to have a test where it does occur in an augmentation library, and still there is no original declaration to augment.

eernstg: (This might be covered by tests further down, I haven't looked at them all).

LanguageFeatures/Augmentation-libraries/augmenting_declarations_A01_t03.dart
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if:
+/// - An augmenting declaration has no corresponding original declaration to
+///   apply to.
+///
+/// @description Checks that it is not an error if an augmenting
+/// declaration do has a corresponding original declaration to apply to
```
eernstg, line 10: Perhaps?:
```suggestion
/// declaration has a corresponding original declaration to apply to
```

My understanding is that it is still an error for the augmenting declaration to occur before (textually) the corresponding original declaration in the same library. When in different libraries, the original one just needs to occur in a library which is earlier than the one that contains the augmenting declaration, according to the pre-order depth-first traversal of the `import augment` tree.

LanguageFeatures/Augmentation-libraries/augmenting_declarations_A02_t01.dart
```diff
@@ -0,0 +1,25 @@
...
+/// @description Checks that it is a compile-time error if an augmenting
+/// declaration appears before non-augmenting one. Test augmented declaration in
+/// a main library
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'augmenting_declarations_A02_t01_lib.dart';
+
+/**/augment class C {}
```
eernstg, line 19: If they accept #3647 then this would be a different error, in which case the description would need to be adjusted.

eernstg (commented):
Looks good!

`LanguageFeatures/Augmentation-libraries/merge_order_A01_t01.dart` wouldn't be testable for a while because `augmented` isn't likely to be implemented during the next couple of months.

But the merging order could be tested without `augmented` by having the same graph (main plus lib1, lib2, lib3) and letting main introduce class A, lib1 augment class A and introduce class B, lib2 augment class B and introduce class C, and finally lib3 augment class C.

This could be done without using `augmented` anywhere, and it would test the ordering.

sgrekhov:
Merge order tests updated and moved to https://github.com/dart-lang/co19/pull/2566. This one is blocked by https://github.com/dart-lang/language/issues/3647 for now

eernstg:
> This one is blocked by https://github.com/dart-lang/language/issues/3647 for now

I'm not sure we have a clear policy, but perhaps such PRs usually have a 'Draft' status?

LanguageFeatures/Augmentation-libraries/augmenting_declarations_A02_t01.dart
sgrekhov: > If they accept #3647 then this would be a different error, in which case the description would need to be adjusted.

#3647 was accepted. Why this is a different error now?

eernstg:
I think we can push this one ahead: https://github.com/dart-lang/language/issues/3647 has been resolved according to https://github.com/dart-lang/language/issues/3647#issuecomment-2145621278, and the result is that it _is_ allowed to have an augmenting declaration in a file which is not a part, that is, we can have it in the library as well as any of its direct or indirect parts.

sgrekhov:
Marked this one as "Ready for review"

LanguageFeatures/Augmentation-libraries/augmenting_declarations_A01_t01.dart
eernstg: It is now OK to have an augmenting declaration in a library, so I'll close this thread.

LanguageFeatures/Augmentation-libraries/augmenting_declarations_A02_t01.dart
eernstg: The outcome of #3647 would make a difference for the error message (we'd have "error: augmenting declaration in library" if rejected and "error: augmenting declaration before corresponding original declaration" if accepted).

But I got it wrong: The description fits the 'accepted' case, which is the one that we got.

eernstg (commented):
Closing threads.

eernstg (commented):
Perhaps this one was never submitted (I just wrote a comment, but forgot to press "Review changes")?

Anyway, here we go.

sgrekhov:
Updated. PTAL

eernstg (approved):
LGTM

PR #2568, opened 2024-03-13
#2559. Add augmenting types tests. Part 1

sgrekhov:


eernstg (approved):
LGTM!

I created https://github.com/dart-lang/co19/issues/2574 to mention a test case which might be missing.

PR #2579, opened 2024-03-26
Fixes #2576. Replace `library augment` by `augment library`

### LINKED ISSUE #2576: Some changes are needed in augmentation library tests
It is now `augment library <uri>;` rather than `library augment <uri>;`, so there will be unintended syntax errors basically in every test. Also, some `enum` declarations have an empty body, they must have at least one enum value (so that's a syntax error as well).

sgrekhov:


eernstg (approved):
LGTM

PR #2581, opened 2024-03-27
Fixes #2574. Add more augmenting types tests

### LINKED ISSUE #2574: [augmentation-libraries] Cover success case for aliased augmenting declarations
PR https://github.com/dart-lang/co19/pull/2568 contains a test LanguageFeatures/Augmentation-libraries/augmenting_types_A01_t02.dart checking that we get the expected compile-time errors in the case where a type alias of a type declaration of a mismatched kind is augmented.

Do we have a test which is testing the same situation, except that the type aliases denote types of the same kind? (That is, the case where we should not get an error because there is no kind mismatch, but there might still be an error because a tool mistakenly treats the type alias as a new and different kind and fails to see that the type alias denotes a declaration of the right kind).

sgrekhov:


eernstg (approved):
LGTM

LanguageFeatures/Augmentation-libraries/augmenting_types_A01_t03_lib.dart
```diff
@@ -0,0 +1,28 @@
...
+
+augment class CAlias {
+  String foo() => "C";
+}
+
+augment mixin MAlias {
+  String foo() => "M";
+}
+
+augment enum EAlias {
```
eernstg, line 26: I think this one needs a value.

eernstg (commented):
One fix needed, after all.

LanguageFeatures/Augmentation-libraries/augmenting_types_A01_t03_lib.dart
sgrekhov: Where it is required? I can't find this statement in the spec. Should it be a new value or an existing one?

eernstg: I think this was clarified at the meeting: There's no support for additional syntax in the feature specification.

eernstg:
Concluded at the meeting: `augment enum EAlias` needs a value in order to be a syntactically correct `enum` declaration.

sgrekhov:
Thank you. Fixed

LanguageFeatures/Augmentation-libraries/augmenting_types_A01_t03_lib.dart
```diff
@@ -0,0 +1,29 @@
...
+  e1;
```
eernstg, line 27: This is going to be interesting! ;-)

I would expect that this is an error because there is an `e1` in `E` already (so it would fail when we try to add another declaration named `e1`).

However, the feature specification does use a construct which would be `augment e1` in this case, which would then be implied to be OK.

eernstg (commented):
I can see in the feature specification that we can do `augment e1` in an `augment enum` declaration in the case where the corresponding `enum` declaration already has a value named `e1`.

I do think we should get an error from `enum E { e1 }` and `augment enum E { e1 }` (which would include the `EAlias` case as well).

LanguageFeatures/Augmentation-libraries/augmenting_types_A01_t03_lib.dart
sgrekhov: Changed to `augment e1;`

eernstg (approved):
LGTM

PR #2582, opened 2024-03-28
#2559. Add augmenting types tests. Part 5

sgrekhov:
This PR doesn't include test for augmenting unnamed extension (`augment extension {...}`). I'm expecting it to be a compile-time error. But probably a separate statement will be added for this case. Anyway I'll add an appropriate test later


eernstg (approved):
LGTM

PR #2587, opened 2024-04-03
#2559. Rename and regroup augmentation types tests

sgrekhov:
Just rename existing tests. Sometimes understanding of how the tests should be organized appears when the work is in progress.  ` All regular rules apply after this appending process` assumes that we should check at least bunch of base cases. It'll be much easier to mantain if we'll have `_A06_` for `extends`, `_A07_` for `implements` etc. Sorry for inconvinience


eernstg (approved):
LGTM

eernstg:
> Sometimes understanding of how the tests should be organized appears when the work is in progress.

Of course, and thanks for taking this kind of step when needed!

PR #2591, opened 2024-04-04
#2559. Add more augmenting types tests for `extends` and `implements` clauses

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_types_A07_t07.dart
```diff
@@ -0,0 +1,48 @@
...
+/// @assertion A class, enum, extension type, or mixin augmentation may specify
+/// extends, implements, on, and with clauses (when generally supported). The
+/// types in these clauses are appended to the original declarations clauses of
+/// the same kind, and if that clause did not exist previously then it is added
+/// with the new types. All regular rules apply after this appending process, so
+/// you cannot have multiple extends on a class, or an on clause on an enum, etc
+///
+/// @description Checks that it is a compile-time error if a class, extension
+/// type, mixin or enum augment specifies an interface in an `implements` clause
+/// which is already exists
```
eernstg, line 14:
```suggestion
/// which already exists
```

LanguageFeatures/Augmentation-libraries/augmenting_types_A07_t07_lib.dart
```diff
@@ -0,0 +1,44 @@
...
+/// @assertion A class, enum, extension type, or mixin augmentation may specify
+/// extends, implements, on, and with clauses (when generally supported). The
+/// types in these clauses are appended to the original declarations clauses of
+/// the same kind, and if that clause did not exist previously then it is added
+/// with the new types. All regular rules apply after this appending process, so
+/// you cannot have multiple extends on a class, or an on clause on an enum, etc
+///
+/// @description Checks that it is a compile-time error if a class augment
+/// specifies an interface in an `implements` clause which is not compatible
+/// with the superclass
```
eernstg, line 14: Perhaps:
```suggestion
/// specifies an interface in an `implements` clause which already exists
```

LanguageFeatures/Augmentation-libraries/augmenting_types_A07_t11_lib.dart
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
```
eernstg, line 1: Perhaps this entire augmentation library needs some updates? Looks like the description should be identical to the description in the corresponding non-`_lib` companion, and `A` should be defined somewhere.

(or, maybe: `A` should simply have been `I`?)

LanguageFeatures/Augmentation-libraries/augmenting_types_A07_t08_lib.dart
```diff
@@ -0,0 +1,41 @@
...
+/// @assertion A class, enum, extension type, or mixin augmentation may specify
+/// extends, implements, on, and with clauses (when generally supported). The
+/// types in these clauses are appended to the original declarations clauses of
+/// the same kind, and if that clause did not exist previously then it is added
+/// with the new types. All regular rules apply after this appending process, so
+/// you cannot have multiple extends on a class, or an on clause on an enum, etc
+///
+/// @description Checks that it is a compile-time error if a class augment
+/// specifies an interface in an `implements` clause which is not compatible
+/// with existing interfaces
```
eernstg, line 14: Perhaps copy the description from LanguageFeatures/Augmentation-libraries/augmenting_types_A07_t08.dart? It looks like that one was updated at some point in time, and this one wasn't.

LanguageFeatures/Augmentation-libraries/augmenting_types_A07_t11.dart
```diff
@@ -0,0 +1,42 @@
...
+/// @assertion A class, enum, extension type, or mixin augmentation may specify
+/// extends, implements, on, and with clauses (when generally supported). The
+/// types in these clauses are appended to the original declarations clauses of
+/// the same kind, and if that clause did not exist previously then it is added
+/// with the new types. All regular rules apply after this appending process, so
+/// you cannot have multiple extends on a class, or an on clause on an enum, etc
+///
+/// @description Checks that it is a compile-time error if a class, mixin or
+/// enum augment specifies an interface in an `implements` clause which is not
+/// compatible with existing members
```
eernstg, line 14: Extension type?

eernstg (commented):

LanguageFeatures/Augmentation-libraries/augmenting_types_A07_t07.dart
sgrekhov: Thank you. Updated

LanguageFeatures/Augmentation-libraries/augmenting_types_A07_t07_lib.dart
sgrekhov: Updated

LanguageFeatures/Augmentation-libraries/augmenting_types_A07_t08_lib.dart
sgrekhov: Yes, sometimes I'm forgetting to update description in `_lib`. Thank you.

LanguageFeatures/Augmentation-libraries/augmenting_types_A07_t11.dart
sgrekhov: Added `augmenting_types_A07_t12.dart` with another expected result

LanguageFeatures/Augmentation-libraries/augmenting_types_A07_t11_lib.dart
sgrekhov: `A` changed to `I` and description updated

sgrekhov (commented):

LanguageFeatures/Augmentation-libraries/augmenting_types_A07_t07_lib.dart
```diff
@@ -0,0 +1,44 @@
...
+//            ^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+augment abstract class C2 implements I {}
+//                     ^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+augment extension type ET implements I {}
```
eernstg, line 31: I'm not sure what we can do about this:

With `extension type ET` we have a syntax error (it needs to have a `(T name)` both without and with the keyword `augment`). However, if we include something like `(int i)` here to make it

```dart
augment extension type ET(I _) implements I {}
```

then we'd introduce a different error: Both the initial declaration of `ET` and the augmenting declaration of `ET` introduce a final instance variable `final I _` and a constructor `ET(this._)`, and both of those facts causes a compile-time error.

I created https://github.com/dart-lang/language/issues/3694 to get this issue on the table.

I think it would be useful to just postpone the issue right now and delete this particular test case, and then we can add it later.

eernstg (commented):
I discovered one issue after all:

```dart
augment extension type ET implements I {}
```

is a syntax error (which means that we probably never get around to testing the thing which is under test here, namely that `implements I` creates a duplicate in the implements clause which is an error).

sgrekhov:
Forgot about `augment enum` syntax. Fixed. Please review

eernstg (commented):
Looks good!

I think we can land this as soon as the example related to https://github.com/dart-lang/language/issues/3694 have been taken out (I don't think we should pretend to know the outcome of that issue).

eernstg (approved):
LGTM

PR #2595, opened 2024-04-08
#2559. Update/add new defining augmentation tests

sgrekhov:


LanguageFeatures/Augmentation-libraries/defining_augmentation_A08_t01.dart
```diff
@@ -0,0 +1,23 @@
...
+///   a library
+///
+/// @description Checks that it is a compile-time error if an `import` refers to
+/// an entity which is not a library
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'defining_augmentation_A08_t01_lib.dart';
+import 'defining_augmentation_A08_t01_lib.dart';
```
eernstg, line 17: We do get a warning from some tools if we have multiple `import` directives with the same URI. Hence, it's actually not obvious what to do when we have an `import` and an `import augment` with the same name.

We could use an import prefix with the `import` directive (or one prefix with the `import` directive and another prefix with the `import augment` directive), this should eliminate any such warnings (and we don't need to think about the question "should we have a warning for that or not?" which is good because this is only about importing a library augmentation, not about clashes between `import` and `import augment` directives) .

We could also have a 3 file configuration: `other.dart` imports `augment.dart`, `augment.dart` has a `augment library 'main.dart';` directive, and `main.dart` has an `import augment 'augment.dart';` directive.

This means that the relationship between `main.dart` and `augment.dart` is as usual, and the third one (`other.dart`) would be the entry point. This test would focus very cleanly on making sure that it is an error to import a library augmentation.

LanguageFeatures/Augmentation-libraries/defining_augmentation_A08_t03.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if:
+/// ...
+/// - An export or import (not import augment) refers to an entity which is not
+///   a library
+///
+/// @description Checks that it is a compile-time error if an `import` refers to
```
eernstg, line 10:
```suggestion
/// @description Checks that it is a compile-time error if an `export` refers to
```

LanguageFeatures/Augmentation-libraries/defining_augmentation_A08_t01.dart
sgrekhov: Makes sense. Updated, thank you

LanguageFeatures/Augmentation-libraries/defining_augmentation_A08_t03.dart
sgrekhov: Fixed

LanguageFeatures/Augmentation-libraries/defining_augmentation_A08_t04_lib2.dart
```diff
@@ -0,0 +1,21 @@
...
+///
+/// @description Checks that it is a compile-time error if an `export` refers to
+/// an entity which is not a library
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'defining_augmentation_A08_t04.dart';
+
+import 'defining_augmentation_A08_t04_lib1.dart';
```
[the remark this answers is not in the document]
sgrekhov: Fixed, thanx

sgrekhov (commented):
Thank you. Updated. Please take another look

eernstg (approved):
LGTM

PR #2597, opened 2024-04-09
#2559. Make enum augmenting types tests stronger

sgrekhov:
Add enum values to avoid false-positive outcome because of missing values


eernstg (approved):
LGTM

PR #2600, opened 2024-04-11
#2559. Add extensions to augmenting types tests

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_types_A10_t02_lib.dart
```diff
@@ -49,3 +49,23 @@ augment enum E {
...
+  int get foo => 42;
+//        ^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  void set bar(String _) {}
+//         ^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  int operator -(int other) => other;
```
eernstg, line 63: Same thing here: No compile-time errors here, but this operator `-` is not so easy to call outside `ExtA`.

LanguageFeatures/Augmentation-libraries/augmenting_types_A10_t02_lib.dart
```diff
@@ -49,3 +49,23 @@ augment enum E {
 // [analyzer] unspecified
 // [cfe] unspecified
 }
+
+augment extension ExtA {
+  int get foo => 42;
```
eernstg, line 54: I wouldn't expect an error here, we just can't call it outside `ExtA` using implicit resolution by referring to the name `foo` because that name is considered to denote `A.foo`.

We could still invoke it using `ExtA(A()).foo`.

LanguageFeatures/Augmentation-libraries/augmenting_types_A10_t04_lib.dart
```diff
@@ -7,9 +7,9 @@
 /// corresponding type in the augmented library. In other words, the
 /// augmentation can add new members to an existing type.
 ///
-/// @description Checks that static members defined in the body of an augment of
-/// a class, mixin or enum are added to a static namespace of the corresponding
-/// type in the augmented library
+/// @description Checks that instance members defined in the body of an augment
+/// of a class, mixin, extension, or enum are added to an instance namespace of
+/// the corresponding type in the augmented library
```
eernstg, line 12: Wouldn't that still be 'static'?
```suggestion
/// @description Checks that static members defined in the body of an augment
/// of a class, mixin, extension, or enum are added to the static namespace of
/// the corresponding type in the augmented library
```

eernstg: We could say that the augmented library contains the corresponding type in any case, semantically, but the actual declaration could also be in an augmentation library.

eernstg (commented):
Looks good.

A couple of locations are expecting a conflict that I don't recognize, PTAL.

sgrekhov:
Yes, missed these. Thank you. Fixed

eernstg (approved):
LGTM

PR #2601, opened 2024-04-11
#2559. Add tests for `augmented()` expression

sgrekhov:
Again, appropriate extension types tests will be added later in a separate PR

### DIFF
--- a/LanguageFeatures/Augmentation-libraries/augmenting_functions_A02_t01.dart
+++ b/LanguageFeatures/Augmentation-libraries/augmenting_functions_A02_t01.dart
@@ -56,7 +56,7 @@ main() {
   clearLog();
 
   Expect.equals("augment topLevelFunction2(A);", topLevelFunction2("A"));
-  Expect.equals("topLevelFunction1();Original v=A;augmented;", _log);
+  Expect.equals("topLevelFunction2(A);Original v=A;augmented;", _log);
   clearLog();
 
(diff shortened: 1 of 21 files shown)


eernstg:
Oh, I think we got the message a few weeks ago that `augmented` hasn't been implemented at all. Do you remember hearing anything like that? I think we should check out the status and hear what the tool teams are thinking.

@johnniwinther, @srawlins, is it fair to say that support for `augmented` is a front end / analyzer only feature?

If so, what do you think about having co19 tests for that topic at this time? Is there an implementation, perhaps partial? It looks like there is no implementation at all (`augmented()` elicits 'Method not found'), but I'm just guessing.

Do we need to wait landing those tests in order to avoid having a lot of failures with no other meaning than "we haven't even started implementing this yet"? We could also land them and approve all failures?

sgrekhov:
> Do you remember hearing anything like that?

Now I started remembering something about that. But my analyzer doesn't produce errors on `augmented()`. 
```
$ dart analyze ./LanguageFeatures/Augmentation-libraries/augmenting_functions_A02_t01_lib.dart
Analyzing augmenting_functions_A02_t01_lib.dart...
No issues found!
```
Looks like analyzer does support this syntax. CFE doesn't

sgrekhov:
Thank you. Typos fixed

eernstg (commented):
LGTM.

Johnni said it would not be a problem to start landing these tests, but I think we should wait a bit and hear the response from others, if any.

eernstg (approved):
LGTM

PR #2611, opened 2024-04-16
#2559. Add more augmenting functions tests

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_functions_A03_t01.dart
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion The augmenting function does not have to pass the same arguments
+/// to augmented() as were passed to it. It may call it once, more than once, or
+/// not at all.
+///
+/// @description Checks that it's not an error if augmenting function passes
```
eernstg, line 9: I think the following would be more natural (and in probably occurs in every library but I'll just mention it here):
```suggestion
/// @description Checks that it's not an error if an augmenting function passes
```

eernstg (commented):
Looks good!

There is just one tiny typo (same thing in every library), and I do think it would be nice to fix it before landing. ;-)

eernstg (approved):
Thanks! LGTM.

PR #2615, opened 2024-04-22
#2559. Add more augmenting functions tests

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_functions_A05_t02.dart
```diff
@@ -0,0 +1,137 @@
...
+///
+/// @description Checks that it is a compile-time error if a function
+/// augmentation specifies any default values.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'augmenting_functions_A05_t02_lib.dart';
+
+void topLevelFunction1([int i]) {}
```
eernstg, line 17: Perhaps it would be safer to use a parameter type like `int?` (and everywhere, for this test) such that it is not an error to have this declaration in the first place?

It seems like we could easily misinterpret the situation when we have both "errors that we are testing" and "errors that just happen because we've written something that is wrong and isn't being tested".

LanguageFeatures/Augmentation-libraries/augmenting_functions_A06_t01.dart
```diff
@@ -0,0 +1,50 @@
...
+/// declaration uses `augmented` when the original declaration has no concrete
+/// implementation.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'augmenting_functions_A06_t01_lib.dart';
+
+class C1 {
+  void foo();
```
eernstg, line 22: This is interesting!

We could say that nothing except syntax errors counts as a compile-time error in the main library or in any library augmentation, we're _always_ waiting to see whether there is an error in the final, merged library.

That would be sufficient to accept this declaration, assuming that it would receive an implementation from an augmentation. We could claim that this doesn't happen after all because the augmentation has an error; however, I think it makes sense to say that we do receive an implementation of `foo` from an augmentation, and it's a separate issue that it has an error because it uses `augmented`.

OK, that's good!

LanguageFeatures/Augmentation-libraries/augmenting_functions_A06_t01.dart
```diff
@@ -0,0 +1,50 @@
...
+}
+
+mixin M {
+  void foo();
+}
+
+enum E {
+  e1;
+  void foo();
```
eernstg, line 35: This looks weird, but it should be fine based on the same argument as in line 22.

LanguageFeatures/Augmentation-libraries/augmenting_functions_A06_t02.dart
```diff
@@ -0,0 +1,74 @@
...
+/// @assertion It is a compile-time error if:
+/// ...
+/// - An augmenting declaration uses augmented when the original declaration has
+///   no concrete implementation. Note that all external declarations are
+///   assumed to have an implementation provided by another external source, and
+///   they will throw a runtime exception when called if not.
+///
+/// @description Checks that it is not an error if an augmenting
+/// declaration uses `augmented` when the original declaration has no concrete
+/// implementation but is an `external` declaration
```
eernstg, line 14: I think it's fair to say that an external declaration is assumed to have an implementation. It's written in some other language, but it is assumed to exist (so, for example, it's ok to call an external instance member `foo` in a superinvocation like `super.foo()`).

It is a separate matter that the implementation can be something like `throw "Not bound to an implementation"` in the case where no implementation is provided in any other way.

Perhaps we should just talk about the fact that the declaration is external, and avoid saying anything about whether or not it has an implementation?

eernstg (commented):
Looks good, with a typo or two!

Just one proposal goes a bit further: It looks like LanguageFeatures/Augmentation-libraries/augmenting_functions_A05_t02.dart gives rise to some error messages that are expected (so they are treated correctly), but will also potentially cause some confusion.

sgrekhov:
Updated. Please teke another look

eernstg (approved):
LGTM!

LanguageFeatures/Augmentation-libraries/augmenting_functions_A06_t02_lib.dart
```diff
@@ -0,0 +1,60 @@
...
+///
+/// @description Checks that it is not an error if an augmenting
+/// declaration uses `augmented` when the original declaration is an `external`.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmenting_functions_A06_t02.dart';
+
+augment external void topLevelFunction() {
```
eernstg, line 20: One more thing! The grammar doesn't allow `augment external`.

I think it's supposed to be written like this:
```suggestion
augment void topLevelFunction() {
```
and it is not intended to be an error that the augmenting declaration doesn't have `external`.

In that sense, `external` plays a role which is similar to a body (so the original and the augmenting declarations can differ), not a modifier (where the original and augmenting declarations must be identical).

eernstg (commented):
One more thing came up.

LanguageFeatures/Augmentation-libraries/augmenting_functions_A06_t02_lib.dart
eernstg: Here's the error: https://github.com/dart-lang/language/blob/main/working/augmentation-libraries/feature-specification.md#:~:text=A%20declaration%20marked%20augment%20is%20also%20marked%20external.

sgrekhov:
Thank you for noticing! Fixed

eernstg (approved):
Thanks! LGTM.

PR #2618, opened 2024-04-24
#2559. Add augmented expression tests for setters

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t02.dart
```diff
@@ -0,0 +1,69 @@
...
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'augmented_expression_A02_t02_lib.dart';
+import '../../Utils/expect.dart';
+
+String _log = "";
+
+void set augmented(String value) {
+  throw Exception("Wrong setter is called");
```
eernstg, line 26: Nice!

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t06.dart
```diff
@@ -0,0 +1,95 @@
...
+/// being augmented, but it generally follows the same rules as any normal
+/// identifier:
+/// ...
+/// - Augmenting setters: Within an augmenting setter `augmented` must be
+///   followed by an `=` and will directly invoke the augmented setter. If
+///   augmenting a field with a setter, this will invoke the implicit setter
+///   from the augmented field.
+///
+/// @description Checks that within an augmented setter `augmented=` can be used
+/// to invoke appropriate setter in the current scope.
```
eernstg, line 15: It's confusing that the invocations in this library correspond to the description above, but the invocations in the library augmentation LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t06_lib.dart are invocations of the augmented declarations in this library.

Perhaps add one more sentence or so to explain this? (I guess it's still a good idea to keep the description identical in this library and in the corresponding library augmentation, but the description could say something like "in the library" vs. "in the augmentation" in order to make it unambiguous).

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t06.dart
```diff
@@ -0,0 +1,95 @@
...
+import '../../Utils/expect.dart';
+
+String _log = "";
+
+void set augmented(String value) {
+  _log = "Setter augmented: $value";
+}
+
+void set topLevelSetter(String value) {
+  augmented = value;
```
eernstg, line 30: Perhaps?:
```suggestion
  augmented = '$value, via original declaration';
```
in order to detect that the invocations performed in the augmentations will call this declaration, they will not call the top-level setter named `augmented` directly.

eernstg (commented):
Looks good! Just one thing: LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t06.dart seems to imply that `augmented = e` in an augmentation will invoke a top-level setter in scope; it should actually invoke the original declaration (which will in turn invoke that top-level setter named `augmented=`). I'm suggesting that the log is adjusted a bit such that we can see that the invocation goes via the original declaration, and the description is adjusted to reflect this behavior.

sgrekhov:

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t02.dart
eernstg: Thinking about this one more time, it's not obvious which is nicer for a developer who is encountering some test failures. The approach where line 26 assigns a value to `_log` which is not expected in some `Expect` statement is perhaps more consistent: We get to see what went wrong in both cases, but having a wrong `_log` value is more like several other buggy behaviors.

Anyway, I don't see a reason to change anything here now.

eernstg (approved):
LGTM. I left one thread open because I wanted to comment on it, but I don't think it's a very good idea to make changes because of that, so I'll just land this PR now.

PR #2625, opened 2024-04-29
Fixes #2624. Add more extension type tests for union types

### LINKED ISSUE #2624: Add more tests for extension types and union types
Add more co19 tests for extension types and union types. See examples in https://github.com/dart-lang/sdk/issues/55578

sgrekhov:
For subtyping tests it's enough to review `TypeSystem/subtyping/test_types/extension_type_A11.dart`. All other tests are generated ones

### DIFF
--- a/LanguageFeatures/Extension-types/nullability_A01_t05.dart
+++ b/LanguageFeatures/Extension-types/nullability_A01_t05.dart
@@ -1,0 +1,23 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion An extension type `V` is a proper subtype of `Object?`. It is
+/// potentially non-nullable, unless it implements `Object` or a subtype thereof
+///
+/// @description Checks that an extension type is assignable to a type `T?` when
+/// `T` is its representation type
+/// @author sgrekhov22@gmail.com
+
+import "../../Utils/expect.dart";
+
+extension type ET(int _) implements int {}
+
+main() {
+  int? et = null;
+  if (2 > 1) {
+    et = ET(1);
+  }
+  int x = et is int ? et : 2;
+  Expect.equals(1, x);
+}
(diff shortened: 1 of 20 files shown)


LanguageFeatures/Extension-types/nullability_A01_t05.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion An extension type `V` is a proper subtype of `Object?`. It is
+/// potentially non-nullable, unless it implements `Object` or a subtype thereof
+///
+/// @description Checks that an extension type is assignable to a type `T?` when
+/// `T` is its representation type
```
eernstg, line 9: Maybe add a few words:
```suggestion
/// `T` is its representation type and it implements `T` 
```
It wouldn't be true without `implements T`. (But `implements T` couldn't be there unless we had `T` or a subtype as the representation type, so it does make sense to mention the representation type as well.|

eernstg: On second thought, I actually think it's overkill to mention the representation type. We don't want to say anything about the representation type, we just want to ensure that `ET <: int`, and then the representation type just has to be something that doesn't make this an error.

So, perhaps:
```suggestion
/// it implements `T`
```

TypeSystem/subtyping/test_types/extension_type_A11.dart
```diff
@@ -0,0 +1,27 @@
...
+
+/// @assertion We say that a type T0 is a subtype of a type T1 (written
+/// T0 <: T1) when:
+/// Assume that DV declares an extension type declaration named Name with type
+/// parameters X1 .. Xs, and V1 is a superinterface of DV. Then Name<T1, .. Ts>
+/// is a subtype of [T1/X1 .. Ts/Xs]V1 for all T1, .. Ts.
+///
+/// @description Check that an extension type `ET<T1, ..., Ts>` is a subtype of
+/// an extension type `ET<X1, ..., Xs>` if `Ti` is a subtype of `Xi` for all
+/// `i`. Test union types.
```
eernstg, line 13: Wouldn't something like this be a better fit?
```suggestion
/// @description Check that an extension type `ET` that implements `S` is a
/// subtype of `FutureOr<S>`.
```

eernstg (commented):
Looks good, just a couple of comments.

(That was quick, too ;-)

sgrekhov:
Descriptions updated

eernstg (approved):
LGTM

PR #2626, opened 2024-04-29
#2559. Add augmenting expression tests for fields with no initializers

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmented_expression_A03_t07.dart
```diff
@@ -0,0 +1,63 @@
...
+/// initializer and the member being augmented is a nullable field with no
+/// initializer then `augmented` evaluates to `null`.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'augmented_expression_A03_t07_lib.dart';
+import '../../Utils/expect.dart';
+
+final String augmented = "Augmented variable";
```
eernstg, line 28: Perhaps use a value like `"Augmented variable, should not be used"` to help the reader? I know the test will fail as it should now, but a casual reader might need to stare at the code for a little while to discover what it's about. ;-)

eernstg (commented):
I can see that we have some late final variables here, and we will have to wait for some input from the language team, in response to https://github.com/dart-lang/language/issues/3735.

So perhaps we should just take the `late final` related tests out of this PR and do it later?

LanguageFeatures/Augmentation-libraries/augmented_expression_A03_t07.dart
sgrekhov: A very good idea! Thank you! Updated.

sgrekhov:
Updated. PTAL

eernstg (approved):
LGTM.

(I adjusted a couple of occurrences of the same situation as in the first test).

PR #2627, opened 2024-04-30
Fixes #2574. Expect an error in case of augmenting a type alias

### LINKED ISSUE #2574: [augmentation-libraries] Cover success case for aliased augmenting declarations

sgrekhov:


sgrekhov:
Thank you for the detailed explanation. Tests updated and hopefully now it's clearer what exactly is tested. Please take another look

LanguageFeatures/Augmentation-libraries/augmenting_types_A01_t03.dart
```diff
@@ -26,12 +25,8 @@ typedef CAlias = C;
 typedef MAlias = M;
 typedef EAlias = E;
```
eernstg, line 26: Is it true that these three declarations could just be removed? I don't see them in use anywhere.

eernstg (commented):
Looks good!

Just one thing: If I'm right that three declarations in LanguageFeatures/Augmentation-libraries/augmenting_types_A01_t03.dart are never used then they might as well be removed.

LanguageFeatures/Augmentation-libraries/augmenting_types_A01_t03.dart
sgrekhov: Right. Thank you! Removed

eernstg (approved):
LGTM

PR #2630, opened 2024-05-01
#2559. Add augmented expression tests for functions

sgrekhov:


sgrekhov:
> I don't see any tear-offs in this test. Perhaps it is a duplicate of some earlier test? I think we did have some tests previously to confirm that a local variable declaration named augmented is an error in an augmenting declaration.

Forgot to update (copy-paste caused error). Fixed. 

> By the way, did we check that a local function declaration is an error as well? And names declared by patterns? ;-)

That's cool! We didn't check it! Thank you for noticing! Added to this part, will add to other `augmented_expression_*` tests in a separate PR.

LanguageFeatures/Augmentation-libraries/augmented_expression_A04_t04_lib.dart
```diff
@@ -0,0 +1,107 @@
...
+/// @assertion The exact result of an `augmented` expression depends on what is
+/// being augmented, but it generally follows the same rules as any normal
+/// identifier:
+/// ...
+/// - Augmenting functions: When augmenting a function, augmented refers to the
+///   augmented function. Tear offs are not allowed, so this function must
+///   immediately be invoked.
+///
+/// @description Checks that it is a compile-time error to declare a local
+/// function named `augmented` inside of an augmenting function. Test a
```
eernstg, line 14: I guess this would be:
```suggestion
/// variable named `augmented` inside of an augmenting function. Test a
```
because pattern declarations introduce local variables.

eernstg: I haven't seen them all, yet, but I think this change is needed in all the pattern tests.

eernstg (commented):
Looks very good!

There is one word in several tests that I think should be changed, PTAL..

sgrekhov:
Typo fixed. PTAL

eernstg (approved):
LGTM

eernstg (approved):
LGTM

(added a couple of missing semicolons)

PR #2633, opened 2024-05-02
#2559. Add more augmented expression tests for local variables inside of getters and setters

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t11_lib.dart
```diff
@@ -0,0 +1,150 @@
...
+/// @description Checks that it is a compile-time error to declare a local
+/// function named `augmented` within an augmenting getter
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmented_expression_A01_t11.dart';
+
+augment String get topLevelGetter {
+  var augmented() => "x";
```
eernstg, line 21: Perhaps there's a point in using `var` because the grammar somehow allows it?

... looking ...

Nope, `⟨functionSignature⟩` is `⟨type⟩? ⟨identifier⟩ ⟨formalParameterPart⟩`, there is no way we can start with the token `'var'`.

Perhaps use `String` as the return type, so that we obviously don't get a spurious error because it's a type incorrect function?

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t11_lib.dart
```diff
@@ -0,0 +1,150 @@
...
+//         ^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  };
+  return "Augmented";
+}
+
+augment class C {
+  augment static String get staticGetter {
+    var augmented() => "x";
```
eernstg, line 36: Same issue. Probably a bunch of cases below, too.

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t13_lib.dart
```diff
@@ -0,0 +1,97 @@
...
+/// variable named `augmented` within an augmenting getter. Test a parenthesized
+/// pattern.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmented_expression_A01_t13.dart';
+
+augment String get topLevelGetter {
+  var (augmented) = (42);
```
eernstg, line 22: It should work exactly the same with `(42)` and with `42`, so either of those can be used.

It disturbs me slightly that the use of `(42)` as the initializing expression seems to spuriously imply that it is needed (perhaps because we've got some parentheses in the pattern "so we probably need parentheses on the right hand side as well").

Perhaps use `(42)` in some cases and `42` in others? ;-)

eernstg (commented):
Looks good!

One thing, I think we have to change `var augmented() => "SomeString";` which occurs many times, because `var` can't be the return type.

Then there's a real nit: In some cases where a parenthesized pattern is used in a declaration, the initializing expression is parenthesized as well. That looks like a misunderstanding, even though I do understand that it doesn't hurt anybody. Anyway, I wrote a comment on that, too. ;-)

sgrekhov:
Updated. PTAL

eernstg (approved):
Thanks, LGTM!

Issue #2631, opened 2024-05-01
New tests needed for wildcarded local variables and parameters


eernstg:
The [wildcards feature specification](https://github.com/dart-lang/language/blob/main/working/wildcards/feature-specification.md) is being implemented, so we need tests.

sgrekhov:
Closing this issue and go on as https://github.com/dart-lang/co19/issues/2641

PR #2639, opened 2024-05-06
#2631. Add tests checking access to local declaration named `_`

sgrekhov:


eernstg:
Just checking, is this about wildcards? In that case it wouldn't exactly be an error to access the value of a local declaration (or class type parameter declaration) named `_`, the point is that no such name is added to the namespace of that scope. This makes a difference if some other declaration in scope has the name `_` (say, an instance variable or a top-level declaration).

sgrekhov:
Yes, it's about wildcards. Tests for the case that you talking abiut now will be added later. They will cover "[Wildcards don't shadow](https://github.com/dart-lang/language/blob/main/working/wildcards/feature-specification.md#wildcards-do-not-shadow)" paragraph

LanguageFeatures/Wildcards/binding_A01_t01.dart
```diff
@@ -0,0 +1,201 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A local declaration whose name is `_` does not bind that name to
+/// anything.
+///
+/// @description Checks that it is a compile-time error to access the value of
+/// a local declaration named `_`. Test parameters of functions and methods.
```
eernstg, line 9: I think it's slightly dangerous to use this wording because it suggests that a local declaration introduces the name `_` into the enclosing scope, but then it is an error to find that entry in a lookup. Perhaps:
```suggestion
/// @description Checks that no entry named `_` is introduced into the
/// enclosing scope by a wildcarded declaration. Test parameters of
/// functions and methods.
```
This kind of wording would make it explicit that `_` isn't added to any namespace by the wildcarded declaration, and I think it's important to make that point explicitly everywhere.

I avoided saying 'local' because we have class type parameters as well as the declarations that we otherwise denote as 'local'. It should be enough to say that the declaration is wildcarded: this implies that the syntactic declared name is `_`, and also that it does not introduce a name into any scope.

LanguageFeatures/Wildcards/binding_A01_t01.dart
```diff
@@ -0,0 +1,201 @@
...
+  }
+  void instanceMethod2([int _ = 7]) {
+    print(_);
+//        ^
+// [analyzer] unspecified
+// [cfe] unspecified
+  }
+}
+
+extension type ET(int _) {
```
eernstg, line 108: We shouldn't have anything named `_` in scope:
```suggestion
extension type ET(int it) {
```

LanguageFeatures/Wildcards/binding_A01_t02.dart
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A local declaration whose name is `_` does not bind that name to
+/// anything.
+///
+/// @description Checks that it is a compile-time error to access the value of
+/// a local declaration named `_`. Test parameters of constructors.
```
eernstg, line 9: Same consideration as earlier `@description`.

eernstg: Same for several others.

LanguageFeatures/Wildcards/binding_A01_t02.dart
```diff
@@ -0,0 +1,88 @@
...
+    print(_);
+//        ^
+// [analyzer] unspecified
+// [cfe] unspecified
+    return C();
+  }
+}
+
+extension type ET<T>(int id) {
+  ET.constructor1(this,id, int _) {
```
eernstg, line 50: Probably:
```suggestion
  ET.constructor1(this.id, int _) {
```

LanguageFeatures/Wildcards/binding_A01_t08.dart
```diff
@@ -0,0 +1,120 @@
...
+  }
+  void instanceMethod<_>() {
+    print(_);
+//        ^
+// [analyzer] unspecified
+// [cfe] unspecified
+  }
+}
+
+extension type ET(int _) {
```
eernstg, line 65:
```suggestion
extension type ET(int it) {
```

eernstg (commented):
Looks good! Noted a couple of typos. I also think it would be helpful to future readers to be absolutely explicit about the fact that a wildcarded declaration does not introduce anything into the current scope.

sgrekhov:
Thank you Updated. Please take another look

eernstg:
Thanks!

eernstg (approved):
LGTM

PR #2640, opened 2024-05-06
Fixes #2638. Add extension methods test for type aliases

### LINKED ISSUE #2638: Add more tests that extension is not a type
There is no co19 tests checking that it is an error to create a type alias for extension (because it's not a type)

sgrekhov:


eernstg (approved):
LGTM

Issue #2643, opened 2024-05-07
Get rid of co19 tests using legacy code


sgrekhov:
We need to get rid of the test if it's only about _legacy_ code and not about unsound mode execution. If there is a need to obtain a value `null` with a non-nullable type `T` it should be possible to use a simple `null as T`. So it should be possible to simply change the legacy library to a more recent language version and then use things like `null as T` to do the job which was previously done by the legacy library

sgrekhov:
Completed. The new roll with these changes is in progress https://dart-review.googlesource.com/c/sdk/+/365821

PR #2644, opened 2024-05-07
#2643. Remove override checking tests in case a legacy class extending an opted-in class

sgrekhov:


eernstg (approved):
LGTM

eernstg:
Wow, that's great! It removes most of these tests from further consideration in one fell swoop.

PR #2646, opened 2024-05-07
#2643. Remove legacy tests from TypeSystem/type-normalization

sgrekhov:


eernstg (approved):
I think each of these tests can be deleted because it is dealing with types of the form `T*`. The two `*lib.dart` libraries are deleted because they are now unused.

With that in mind, LGTM!

PR #2650, opened 2024-05-08
#2643. Remove some of 2.6 libraries, update/remove tests using them

sgrekhov:


eernstg (commented):
Looks good!

However, I couldn't spot any real dependency on legacy types in a couple of cases, PTAL:

I can't see why the `List` length setter test LanguageFeatures/nnbd/weak/list_length_t02.dart shouldn't be valid today. Similarly for LanguageFeatures/nnbd/weak/list_length_t04.dart, using just unsound execution but no legacy code. Is that true?

I also got the impression that LanguageFeatures/nnbd/weak/nullability_definitions_A01_t08.dart could be updated to work the same, but not use legacy code. Or would that just duplicate some other test?

sgrekhov:
Thank you for the detailed review. `list_length_*` tests reverted and then updated.
`LanguageFeatures/nnbd/weak/nullability_definitions_A01_t08.dart` can be updated indeed but in this case it'll became a duplicate of `LanguageFeatures/nnbd/weak/nullability_definitions_A01_t06.dart`. Please take another look

eernstg (approved):
LGTM

PR #2652, opened 2024-05-08
#2643. Remove legacy types from subtyping tests

sgrekhov:


eernstg (approved):
I can see that all tests except TypeSystem/subtyping/static/test_types/right_object_A03.dart and TypeSystem/subtyping/static/test_types/right_object_fail_A05.dart are clearly concerned with legacy-only features (so it's fine to delete them).

Those remaining two tests, 'right_object_A03' and 'right_object_fail_A05', do not obviously involve any legacy-only features according to their description, but they do use a legacy type `X` to test the situation in the description. So it should be fine to delete those as well, and there will be other tests testing the same thing with purely null-safe tests.

TypeSystem/subtyping/static/test_types/left_null_fail_A04.dart is tricky because it says 'Check that otherwise, the query is false', and `T1` is `X`, and `X` is declared in a legacy library. That seems to match the case where `T1` is of the form `S*`.

Pondering this one more time, I think this is OK: A legacy class used as a type in a null-safe library is a null-safe type, not a legacy type.

OK, LGTM!

Issue #2641, opened 2024-05-06
[Wildcard Variables] co19 Tests


kallentu:
Write up tests for co19. "Possible Test Cases" mirrors the test list for https://github.com/dart-lang/sdk/issues/55652.
Examples pulled from [the spec](https://github.com/dart-lang/language/blob/main/working/wildcards/feature-specification.md).

## Possible Test Cases
May not be exhaustive.
- [ ] Multiple `_` are allowed in function parameters. This includes top-level functions, local functions, function expressions ("lambdas"), instance methods, static methods, constructors, etc. It includes all parameter kinds: simple, field formals, and function-typed formals, etc.:
```
Foo(_, this._, super._, void _(), {_}) {}

list.where((_) => true);
```
- [ ] Multiple `_` are allowed in local variable declaration statement variables.
- [ ] Multiple `_` are allowed in for loop variable declarations.
- [ ] Multiple `_` are allowed in catch clause parameters.
- [ ] Multiple `_` are allowed in generic type and generic function type parameters.
- [ ] Other declarations can still be named `_` as they are today: top-level variables, top-level function names, type names, member names, etc. are unchanged.
- [ ] `_` is still a wildcard in patterns
```
int a;
(_, a) = (1, 2); // Valid.
```
- [ ] Error if we cannot resolve `_` to a member or top-level declaration.
```
main() {
  _ = 1; // Error.
}

class C {
  var _;

  test() {
    _ = 2; // OK.
  }
}
```
- [ ] Wildcards do not shadow.
```
class C {
  var _ = 'field';

  test() {
    var _ = 'local';

    _ = 'assign'; // Assigns to the field and not the local.
  }
}
```
- [ ] A positional initializing formal named `_` does still initialize a field named `_ `(and you can still have a field with that name)
```
class C {
  var _;

  C(this._); // OK.
}
```
- [ ] `_` can't be accessed inside an initializer list.
```
class C {
  var _;
  var other;

  C(this._): other = _; // Error, cannot access `this`.
}
```
- [ ] The name `_` can be used in the body, but this is a reference to the field, not the parameter.
```
class C {
  var _;

  C(this._) {
    print(_); // OK. Prints the field.
  }
}
```
- [ ] Error to have two initializing formals named `_`.
```
class C {
  var _;
  C(this._, this._); // Error.
}
```
- [ ] `super._` will pass on the given actual argument to the corresponding superconstructor parameter.
```
class B {
  final _;
  B(this._);
}

[...]

PR #2656, opened 2024-05-13
#2641. Add tests for multiple declarations named `_`

sgrekhov:


LanguageFeatures/Wildcards/binding_A02_t02.dart
```diff
@@ -0,0 +1,77 @@
...
+  ET.constructor1(1, 2, "3");
+  ET.constructor2(1, 2, true, "3", 4, "5");
+  ET<int>.constructor3(1, 2, 3, 4);
+  const ET.constConstructor1("1", 2, true);
+  const ET.constConstructor2("1", 2, true, "3", 4, "5");
+  const ET<int>.constConstructor3(1, 2, null, 4);
+  ET.factoryConstructor1("1", 2, true);
+  ET.factoryConstructor2("1", 2, true, "3", 4, "5");
+  ET<String>.factoryConstructor3(1, "2", null, "4");
+}
```
eernstg, line 77: I think it would be good to cover redirecting generative constructors (`: this(...)`), too.

Another thing that would probably touch code which isn't covered by these tests is constructor tear-offs. The weird thing is that they would give rise to a situation where we are creating a function literal that calls the constructor, but we can't just reuse the parameter names of the constructor because that would yield a situation where the function literal parameters are wildcarded, and the occurrences of `_` as actual arguments in the invocation of the constructor would (probably) be undefined.

LanguageFeatures/Wildcards/binding_A02_t03.dart
```diff
@@ -0,0 +1,39 @@
...
+/// @description Checks that it is not an error to have a multiple local
+/// declarations named `_`. Test parameters of typedefs.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=wildcard-variables
+
+import '../../Utils/expect.dart';
+
+typedef int F1(_, int _, var _);
+typedef int F2(_, int _, var _, [_, int _ = 0, var _ = ""]);
```
eernstg, line 18: I think it's a compile-time error for any formal parameter in a type alias to have a default value.

eernstg: But `dynamic` is nullable, so it should be enough to just omit the default value.

LanguageFeatures/Wildcards/binding_A02_t03.dart
```diff
@@ -0,0 +1,39 @@
...
+typedef F5 = int Function(_, int _, var _, [_, int _ = 0, var _ = ""]);
+typedef F6 = int Function<T>(T _, T _, [T? _, T? _]);
+
+main() {
+  F1 f1 = (_, int _, var _) => 1;
+  F2 f2 = (_, int _, var _, [_, int _ = 0, var _ = ""]) => 2;
+  F3<int> f3 = (int _, int _, [int? _, int? _]) => 3;
+  F4 f4 = (_, int _, var _) => 4;
+  F5 f5 = (_, int _, var _, [_, int _ = 0, var _ = ""]) => 5;
+  F6<int> f6 = (int _, int _, [int? _, int? _]) => 6;
```
eernstg, line 31: Don't we already cover function literals in LanguageFeatures/Wildcards/binding_A02_t01.dart?

Ah, we probably have these because they are testing that the meaning of the type aliases is as intended. So that's good (and more informative than, say, `F1? f1 = null;`).

eernstg: Perhaps add a few words of explanation in a comment in line 26?

eernstg (commented):
Looks good! I think one location needs to change (it's an unwanted compile-time error), and it would be good to have a bit more on constructors.

LanguageFeatures/Wildcards/binding_A02_t02.dart
sgrekhov: Added tests for generative constructors.

As for the tear-offs, please clarify. What exactly is meant?
```dart
class C {
  C(_, int _);
}

main() {
  var f = C.new;
  f("a", 0); // What else can we check?
}
```

eernstg: Thanks!

About the tear-offs: The point is that they could be implemented in a way which is similar to the following:

```dart
class C {
  C(_ int _);
}

void main() {
  var f = (_, int _) => C(_, _); // Error!
}
```

and we just want to make sure the need to create a function object whose body is a constructor invocation doesn't get confused about wildcarded parameters.

So creating the tear-off and calling it should be fine!

LanguageFeatures/Wildcards/binding_A02_t03.dart
eernstg: Thx!

eernstg (commented):
Looks good! With some tear-off related test cases, it should be fine!

LanguageFeatures/Wildcards/binding_A02_t03.dart
eernstg: (I think the default values are still there, and also that the compile-time errors caused by those default values would be unwanted.)

sgrekhov:
Added `binding_A01_t11.dart`. PTAL

LanguageFeatures/Wildcards/binding_A02_t03.dart
sgrekhov: > (I think the default values are still there, and also that the compile-time errors caused by those default values would be unwanted.)

Probably you are looking at the outdated diff. I don't see any default values on line 18 and other lines except 28 and  31. But these default values should be Ok

LanguageFeatures/Wildcards/binding_A02_t02.dart
eernstg: Thanks, it looks like all those tear-offs are in LanguageFeatures/Wildcards/binding_A01_t11.dart.

LanguageFeatures/Wildcards/binding_A01_t11.dart
```diff
@@ -0,0 +1,403 @@
...
+class MA = Object with M;
+
+main() {
+  void localFunction1(int _) {}
+  void localFunction2([int _ = 12]) {}
+  var functionExpression1 = (int _) {};
+  var functionExpression2 = ([int _ = 13]) {};
+
+  var to1 = A.new;
+  var f1 = (_) {
```
eernstg, line 69: Hmmm, thinking... ;-)

I don't think all these function literals will test anything new, we already have other tests about function literals.

All we need to check for the tear-offs is that we can perform the tear-off and the tear-off can be called (and it is accepted at compile time as well as running without throwing at run time).

When I mentioned a function literal it was a reference to a possible implementation technique. The expression `A.new` evaluates to a function object at run time. A constructor is not a function object, and this means that the operation that tears off a constructor is an operation that creates a function object (from something that we don't need to know anything about, because the representation of a constructor as such is an implementation specific property).

However, with `A.new`, that function object would be very similar to the function objects that we can obtain by evaluating the function literal `(dynamic d) => A(d)`, and `A.constr` would be similar to `([int i = 1]) => A.constr(i)`.

However, I've renamed the variables here in order to make sure that the result has the right behavior. An implementation could (if someone forgot to think about it ;-) model those constructor tear-offs using the same variable names as in the declaration of the constructor:

If so, then `A.new` would be modeled by `(_) => A(_)`, which is an error. This could be an internal error at compile time, or it could be a run-time error.

All in all, we should have what we need by means of something like the following code:

```dart
void main() {
  ...
  var to1 = A.new;
  to1('Hello')..checkThatTheArgumentWasReceivedCorrectly;
  var to2 = A.constr;
  to2(15)..checkThatTheArgumentWasReceivedCorrectly;
  ...
}
```

I think this test could me adjusted to do this such that it has no compile-time errors, and such that we can see that the tear-off can be performed (no compile-time errors, no run-time errors), and the behavior of the constructor is as expected.

Most of it would amount to deleting a bunch of lines from this `main`, but it would be very good if it is also tested that the transfer of actual arguments from the function object to the constructor works as expected.

LanguageFeatures/Wildcards/binding_A02_t03.dart
eernstg: OK, thanks!

eernstg (commented):
Looks good! Just one thing remaining: The testing of constructor tear-offs will not reveal the specific faults that I mentioned, as far as I can see. I wrote a rather long comment about this, PTAL.

eernstg (approved):
LGTM

eernstg (approved):
LGTM (fixed a typo).

eernstg:
Issue about `super._`: https://github.com/dart-lang/language/issues/3792. There is no separate issue about `this._` as far as I know.

sgrekhov:
> (I don't remember, but if we already have some tests about those things then we should just make sure they don't create any noise, e.g., by causing tools to be changed such that they pass a test that may need to be changed.)

No, we don't have such tests yet. They will be added later as test for [Initializing formals](https://github.com/dart-lang/language/blob/main/working/wildcards/feature-specification.md#initializing-formals) part of the spec

eernstg:
Sounds good, thanks! The language team just decided that `super._` and `this._` are allowed (`super._` when it occurs as a declaration of a positional parameter, not a named one). `super._` gives rise to a forwarding semantics in that the actual argument passed to that parameter is also passed to the superconstructor (even though we couldn't write the code to do that because there is no name for that parameter in scope). `this._` will initialize an instance variable named `_`, and it is not possible to reference that parameter by means of `_` in the initializer list (but it could be a top-level or static thing, as usual).

PR #2657, opened 2024-05-13
#2119. Remove some accidental comments

sgrekhov:


eernstg (approved):
LGTM.

So `check` was misnamed, it should actually have been `dontCheck` for a few years? ;-)

sgrekhov:
Now it's named right (and even may fall on web) :)

PR #2669, opened 2024-05-16
#2641. Check that it is still an error to declare more than one class member named `_`

sgrekhov:


LanguageFeatures/Wildcards/unchanged_A05_t06.dart
```diff
@@ -0,0 +1,92 @@
...
+
+class C4 {
+  static int _ = 1;
+  static void set _(int v) {}
+//                ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+class C5 {
```
eernstg, line 47: Perhaps create a variant?: `C5Base` could have the instance member and we could have `class C5 extends C5Base` or `class C5 implements C5Base`. The point would be to verify that the error is detected also in the case where the instance member isn't in the same lexical scope.

LanguageFeatures/Wildcards/unchanged_A05_t07.dart
```diff
@@ -0,0 +1,89 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+class C4 {
+  static int get _ => 1;
+  static void set _(int v) {} // Ok
+}
+
+class C5 {
```
eernstg, line 44: Same here: We could change a couple of these cases such that the instance member is in a superinterface.

LanguageFeatures/Wildcards/unchanged_A05_t10.dart
```diff
@@ -0,0 +1,88 @@
...
+
+// SharedOptions=--enable-experiment=wildcard-variables
+
+class C1 {
+  int _ = 1;
+  static int _ = 1;
+//           ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
```
eernstg, line 21: Isn't this a duplicate of something like `C5` in LanguageFeatures/Wildcards/unchanged_A05_t06.dart?

eernstg: Perhaps the easiest way to avoid duplicates would be to have instance/static clashes in their own tests (like this one) and omit the instance/static clashes in those other tests (like A05_t06)? I'm not quite sure which approach is easier, but it does look like the current system will generate some duplicates.

(In some cases, e.g., augmenting declarations, it does make sense to test different orderings of elements; but in general we'll just have to assume that it has already been tested "somewhere else" that different textual orderings of declarations will give rise to the same name clash errors, otherwise everything will explode ;-).

eernstg (commented):
Looks good!

I suggested a small extension, in order to ensure that name clashes involving superinterfaces is covered as well.

I also noticed something that looks like a duplication of some test cases. I'm not quite sure how much duplication we have (or if there's actually a reason to have both that I didn't see), but PTAL.

sgrekhov:
A very good advices! Thank you! Implemented, PTAL

eernstg (approved):
Thanks, LGTM!

PR #2674, opened 2024-05-20
#2641. Check that it is still an error to declare more than one enum member named `_`

sgrekhov:
There are no "implements" tests because `enum E implements ContainsWildcardMethod` is an error itself without implementation of `_`

### DIFF
--- a/LanguageFeatures/Wildcards/other_declarations_A05_t26.dart
+++ b/LanguageFeatures/Wildcards/other_declarations_A05_t26.dart
@@ -1,0 +1,86 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Other declarations: top-level variables, top-level function
+/// names, type names, member names, etc. are unchanged. They can be named `_`
+/// as they are today.
+///
+/// @description Checks that it is a compile-time error to declare more than one
+/// member named `_`. Test an instance variable vs other enum members conflict.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=wildcard-variables
+
+mixin class ContainsWildcardMethod {
+  int _() => 0;
+}
+
+mixin class ContainsWildcardGetter {
+  int get _ => 0;
+}
+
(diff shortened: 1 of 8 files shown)


LanguageFeatures/Wildcards/other_declarations_A05_t26.dart
```diff
@@ -0,0 +1,86 @@
...
+  final int _ = 1;
+  int _() => 6;
+//    ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+enum E2With with ContainsWildcardMethod {
+  e0;
+  final int _ = 1;
```
eernstg, line 47: This one is unusual, it's an error because there is an incorrect override relationship (an implicitly induced getter would override a method). This means that there is no name clash, it's just a mismatch with respect to the kind of declaration (getter vs. method).

Anyway, I don't think there is anything wrong about that.

LanguageFeatures/Wildcards/other_declarations_A05_t27.dart
```diff
@@ -0,0 +1,86 @@
...
+  int get _ => 1;
+  int _() => 6;
+//    ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+enum E2With with ContainsWildcardMethod {
+  e0;
+  int get _ => 1;
```
eernstg, line 47: This is again an override kind error, not a name clash.

LanguageFeatures/Wildcards/other_declarations_A05_t28.dart
```diff
@@ -0,0 +1,92 @@
...
+  int get _ => 7;
+//        ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+enum E3With with ContainsWildcardGetter {
+  e0;
+  int _() => 1;
+//    ^
```
eernstg, line 62: Once again an override kind error, not a name clash.

eernstg (approved):
LGTM. In a couple of situations there is an override kind error rather than a name clash (e.g., a getter overrides a method), which means that it isn't an error to have _some_ declarations with the name `_` as written, it's just an error that one of them is a method and the other is a getter.

However, I don't think it's a problem, so I'll just land this now.

PR #2680, opened 2024-05-22
#2641. Add local functions test

sgrekhov:


eernstg (approved):
I guess this will fail because the warning hasn't been implemented, but that shouldn't matter much. LGTM.

PR #2681, opened 2024-05-22
#2641. Add import prefix tests

sgrekhov:


LanguageFeatures/Wildcards/imports_A01_t01.dart
```diff
@@ -0,0 +1,19 @@
...
+/// provides access to the imported library.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=wildcard-variables
+
+import 'imports_lib1.dart' as _;
+import '../../Utils/expect.dart';
+
+main() {
+  Expect.equals(42, theAnswer);
```
eernstg, line 18: That won't work. The import has prefix `_`, and an unqualified reference like `theAnswer` will just be an undefined identifier. The fact that the import has a wildcarded prefix doesn't mean that it doesn't have a prefix, it means that we can't denote the prefix even if we want to (you could say that it works just like a compiler-generated fresh name, `_prefix9283798`, and we don't know what it is so we can't use it).

This means that the only thing we can access for an import with a wildcarded prefix is extensions.

(You could say that the compiler _can_ use the secret prefix and turn `42.myExtensionGetter` into `_prefix9283798.MyExtension(42).myExtensionGetter`.)

LanguageFeatures/Wildcards/imports_lib1.dart
```diff
@@ -0,0 +1,16 @@
...
+///
+/// @description Checks that an import prefix named `_` is non binding and
+/// provides access to the imported library.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=wildcard-variables
+
+library imports_lib1;
+
+int theAnswer = 42;
```
eernstg, line 16: We'd need to have an extension here. Also, it would make sense to have a wildcarded import and an attempt to use `theAnswer` as if it had been imported without a prefix (`theAnswer`) and as if it had been imported with the prefix `_` (`_.theAnswer`), with the expectation that it is a compile-time error in both cases.

LanguageFeatures/Wildcards/imports_lib2.dart
```diff
@@ -0,0 +1,14 @@
...
+/// @assertion Import prefixes named `_` are non-binding and will provide access
+/// to the extensions in that library.
+///
+/// @description Checks that an import prefix named `_` is non binding and
+/// provides access to the imported library.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=wildcard-variables
+
+int one = 1;
```
eernstg, line 14: We should again have an extension here.

It would probably be a good idea to have an extension conflict, such that we can do `42.foo()` and get an error because `foo` is ambiguous on `int`, because we have an extension from `lib1` as well as an extension from `lib2`.

eernstg (commented):
Looks good, but needs a couple of adjustments!

sgrekhov:
Thank you! I misread the spec. Updated.
Please also note that there is a typo in the spec https://github.com/dart-lang/language/pull/3827

LanguageFeatures/Wildcards/imports_lib1.dart
eernstg: Thx!

eernstg (approved):
LGTM

PR #2682, opened 2024-05-23
#2641. Add patterns tests

sgrekhov:


eernstg (commented):
Looks good! However, it isn't entirely obvious to me why these tests have a relationship with the wildcarded variable feature, and why they aren't just duplicates of existing tests about patterns?

(There's a lot of similar elements in 'LanguageFeatures/Patterns/wildcards_A*.dart').

I think there was an interaction between wildcard patterns and wildcarded variable declarations that came up recently, but I can't remember it right now. Is there something which is actually new here?

sgrekhov:
Yes, there are duplicated elements but not entierly. These tests check wildcards in more details, in particular wildcards assignment. Patterns tests more tests `switch` statement and no fully tests the assignment. There are also no tetst that check that initializer is executed anyway in case of wildcarded assignment etc. I think it's Ok to have these, partially intersecting tests.

But if you insists I can move them to the `Patterns` folder

eernstg:
OK, thanks!

I think it would indeed be helpful to move them to `Patterns`, considering that there are no wildcarded variables in these tests.

PR #2686, opened 2024-05-24
#2641. Add initializing formals tests

sgrekhov:


Language/Functions/Formal_Parameters/Optional_Formals/name_t02.dart
```diff
@@ -2,23 +2,32 @@
...
+  String? _;
+  C({this._});
+//        ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
 main() {
   new A(_p: "Optional parameters names must not begin with an underscore!");
+  new C(_: "Optional parameters names must not begin with an underscore!");
```
eernstg, line 32: Just a typo:
```suggestion
  new A(_p: "Optional parameter names must not begin with an underscore!");
  new C(_: "Optional parameter names must not begin with an underscore!");
```

LanguageFeatures/Wildcards/initializing_formals_A02_t02.dart
```diff
@@ -0,0 +1,30 @@
...
+import '../../Utils/expect.dart';
+
+class C1 {
+  int _, a;
+  C1(this.a) : _ = a;
+}
+
+class C2 {
+  final int _, a;
+  const C2(this.a) : _ = a;
```
eernstg, line 24: How about this case, is that already covered?:
```suggestion
  const C2(this.a) : _ = a;
  const C2(this.a, bool _) : _ = a; // OK, there is no name clash with the wildcard parameter.
```

LanguageFeatures/Wildcards/initializing_formals_A03_t01.dart
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Even though the parameters no longer collide, it is still an
+/// error to have two initializing formals named `_`
+///
+/// @description Checks that it is a compile-time error to declare more than one
+/// initializing formals named `_`
```
eernstg, line 9: Yeah, that's a tricky choice of singular/plural, but I think the most natural choice is singular:
```suggestion
/// initializing formal named `_`
```

eernstg (commented):
Looks good, couple of typos.

Language/Functions/Formal_Parameters/Optional_Formals/name_t02.dart
sgrekhov: Updated

LanguageFeatures/Wildcards/initializing_formals_A02_t02.dart
sgrekhov: Added

LanguageFeatures/Wildcards/initializing_formals_A03_t01.dart
sgrekhov: Several `super._` are allowed, because changed to "...initializing formal `this._`."

sgrekhov:
Updated. PTAL

LanguageFeatures/Wildcards/initializing_formals_A02_t02.dart
```diff
@@ -16,12 +16,12 @@ import '../../Utils/expect.dart';
 
 class C1 {
   int _, a;
-  C1(this.a) : _ = a;
+  C1(this.a) : _ = a; // OK, there is no name clash with the wildcard parameter.
```
eernstg, line 19: I'm not quite sure what this means. We don't have any wildcard parameters in this declaration.

I intended to suggest that we could have a case where there is another constructor of `C1` that _does_ have a wildcard parameter (unused), and we would still be able to initialize the instance variable named `_` because that wildcard parameter does not disrupt the resolution of `_` in the initializer element: `_ = a` always means "initialize the instance variable named `_` to the value of `a`".

```suggestion
  C1(this.a) : _ = a;
  C1.otherCase(this.a, String _) : _ = a;
```

Perhaps we have this already, it just occurred to me that we might not have a test that covers this situation (it is similar to "a wildcard doesn't shadow anything", but not quite the same).

eernstg (commented):
I think one of my comments was unclear.

LanguageFeatures/Wildcards/initializing_formals_A02_t02.dart
sgrekhov: Ah... I misunderstood. Thank you, we had no such test. Updated

LanguageFeatures/Wildcards/initializing_formals_A02_t02.dart
eernstg: Thx!

LanguageFeatures/Wildcards/initializing_formals_A02_t02.dart
eernstg: Thanks!

eernstg (approved):
LGTM

PR #2687, opened 2024-05-27
#2641. Add super parameters tests

sgrekhov:


LanguageFeatures/Wildcards/super_parameters_A01_t01.dart
```diff
@@ -0,0 +1,44 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion An occurrence of `super._` as a declaration of a formal parameter
+/// in a constructor is a compile-time error. This error also occurs in the case
+/// where the super parameter has an explicitly declared type and/or default
+/// value.
+///
+/// @description Checks that it is a compile-time error to refer `super._` as a
+/// formal parameter of a constructor.
```
eernstg, line 11: Oh, this was exactly the thing which was changed: It is _no longer_ an error to use `super._`.

The meaning of `super._` is that
- This is a wildcarded parameter declaration, and `_` in an initializer list entry will resolve to an instance/static/top-level declaration, not to the parameter.
- Nevertheless, the compiler is able to generate code for the superinitialization such that the actual argument is passed on to the constructor of the superclass which is called in that manner.

```dart
class A {
  final int i;
  A(this.i);
}

class B extends A {
  B(super._); // Means something like this:
  B(int _freshName2398): super(_freshName2398);
  B.name(super._): assert(_ > 0); // Error, unknown identifier `_`.
}
```

eernstg (commented):
This one needs to be adjusted: As mentioned in https://github.com/dart-lang/language/issues/3792#issuecomment-2112869245, it is no longer an error to have `super._` as a constructor parameter declaration.

sgrekhov:
Ahh... I marked it as "Blocked" and want to go on with this after spec update

eernstg:
Ok, thanks!

sgrekhov:
Updated according to the changed specification. Please review

LanguageFeatures/Wildcards/super_parameters_A04_t01.dart
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Multiple `super._` and `this._` parameters can occur in the same
+/// constructor without any name collision.
+///
+/// @description Checks that multiple `super._` and `this._` parameters can
```
eernstg, line 8: Great, I hadn't thought of that! ;-)

LanguageFeatures/Wildcards/super_parameters_A04_t01.dart
```diff
@@ -0,0 +1,31 @@
...
+/// occur in the same constructor without any name collision.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=wildcard-variables
+
+import '../../Utils/expect.dart';
+
+class A {
+  final int x, y;
+  A(this.x, this.y);
```
eernstg, line 18: Shouldn't we have enough parameters to avoid extra errors?:
```suggestion
  final int x, y;
  A(this.x, this.y, int _);
```

LanguageFeatures/Wildcards/super_parameters_A04_t01.dart
eernstg: Maybe just say that there is more than one `super._`, and then we can have test cases with 0 and 1 occurrences of `this._`.
```suggestion
/// @description Checks that multiple `super._` parameters can
```

LanguageFeatures/Wildcards/super_parameters_A04_t01.dart
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Multiple `super._` and `this._` parameters can occur in the same
```
eernstg, line 5: Does the spec say 'multiple ... `this._`'? That doesn't sound right.

eernstg: Indeed, the spec says precisely that.

eernstg (commented):
Looks good! However, I think LanguageFeatures/Wildcards/super_parameters_A04_t01.dart has a spurious error.

sgrekhov (commented):
Test descriptions updated. But I don't see an error in `super_parameters_A04_t01.dart`. Please clarify

LanguageFeatures/Wildcards/super_parameters_A04_t01.dart
eernstg: Oops, you're right, sorry!

LanguageFeatures/Wildcards/super_parameters_A04_t01.dart
eernstg: Thanks!

eernstg (approved):
LGTM

PR #2690, opened 2024-05-28
Fixes #2689. Add more extension methods tests

### LINKED ISSUE #2689: Add more extension member conflict tests
Add static vs. instance member conflict tests for extensions

sgrekhov:


LanguageFeatures/Extension-methods/extension_conflict_resolution_t16.dart
```diff
@@ -0,0 +1,48 @@
...
+///      non-platform libraries, and
+///   3. the instantiated type (the type after applying type inference from the
+///      receiver) of [T1] is a subtype of the instantiated type of [T2] and
+///      either
+///   4. not vice versa, or
+///   5. the instantiate-to-bounds type of [T1] is a subtype of the
+///      instantiate-to-bounds type of [T2] and not vice versa.
+///
+/// @description Check that it is not an error if two extensions add static and
+/// instance members with the same basename
```
eernstg, line 25: The basename is just the name unless the member is a setter. So perhaps change the getters to setters? And also cover static method vs. instance setter?

eernstg (commented):
Looks good!

I'm suggesting an adjustment of a test where 'basename' is mentioned, because it doesn't actually use anything that differs from the plain names.

Another thing could be slightly confusing: Several description paragraphs talk about "it is not an error" and then something about the existence of declarations in extensions. However, that's never an error, because we don't have name clashes among separate extension declarations as such.

The situation where we may get an error is when there is a call site whose resolution fails, because there are multiple accessible and applicable extensions with an extension instance member whose basename is the same as the basename of the invocation.

So, in order to avoid the confusion, it could be safer to say something like "it does not give rise to an ambiguity error at the call site" rather than "it is not an error". I'm sure there's a more elegant way to say this, but the point is just that we shouldn't imply that there is an error in one or the other extension, it's all about errors at the call site.

sgrekhov:
Thank you. Updated, PTAL

eernstg (approved):
LGTM, thanks!

PR #2691, opened 2024-05-28
#2641. Add external functions and covariant parameters tests

sgrekhov:


LanguageFeatures/Wildcards/other_declarations_A01_t09.dart
```diff
@@ -0,0 +1,26 @@
...
+///
+/// @description Checks that an external top-level variable called `_` can be
+/// declared and accessed.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=wildcard-variables
+
+import '../../Utils/expect.dart';
+
+external int _;
```
eernstg, line 17: This would be a compile-time error with 'dart compile js', right?

I would assume that we have a standard way to specify that a particular test should not be executed on some configurations, but I can't immediately see any examples (grepping for 'SharedOptions', 'requirements', 'web-numbers', and a few others).

Perhaps it's going to be an entry in co19/*.status?

eernstg (commented):
Looks good!

Just one question: Do we need to prevent configurations like `dart compile js` from running the `external` related tests?

LanguageFeatures/Wildcards/other_declarations_A01_t09.dart
sgrekhov: Yes, we are doing it in a `.status` files with a comment `SkipByDesign # Non-JS-interop external members are not supported`

eernstg: OK, thanks!

eernstg (approved):
LGTM

PR #2692, opened 2024-05-29
Fixes #2678. Add more extension types static vs instance conflict tests

### LINKED ISSUE #2678: Add extension type preclude tests static vs instance
See https://github.com/dart-lang/language/issues/3822 for the details

sgrekhov:


LanguageFeatures/Extension-types/static_analysis_member_invocation_A06_t08.dart
```diff
@@ -0,0 +1,44 @@
...
+
+/// @assertion We say that an extension type declaration DV has an extension
+/// type member named n in the cases where:
+/// - DV declares a member named n.
+/// - DV has no such declaration, but DV has a direct extension type
+///   superinterface V that has an extension type instance member named n due to
+///   a member declaration DM2, and DV does not declare an instance member that
+///   precludes DM2.
+///
+/// @description Checks that an instance members don't preclude a static method
```
eernstg, line 13:
```suggestion
/// @description Checks that an instance member doesn't preclude a static method
```

This one is somewhat tricky to understand. 'Preclusion' refers to the fact that an extension type does not 'have' a particular instance member, in spite of the fact that a superinterface does have it. In other words, 'to preclude `E0.m`' means 'to prevent inheritance of `E0.m` into `E1`", where `m` is an instance member (an extension type instance member or a non-extension type member signature).

So it means that it just doesn't make sense to say that a static method is precluded (we can't prevent inheritance of something which isn't inherited anyway).

We're still testing _something_, though. I guess we can say that the test confirms that it does not give rise to a compile-time error that `ET1` declares a method named `n` and `ET1` has a superinterface `V1` that declares a static member named `n`.

This is kind of a "no error in this case" test for the section 'Class Member Conflicts' in the language specification.

Do we cover that already, for extension types? Should it be in some other directory, or have a different name?

I think it's worth taking a look, because we should take care to ensure that each test is structured/named/commented in a way that matches what is actually being tested.

LanguageFeatures/Extension-types/static_analysis_member_invocation_A06_t09.dart
```diff
@@ -0,0 +1,44 @@
...
+
+/// @assertion We say that an extension type declaration DV has an extension
+/// type member named n in the cases where:
+/// - DV declares a member named n.
+/// - DV has no such declaration, but DV has a direct extension type
+///   superinterface V that has an extension type instance member named n due to
+///   a member declaration DM2, and DV does not declare an instance member that
+///   precludes DM2.
+///
+/// @description Checks that an instance members don't preclude a static getter
```
eernstg, line 13: Similar issue here.

LanguageFeatures/Extension-types/static_analysis_member_invocation_A06_t10.dart
```diff
@@ -0,0 +1,47 @@
...
+
+/// @assertion We say that an extension type declaration DV has an extension
+/// type member named n in the cases where:
+/// - DV declares a member named n.
+/// - DV has no such declaration, but DV has a direct extension type
+///   superinterface V that has an extension type instance member named n due to
+///   a member declaration DM2, and DV does not declare an instance member that
+///   precludes DM2.
+///
+/// @description Checks that an instance members don't preclude a static setter
```
eernstg, line 13: Similar issue here.

LanguageFeatures/Extension-types/static_analysis_member_invocation_A06_t11.dart
```diff
@@ -0,0 +1,58 @@
...
+
+/// @assertion We say that an extension type declaration DV has an extension
+/// type member named n in the cases where:
+/// - DV declares a member named n.
+/// - DV has no such declaration, but DV has a direct extension type
+///   superinterface V that has an extension type instance member named n due to
+///   a member declaration DM2, and DV does not declare an instance member that
+///   precludes DM2.
+///
+/// @description Checks that a static members doesn't preclude each others
```
eernstg, line 13: Similar issue here, this is actually also a case which does not give rise to a compile-time error in 'Class Member Conflicts'.

LanguageFeatures/Extension-types/static_analysis_member_invocation_A08_t08.dart
```diff
@@ -0,0 +1,47 @@
...
+/// type member named n in the case where DV does not declare a member named n,
+/// and one of the following criteria is satisfied:
+/// - DV has a direct extension type superinterface V that has a non-extension
+///   type member with signature m and name n, and DV does not declare a member
+///   that precludes m.
+/// - DV has a direct non-extension type superinterface whose interface contains
+///   a member signature m named n, and DV does not declare a member that
+///   precludes m.
+///
+/// @description Checks that an instance members don't preclude a non-extension
```
eernstg, line 15: Seems like a fix like this one is needed in a few more places:
```suggestion
/// @description Checks that an instance member doesn't preclude a non-extension
```

We could also use plural ('that instance members don't preclude'), but singular is usually more precise (that is, less open to misreadings).

eernstg (commented):
Looks good, modulo a couple of typos!

We do need a slightly bigger adjustment in some cases, though. There are several tests whose topic seems to be some variation of "... doesn't preclude a static ...". Those tests are somewhat misleading because the concept of preclusion is associated with instance members, and it is simply never relevant for a static member declaration.

So we cannot possibly test anything about the correctness of the implementation of preclusion by testing the treatment of a static member.

We are testing some other properties, though, and the best way ahead would probably be to look for the right group of tests (the right directory and the right name) to use for these tests, and then check that they aren't redundant, and then land them there.

sgrekhov:
Thanx! Updated. PTAL

eernstg (approved):
Looks great, thanks!

I fixed one typo: The descriptions contain the phrase 'an instance members' six times, it should be 'an instance member'.

Issue #2694, opened 2024-05-30
Testing null aware elements


eernstg:
The [null aware elements](https://github.com/dart-lang/language/blob/main/working/0323-null-aware-elements/feature-specification.md) feature is being implemented, so we need some tests on that topic.

eernstg:
Thanks!

sgrekhov:
Closing as completed. Expressions checklist below

Expression | Test
-- | --
?a.b = e | https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_assignment_A02_t01.dart
?a?.b = e | https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_assignment_A02_t01.dart
?a[e1] = e2 | https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_assignment_A02_t01.dart
?a?[e1] = e2 | https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_assignment_A02_t01.dart
?super.b = e | https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_assignment_A02_t01.dart
?super[e1] = e2 | https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_assignment_A02_t01.dart
?this | https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_this_super_A02_t01.dart
?super.b | https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_this_super_A01_t01.dart
?super[e1] | https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_this_super_A01_t01.dart
?super(42) | https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_this_super_A01_t01.dart
?(e) | https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_parenthesized_A01_t01.dart
[...]

PR #2696, opened 2024-06-03
#2694. Add the first bunch of null-aware elements tests

sgrekhov:


eernstg (commented):
Looks good!

Just one typo that it would be nice to eliminate: It says 'if leaf elements has' rather than 'if leaf elements have', about 5 times. Another way to fix it would be to say that 'if the set of leaf elements has'.

The feature specification does have phrases like '_leaf elements_ has', but in this situation _leaf elements_ is actually a meta-variable, so it's more like "if theVariableNamedLeafElements has". So yet another way to handle it would be to make sure that "leaf elements" is in italic everywhere.

(Actually "leaf elements" is in italic almost everywhere in the spec, but there is one missing occurrence, at the end. ;-).

sgrekhov:
Updated. PTAL

eernstg (approved):
LGTM

PR #2697, opened 2024-06-03
#2694. Add type inference tests for sets

sgrekhov:


eernstg (commented):
I think the 2nd test needs a small adjustment.

LanguageFeatures/Control-flow-collections/Null-aware-elements/type_inference_set_A02_t01.dart
```diff
@@ -30,7 +30,7 @@ main() {
   String? e1 = "e1";
 
   var set1 = <String>{
-    ?(e1.expectStaticType<Exactly<String?>>()),
+    ?(e1..expectStaticType<Exactly<String?>>()),
```
eernstg, line 33: I don't think that will suffice, it will succeed even if the context type is `Object` or a bunch of other types, it still only confirms that `e1` has the type `String?`.

eernstg (commented):
That's still not the context type.

LanguageFeatures/Control-flow-collections/Null-aware-elements/type_inference_set_A02_t01.dart
sgrekhov: @eernstg please clarify why? My logic is the following
```dart
  String? e1 = "e1"; // promoted by initialization to static type String
  e1.toString(); // Ok
  e1.expectStaticType<Exactly<String>>(); // Ok
  e1.expectStaticType<Exactly<String?>>(); // Error

  var set1 = <String>{
    ?(e1..expectStaticType<Exactly<String?>>()) // if this is Ok that means that e1 is downgraded to String? because of the context
  };
```
Where am I wrong?

sgrekhov: May be this code is better?
```dart
 String? e1 = "e1";

  var set1 = <String>{
    ?(contextType(e1)..expectStaticType<Exactly<String?>>())
  };
```

eernstg: Ah, now I see where we're going down different lanes.

The description says that we're testing that the inferred type of the element `?(e1..something)` is `Nonnull(U)` where `U` is the inferred type of the expression `e1` in the context `Ps?`.

The invocation of `expectStaticType` doesn't actually test that type (but we can't do `(?e1)..expectStaticType<...>()` because `?e1` is not an expression), so it seems reasonable that we'd test the type of `e1` and check that it is of the form `T?` and the given `T` makes sense.

However, the type of `e1` doesn't depend on the use of the null-aware element operator at all: `e1` has the declared type `String?` and it is immediately promoted to `String`, based on the initializing expression. After the declaration the inferred type of `e1` is `String`, also when it occurs in the context `<String>{?e1}`, and this implies that we aren't testing anything at all about null-aware elements, we're just checking that `e1` has the type we'd expect based on some other, unrelated language features.

That type is `String`, so `e1..expectStaticType<Exactly<String?>>()` should fail.

So I tried to find something which could be tested and which would say something about null-aware elements. I should have made that more explicit.

I thought we might want to test that the context type of `e1` is `Ps?`, that is, `String?`.

We can do that by means of

```dart
contextType(e1)..expectStaticType<Exactly<String>>()`
```

`contextType` takes an argument of type `Object?` (which must have the required type, and the type system isn't going to help us, but we know that a `String` will work). Moreover, `contextType` returns a result of type `T` which is a type parameter of that function. So we'd infer the value of that type argument (and that's going to yield the context type (the greatest closure, if it's a context type schema) because `T` isn't used in any other way). So `T` will be inferred as `String?` and the type of the expression `contextType(e1)`  is then `String?`. The next step is to test that using `expectStaticType<Exactly<String>>()`.

I think it would be helpful to adjust the description to say something about the context type, because that's the thing which is directly testable and related to the new language feature, and then adjust the code to test the context type using `contextType`.

eernstg (commented):
Clarified the context type thing.

eernstg:
> May be this code is better?

Indeed!

sgrekhov:
Updated. PTAL. Note that

>  So T will be inferred as String? and the type of the expression contextType(e1) is then String?. The next step is to test that using `expectStaticType<Exactly<String>>()`.

I believe  there should be `expectStaticType<Exactly<String?>>().`

LanguageFeatures/Control-flow-collections/Null-aware-elements/type_inference_set_A02_t01.dart
```diff
@@ -19,9 +19,8 @@
 ///   - The inferred set element type is `NonNull(U)`. The value added to the
 ///     set will never be `null`.
 ///
-/// @description Check that the inferred type of element `?e1` in the context
-/// `Set<Ps> is `NonNull(U)`, where `U` is the inferred type of the expression
-/// `e1` in the context `Ps?`
+/// @description Check that if the context type of element `?e1` is `Ps`, then
+///  the inferred type of the expression `e1` in the context `Ps?`.
```
eernstg, line 23: That's not quite a sentence. Perhaps?:
```suggestion
/// @description Check that if the context type of the element `?e1` is `Ps`
/// then the context type of `e1` is `Ps?`.
```
```suggestion
/// the context type of `e1` is `Ps?`.
```

LanguageFeatures/Control-flow-collections/Null-aware-elements/type_inference_set_A02_t01.dart
```diff
@@ -30,8 +29,7 @@ main() {
   String? e1 = "e1";
 
   var set1 = <String>{
-    ?(e1..expectStaticType<Exactly<String?>>()),
-    ?null
+    ?(contextType(e1)..expectStaticType<Exactly<String?>>())
```
eernstg, line 32: Exactly!

eernstg (commented):
Looks good! Just one comment on the wording of the description.

sgrekhov:
Description updated. Thank you.

LanguageFeatures/Control-flow-collections/Null-aware-elements/type_inference_set_A02_t01.dart
eernstg: And yes, it should be `String?`, not `String`.

eernstg (approved):
LGTM!

PR #2700, opened 2024-06-04
#2694. Add constants and runtime semantics tests

sgrekhov:


LanguageFeatures/Control-flow-collections/Null-aware-elements/constants_A01_t03.dart
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A `nullAwareExpressionElement` or `nullAwareMapElement` is
+/// constant if its inner expression or map entry is constant.
+///
+/// @description Check that it is a compile-time error if not a constant
```
eernstg, line 8: We'd probably say 'which is not a constant expression' in spec documents, but 'a non-constant expression' should be fine here.

```suggestion
/// @description Check that it is a compile-time error if a non-constant
```

LanguageFeatures/Control-flow-collections/Null-aware-elements/runtime_A01_t01.dart
```diff
@@ -0,0 +1,62 @@
...
+
+  var set1 = {
+    ?e1,
+    ?null,
+    ?e2
+  };
+  Expect.setEquals({e1}, set1);
+
+  var set2 = {
+    ?e1 + 1,
```
eernstg, line 28: Isn't this just going to be a compile-time error? (`e1` has type `int?`).

Perhaps there is an extension operator `+` that works on an `int?`? Otherwise we could declare it, in order to enable `e1 + 1` (and, in general, `?e` where `e` is various kinds of expressions).

eernstg (commented):
Looks good! Couple of typos.

Nit: Perhaps we should use `nil` rather than `Nil`, just because that's the standard capitalization, and we probably aren't testing anything relevant or interesting by using `Nil`.

Another thing, probably for a different PR if needed:

Do we cover null-aware expressions of different syntactic forms? I can see `<identifier>` as in `?e1`, we have one example in #2697 about an element of the form `?(e)`, and we have `?e1 + 1`.

But I'm not sure we have anything like `?e1 >>> e2` and a bunch of other syntactic forms. We should make sure they are parsed correctly (as `?e` where `e` is `e1 >>> e2`, as opposed to being rejected as a syntax error).

It should be sufficient to have just one test with one case for each form of expression. That is, each right hand side of a non-terminal which can be derived from `<expression>` where there is at least one terminal. For example, `<expression>` derives `<shiftExpression>` which derives `<additiveExpression> '>>>' <additiveExpression>`, so we need to test `?e1 >>> e2`. Similarly, `<expression>` derives `<conditionalExpression>` which derives `<ifNullExpression> '?' <expressionWithoutCascade> ':' <expressionWithoutCascade>`, so we need to test `?e1 ? e2 : e3`.

Do we have that already?

eernstg:
I can see that we have some of the expression forms already (such as `? 2 > 1 ? 2 : 3: 0` in https://github.com/dart-lang/co19/pull/2701), but it would still be nice to check the coverage if it might be incomplete.

eernstg:
This one also contributes several additional forms: https://github.com/dart-lang/co19/pull/2706, but there may still be room for a test that systematically covers all remaining expression forms.

LanguageFeatures/Control-flow-collections/Null-aware-elements/constants_A01_t01.dart
```diff
@@ -0,0 +1,45 @@
...
+  ];
+  Expect.listEquals([e1], list);
+
+  const map = {
+    ?e1: 1,
+    e1: ?null,
+    e1: ?Nil,
+    ?e1: ?e2,
+    ?null: e2,
+    ?Nil: e2,
```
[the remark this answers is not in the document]
sgrekhov: Indeed. Added

sgrekhov (commented):
Updated. I'll add tests for all other expressions (like `?e >> 1`) in a separate PR.

eernstg:
> other expressions

Sounds good, thanks!

PR #2701, opened 2024-06-05
#2694. Add unnecessary null-aware elements tests

sgrekhov:


LanguageFeatures/Control-flow-collections/Null-aware-elements/warnings_A01_t01.dart
```diff
@@ -0,0 +1,109 @@
...
+// [cfe] unspecified
+      else
+        ?e3, // Ok
+      ?e4
+//    ^
+// [analyzer] unspecified
+// [cfe] unspecified
+    };
+
+    <Object>[
```
eernstg, line 43: We should probably put the set tests into a separate function `testSets` and the list tests into `testLists`, or something like that, in order to avoid any issues with flow-analysis (promotions) that arise in one test and changes the meaning of the next one.

LanguageFeatures/Control-flow-collections/Null-aware-elements/warnings_A01_t01.dart
```diff
@@ -0,0 +1,109 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+      else
+        3: ?e3, // Ok
+      4: ?e4
+//       ^
+// [analyzer] unspecified
+// [cfe] unspecified
+    };
+  }
```
eernstg, line 108: For the map tests we should probably even keep the map-key and the map-value tests separate.

LanguageFeatures/Control-flow-collections/Null-aware-elements/warnings_A01_t02.dart
```diff
@@ -0,0 +1,106 @@
...
+// [cfe] unspecified
+    ?true,
+//  ^
+// [analyzer] unspecified
+// [cfe] unspecified
+    ?"",
+//  ^
+// [analyzer] unspecified
+// [cfe] unspecified
+    ?'''  Raw string   '''
```
eernstg, line 32: Perhaps:
```suggestion
    ?'''  Triple quoted string   '''
```
because 'raw string' has a specific meaning. However, we might also want to have `? r'Actually a raw string'`.

eernstg (commented):
Looks good!

However, we should probably avoid multiple uses of a local variable that may be promoted: We aren't testing chains of promotions, we're testing null-aware elements, and it's confusing if we suddenly have an unexpected warning because `e2` has type `int?` during the first test and is promoted to `int` in the next (very similar) test.

LanguageFeatures/Control-flow-collections/Null-aware-elements/warnings_A01_t01.dart
sgrekhov: Moved to a separate function

LanguageFeatures/Control-flow-collections/Null-aware-elements/warnings_A01_t01.dart
sgrekhov: Done

LanguageFeatures/Control-flow-collections/Null-aware-elements/warnings_A01_t02.dart
sgrekhov: Fixed and added true raw string

sgrekhov (commented):
Updated. PTAL

LanguageFeatures/Control-flow-collections/Null-aware-elements/warnings_A01_t01.dart
eernstg: Thanks!

eernstg (approved):
LGTM

LanguageFeatures/Control-flow-collections/Null-aware-elements/warnings_A01_t04.dart
```diff
@@ -0,0 +1,57 @@
...
+  <int, int>{
+    ? f(1) ?? 1: 0,
+//  ^
+// [analyzer] unspecified
+// [cfe] unspecified
+    ? 2 > 1 ? 2 : 3: 0,
+//  ^
+// [analyzer] unspecified
+// [cfe] unspecified
+    4: ? f(1) ?? 1: 0,
```
eernstg, line 48: Perhaps?:
```suggestion
    4: ? f(1) ?? 1,
```

eernstg:
Just one more thing: LanguageFeatures/Control-flow-collections/Null-aware-elements/warnings_A01_t04.dart seems to have an unintended syntax error in line 48. (And the spec parser is complaining about line 40, too, not sure why).

eernstg (commented):
One more thing came up.

sgrekhov:
Fixed. There were a typo on line 50 as well.

eernstg (approved):
LGTM

LanguageFeatures/Control-flow-collections/Null-aware-elements/warnings_A01_t02.dart
eernstg: Thx!

PR #2706, opened 2024-06-10
#2694. Add more null-aware elements syntax tests

sgrekhov:


LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_A03_t01.dart
```diff
@@ -0,0 +1,62 @@
...
+///
+/// @description Check that `nullAwareExpressionElement` `?` affects the
+/// whole expression, not just the first token
+/// @author sgrekhov22@gmail.com
+
+main() {
+  int? e = 2 > 1 ? 1 : null;
+
+  <int>[
+    ? e + 1,
```
eernstg, line 27: I'm not quite sure what we're testing here:

If it is parsed as `? (e + 1)` then `e + 1` is an expression that has a compile-time error (because `int?` doesn't have an operator `+`).

On the other hand, if it's parsed as `(?e) ...` then it's a syntax error because `?e` is an element, not an expression, and elements can't be parenthesized.

So both of the obvious parsing approaches will yield an error, and we're not really detecting anything by confirming that there is an error.

I think it would be better to use an extension operator `+` that has on-type `int?`, and then confirm that it is _not_ an error to have `? e + 1`.

The point is that that one would still be an error if it is parsed as the null-aware element `?e` followed by anything (that is, if `?` is taken to apply to the first token rather than the whole expression).

LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_A04_t02.dart
```diff
@@ -0,0 +1,85 @@
...
+///   | // Existing productions...
+///
+/// nullAwareExpressionElement ::= '?' expression
+///
+/// nullAwareMapElement ::=
+///   | '?' expression ':' '?'? expression // Null-aware key or both.
+///   |     expression ':' '?' expression  // Null-aware value.
+///
+/// @description Check that it is a syntax error if a `nullAwareMapElement`
+/// or a `nullAwareExpressionElement` is used in a string expression
```
eernstg, line 20: This one is delicate. ;-)

We don't want to test systematically that it is an error (that is, probably in every case, a syntax error) if a null-aware element is used in every syntactic position where we do not expect an element (so we'd test that it is an error to have `?Object` as a superclass or in `implements` or as an extension on-type, `?2` as an actual argument in a function invocation, `?int` as the declared type of a local variable and parameter, etc.etc.etc.) That's a job for a fuzzer. ;-)

So why should we then test specifically that it's a syntax error to have an element (in particular, a null-aware element) as an interpolated expression of a literal string?

Or did I miss something? Isn't this just another syntax error? The problem is that there are too many possible syntax errors, so we shouldn't even start covering them.

eernstg (commented):
This one might not be testing precisely enough, PTAL.

LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_A03_t01.dart
sgrekhov: The idea here is to check that `?e + 1` is parsed as `?(e + 1)` and this is a compile-time error because type of `e` is `int?`. If `?` will be applied to `e` first, then the type of  `?e` became `int` and then there will be no error to call `+ 1` on it. This test checks it. 

Tests that add extension on `int?` and check that `?e + 1` is not an error were just added to https://github.com/dart-lang/co19/pull/2700

LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_A04_t02.dart
sgrekhov: This test is mainly based on the sentence from the spec "The `?` character is already overloaded in Dart for nullable types, conditional expressions, null-aware operators, and null-check patterns". We don't have much tests that use '?' character in string expression, so just wanted to check that it is parsed as should be. If you think that here were are testing just one piece of endless possible random data then I don't mind to delete it.

sgrekhov (commented):
Added some comments

LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_A03_t01.dart
eernstg: Ah, 'elements can't be parenthesized' was wrong (those parentheses were added by me to imply a specific AST structure).

But it is still true that it is a syntax error if an element occurs in a position where an expression is expected. For example, `[1 + if (true) 2 else 3]` is a syntax error.

Parsing `?e + 1` such that the subterms are `?e` and `+` and `1` will give rise to a syntax error.

So we still can't see the difference between the correct parse and an incorrect one (that seems likely to be worth testing for&mdash;there's absolutely no limit on the number of possible incorrect parsings, so we have to focus on a likely one).

This basically means that we don't need help from the type system, the very fact that `?e + 1` is not a compile-time error demonstrates that it cannot possibly be parsed in a way that makes `?e` a subterm, it _must_ have been parsed as a term with the subterms `?` and `e + 1`.

This means that the approach taken in #2700 is perfectly fine when the goal is to confirm that the parsing structure is appropriate (but we don't need to cover "is actually null"/"is actually non-null" if we're just testing the parse).

LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_A04_t02.dart
eernstg: I do think the value of testing that a particular syntactic construct can be parsed correctly is high. We should have good coverage when it comes to new (and surprising ;-) syntactic forms or combinations.

But that's the positive case: "This construct can be used as intended" (also: if there's an issue, like grammar ambiguity, then perhaps the exploration of combinations will bring it out).

The negative case is much less valuable: "This is wrong" is something that tools should report reliably, for syntax as well as other properties, but there are so many ways to be wrong, in particular syntactically, that we should abstain from trying to cover them all.

This means that we should be careful whenever we add to the set of tests doing that sort of thing. We do have some tests of this nature (typically named `syntax_*.dart`, I think), but we should at least take a deep breath before writing more tests like that, especially if they have a different name (so they're sort of escaping the small closed box that we usually keep this kind of test in, namely having a name of the form `syntax_*.dart`).

We do have some efforts in some teams about fuzzing, and I think creating wrong syntax (really: random sequences of tokens that occur in designated "blank spots" in an otherwise correct library) would be a very useful thing to do systematically.

But for manually written syntax error tests we have a quite small number of libraries (often named `syntax_*.dart`), and we shouldn't

eernstg (commented):
One more iteration on LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_A03_t01.dart.

sgrekhov:
Updated `syntax_A03_t01` and deleted `syntax_A04_to2`. PTAL

eernstg (approved):
LGTM, thanks!

PR #2715, opened 2024-06-17
#2694.Add syntax tests for different expressions. Part 1

sgrekhov:


eernstg (commented):
Looks good!

This kind of test ("try this with all expressions") is rather delicate, because it's difficult to draw the line between cases to try and excessive detail.

I looked at the grammar and came up with a bunch of cases. I know that at least a few of them are covered in other tests (not in this PR). Some cases are difficult because they are by their very nature non-nullable.

I don't have a good solution for the non-nullable cases, but if we could `// ignore_for_file` the warning about `?e` where `e` is non-nullable then we would be able to include all these syntactic forms, just so we can see that the parser doesn't crash (and we don't get any weird error messages, indicating that the parser does accept the construct, but it gets the structure of the construct wrong).

So here is the list I came up with:
  ?a.b = e
  ?a?.b = e
  ?a[e1] = e2
  ?a?[e1] = e2
  ?super.b = e
  ?super[e1] = e2
  ?this
  ?super.b
  ?super[e1]
  ?super<int>(42)
  ?(e)
  ?null
  ?1
  ?0x1
  ?true
  ?false
  ?'Hello'
  ?r'Hello'
  ?{1}
  ?<int>{1}
  ?{1: 2}
  ?<int, int>{1: 2}
  ?[1]
  ?<int>[1]
  ?const {1}
  ?const <int>{1}
  ?const {1: 2}
  ?const <int, int>{1: 2}
  ?const [1]
  ?const <int>[1]
  ?(1, name: 2)
  ?const (1, name: 2)
  ?MyClass.new
  ?switch (1) { _ => 2 }
  ?throw 0
  ?() => 1
  ?() async => 1
  ?() {}
  ?() async {}
  ?new MyClass()
  ?const MyClass()
  ?a..toString()
  ?a?..toString()
  ?true ? 1 : 2
  ?1 ?? 2
  ?false || true
  ?true && false
  ?super == 1
  ?1 is int
  ?1 as int
  ?1 < 2
  ?super < 2
  ?1 | 1
  ?super | 1
  ?1 ^ 1
  ?super ^ 1
  ?1 & 1
  ?super & 1
  ?1 << 1
  ?super << 1
  ?1 + 1
  ?super + 1
  ?1 * 1
  ?super * 1
  ?-1
  ?!true
  ?~1
  ?await 1
  ?a++
  ?a.b++
  ?super.b++
  ?(a, b) = (2, 3)
  ?[a, b] = [2, 3]
  ?{true: a, false: b} = {true: 2, false: 3}

The idea is that each of these correspond to a grammar rule, and they are only concerned with the top level structure (we never have one construct inside another one, e.g., like `a + b + c` which has two `+` nodes in the AST, one as a child of the other one), just so we know that `? .. + ..` doesn't fail for the wrong reasons.

`super` is rather special because it's a separate syntactic construct. So that's the reason why I include it in so many cases.

I don't think we have to have null-aware map elements for all combinations, we just need to see that the null-aware expression elements are working correctly, and then we'll trust the parser to run the same algorithm on each of the key and value part of a map element.

It would be great if you could use the list above and basically give each of them a check mark when you have detected that it is already covered, or a new case has been added to cover it.

I know, this is rather tedious! ... But I do think we should make sure we have good coverage because this kind of test is really an exception. ;-)

sgrekhov:
Thank you! I’ll use the above as a checklist. I used the “Operator Precedence” table from the spec when working on these tests and, therefore, missed record types and switch expressions altogether! But, on the other hand, why doesn’t this list contain `+=, -=,…`? Do we need to test these expressions as well?

I’ll add more tests for all of these cases. Meanwhile, I think you can review my PRs despite some cases being missing. They will be added later. All the tests that require `// ignore_for_file: TODO: add lint_name` will be added in the very last PR, probably in Draft status while there is no real `lint_name`. That’s the plan!

eernstg:
>  `+=`, `-=`,…?

Good catch! I missed them because I forgot that `assignmentOperator` is much more than `=`. ;-)

eernstg:
> Draft status while there is no real `lint_name`.

Sounds good! Here is a possible alternative:

We could also expect the warnings from the analyzer, and rely on the fact that the CFE does not report warnings: So the CFE would be expected to parse them all (without any diagnostic messages of any kind), and the analyzer we be expected to complain about many of them (the ones that are non-nullable by nature), but we'd just have `// ^^^^` for the analyzer to allow that.

We would then rely on the fact that the CFE and the analyzer are sharing the parser, so everything should be parsed correctly if everything is accepted by the CFE.

sgrekhov:
> > Draft status while there is no real `lint_name`.
> 
> Sounds good! Here is a possible alternative:
> 
> We could also expect the warnings from the analyzer, and rely on the fact that the CFE does not report warnings: So the CFE would be expected to parse them all (without any diagnostic messages of any kind), and the analyzer we be expected to complain about many of them (the ones that are non-nullable by nature), but we'd just have `// ^^^^` for the analyzer to allow that.
> 
> We would then rely on the fact that the CFE and the analyzer are sharing the parser, so everything should be parsed correctly if everything is accepted by the CFE.

Yes, I will work for the analyzer and CFE. But this tests will be ignored by VM. If some test contains
```
//      ^
// [analyzer] unspecified
```
then `vm-...` test configurations ignore it. So it's better to use `// ignore_for_file: ...`

eernstg:
I don't think it matters: For this kind of test we're concerned with parsing. As long as the front end will be executed, it does not cause any harm that the backends aren't being executed.

So we should just get a result from configurations like 'cfe-strong-linux' and 'cfe-strong-mac', and then it isn't a problem that there is no result from configurations like 'vm-*'.

sgrekhov:
Ah...I was a little bit wrong. If some test contains
```
//      ^
// [analyzer] unspecified
```
it is really ignored on `vm-...` configurations. But if a tests contains
```
//                   ^^
// [analyzer] STATIC_WARNING.INVALID_NULL_AWARE_OPERATOR
```
it is not ignored and executed on `vm-...` configurations. Just for your (and my) information about how the test runner works

eernstg:
OK, that makes sense! Still, it shouldn't actually matter in this particular case if the `vm-...` configurations are skipped, as long as the `cfe-...` configurations are being executed. A `cfe-...` configuration execution is sufficient to confirm that the given syntactic form is being parsed without errors, and the resulting program has no compile-time errors.

PR #2717, opened 2024-06-18
#2694. Rename syntax tests for expressions

sgrekhov:
It makes sense to rename syntax tests for expressions to make maintenance easier


eernstg (approved):
LGTM

PR #2721, opened 2024-06-21
#2694. Add syntax tests for different expressions. Part 5

sgrekhov:


LanguageFeatures/Control-flow-collections/Null-aware-elements/syntax_assignment_A03_t01.dart
```diff
@@ -0,0 +1,169 @@
...
+
+// TODO(sgrekhov): replace unspecified by the actual lint name
+
+import '../../../Utils/expect.dart';
+
+main() {
+  int? a, b, c, d, e, f;
+
+  var list = [
+    ?(a) = 1,
```
eernstg, line 33: Just wondering, is it true that this gives rise to a warning from the analyzer because `(a) = 1` is non-nullable? If so, then we could avoid the warning by assigning an expression of type `int?` to `(a)`. That might be nice in the cases where it's actually not difficult to avoid that warning.

Otherwise, it's a different warning, and I'm just missing something.

eernstg: Bad idea after all, I think we can only avoid those warnings in very few cases. So I'll close this thread again.

eernstg (approved):
LGTM!

(I considered whether it would be feasible to avoid some of the warnings from the analyzer, but I don't think it's going to work with more than very few cases. So it makes more sense to simply have the warnings everywhere.)

PR #2722, opened 2024-06-24
#2559. Add more augmented expression tests

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t18.dart
```diff
@@ -0,0 +1,54 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion The exact result of an `augmented` expression depends on what is
+/// being augmented, but it generally follows the same rules as any normal
+/// identifier:
+/// - Augmenting getters: Within an augmenting getter `augmented` invokes the
+///   getter and evaluates to the return value. If augmenting a field with a
+///   getter, this will invoke the implicit getter from the augmented field.
+///
+/// @description Checks that it is a compile-time error to if an augmenting
```
eernstg, line 12:
```suggestion
/// @description Checks that it is a compile-time error if an augmenting
```

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t18_lib.dart
```diff
@@ -0,0 +1,73 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion The exact result of an `augmented` expression depends on what is
+/// being augmented, but it generally follows the same rules as any normal
+/// identifier:
+/// - Augmenting getters: Within an augmenting getter `augmented` invokes the
+///   getter and evaluates to the return value. If augmenting a field with a
+///   getter, this will invoke the implicit getter from the augmented field.
+///
+/// @description Checks that it is a compile-time error to if an augmenting
```
eernstg, line 12:
```suggestion
/// @description Checks that it is a compile-time error if an augmenting
```

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t19.dart
```diff
@@ -0,0 +1,54 @@
...
+
+/// @assertion The exact result of an `augmented` expression depends on what is
+/// being augmented, but it generally follows the same rules as any normal
+/// identifier:
+/// - Augmenting getters: Within an augmenting getter `augmented` invokes the
+///   getter and evaluates to the return value. If augmenting a field with a
+///   getter, this will invoke the implicit getter from the augmented field.
+///
+/// @description Checks that it is a compile-time error for an augmenting
+/// function to have as a parameter of a type with the name `augmented`.
```
eernstg, line 13: Perhaps the members in this library should have been methods? They would allow `...A01_t19_lib.dart` to have some augmenting method declarations where a parameter has a type whose name is `augmented`. I don't really see how the getter declarations below can be useful in this context.

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t19_lib.dart
```diff
@@ -0,0 +1,167 @@
...
+
+/// @assertion The exact result of an `augmented` expression depends on what is
+/// being augmented, but it generally follows the same rules as any normal
+/// identifier:
+/// - Augmenting getters: Within an augmenting getter `augmented` invokes the
+///   getter and evaluates to the return value. If augmenting a field with a
+///   getter, this will invoke the implicit getter from the augmented field.
+///
+/// @description Checks that it is a compile-time error to if an augmenting
+/// getter has return type with the name `augmented`.
```
eernstg, line 13: Shouldn't the description in this library be the same as in `...A01_t19.dart`? So it would be about augmenting function declarations where a parameter has a type whose name is `augmented`.

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t19_lib.dart
```diff
@@ -0,0 +1,167 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmented_expression_A01_t19.dart';
+
+augment String get topLevelGetter {
+  augmented? x;
```
eernstg, line 21: How would this help testing augmenting function declarations where a parameter has a type whose name is `augmented`?

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t19_lib.dart
```diff
@@ -0,0 +1,167 @@
...
+//^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  var f = (augmented x) {};
```
eernstg, line 25: OK, here we have the parameter whose type has the name `augmented`! However that isn't an augmenting declaration, it's a function literal (which is an expression and a declaration in one package, but it doesn't augment anything).

Perhaps the description should just talk about using `augmented` as a parameter type, and not promise that it will be an augmenting declaration of a function?

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t18.dart
```diff
@@ -0,0 +1,55 @@
...
+/// setter to have as a parameter of a type with the name `augmented`.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'augmented_expression_A02_t18_lib.dart';
+
+class augmented {}
+
+void set topLevelSetter(String value) {}
```
eernstg, line 24: Probably:
```suggestion
void set topLevelSetter(augmented value) {}
```
Similarly for 8 locations below.

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t18_lib.dart
```diff
@@ -0,0 +1,162 @@
...
+/// identifier:
+/// ...
+/// - Augmenting setters: Within an augmenting setter `augmented` must be
+///   followed by an `=` and will directly invoke the augmented setter. If
+///   augmenting a field with a setter, this will invoke the implicit setter
+///   from the augmented field.
+///
+/// @description Checks that it is a compile-time error to declare local
+/// variable named `augmented` within an augmenting setter. Test an object
+/// pattern.
```
eernstg, line 16: This description is again different from the one in LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t18.dart. Is that intended? I'll assume that it is the one in `...A02_t18.dart` that is valid.

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t18_lib.dart
```diff
@@ -0,0 +1,162 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmented_expression_A02_t18.dart';
+
+augment void set topLevelSetter(String value) {
+    augmented? x;
```
eernstg, line 24: I'm afraid I still don't see any parameters here whose type has the name `augmented`. A local variable doesn't induce a setter, so we don't even have an implicitly induced setter whose parameter has that type.

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t18_lib.dart
```diff
@@ -0,0 +1,162 @@
...
+//  ^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    var f = (augmented x) {};
```
eernstg, line 28: Right, we do want to test that it is an error to use `augmented` as the name of a type in an augmenting declaration, also in the case where it is used to specify the parameter type of a function literal. But I don't see the connection to the parameter type of an augmenting setter declaration. So perhaps `(augmented x) {}` should occur in some other test?

LanguageFeatures/Augmentation-libraries/augmented_expression_A03_t21.dart
```diff
@@ -0,0 +1,57 @@
...
+///   in an initializer expression, and refers to the original field's
+///   initializer expression, which is immediately evaluated.
+///
+/// If augmented refers to a variable declaration (as defined by a declaration
+/// and a number of prior augmentations) with no initializer expression, and the
+/// variable's type is nullable, augmented evaluates to null. If the variable's
+/// type is not nullable, then it's a compile-time error.
+///
+/// @description Checks that it is a compile-time error if an augmenting field
+/// has `augmented` type name.
```
eernstg, line 19: Perhaps?:
```suggestion
/// has has a type whose name is `augmented`.
```

eernstg: Same issue in a few other libraries.

LanguageFeatures/Augmentation-libraries/augmented_expression_A04_t20_lib.dart
```diff
@@ -0,0 +1,160 @@
...
+/// @description Checks that it is a compile-time error to use type named
+/// `augmented` in an augmenting function body.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmented_expression_A04_t20.dart';
+
+augment void topLevelFunction() {
+  augmented? x;
```
eernstg, line 22: Yes, in this test it's perfectly well justified! All usages of the word `augmented` as a type should be flagged as an error in here.

Also, this is the place where `(augmented? x) {}` fits.

I suspect line 26 is perfect here, but shouldn't occur in any of those several other tests where it does appear (because those other tests are about other things, if I understand their description correctly).

eernstg (commented):
Looks good! However, I have commented on several tests where the description doesn't seem to be in sync (`*A0?_t??.dart` and `A)?_t??_lib.dart` have different descriptions), and also a few cases where the test doesn't do what the description says.

I added several comments, PTAL.

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t19.dart
sgrekhov: The test checks that it is a compile-time error to use a class named `augmented` in a getter body. Updated the description accordingly

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t19_lib.dart
sgrekhov: It should! Updated, thank you!

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t19_lib.dart
sgrekhov: It was just a typo caused by a copy-paste. The test checka use of a type with the name `augmented` in a body of a getter

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t18.dart
sgrekhov: No, we tested it in a previous test (`...A02_t17.dart`). Here we are testing use a class  `augmented` in a body of a setter. Description updated

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t18_lib.dart
sgrekhov: Fixed

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t18_lib.dart
sgrekhov: There is no connection with the parameter, the description was wrong (result of the copy-paste). Now description is updated.

sgrekhov (commented):
Lots of wrong tests descriptions as a result of copy-paste. Sorry for that. Updated. Please take another look

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t19.dart
eernstg: OK, thanks!

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t18.dart
eernstg: Makes sense, thanks!

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t18_lib.dart
eernstg: (Fixed in description)

LanguageFeatures/Augmentation-libraries/augmented_expression_A04_t20_lib.dart
eernstg: (All fixed in description)

eernstg (approved):
LGTM!

PR #2723, opened 2024-06-25
#2559. Add augmented expression tests

sgrekhov:


eernstg (approved):
Good catch, I hadn't thought about several of those cases!

LGTM.

eernstg (commented):
There was a handful of syntax errors after all: An `augment extension` cannot have an `on` clause.

sgrekhov:
> An augment extension cannot have an on clause.

Forgot about it. Thank you for noticing! Fixed.

eernstg (approved):
LGTM

PR #2730, opened 2024-06-28
Fixes #2529. Add test checking wildcard parameter with no default value

### LINKED ISSUE #2529: Remove LibTest containg compile-time errors

sgrekhov:


LanguageFeatures/Wildcards/other_declarations_A05_t45.dart
```diff
@@ -0,0 +1,142 @@
...
+//                          ^
+// [analyzer] unspecified
+// [cfe] unspecified
+  void instanceMethod2({int _}) {}
+//                          ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+exnetsion type ET(int i) {
```
eernstg, line 84:
```suggestion
extension type ET(int i) {
```

eernstg (commented):
Looks good, with a single typo.

I'm wondering about the named parameters, though: They will give rise to a compile-time error because named parameters can't have a name that starts with `_`, no matter whether they're `required` or optional, and no matter their type.

(The current specification has it [here](https://github.com/dart-lang/language/blob/314434dc1bdea2cba38a74b0dbe497bbe4473755/specification/dartLangSpec.tex#L2416), and the null-safety update https://github.com/dart-lang/language/pull/2605 makes the necessary adjustment that it applies to all named parameters including the `required` ones).

So you could claim that the named-parameter cases do not test the topic of the description.

Perhaps just mention this somewhere (in the description or in a separate comment), and then omit the named cases?

It is also possible that we actually do get the error about the need for a default value (with several errors in the same location you never know exactly which ones we'll get), but it seems likely to me that the actual error message would be about `_` being a bad name, not about the missing default value.

sgrekhov:
Ah, thank you! I forgot that named paremeter started with `_` is an error itself. We have appropriate tests, so I simply removed named parameters from this test. PTAL

eernstg (approved):
LGTM!

PR #2750, opened 2024-07-03
#2559. Add more test cases for augmented expression

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmented_expression_A03_t28.dart
```diff
@@ -0,0 +1,81 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'augmented_expression_A03_t28_lib.dart';
+import '../../Utils/expect.dart';
+
+String augmented() => "Top-level augmented(), shouldn't be invoked";
+
+String foo() => "Original";
```
eernstg, line 27: I think it would be useful for `foo` to have a side effect (incrementing a counter or something like that), such that we can see that it evaluates the initializing expression. This would show that we aren't just reading the value of a storage location, we are actually evaluating the initializing expression.

But this means that `foo` should not be a tear-off, it should be a getter that returns a function object (or we could just make it a getter and then omit `()` from all the occurrences of `augmented` below).

We might want to have at least a couple of cases where `augmented` or `augmented()` is invoked more than once, again to detect that a non-first evaluation is still an evaluation of the expression, not a stored value.

LanguageFeatures/Augmentation-libraries/augmented_expression_A03_t29.dart
```diff
@@ -0,0 +1,81 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'augmented_expression_A03_t29_lib.dart';
+import '../../Utils/expect.dart';
+
+String augmented() => "Top-level augmented(), shouldn't be invoked";
+
+String foo() => "Original";
```
eernstg, line 27: Same considerations about side effects.

LanguageFeatures/Augmentation-libraries/augmented_expression_A03_t29_lib.dart
```diff
@@ -0,0 +1,97 @@
...
+/// ...
+/// - Augmenting fields: Within an augmenting field, augmented can only be used
+///   in an initializer expression, and refers to the original field's
+///   initializer expression, which is immediately evaluated.
+///
+/// It is a compile-time error to use augmented in an augmenting field's
+/// initializer if the member being augmented is not a field with an initializer
+///
+/// @description Checks that within an augmenting field `augmented` invokes the
+/// original field's initializer expression. Test a local function.
```
eernstg, line 17:
```suggestion
/// original field's initializer expression. Test a function literal.
```

LanguageFeatures/Augmentation-libraries/augmented_expression_A04_t26.dart
```diff
@@ -0,0 +1,80 @@
...
+/// @assertion The exact result of an `augmented` expression depends on what is
+/// being augmented, but it generally follows the same rules as any normal
+/// identifier:
+/// ...
+/// - Augmenting functions: When augmenting a function, `augmented` refers to
+///   the augmented function. Tear offs are not allowed, so this function must
+///   immediately be invoked.
+///
+/// @description Checks that within an augmenting function `augmented()` invokes
+/// the augmented function with appropriate default values.
```
eernstg, line 14: Good catch!

We also have another rule about default values: An augmenting function or constructor declaration cannot declare default values, they _must_ all be declared in the original declaration. So we should make sure the augmenting declaration will itself have the proper default value (in this case it would presumably be invoked from a 2nd augmenting declaration that calls `augmented()`).

LanguageFeatures/Augmentation-libraries/augmented_expression_A04_t26_lib.dart
```diff
@@ -0,0 +1,53 @@
...
+/// @description Checks that within an augmenting function `augmented()` invokes
+/// the augmented function with appropriate default values.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmented_expression_A04_t26.dart';
+import '../../Utils/expect.dart';
+
+augment String topLevelFunction1() => augmented();
```
eernstg, line 22: I think this declaration _must_ declare the same formal parameters, but the must also omit the default values. That is, `([String value])`, which will be turned into `([String value = "default"])` by the merging process (so there's no error because of the missing default value).

https://github.com/dart-lang/language/blob/main/working/augmentation-libraries/feature-specification.md#augmenting-functions

eernstg (commented):
Looks good!

I commented on a couple of typos in descriptions.

A bigger thing is that some tests using `augmented` in an augmenting variable declaration do not show that we're actually evaluating the initializing expression from the augmented declaration each time we evaluate `augmented` (that is, we don't just initialize a variable somewhere and then reuse the value). I think it would be great if those tests could be enhanced to detect the side effects.

One test omits formal parameter declarations in the augmenting declarations even though the augmented declaration is a function with some parameters. That would be a compile-time error, but no errors are expected. I think those parameters should just be added.

There is also a rule about default values: If the original declaration has a named parameter `String value = "default"` and a 1st augmenting declaration has `String value` then an invocation from a 2nd augmenting declaration should be able to show that the 1st augmenting declaration actually has the proper default value (which is a copy of the one in the original declaration).

Finally, we have a funny case where the original declaration has `String v = "x";` and we have `@someMetadata augment String v;`. That should be an error according to the spec ("Since the initializer expression is the only meaningful part of the augmenting declaration, an initializer expression must be provided"), but I created https://github.com/dart-lang/language/issues/3957 in order to suggest that it should not be an error.

So we might want to cover that case now (as an error), just so we don't forget to cover it entirely. We probably wouldn't forget to update it because an update to the specification would give rise to changes to the implementation, and then we'll get the failing tests as a reminder.

Another case that doesn't seem to be covered is the case where an original variable declaration is augmented by a 1sh level of augmentation declarations augmenting the getter, or the setter, or both, and a 2nd augmenting variable declaration changes the initializing expression. (Or, modulo #3957, adds a comment or metadata.)

LanguageFeatures/Augmentation-libraries/augmented_expression_A04_t26_lib.dart
sgrekhov: Ah, of course. Just a typo, thank you for noticing. Fixed.

LanguageFeatures/Augmentation-libraries/augmented_expression_A04_t26.dart
sgrekhov: Added `augmented_expression_A04_t32.dart` testing default value after the seconf augmentation.

sgrekhov (commented):
Updated, additional tests added. PTAL

LanguageFeatures/Augmentation-libraries/augmented_expression_A03_t28.dart
eernstg: Thanks!

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t26.dart
```diff
@@ -0,0 +1,67 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'augmented_expression_A01_t26_lib.dart';
+import '../../Utils/expect.dart';
+
+String get augmented => "Global getter, shouldn't be invoked";
+int counter = 1;
+String get topLevelGetter => "Original ${counter++}";
```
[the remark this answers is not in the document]
sgrekhov: Looks like writing of this tests was  not finished. Change it to actually test evaluating of the `augmented` more than once and dropped side effects. Thank you for noticing!

LanguageFeatures/Augmentation-libraries/augmented_expression_A03_t28.dart
```diff
@@ -19,12 +19,13 @@
 
 // SharedOptions=--enable-experiment=macros
 
-import augment 'augmented_expression_A03_t28_lib.dart';
+//import augment 'augmented_expression_A03_t28_lib.dart';
```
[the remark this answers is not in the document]
sgrekhov: That was an accident. Uncommented, thank you!

LanguageFeatures/Augmentation-libraries/augmented_expression_A03_t31.dart
```diff
@@ -0,0 +1,86 @@
...
+/// - Augmenting fields: Within an augmenting field, augmented can only be used
+///   in an initializer expression, and refers to the original field's
+///   initializer expression, which is immediately evaluated.
+///
+/// It is a compile-time error to use augmented in an augmenting field's
+/// initializer if the member being augmented is not a field with an initializer
+///
+/// @description Checks that within an augmenting field `augmented` invokes the
+/// original field's initializer expression. Test a  case when `augmented` is
+/// called more than once.
```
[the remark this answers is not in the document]
sgrekhov: Yes, there is. It's `augmented_expression_A03_t01.dart`. Replaced it by the current test and removed this one.

sgrekhov (commented):
Updated once again, all recommendations are implemented. PTAL

LanguageFeatures/Augmentation-libraries/augmented_expression_A04_t26.dart
eernstg: Thanks!

eernstg (approved):
LGTM!

eernstg:
Thanks for tolerating all the churn! ;-)

PR #2761, opened 2024-07-08
#2559. Update assertion text and rename some tests

sgrekhov:
Tests that renamed to `*_A10_*` check another statement in the spec. So they were renamed and assertion text updated. There will be more tests. We need tests for initializing expression, for example.


eernstg (approved):
LGTM

PR #2765, opened 2024-07-09
#2559. Remove erroneous test cases, add more tests for augmented expression

sgrekhov:
In case of enum `final f = () {...};` is not a constant exptession and therefore an error. Removed.


eernstg (approved):
LGTM

(Fixed a typo 'ans' --> 'and' in two locations)

sgrekhov:
Thank you!

PR #2767, opened 2024-07-11
#2559.Update augmented expression tests for variable

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmented_expression_A03_t12.dart
```diff
@@ -0,0 +1,81 @@
...
+  static String get augmented => "Ext.augmented, shouldn't be invoked";
+}
+
+class MA = Object with M;
+
+main() {
+  Expect.equals("1: Original 1, 2: Original 2", topLevelVariable);
+  Expect.equals("1: Original 3, 2: Original 4", finalTopLevelVariable);
+  Expect.equals("1: Original 5, 2: Original 6", C.staticVariable);
+  Expect.equals("1: Original 6, 2: Original 8", C.finalStaticVariable);
```
eernstg, line 70: Really?:
```suggestion
  Expect.equals("1: Original 7, 2: Original 8", C.finalStaticVariable);
```

eernstg (commented):
Looks good!

I noted one thing that might be a typo.

eernstg (approved):
LGTM

eernstg:
Just wondering: We used to have an assertion containing this:

/// If augmented refers to a variable declaration (as defined by a declaration
/// and a number of prior augmentations) with no initializer expression, and the
/// variable's type is nullable, augmented evaluates to null. If the variable's
/// type is not nullable, then it's a compile-time error.

and the new text in the assertion is this:

/// It is a compile-time error to use `augmented` in an augmenting field's
/// initializer if the member being augmented is not a field with an initializer

I can find the latter in the spec, but not the former. Do you remember where it came from? It almost looks like there was a change whereby https://github.com/dart-lang/language/issues/3725 was accepted, and then it was later undone (by changing from the old to the new assertion). Do you remember what happened in this area?

sgrekhov:
At least I found it. I came not from the spec but from the discussion. See https://github.com/dart-lang/language/issues/3725#issuecomment-2077120808. Seems that I decided that spec will be soon  updated as Lasse suggested and I used this sentense as part of the specification. But I was wrong, sorry!

eernstg:
Aha, thanks!!

I suspect that dart-lang/language#3725 might be accepted, but we currently have that other rule, so it's fine that the tests are strictly based on the current specification and then we'll revisit them if dart-lang/language#3725 is accepted.

(It is somewhat inconvenient that the analyzer team will need to change this part of the implementation multiple times, because they are already doing what dart-lang/language#3725 proposes, but I think it's worse to have testing that contradicts the spec, even if the spec is perhaps going to change soon).

PR #2768, opened 2024-07-12
#2559. Update augmenting declarations tests

sgrekhov:
Assertions updated according to the current spec. New tests added.


LanguageFeatures/Augmentation-libraries/augmenting_declarations_A03_t01.dart
```diff
@@ -0,0 +1,30 @@
...
+/// @assertion The augmentation application order for a single base
+/// declaration’s (validly positioned) augmentation declarations is then in
+/// before order: An augmentation declaration is applied after any augmentation
+/// declarations that are before it, and before augmentation declarations that
+/// it is before.
+///
+/// This applies both to top-level declarations and to member declarations of,
+/// for example, class declarations.
+///
+/// @description Checks augmentation application order.
```
eernstg, line 14: It is not obvious to me that this test would reveal that the order is wrong in any particular way. Is there an ordering dependency that I've missed?

eernstg: Aha, I think I can see it now: We're checking that the traversal of the tree of library augmentations (soon to be parts) uses the pre-order depth-first left-to-right ordering as it visits all those files, it doesn't use some other order (because we would then have some misordered pair of an introductory declaration and an augmenting declaration).

I think it would be helpful to add a sentence or two about this in the description.

eernstg (commented):
Looks good!

Just one tiny thing: There is one test where I couldn't see that it was testing anything. I think it would be useful to extend the description a bit.

sgrekhov:
I've updated the description. PTAL

eernstg (approved):
LGTM!

PR #2773, opened 2024-07-15
#2559. Update operators tests according to the new spec

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmented_expression_A05_t01.dart
```diff
@@ -6,11 +6,11 @@
 /// being augmented, but it generally follows the same rules as any normal
 /// identifier:
 /// ...
-/// - Augmenting operators: When augmenting an operator, `augmented` must be
-///   followed by the operator. For example when augmenting `+` you must do
-///   `augmented + 1`, and when augmenting `[]` you must do `augmented[<arg>]`.
-///   These constructs invoke the augmented operator, and are the only valid
-///   uses of `augmented` in these contexts.
+/// - Augmenting operators: When augmenting an operator, augmented refers to the
```
eernstg, line 9: Unfortunately, this will come up in many locations&mdash;but I think it would be helpful to future readers that the first and the second occurrence of the word 'augmented' are clearly identifier: The first one is a reference to the syntactic work 'augmented', which is usually written as `augmented`, and the second one is the semantic notion of "the declaration which is being declaration".

```suggestion
/// - Augmenting operators: When augmenting an operator, `augmented` refers to the
```

eernstg (commented):
Looks good!

Just one single thing (.. though, unfortunately, it occurs in about 50 places ..): I think we should preserve the backticks around the word `augmented` in a snippet of 'assertion' text that came from the feature spec.

sgrekhov:
Assertion text updated. PTAL

eernstg (approved):
LGTM, thanks!

PR #2774, opened 2024-07-16
#2559. Add type inheritance tests. Part 1

sgrekhov:


LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t01_lib.dart
```diff
@@ -0,0 +1,128 @@
...
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'type_inheritance_A02_t01.dart';
+
+augment int get topLevelGetter => 0;
+//      ^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+augment Object topLevelFunction() => 0;
```
eernstg, line 29: Very good! It's a subtype in line 25 and a supertype here, and we should keep an eye on both.

It's probably not going to catch any bugs, but we could have a case where the augmented return type and the augmenting return type are unrelated (like `num` and `String`).

LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t04.dart
```diff
@@ -0,0 +1,75 @@
...
+///
+/// When applying an augmenting declaration that contains a type annotation at
+/// one of these positions, to a definition to be augmented, it's a compile-time
+/// error if the type denoted by the augmenting declaration is not the same type
+/// as the type that the augmented definition has at the corresponding position.
+/// An augmenting declaration can omit type annotations, but if it doesn't, it
+/// must repeat the type from the augmented definition.
+///
+/// @description Check that it is a compile-time error if augmenting declaration
+/// specifies a different return type than the augmented declaration.Test a
```
eernstg, line 18: Typo
```suggestion
/// specifies a different return type than the augmented declaration. Test a
```

eernstg (commented):
Looks good! One typo, just a missing space.

The main comment is that it would be helpful to test the actual inherited type, not just the returned value.

The point is that there could be a bug whereby the augmenting declaration gets to decide (so `int i; augment var i;` causes `i` to have type `dynamic` because a lone `var j;` causes `j` to have type `dynamic`), or the resulting type is just the least upper bound of all the types encountered, or the greatest lower bound.

So it would be great if we can directly check that the augmenting declaration gets a signature with the expected types (return type of getter and method, type of variable, parameter types of method and setter).

My suggestion is that we can use `expectStaticType` to do this at a call site (for the return type) or in the body of the augmenting declaration (for parameter types).

sgrekhov:
Updated. PTAL

eernstg (approved):
LGTM!

PR #2775, opened 2024-07-17
#2559. Add type inheritance tests. Part 2

sgrekhov:


LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t05.dart
```diff
@@ -0,0 +1,63 @@
...
+/// @description Check that it is a compile-time error if augmenting declaration
+/// specifies a different return type than the augmented declaration. Test a
+/// wrong top type.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'type_inheritance_A02_t05_lib.dart';
+
+topLevelFunction() {}
```
eernstg, line 26: Don't we have a test somewhere to verify that the return type of a top-level function with no explicit return type is inferred as `dynamic`? This would be a test that has nothing to do with augmentations, because it's a much more basic property.

With that in place, I think it would be safer to declare the return type explicitly here: We want to verify that two different top types (still syntactically different after normalization, that is) will give rise to a compile-time error, and the safest way we can conclude that this is actually what we are testing is to declare the return type explicitly everywhere.

eernstg: Thinking about this again, we do have a special (augmentation-related) rule specifying that the type of the introductory declaration should be computed based on the introductory declaration alone, which means that we _do_ rely in a crucial way on inferring the types that are omitted.

So let's keep the version that omits the return type etc.

But I think we should have a direct test of explicitly different return types as well. It might be somewhere among the tests in this PR, I haven't seen all of it yet, otherwise I think it would be helpful to add it.

LanguageFeatures/Augmentation-libraries/type_inheritance_A03_t03.dart
```diff
@@ -0,0 +1,84 @@
...
+  Expect.equals("Augmented", MA().instanceVariable);
+  Expect.equals("Augmented", MA().finalInstanceVariable);
+  Expect.equals("Augmented", E.staticVariable);
+  Expect.equals("Augmented", E.finalStaticVariable);
+  Expect.equals("Augmented", E.e0.instanceVariable);
+  Expect.equals("Augmented", E.e0.finalInstanceVariable);
+  Expect.equals("Augmented", Ext.staticVariable);
+  Expect.equals("Augmented", Ext.finalStaticVariable);
+  Expect.equals("Augmented", ET.instanceVariable);
+  Expect.equals("Augmented", ET.finalInstanceVariable);
```
eernstg, line 82: One more case is `ET.id`.

LanguageFeatures/Augmentation-libraries/type_inheritance_A04_t03.dart
```diff
@@ -0,0 +1,160 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'type_inheritance_A04_t03_lib.dart';
+import '../../Utils/expect.dart';
+
+String _log = "";
+
+void topLevelFunction(var v) {
```
eernstg, line 28: I think it would be useful to verify that the inferred type of `v` is `dynamic`, because the parameter type isn't explicit. (I guess that could be something like `if (1 > 2) v.memberThatDoesNotExist;`)

Otherwise, for example, this test might break for reasons that aren't obvious if the language is updated at some point to use the type `Object?` rather than `dynamic` for parameters whose type isn't declared.

LanguageFeatures/Augmentation-libraries/type_inheritance_A04_t04.dart
```diff
@@ -0,0 +1,160 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'type_inheritance_A04_t04_lib.dart';
+import '../../Utils/expect.dart';
+
+String _log = "";
+
+void topLevelFunction(v) {
```
eernstg, line 28: Again `if (1 > 2) v.memberThatDoesNotExist;`.

LanguageFeatures/Augmentation-libraries/type_inheritance_A04_t05.dart
```diff
@@ -0,0 +1,74 @@
...
+///
+/// @description Check that it is a compile-time error if augmenting declaration
+/// specifies a formal parameter type other than the augmented declaration.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+import augment 'type_inheritance_A04_t05_lib.dart';
+
+void topLevelFunction(v) {}
```
eernstg, line 25: `if (1>2) v.memberThatDoesNotExist;` would help here, too.

eernstg (commented):
Looks good!

I think it would be helpful to verify the inferred types in a number of situations where a return type or parameter type or variable type hasn't been declared, but is inferred. It is a safer bet to say that we're testing that it is an error for one type to be `dynamic` and another type to be `Object?` if we know for sure that the former one is actually `dynamic`. If the first one was inferred as `Object?` in the first place then we'd get a confusing error situation (it looks like 'missing compile-time error', but the actual problem was that we had a wrong assumption about which type is inferred in some situation where that type isn't specified explicitly).

So I'm suggesting that we add local type verification code (`expectStaticType` and such).

LanguageFeatures/Augmentation-libraries/type_inheritance_A03_t03.dart
sgrekhov: This test checks the case `var x = 1;` and then `augment int x = 2;`. We can't have `extension type ET(var/final id) {...}` and then to augment it by the exact type. So, I believe, this is not applicable to this test. But I added the cases to `type_inheritance_A01_t*.dart` checking augmenting of representation variable and getter. Thank you!

I also want to add some more tests for extension types checking augmentation of the representation variable but as a separate tests in the next PR. it seems to me that it slightly different case which may be handled separately. Therefore I want to have a separete tests for it, not just one test case among dozens of similar checks.

Thank you for pointing!

LanguageFeatures/Augmentation-libraries/type_inheritance_A04_t03.dart
sgrekhov: Added

LanguageFeatures/Augmentation-libraries/type_inheritance_A04_t04.dart
sgrekhov: Added

LanguageFeatures/Augmentation-libraries/type_inheritance_A04_t05.dart
sgrekhov: Added

LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t05.dart
sgrekhov: > Don't we have a test somewhere to verify that the return type of a top-level function with no explicit return type is inferred as dynamic?

https://github.com/dart-lang/co19/blob/master/Language/Functions/Type_of_a_Function/return_type_t01.dart does this. But it can be improved. I'll do that.

sgrekhov (commented):
Updated. Thank you for pointing to representation variable of an extension type. I didn't thought of it. I'm going to add a bunch of separate tests fro it in a separate PR.

LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t05.dart
eernstg: Sounds good!

LanguageFeatures/Augmentation-libraries/type_inheritance_A04_t03.dart
eernstg: Thanks!

LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t05.dart
eernstg: I took another look, is it true that there is no test where it is detected that it is a compile-time error to have an introductory function declaration with explicitly declared return type `T1` and an augmenting declaration with the same name and explicitly declared return type `T2`, where `T1` and `T2` are not the same type (according to the usual test)? For example, `T1` could be `List<int>` and `T2` could be `Map<int, int>`.

It seems reasonable that we should have a test that will continue to succeed (that is, the compile-time error is still reported) if the language is modified such that the inferred return type of one or more of these function declarations is different from today (for example, if `topLevelFunction() {}` is inferred to have return type `void` rather than `dynamic`).

LanguageFeatures/Augmentation-libraries/type_inheritance_A03_t03.dart
eernstg: I was thinking that the extension type has a single, final instance variable which is the representation variable.

So it would presumably be possible to augment the representation variable by a getter.

```dart
extension type ET(int i) {
  augment get i => augmented + 1;
}
```

Similarly, it should be possible to augment it by a variable declaration that adds metadata.

```dart
extension type ET(int i) {
  @metadata
  augment var i;
}
```

I'm just about to go home, so I haven't checked anything ;-), but I thought it could be a corner which isn't otherwise covered.

eernstg (commented):
Couple of things...

sgrekhov:
Added test cases for augmenting extension type representation variable. Please review.

eernstg (approved):
LGTM

PR #2776, opened 2024-07-18
#2559. Add more augmented declarations tests. Update assertions.

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_declarations_A04_t05.dart
```diff
@@ -0,0 +1,50 @@
...
+class A {
+  String method() => "Method";
+  String get getter => "Getter";
+  void set setter(String _) {}
+  String operator +(Object? other) => "Operator";
+}
+
+extension type ET(A id) implements A {}
+
+augment extension Ext {
```
eernstg, line 29: I think this should be an extension type augmentation, not an extension augmentation.

eernstg (commented):
Looks good! Just one thing: It looks like an `extension` declaration should have been an `extension type` declaration.

LanguageFeatures/Augmentation-libraries/augmenting_declarations_A04_t05.dart
sgrekhov: Of course! Thank you! Fixed

eernstg (approved):
LGTM.

PR #2777, opened 2024-07-18
#2559. Update assertions in augmented expressions tests

sgrekhov:
Only assertions text is updated according to the changed spec


eernstg (approved):
LGTM

Thanks for keeping it simple such that it is easy to see what's going on, even with almost 200 modified files!

PR #2779, opened 2024-07-18
#2559. Update assertions in augmenting types tests. Add more cases

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_types_A01_t01.dart
```diff
@@ -2,28 +2,42 @@
...
 // BSD-style license that can be found in the LICENSE file.
 
-/// @assertion It is a compile-time error if:
-///
-/// - The augmenting type and corresponding type are not the same kind: class,
-///   mixin, enum, or extension. You can't augment a class with a mixin, etc.
+/// @assertion It's a compile-time error if a library contains two top-level
+/// declarations with the same name, and:
+/// - Neither is an augmenting declaration, or
+/// - one of the declarations is a class-like declarations and the other is not
```
eernstg, line 8: Yup, it actually says 'a class-like declarations'. OK, so we keep it just like the feature spec here, and then it'll be fixed later.

eernstg (approved):
LGTM

PR #2791, opened 2024-07-26
#2559. Add augmenting variables tests. Part 1

sgrekhov:


sgrekhov:
Spec parser is not happy here but it looks like a spec parser issue

LanguageFeatures/Augmentation-libraries/augmenting_variables_A05_t04_lib.dart
```diff
@@ -0,0 +1,47 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmenting_variables_A05_t04.dart';
+
+augment dynamic topLevelVariable;
+
+augment class C {
+  augment static Object? staticVariable;
```
chloestefantsova, line 27: What is the difference between the variable in this line and `C.staticVariable` from `LanguageFeatures/Augmentation-libraries/augmenting_variables_A05_t03_lib.dart`, where it's marked as a compile-time error. I might miss something, but they look the same and should possibly be treated similarly by the compiler.

chloestefantsova (changes requested):

LanguageFeatures/Augmentation-libraries/augmenting_variables_A05_t04_lib.dart
sgrekhov: It was a copy-paste issue. There should be `dynamic`, of course. Thank you for noticing. Fixed. PTAL

chloestefantsova (approved):

PR #2795, opened 2024-07-31
#2559. Representation variable cannot be augmented

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_getters_A01_t02.dart
```diff
@@ -80,17 +80,14 @@ augment extension Ext {
   }
 }
 
-extension type ET(String instanceVariable) {
+extension type ET(String _) {
   static String staticVariable = "Original";
 }
 
-augment extension type ET(String instanceVariable) {
+augment extension type ET(String _) {
```
eernstg, line 87: This will be a syntax error, shouldn't we expect an error somewhere?

eernstg (commented):
One case needs to be adjusted, or an error should be expected.

sgrekhov:
Updated. PTAL

eernstg (approved):
LGTM

PR #2798, opened 2024-08-01
#2559. Add `augmented` expression tests for control-flow elements

sgrekhov:


eernstg (commented):
I'm afraid we can't land this directly: It embodies the assumption that it is an error to use `augmented` as a regular identifier in some cases where Lasse wishes to allow it. I created https://github.com/dart-lang/language/issues/4009 in order to resolve that discussion.

What's more practical?: Delaying this PR entirely until 4009 has been resolved, or splitting it in two (1: `augmented` is only used in a member declaration which is augmented or in a context where no declaration at all is augmenting; and 2: `augmented` is used inside a non-augmenting member declaration inside an augmenting declaration).

sgrekhov:
Let's wait for 4009 here. Thank you!

eernstg:
Very good!

eernstg:
I used the label 'status-blocked' to characterize the situation. Does that work well for you?

sgrekhov:
Yes, perfect!

eernstg (approved):
LGTM.

So nice to get this topic settled!

LanguageFeatures/Augmentation-libraries/augmented_expression_A10_t21_lib.dart
```diff
@@ -0,0 +1,543 @@
...
+      if (1 > augmented) 42
+//            ^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    };
+    return 0;
+  }
+}
+
+extension type ET {
```
eernstg, line 457: Wouldn't this be?:
```suggestion
extension type ET(int _) {
```

LanguageFeatures/Augmentation-libraries/augmented_expression_A10_t22_lib.dart
```diff
@@ -0,0 +1,454 @@
...
+    var set = {
+      if (1 > augmented) 42
+//            ^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    };
+  }
+}
+
+augment extension type ET {
```
eernstg, line 369:
```suggestion
augment extension type ET(int _) {
```

eernstg (commented):
There was a typo after all.

eernstg (commented):
Thanks! I think one `augment extension type` typo is still there.

eernstg:
Funny. I can see `augment extension type ET {` in LanguageFeatures/Augmentation-libraries/augmented_expression_A10_t22_lib.dart, but the checks, including the spec parser, claim that there are no problems.

sgrekhov:
> Funny. I can see `augment extension type ET {` in LanguageFeatures/Augmentation-libraries/augmented_expression_A10_t22_lib.dart, but the checks, including the spec parser, claim that there are no problems.

Why `augment extension type ET {` should cause any problems? It's a compile-time error to specify a primary constructor (representation field) in the extension type augmentation.

eernstg:
Ah, I'm sorry, the grammar does indeed require that the representation declaration is omitted.

I couldn't remember, and then I grepped for `augment extension type` in LanguageFeatures/Augmentation-libraries and found 116 of them, every one of them including the `(T v)` part, so I thought that they had to be there. But I can see that some of them are actually testing that it's an error.

So this means that all the remaining ones will need to be updated by removing the representation declaration? Ex: augmenting_functions_A06_t01_lib.dart, type_inheritance_A03_t02_lib.dart, type_inheritance_A04_t02_lib.dart.

(I'm landing this one now.)

LanguageFeatures/Augmentation-libraries/augmented_expression_A10_t22_lib.dart
eernstg: Wrong suggesting, closing the thread.

eernstg (approved):

PR #2799, opened 2024-08-01
#2559. Add augmenting members tests. Part 2

sgrekhov:


sgrekhov:
@eernstg @chloestefantsova please review

LanguageFeatures/Augmentation-libraries/augmenting_members_A02_t01_lib.dart
```diff
@@ -0,0 +1,69 @@
...
+}
+
+augment extension Ext {
+  augment static void set staticVariable(String _) {}
+//^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+augment extension type ET {
```
eernstg, line 64: Here we could also add a setter that matches up with the representation variable.

LanguageFeatures/Augmentation-libraries/augmenting_members_A02_t02.dart
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if:
+/// ...
+/// A non-writable variable declaration is augmented with a setter.
+///
+/// @description Checks that it is a compile-time error if a non-writable
+/// variable declaration is augmented with a setter. Test `late final` variables
```
eernstg, line 10: Oh, that should not be an error: A `late final` variable with no initializing expression _does_ have a setter, so we should be able to augment that setter.

But it's true for a `late final` variable with an initializing expression. One way ahead would be to change the expected response, another could be to add an initializing expression to all these variables.

LanguageFeatures/Augmentation-libraries/augmenting_members_A03_t01.dart
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if:
+/// ...
+/// A non-final variable is augmented with a final variable.
```
eernstg, line 7: OK, this is one of those cases that we're discussing in https://github.com/dart-lang/language/issues/4017.

As usual, we will follow the current spec and update the test later if needed.

LanguageFeatures/Augmentation-libraries/augmenting_members_A04_t01.dart
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A final variable can be augmented with a non-final augmenting
+/// variable, and that will not add any setter. An augmenting variable
+/// declaration only affects the initializer expression, not setters.
+///
+/// @description Checks that a final variable can be augmented with a non-final
+/// augmenting variable.
```
eernstg, line 10: Again, I find it difficult to explain why certain things are allowed and other things are errors in this domain. But the outcome is that final/non-final makes no difference in this case at all, the only thing we're using from the augmenting declaration is the initializing expression.

eernstg (commented):
I really don't know where this is going, I hope we can arrive at a clearer model in, for instance, https://github.com/dart-lang/language/issues/4017.

However, right now we'll follow the current spec wording as usual.

Please take a look at the comments. Most of them are background material. Then I'll take another look at the whole thing tomorrow and then we can take the next step.

LanguageFeatures/Augmentation-libraries/augmenting_members_A02_t01.dart
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if:
+/// ...
+/// A non-writable variable declaration is augmented with a setter.
+///
+/// @description Checks that it is a compile-time error if a non-writable
+/// variable declaration is augmented with a setter.
```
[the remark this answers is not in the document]
sgrekhov: > I suppose we already cover that?
No, some cases were not covered. Added, thank you! Now all of these cases are covered.

LanguageFeatures/Augmentation-libraries/augmenting_members_A02_t01_lib.dart
sgrekhov: Added.

LanguageFeatures/Augmentation-libraries/augmenting_members_A02_t02.dart
sgrekhov: Indeed. Fixed by adding an initializer and also added a separate test that checks that it is not an error to augment `late final` variable with no initializer with a setter.

LanguageFeatures/Augmentation-libraries/augmenting_members_A02_t01.dart
```diff
@@ -0,0 +1,52 @@
...
+  final String instanceVariable = "instanceVariable";
+}
+
+class A {}
+
+extension Ext on A {
+  static final String staticVariable = "staticVariable";
+}
+
+extension type ET(String _) {
```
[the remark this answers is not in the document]
sgrekhov: Thank you very much for the interesting and detailed explanation! I didn't think about all of this!

sgrekhov (commented):
Thank you! Updated. PTAL.

eernstg:
This one is on my TODO list, but it is full of situations that give rise to questions that haven't been clarified. I'll look into it on Monday.

sgrekhov:
No problem, thank you!

LanguageFeatures/Augmentation-libraries/augmenting_members_A02_t01.dart
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if:
+/// ...
+/// A non-writable variable declaration is augmented with a setter.
+///
+/// @description Checks that it is a compile-time error if an implicitly induced
+/// getter of a final variable declaration is augmented with a setter.
```
eernstg, line 10: Note that there's nothing new about this, it is only targeting the situation where there is no setter to augment (in particular, there is no implicitly induced setter, but also no explicitly written setter).

So there's absolutely nothing in this error message other than the usual "can't augment something that doesn't exist" error.

LanguageFeatures/Augmentation-libraries/augmenting_members_A02_t01_lib.dart
eernstg: Thanks!

eernstg (approved):
OK, let's land it!

PR #2802, opened 2024-08-02
Fixes #2800. Change expectations for `IOSink.writeln()` tests

### LINKED ISSUE #2800: Assumption in tests about the implementation of writeln cause them to break when the implementation is changed
The following tests make assumptions about how `writeln` is implemented in the VM
```
co19/src/LibTest/io/IOSink/IOSink_class_A01_t01.dart
co19/src/LibTest/io/IOSink/writeln_A01_t01.dart
```
The VM currently implements `writeln`  in `IOSink` as

```
void writeln([Object? object = ""]) {
  write(object);
  write("\n");
}
```

but we have had some issues filed about how this is not atomic and causes issues when multiple isolates issue `writeln` requests, see https://github.com/dart-lang/sdk/issues/55345

If the implementation in the VM is changed to make the `writeln` atomic (https://dart-review.googlesource.com/c/sdk/+/378024) the above two tests break as it has assumptions about the implementation. Maybe the tests should be changed to not make such assumptions.

sgrekhov:
Now these tests assept both `write(object);write('\n');` and `write('$object\n');`

### DIFF
--- a/LibTest/io/IOSink/IOSink_class_A01_t01.dart
+++ b/LibTest/io/IOSink/IOSink_class_A01_t01.dart
@@ -7,15 +7,14 @@
 /// An [IOSink] combines a [StreamSink] of bytes with a [StringSink], and allows
 /// easy output of both bytes and text.
 /// Writing text ([write]) and adding bytes ([add]) may be interleaved freely.
+///
 /// @description Checks that the different targets can be added and written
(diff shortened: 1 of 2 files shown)


a-siva (approved):

eernstg (approved):
LGTM

PR #2805, opened 2024-08-06
Fixes #2804. Remove excessive flaky code from writeByte_A01_t01.dart

### LINKED ISSUE #2804: Incorrect logic in writeByte_A01_t01
https://github.com/dart-lang/co19/blob/388f80d26573f6a234da2d7ec8448998e10fb8a5/LibTest/io/RandomAccessFile/writeByte_A01_t01.dart#L19-L53

I think logic in this test is incorrect: 

- it starts 10 concurrent invocations of `check(i, list)` with `i = 0, ...,9`. 
- each of these invocations fills one element of the list at index `i`.
- but each invocations validates contents of `list` sequentially:
    ```dart
    for (int i = 0; i <= num; i++) {
      Expect.equals(i, list[i]);
    }
    ```

That does not seem correct - because there is no guarantee that these invocations complete sequentially. If that's expected we should introduce ordering by `await`ing `check` in `main`.

sgrekhov:
Removed not thread-safe buffer from this test.

### DIFF
--- a/LibTest/io/RandomAccessFile/writeByte_A01_t01.dart
+++ b/LibTest/io/RandomAccessFile/writeByte_A01_t01.dart
@@ -6,9 +6,9 @@
 /// Writes a single byte to the file. Returns a Future<RandomAccessFile> that
 /// completes with this RandomAccessFile when the write completes.
 ///
-/// @description Checks that method writeByte writes a single byte to the file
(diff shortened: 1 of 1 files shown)


mraleph (approved):

eernstg (approved):
LGTM

PR #2809, opened 2024-08-08
Fixes 2808. Add additional extension type constructor test

sgrekhov:


eernstg (approved):
LGTM

eernstg:
It's surprising that `WARNING` would be replaced by `HINT`, I thought we were moving in the opposite direction.

PR #2811, opened 2024-08-08
#2559. Add constructor augmentation tests. Part 1.

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_constructors_A01_t08_lib.dart
```diff
@@ -0,0 +1,40 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+  augment C1.foo([super.y, super.x]);
+//                      ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+augment class C2 {
+  augment C2({super.y, super.x});
```
eernstg, line 32: This one is completely benign, the ordering of named parameters doesn't matter. (As long as we don't have optionally named parameters ;-).

So it needs to be adjusted such that it has an inconsistency with respect to the "superness" of one or more parameters (as in `{super.x, super.y}` is augmented with `{int x, int y}`).

eernstg (commented):
Looks good, but a few things need to be adjusted.

sgrekhov:
Thank you!
- `augmenting_constructors_A01_t05.dart` updated to check the optionality only
- `augmenting_constructors_A01_t06.dart` deleted
- `augmenting_constructors_A01_t08.dart` - fixed.
Please take another look.

eernstg (approved):
LGTM

PR #2813, opened 2024-08-12
#2559. Add constructor augmentation tests. Part 2.

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_constructors_A04_t01_lib.dart
```diff
@@ -0,0 +1,30 @@
...
+/// @description Checks that it is a compile-time error if the introductory
+/// constructor is marked `factory` and the augmenting constructor is not.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmenting_constructors_A04_t01.dart';
+
+augment class C {
+  augment C.foo();
```
eernstg, line 19: Do we have a test showing that it is _not_ an error for this constructor to be minimal (e.g., because we just want to add metadata). For example, given that `factory C.foo() = C;` can be augmented by `@metadata augment factory C.foo() = C;`, can we also use `@metadata augment factory C.foo();`?

That would require a grammar change, in line with https://github.com/dart-lang/language/issues/4025.

Let's just stick to the feature specification for now, and update tests if 4025 causes changes to the spec.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A04_t02_lib.dart
```diff
@@ -0,0 +1,23 @@
...
+/// @description Checks that it is a compile-time error if the introductory
+/// constructor is marked `factory` and the augmenting constructor is not.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmenting_constructors_A04_t02.dart';
+
+augment class C {
+  augment C(): super(0);
```
eernstg, line 19: This should be a `factory` constructor, right?

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A04_t02_lib.dart
```diff
@@ -0,0 +1,23 @@
...
+//             ^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
```
eernstg, line 23: No augmentations of `ET`?

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A07_t01.dart
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if:
+/// ...
+/// - A non-redirecting constructor augments a constructor which is not
+///   potentially non-redirecting.
+///
+/// @description Checks that it is a compile-time error a non-redirecting
```
eernstg, line 10:
```suggestion
/// @description Checks that it is a compile-time error if a non-redirecting
```

eernstg: A similar typo exists in a few other places.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A07_t01_lib.dart
```diff
@@ -0,0 +1,20 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is a compile-time error if:
+/// ...
+/// - A non-redirecting constructor augments a constructor which is not
+///   potentially non-redirecting.
+///
+/// @description Checks that it is a compile-time error a non-redirecting
+/// constructor augments a constructor which is not potentially non-redirecting.
```
eernstg, line 11: There would be a `factory` case here, as well. Do we already test in some other test that it is an error to have a non-redirecting factory constructor which is augmented by a redirecting one, and vice versa?

eernstg (commented):
Looks good, couple of comments added!

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A04_t01_lib.dart
sgrekhov: No, we don't have such test yet. Keeping my eye on 4025.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A04_t02_lib.dart
sgrekhov: Right, thank you!

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A04_t02_lib.dart
sgrekhov: Added

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A07_t01_lib.dart
sgrekhov: Ups! Missed this. Added, thank you for catching this!

sgrekhov (commented):
Updated. PTAL

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A04_t01_lib.dart
eernstg: Good, thanks!

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A07_t01_lib.dart
eernstg: Got it, thanks!

eernstg (approved):
LGTM

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A02_t01_lib.dart
```diff
@@ -0,0 +1,82 @@
...
+//               ^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  augment C.c1({int x = 0});
+//                  ^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  augment factory C.c2([int x = 0]);
```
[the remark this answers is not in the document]
eernstg: The problem would be that we get a syntax error and then it would look like we got the static semantic error which is "can't augment a non-factory constructor with a factory" (that is, the test succeeds for the wrong reason).

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A03_t01_lib.dart
```diff
@@ -0,0 +1,38 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+augment extension type ET {
+  augment ET.foo(this.id);
+//        ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  augment factory ET.bar(int id);
```
eernstg, line 34: Same problem: Succeeds because we expect and error and get an error, but it is not the expected one.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A03_t02_lib.dart
```diff
@@ -0,0 +1,38 @@
...
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmenting_constructors_A03_t02.dart';
+
+augment class C {
+  augment const C();
+//              ^
+// [analyzer] unspecified
+// [cfe] unspecified
+  augment const factory C.foo();
```
eernstg, line 23: Similar to the other cases where we need a decision on dart-lang/language#4025.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A03_t02_lib.dart
```diff
@@ -0,0 +1,38 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+augment extension type ET {
+  augment const ET.foo(this.id);
+//              ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  augment const factory ET.bar(int id);
```
eernstg, line 34: Ditto.

eernstg:
A couple of things came up after all: We can't really tell at this point whether `const factory C();` should be a syntax error or not.

I'm not quite sure which way to go, perhaps a few test cases in this PR should be postponed until we know which way it goes.

eernstg (commented):
We have a couple of situations where the current specification makes something a syntax error (like a `const factory` constructor with no body and no redirection).

I'm not sure how we can best handle this situation. If the language team is happy about introducing special syntax for an omitted body then lots of things would need to change. If they don't do that then we will probably need to introduce/generalize the grammar rules such that it is possible to declare several kinds of things with no body where the body is required in the grammar today.

Perhaps the best approach right now would be to block the tests that need to omit the body, but where that is currently a syntax error, especially in the case where they will expect some other compile-time error (that is, not a syntax error).

sgrekhov:
Removed `constant factory` constructors tests from this PR. Will add them later after 4025.

eernstg:
Very good, thanks!

eernstg (approved):
LGTM

eernstg:
Ah, still a few locations where the same issue comes up: LanguageFeatures/Augmentation-libraries/augmenting_constructors_A02_t01_lib.dart has an `augment factory C.c2([int x = 0]);`, and more like that. I think there's nothing in any other library.

sgrekhov:
Indeed. Ok, augmenting factory constructors with no body removed from this PR. Will add them later. PTAL.

eernstg (approved):
LGTM

PR #2817, opened 2024-08-14
#2559. Add constructor augmentation tests. Part 4.

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_constructors_A10_t09_lib.dart
```diff
@@ -0,0 +1,37 @@
...
+/// in a non-augmenting constructor body.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmenting_constructors_A10_t09.dart';
+
+augment class C {
+  C() {
+    augmented();
```
eernstg, line 23: Depends on https://github.com/dart-lang/language/issues/4009, but I guess the consistent approach would be to follow the current spec text and then update all tests that are affected by a different spec wording, if the spec is updated on this topic.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A11_t01_lib.dart
```diff
@@ -0,0 +1,30 @@
...
+/// @description Checks that when `augmented()` is called in the body of an
+/// augmenting constructor initializer lists are not re-run.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmenting_constructors_A11_t01.dart';
+
+augment class C {
+  augment C(): x = "Augmented" {
```
eernstg, line 21: Check out 'Add initializers to the initializer list ...' in the augmentation spec.

I think the semantics of initializer lists is cumulative: The introductory constructor (if it can indeed have an initializer list) can introduce an initializer list.

If an initializer list has been introduced then augmenting constructor declarations can _add to_ the existing initializer list (otherwise it can introduce the initializer list, as usual).

But I would expect that it is an error to have one initializer list element that initializes `x` (in the introductory or an augmenting declaration) and then another one that also initializes `x` in a later augmentation. Just like `x = 1, x = 1` is an error in an initializer list which is written in one place.

So we want to make sure that _all_ elements of the initializer list of the constructor _definition_ are evaluated before the first body runs, and they are not re-evaluated later on. Side effects can show that this is indeed the case.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A11_t02_lib.dart
```diff
@@ -0,0 +1,30 @@
...
+/// @description Checks that when `augmented()` is called in the body of an
+/// augmenting constructor initializing formals are not re-run.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmenting_constructors_A11_t02.dart';
+
+augment class C {
+  augment C(this.x) {
```
eernstg, line 21: We can't do that because an initializing formal is concerned with instance variable _initialization_, and initialization never invokes a setter, it directly stores a value in the storage location which was allocated for the given variable.

This is important if a subclass introduces a `set` setter (that is, a setter which isn't implicitly induced by a variable declaration), because we still need to initialize the variable storage, even in the case where nobody else can write to that location because the setter "captures" all attempts to write to that location, and the setter doesn't write to that location itself. So initialization really needs to have a bit of magic power beyond what we can express in code that runs after the initialization has completed.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A11_t03.dart
```diff
@@ -0,0 +1,32 @@
...
+
+/// @assertion A non-redirecting generative constructor marked `augment` may:
+/// - Add or replace the body of the augmented constructor with a new body.
+/// ...
+///   - Initializer lists are not re-run, they have already executed and
+///     shouldn't be executed twice. The same goes for initializing formals and
+///     super parameters.
+///
+/// @description Checks that when `augmented()` is called in the body of an
+/// augmenting constructor super parameters are not re-run.
```
eernstg, line 13: The spec wording isn't too good here. I don't think we can say that a super parameter "runs". It's basically a syntactic sugar mechanism, and the meaning of a super-parameter declaration is a regular parameter plus a type (which may have been written or may come from the superclass) plus possibly a default value (written or from the superclass), and an actual argument in the superinitializer, `super.maybeAName(some, arguments)`.

I think we have to understand "the same goes for ... super parameters" as "a positional super-parameter will not be passed twice in the implicitly modified or induced superinitializer".

eernstg: This could again be tested by detecting that we don't get anything like "wrong number of positional parameters in superinitializer" when `super.x` occurs both in the introductory and an augmenting declaration, but also that it _is_ an error if the number of positional parameters to the superinitializer will only be _correct_ in the case where the super-parameter is passed twice (that is, we would then expect a compile-time error because there are too few positional parameters).

For named parameters we can only test the non-error case: If the same named parameter is passed twice then it's an error, so we just check that there is no error when the super-parameter is specified both in the introductory and an augmenting declaration.

eernstg (commented):
Looks good! Just one thing needs an extra check: I think the initializer list rules are a bit different than what's probably assumed in one test, and similarly for the treatment of super-parameters.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A11_t01_lib.dart
sgrekhov: Rewritten.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A11_t02_lib.dart
sgrekhov: Rewritten. Simply check that final variable is initialized by the augmenting constructor and `augmented()` doesn't try to reinitialize it. I don't think that anything else can be done to tests it.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A11_t03.dart
sgrekhov: Same

sgrekhov (commented):
Updated. PTAL.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A11_t01_lib.dart
eernstg: Thanks!

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A11_t02_lib.dart
eernstg: Right, we probably can't detect whether the value provided to a parameter declared as `this.x` is stored in `x` once or twice: The operation that stores a value in a variable does not have any side effects, and we can't change the value of the parameter declared as `this.x` in the initializer list (it is implicitly final). Otherwise, we could have assigned to it in an initializer list element:

```dart
// It works as long as we're assigning a parameter which isn't initializing.
class A {
  int i, j;
  Object? dummy;
  A(int i): this.i = i, dummy = (i = i + 1), j = i;
  toString() => 'A(i: $i, j: $j)';
}

// But not with an initializing formal.
class B {
  int i, j;
  Object? dummy;
  A(this.i): dummy = (
    i = // Error, can't assign to `i` which is implicitly `final`.
    i + 1), j = i;
  toString() => 'A(i: $i, j: $j)';
}
```

eernstg (approved):
LGTM!

In some cases we basically can't check anything (at run time). E.g., in LanguageFeatures/Augmentation-libraries/augmenting_constructors_A11_t02.dart we can just confirm that there is no compile-time error. But if it is really impossible to detect whether the same value is stored in `x` once or twice, it is probably fair to say that it is an implementation detail. So this is OK anyway, just like it's OK for a compiler to compile `for (int i = 0; i > 0; ++i);` to the empty sequence of instructions.

eernstg:
I noticed that LanguageFeatures/Augmentation-libraries/augmenting_constructors_A10_t07_lib.dart has syntax errors, because `augmented()` isn't a syntactically correct initializer list element.

We should perhaps add a test where everything is OK except that `augmented()` can't be used in an initializer list:

```dart
class A {
  dynamic x;
  A() {
    print('Running introductory!');
  }
  augment A(): x = augmented();
}
```

This one actually doesn't give rise to a compile-time error from the analyzer. ;-)

eernstg:
I created this one: https://github.com/dart-lang/sdk/issues/56477, along with this language repo issue: https://github.com/dart-lang/language/issues/4039.

PR #2818, opened 2024-08-14
Fixes #2815. Update expected warnings in binding_A03_t02.dart

### LINKED ISSUE #2815: [Wildcard Variables] Update DEAD_CODE expectations in `binding_A03_t02.dart`
This test marks a late variable as dead but we're actually reporting a more specific error on the initializer (which is the more precisely dead part).

Ideally, the test would be updated to look like:

```
// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file

/// @assertion A local declaration whose name is `_` does not bind that name to
/// anything. This means you can have multiple local declarations named `_` in
/// the same namespace without a collision error. The initializer, if there is
/// one, is still executed, but the value is not accessible.
///
/// @description Checks that it is a dead code warning in the analyzer to
/// declare late local variable named `_`. Check that at a run time the
/// initializer is not executed.

// SharedOptions=--enable-experiment=wildcard-variables

import '../../Utils/expect.dart';

String _log = "";

int init(int val) {
  _log = "init($val);";
  return val;
}

test1() {
  late var _ = init(1);
//             ^^^^^^^
// [analyzer] unspecified
}

test2() {
  late final _ = init(2);
//               ^^^^^^^
// [analyzer] unspecified
}

test3() {
  late int _ = init(3);
//             ^^^^^^^
// [analyzer] unspecified
}

main() {
  test1();
  Expect.equals("", _log);
  test2();
  Expect.equals("", _log);
  test3();
  Expect.equals("", _log);
}
```

The error message currently reported is `WarningCode.DEAD_CODE_LATE_WILDCARD_VARIABLE_INITIALIZER` and reads:
"Dead code: The assigned-to wildcard variable is marked late and can never be referenced so this initializer will never be evaluated." With a proposed resolution: "Try removing the code, removing the late modifier or changing the variable to a non-wildcard."

/fyi @kallentu @bwilkerson

sgrekhov:
The change suggested in #2815 just updates expected error position, but in case of `unspecified` error position is ignored. Therefore, I believe, the intention was to update error code as well. Updated.

### DIFF
--- a/LanguageFeatures/Wildcards/binding_A03_t02.dart
+++ b/LanguageFeatures/Wildcards/binding_A03_t02.dart
@@ -25,20 +25,20 @@ int init(int val) {
 
 test1() {
   late var _ = init(1);
-//         ^
-// [analyzer] unspecified
+//             ^^^^^^^
+// [analyzer] STATIC_WARNING.DEAD_CODE_LATE_WILDCARD_VARIABLE_INITIALIZER
 }
 
 test2() {
   late final _ = init(2);
-//           ^
-// [analyzer] unspecified
+//               ^^^^^^^
+// [analyzer] STATIC_WARNING.DEAD_CODE_LATE_WILDCARD_VARIABLE_INITIALIZER
 }
 
 test3() {
   late int _ = init(3);
-//         ^
-// [analyzer] unspecified
+//             ^^^^^^^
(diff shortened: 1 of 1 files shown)


eernstg (approved):
LGTM

PR #2819, opened 2024-08-14
Fixes #2816. Don't expect DEAD_CODE warning

### LINKED ISSUE #2816: [Wildcard Variables] `local_functions_A01_t01.dart` DEAD_CODE expectations
The following analyzer expectation in `local_functions_A01_t01.dart` fails:

```dart
test1() {
  _() {}
//^
// [analyzer] unspecified <--HERE
  _();
//^
// [analyzer] unspecified
// [cfe] unspecified
}
```

because:

1. Our highlight range doesn't line up (we highlight the whole function declaration) -- fix for that in flight: https://dart-review.googlesource.com/c/sdk/+/380284 but more significantly, 
2. the reported diagnostic doesn't get detected because it's a `STATIC_WARNING.DEAD_CODE` which is ignored by the test runner...

That warning seems right but expect that if the test runner were updated to allow it, LOTS of other tests would start failing.

Could you advise? Do we flag dead code in other co19 tests? Should we be treating wildcard local functions as *more* than dead code?

/cc @sgrekhov @munificent @kallentu

sgrekhov:
There are no ways to force test runner to report this error for one particular test. We never expect `DEAD_CODE` warning in co19 tests. So, the only thing that we can to do is to remove this expectation.

### DIFF
--- a/LanguageFeatures/Wildcards/local_functions_A01_t01.dart
+++ b/LanguageFeatures/Wildcards/local_functions_A01_t01.dart
@@ -12,9 +12,9 @@
 // SharedOptions=--enable-experiment=wildcard-variables
 
 test1() {
-  _() {}
-//^
-// [analyzer] unspecified
+  _() {} // Analyzer reports DEAD_CODE here, but this warning is ignored by
+         // the test runner. So, we only can check that the invocation of this
+         // function is an error (see the check below).
   _();
 //^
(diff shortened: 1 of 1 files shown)


eernstg (approved):
LGTM

pq:
Thanks for the explanation. I appreciate it!

PR #2826, opened 2024-08-20
#2559. Add constructor augmentation tests. Part 8. Redirecting factory constructors

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_constructors_A20_t02_lib.dart
```diff
@@ -0,0 +1,35 @@
...
+/// @description Checks that it is a compile-time error if the augmenting
+/// constructor references itself.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=macros
+
+augment library 'augmenting_constructors_A20_t02.dart';
+
+augment class C {
+  augment factory C.foo(int x) = C.foo;
```
eernstg, line 19: So these are cycles of length 1, I think it would be nice to have a cycle of length 2. Perhaps just copy this test and introduce  `foo2`, `bar2`, ... and then let the two form a redirection cycle?

The reason why this could matter is that it can be seen locally in the declarations that introduce a cycle of length 1 that there is a problem, but for the cycles of length 2 it is necessary to visit multiple declarations in order to discover that the cycle exists.

sgrekhov: I believe `LanguageFeatures/Augmentation-libraries/augmenting_constructors_A20_t03.dart` does so.

sgrekhov (commented):
Updated. PTAL.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A20_t02_lib.dart
eernstg: OK, thanks!

eernstg (approved):
LGTM!

PR #2832, opened 2024-08-26
#2559. Add metadata tests

sgrekhov:


LanguageFeatures/Augmentation-libraries/metadata_A01_t01.dart
```diff
@@ -0,0 +1,80 @@
...
+  static String staticVariable = "Static variable";
+}
+
+extension type ET(int id) {
+  static String staticVariable = "Static variable";
+}
+
+main() {
+  Symbol emptyName = MirrorSystem.getSymbol('');
+  LibraryMirror libraryMirror = currentMirrorSystem().findLibrary(emptyName);
```
eernstg, line 56: What if there are multiple libraries whose name is empty? Is it really true that there aren't any? Perhaps it is safer to use `library some.name;` in this library and then find that.

eernstg: Same consideration seems to apply to several other tests.

LanguageFeatures/Augmentation-libraries/metadata_A01_t07.dart
```diff
@@ -0,0 +1,82 @@
...
+
+/// @assertion All declarations can be augmented with metadata annotations
+/// and/or doc comments directly preceding an augmenting declaration.
+///
+/// In both cases, these should be appended to existing metadata or doc
+/// comments. For metadata annotations, these may trigger additional macro
+/// applications.
+///
+/// @description Check that augmenting metadata is appended to the introductory
+/// declaration. Test setters.
```
eernstg, line 13: Typo:
```suggestion
/// declaration. Test functions.
```

LanguageFeatures/Augmentation-libraries/metadata_A01_t08.dart
```diff
@@ -0,0 +1,97 @@
...
+
+/// @assertion All declarations can be augmented with metadata annotations
+/// and/or doc comments directly preceding an augmenting declaration.
+///
+/// In both cases, these should be appended to existing metadata or doc
+/// comments. For metadata annotations, these may trigger additional macro
+/// applications.
+///
+/// @description Check that augmenting metadata is appended to the introductory
+/// declaration (not replaces it). Test setters.
```
eernstg, line 13: Typo:
```suggestion
/// declaration (not replaces it). Test functions.
```

eernstg (commented):
Looks good! Just a couple of typos. I did suggest using a non-empty library name, too, because I'w worried that the empty library name would be ambiguous (if not now then later, when somebody wants to "clean up" a couple of SDK libraries with "unused" library names ;-).

sgrekhov:
Updated. PTAL. Do we need a library name in an augmentating file?

eernstg:
No. Every `augment library`  will soon be a part file, and they definitely do not allow for a library name. Also, 'dart:mirrors' should work on the contents of the library at the semantic level ("after merging"), not just the textual contents of the library file, which means that all references to the library should include the result of applying all augmentations, no matter which file they are located in.

eernstg (approved):
LGTM

PR #2834, opened 2024-08-27
Fixes #2833. Initializer of a wildcard variable should be executed

### LINKED ISSUE #2833: [Wildcard Variables] Test expectations are wrong in binding_A03_t02.dart
The value of `_log` should be updated for each call.
I believe that `main` should have the following expectations instead? Since even though we don't use the values returned by `testx(...)`, we still execute each `init` call.
```
main() {
  test1();
  Expect.equals("init(1);", _log);
  test2();
  Expect.equals("init(2);", _log);
  test3();
  Expect.equals("init(3);", _log);
}
```

cc. @eernstg @sgrekhov

sgrekhov:


eernstg (approved):
LGTM

eernstg:
Double-checking the change, I concluded that this change should probably not be landed. The expectations were correct already before the change.

@kallentu and @sgrekhov,did I miss something, do you see any reason why `init` would ever be invoked during the execution of this program?

sgrekhov:
Agreed. Missed that these variables are `late`.

kallentu:
Further discussion in #2833.
We can probably close this PR.

PR #2835, opened 2024-08-27
#2559. Add scoping tests

sgrekhov:


eernstg (commented):
Looks good! Noticed one small typo.

I somewhat bigger thing is the naming of the members when there is a name clash: They are named `introductory...` or `augmenting...`. I think there's a reason for this, but I still suspect that it would be easier for future maintainers of these tests (including ourselves ;-) if the clashing names have a name like `clashingName...` rather than `introductory...`.

So what do you think? Is there a systematic reason for the current naming which would be disrupted if those declarations were renamed?

LanguageFeatures/Augmentation-libraries/scoping_A01_t05.dart
```diff
@@ -0,0 +1,88 @@
...
+const foo = "Global foo";
+const bar = "Global bar";
+
+class C {
+  static const foo = "foo";
+  String bar = "bar";
+
+  void testAugmenting() {
+    Expect.equals("Global baz", baz);
+    Expect.equals("qux", qux);
```
eernstg, line 30: This one should actually yield "Global qux": The lexical lookup does not find any declarations whose name is `qux` in the current scope, the formal parameter scope of `testAugmenting`, or the body scope of this `class C` declaration, but it _does_ find a declaration named `qux` in the library scope (because the library scope contains bindings for all names of declarations at the library level, which includes top-level declarations of part files).

Agreed, it's surprising that the library scope contains everything from all files, but the class body scope only contains bindings for the declarations that occur in the same declaration (so we do not have lexical access to names that are declared in other declarations of the same enclosing entity, whether it's in the same file or in different files). Surprising, but currently true.

See also https://github.com/dart-lang/language/issues/4061. ;-)

eernstg: Still, we should treat this discussion as we've treated other discussions: Follow the spec for now, change tests if and when the spec is modified.

eernstg: Other cases are similar.

eernstg (commented):
Very nice!

One thing: LanguageFeatures/Augmentation-libraries/scoping_A01_t05.dart expects that it is possible to obtain a transformation from `qux` to `this.qux` in a situation where the library scope contains a declaration named `qux` (albeit in a different file). That isn't so (and it couldn't be so because then we would probably be unable to access that top-level declaration from the body of a declaration D in a different file if D has a member with that name, instance or static).

sgrekhov:
Thank you for pointing to this issue. Fixed. PTAL.

LanguageFeatures/Augmentation-libraries/scoping_A02_t01.dart
```diff
@@ -0,0 +1,137 @@
...
+  static void set introductoryStaticSetter(String _) {}
+//                ^^^^^^^^^^^^^^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  static String introductoryStaticMethod() => "introductoryStaticMethod";
+//              ^^^^^^^^^^^^^^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  String introductoryInstance = "introductoryInstance";
```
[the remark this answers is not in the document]
eernstg: (The source changed substantially here, so I'll close the thread.)

eernstg (approved):
LGTM!

PR #2836, opened 2024-08-28
#2559. Add tests for some corner cases

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_declarations_A01_t06.dart
```diff
@@ -0,0 +1,71 @@
...
+  void instanceMethod() {}
+  int get instanceGetter => 0;
+  void set instanceSetter(int _) {}
+}
+
+class C1 extends A {}
+class C2 with A {}
+
+augment class C1 {
+  augment void instanceVariable;
```
eernstg, line 31: Typo?:
```suggestion
  augment int instanceVariable;
```

eernstg: We would probably be able to use the type `void` for the instance variable, but it isn't helpful to have the error caused by the wrong override (of the return type of the getter), in addition to the error that we want to obtain (about not having an introductory declaration to augment).

LanguageFeatures/Augmentation-libraries/augmenting_declarations_A01_t06.dart
```diff
@@ -0,0 +1,71 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+  augment void set instanceSetter(int _) {}
+//                 ^^^^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+augment class C2 {
+  augment void instanceVariable;
```
eernstg, line 50: Ditto.

LanguageFeatures/Augmentation-libraries/augmenting_types_A01_t04_lib.dart
```diff
@@ -0,0 +1,34 @@
...
+/// @assertion It's a compile-time error if a library contains two top-level
+/// declarations with the same name, and:
+/// - Neither is an augmenting declaration, or
+/// - one of the declarations is a class-like declarations and the other is not
+///   of the same kind, meaning that at either one is a class, mixin, enum,
+///   extension or extension type declaration, and the other is not the same
+///   kind of declaration.
+///
+/// @description Checks that it is a compile-time error to augment a mixin
+/// application class with `=` syntax.
```
eernstg, line 14: This is a real corner case!

It could be argued that it could work to augment a class declaration with a mixin application class declaration or vice versa, but I think it's fine to avoid trying to sort out the details. We can always add support for it later.

eernstg (commented):
Looks good, one apparent typo!

sgrekhov:
Typo, of course. Thank you for noticing. Fixed.

eernstg (approved):
LGTM

eernstg:
Thanks! I just discovered one more thing, though: The syntax `augment class C =` is a syntax error no matter what comes next (we just can't augment a mixin application class declaration).

So we shouldn't test anything that uses `augment class C =` and then something, we should just have a single test somewhere (or perhaps we already have it), confirming that this syntax error is reported.

sgrekhov:
Split the test into two parts. The first checks that it is an error to augment `class C = A with M;` with `augment class C {}`.  The second checks that `augment class C = ...` is a syntax error.

PR #2848, opened 2024-09-06
#2559. Replace augmentation libraries by parts

sgrekhov:


eernstg (commented):
Looks good! I noted a couple of locations where it seems to have been transformed by a search and replace, but the result isn't quite natural.

Also, I get about 20 results on a grep for 'library augmentation'. That would probably become 'part' in most cases.

eernstg (approved):

sgrekhov:
Updated. 
There still are  'library augmentation' strings in these tests but all of them are part of obsolete assertions. I'm going to update `defining_augmentations_*` tests to fix all of these assertions (there is no more that part in the spec) but in a separate PR if you don't mind. PTAL.

eernstg:
OK, thanks!

Issue #2825, opened 2024-08-20
Tests dealing with enhanced parts are welcome now!


eernstg:
Here is the specification: https://github.com/dart-lang/language/blob/main/working/augmentation-libraries/parts_with_imports.md.

sgrekhov:
Waiting for https://github.com/dart-lang/language/issues/4081

PR #2851, opened 2024-09-09
#2825. Add scope tests. Part 3.

sgrekhov:


LanguageFeatures/Parts-with-imports/scope_A05_t02_part2.dart
```diff
@@ -0,0 +1,47 @@
...
+/// `loadLibrary()` is called from another part file.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=enhanced-parts
+
+part of 'scope_A05_t02_part1.dart';
+
+testPart2() async {
+  // From scope_lib1.dart
+  Expect.throws(() {print(l1.libVar);});
```
eernstg, line 24: I suspect this could fail with some backends (perhaps DDC?) because they do not implement deferred libraries, in a sense, they just load everything immediately.

I also think it's OK for an implementation to do so: we have to be able to load it using `loadLibrary()` if it hasn't been done, and we have to be able to call `loadLibrary()` and succeed even in the situation where the library has been loaded, but nobody _guarantees_, at any point in time, that the library hasn't already been loaded.

I think it would be fine to test the situation in a more flexible way: It might complete normally, or it might throw.

I don't know if there is a variant of `Expect.throws` that allows for this kind of "multiple-choice outcomes", but otherwise it could be achieved using a regular `try`/`catch` statement.

LanguageFeatures/Parts-with-imports/scope_A06_t01_part2.dart
```diff
@@ -0,0 +1,42 @@
...
+part of 'scope_A06_t01_part1.dart';
+
+import 'scope_lib1.dart' as l;
+
+testPart2() async {
+  // From scope_lib1.dart
+  Expect.equals("scope_lib1 libVar", l.libVar);
+  Expect.equals("scope_lib1 libGetter", l.libGetter);
+  l.libSetter = "x";
+  Expect.equals("libSetter", l.log);
```
eernstg, line 33: Perhaps?:
```suggestion
  Expect.equals("scope_lib1 libSetter", l.log);
```

LanguageFeatures/Parts-with-imports/scope_A08_t02_part2.dart
```diff
@@ -0,0 +1,91 @@
...
+///
+/// @description Check that it is a compile-time error if a part-file declares
+/// an import prefix with the same base name as a top-level declaration.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=enhanced-parts
+
+part of 'scope_A08_t02_part1.dart';
+
+int l1 = 1;
```
eernstg, line 16: We will get a compile-time error here because both part1 and part2 have a declaration with the same name at the top level. So we can't know for sure that the conflict with the import prefix is detected, it might just be the conflict with the other top-level declaration.

Perhaps just remove half of these declarations whose name starts with `l` from part1, and the other half from part2?

eernstg (commented):
Looks good!

One test may still fail on some implementations, I think, because they don't actually defer a deferred import. So we may need to test in a way that tolerates this.

One other issue is that we may have a top-level name clash which shadows the error message that we actually want to test.

LanguageFeatures/Parts-with-imports/scope_A05_t02_part2.dart
sgrekhov: Never heard about this. Thank you, updated.

LanguageFeatures/Parts-with-imports/scope_A06_t01_part2.dart
sgrekhov: Of course!

LanguageFeatures/Parts-with-imports/scope_A08_t02_part2.dart
sgrekhov: The idea was to test transitivity. Let's simply add another branch to this tree.

sgrekhov (commented):
Thank you! Updated. PTAL

LanguageFeatures/Parts-with-imports/scope_A08_t02.dart
```diff
@@ -0,0 +1,32 @@
...
+/// with the same base name as a top-level declaration of the library.
+///
+/// @description Check that it is a compile-time error if a part-file declares
+/// an import prefix with the same base name as a top-level declaration.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=enhanced-parts
+
+part 'scope_A08_t01_part1.dart';
+part 'scope_A08_t01_branch2_part1.dart';
```
eernstg, line 15: I think this should be moved after imports, and adjusted:
```suggestion
part 'scope_A08_t02_part1.dart';
part 'scope_A08_t02_branch2_part1.dart';
```

LanguageFeatures/Parts-with-imports/scope_A08_t02_branch2_part1.dart
```diff
@@ -0,0 +1,19 @@
...
+/// @assertion It’s a compile-time error if any file declares an import prefix
+/// with the same base name as a top-level declaration of the library.
+///
+/// @description Check that it is a compile-time error if a part-file declares
+/// an import prefix with the same base name as a top-level declaration.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=enhanced-parts
+
+part of 'scope_A08_t01.dart';
```
eernstg, line 14: Typo?
```suggestion
part of 'scope_A08_t02.dart';
```

LanguageFeatures/Parts-with-imports/scope_A08_t02_branch2_part1.dart
```diff
@@ -0,0 +1,19 @@
...
+
+part 'scope_A08_t01_branch2_part2.dart';
```
eernstg, line 16:
```suggestion
part 'scope_A08_t02_branch2_part2.dart';
```

LanguageFeatures/Parts-with-imports/scope_A08_t02_part1.dart
```diff
@@ -0,0 +1,66 @@
...
+/// @assertion It’s a compile-time error if any file declares an import prefix
+/// with the same base name as a top-level declaration of the library.
+///
+/// @description Check that it is a compile-time error if a part-file declares
+/// an import prefix with the same base name as a top-level declaration.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=enhanced-parts
+
+part of 'scope_A08_t01.dart';
```
eernstg, line 14:
```suggestion
part of 'scope_A08_t02.dart';
```

LanguageFeatures/Parts-with-imports/scope_A08_t02_part1.dart
```diff
@@ -0,0 +1,66 @@
...
+import 'scope_lib2.dart' as p2;
+import 'scope_lib2.dart' as p3;
+import 'scope_lib2.dart' as p4;
+import 'scope_lib2.dart' as p5;
+import 'scope_lib2.dart' as p6;
+import 'scope_lib2.dart' as p7;
+import 'scope_lib2.dart' as p8;
+import 'scope_lib2.dart' as p9;
+
+part 'scope_A08_t01_part2.dart';
```
eernstg, line 26:
```suggestion
part 'scope_A08_t02_part2.dart';
```

LanguageFeatures/Parts-with-imports/scope_A08_t02.dart
eernstg: This seemed useful:

```mermaid
graph BT;
  lib82["scope_A08_t02.dart"] -->|as l*| scopeLib2["scope_lib2.dart"]
  part821 -->|as p*| scopeLib2
  part821["scope_A08_t02_part1.dart"] -->|part of| lib82
  part82b1["scope_A08_t02_branch2_part1.dart"] -->|part of| lib82
  part82b2["scope_A08_t02_branch2_part2.dart"] -->|part of| part82b1
  part822["scope_A08_t02_part2.dart"] -->|part of| part821
```
```
graph BT;
  lib82["scope_A08_t02.dart"] -->|as l*| scopeLib2["scope_lib2.dart"]
  part821 -->|as p*| scopeLib2
  part821["scope_A08_t02_part1.dart"] -->|part of| lib82
  part82b1["scope_A08_t02_branch2_part1.dart"] -->|part of| lib82
  part82b2["scope_A08_t02_branch2_part2.dart"] -->|part of| part82b1
  part822["scope_A08_t02_part2.dart"] -->|part of| part821
```

eernstg (commented):
I think scope_A08_t02 has several spurious usages of scope_A08_t01 related files.

LanguageFeatures/Parts-with-imports/scope_A08_t02_part1.dart
```diff
@@ -0,0 +1,66 @@
...
+
+int l1 = 1;
```
eernstg, line 28: Oh, but then we still have the top-level name clash, so we won't know for sure whether the conflict between the top-level names and the import prefixes is detected.

eernstg (commented):
We still have a name clash with respect to top-level declarations, which means that we can't reliably confirm that the conflict between a top-level declaration and an import is detected (`p1 .. p9` are fine, but `l1 .. l9` have this name clash).

LanguageFeatures/Parts-with-imports/scope_A08_t02_part1.dart
sgrekhov: Ah! `scope_A08_t02_part1.dart` and  `scope_A08_t02_branch2_part2.dart` both adding `lx` declarations in a top-level scope and therefore it is a name conflict! Of course! Thank you for noticing! I'll fix this in this PR and add a separate test(s) to check it!

eernstg: Great, thanks!

PR #2873, opened 2024-09-13
Fixes #2872. Fix `unreachable_switch_case` warning in shared_case_scope_A02_t02.dart

### LINKED ISSUE #2872: LanguageFeatures/Patterns/shared_case_scope_A02_t02 needs to be modified for bugfix to UNREACHABLE_SWITCH_CASE
https://dart-review.googlesource.com/c/sdk/+/384326 fixed issue https://github.com/dart-lang/sdk/issues/56710, which prevented the analyzer from properly reporting an `UNREACHABLE_SWITCH_CASE` warning for a switch case that contains a `when` clause.

The following co19 test needs to be modified to account for the fixed warning:
- LanguageFeatures/Patterns/shared_case_scope_A02_t02

Probably the best thing to do is ignore the warning with an `// ignore` comment.

sgrekhov:
The updated test passes

### DIFF
--- a/LanguageFeatures/Patterns/shared_case_scope_A02_t02.dart
+++ b/LanguageFeatures/Patterns/shared_case_scope_A02_t02.dart
@@ -53,7 +53,6 @@ String test1(num x) {
 String test2(num x) {
   switch (x) {
     case var a as int when a > 0:
-    case int a:
     case int a when a is Never:
       a.expectStaticType<Exactly<int>>();
       print(a);


LanguageFeatures/Patterns/shared_case_scope_A02_t02.dart
```diff
@@ -53,7 +53,6 @@ String test1(num x) {
 String test2(num x) {
   switch (x) {
     case var a as int when a > 0:
-    case int a:
     case int a when a is Never:
```
eernstg, line 56: Ah, now I understand, it's the case on line 56 (`... Never:`) which isn't reachable when the `case int a:` is included. So we're fixing the test by removing code that has no diagnostics, and keeping the code that used to cause a warning. Tricky! ;-)

eernstg (approved):
LGTM

PR #2881, opened 2024-09-16
Fixes #2877. Catch and ignore exceptions during temp file deletion

### LINKED ISSUE #2877: Catch and ignore exceptions during temp file cleanup deletion
On Windows file deletion might fail with `The process cannot access the file because it is being used by another process.`
I believe these should be ignored during test cleanup like it's done for vm tests(for example https://github.com/dart-lang/sdk/blob/a15be7cffea13889d1d6599e04f3dca92aad4c52/tests/standalone/io/file_system_watcher_large_set_test.dart#L36).

Otherwise, you get flaky tests, for example https://github.com/dart-lang/sdk/issues/56708

sgrekhov:


aam (approved):

eernstg (approved):
I'm assuming that none of these tests should report a different outcome to the test runner in the case where the `deleteSync` operation throws (so if they succeed without the exception then they also succeed with the exception).

With that, LGTM.

PR #2883, opened 2024-09-16
Fixes #2878. Use `pid` as a part of an unique temp name

### LINKED ISSUE #2878: co19 temp file names generation results in collision leading to flaky failing tests
Example of flaky test is https://github.com/dart-lang/sdk/issues/56706

It seems that either the temp name generation method should add some file uniqueness checks before handing out generated result, or https://api.flutter.dev/flutter/package-file_file/Directory/createTempSync.html should be used.

sgrekhov:
I was unable to reproduce the issue. But if several instances of the same test should be runned concurrently to reproduce the issue, then, I think, using `pid` as a part of the name should fix it. Please review.


aam (approved):

eernstg (approved):
LGTM

PR #2884, opened 2024-09-17
#2559. Update `defining_augmentation_*` tests

sgrekhov:
- Assertions updated
- Some of tests deleted, as duplicates of ones from  `Parts-with-imports`
- Some renamed and moved to `Parts-with-imports` folder (in a separate PR)


LanguageFeatures/Augmentation-libraries/augmentation_libraries_lib.dart
```diff
@@ -2,7 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-/// @description Common library for augmentation libraries tests
+/// @description Common library for augmentations tests.
```
eernstg, line 5: I think this would flow more naturally:
```suggestion
/// @description Common library for augmentation tests.
```

LanguageFeatures/Augmentation-libraries/defining_augmentation_A01_t01.dart
```diff
@@ -2,11 +2,12 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-/// @assertion Declarations in any library augmentation or the augmented library
-/// are visible to all of the others, including private ones.
+/// @assertion Augmentation declarations interact with part files mainly in
+/// restrictions on where an augmenting declaration may occur relative to the
+/// declaration it augments.
 ///
 /// @description Checks that declarations in an augmentation are visible in a
```
eernstg, line 9: I can't help asking what this means. The word 'augmentation' could refer to a Dart file, but they would be called 'part files' now, rather than 'augmentation libraries' or anything like that. So maybe it should be adjusted like this?:
```suggestion
/// @description Checks that declarations in a part are visible in the
```

LanguageFeatures/Augmentation-libraries/defining_augmentation_A01_t02.dart
```diff
@@ -2,17 +2,17 @@
...
 
-/// @assertion Declarations in any library augmentation or the augmented library
-/// are visible to all of the others, including private ones.
+/// @assertion Augmentation declarations interact with part files mainly in
+/// restrictions on where an augmenting declaration may occur relative to the
+/// declaration it augments.
 ///
 /// @description Checks that declarations in a main library are visible in an
-/// augmentation library
+/// augmentation.
```
eernstg, line 10: Following the same idea, we could have this (`suggestion` doesn't work here, apparently):
```
/// @description Checks that declarations in the main library are visible in a
/// part file.
```

eernstg (commented):
I might have misunderstood, but it looks like there's a need to adjust the terminology in many, similar phrases: 'Augmentation' could now be 'part file'.

It is also possible that 'augmentation' means 'augmenting declaration' (but in that case I've misread several things).

Anyway, PTAL!

LanguageFeatures/Augmentation-libraries/augmentation_libraries_lib.dart
sgrekhov: Fixed

LanguageFeatures/Augmentation-libraries/defining_augmentation_A01_t01.dart
sgrekhov: May be "Checks that declarations in an augmenting part file are visible in the main library."?

LanguageFeatures/Augmentation-libraries/defining_augmentation_A01_t02.dart
sgrekhov: May be "Checks that declarations in the main library are visible in an augmenting part file."?

sgrekhov (commented):
Thabk you. Updated.

LanguageFeatures/Augmentation-libraries/defining_augmentation_A01_t02.dart
eernstg: Same issue: It should work even if that part file doesn't contain any augmenting declarations at all.

eernstg (commented):
I'm still worried about the mixture of concepts when we seem to assume that "part file" and "augmentation" is somehow connected.

sgrekhov:
Rewritten once again. PTAL

eernstg (approved):
That's great, thanks! Sorry about nitpicking for a while. LGTM.

sgrekhov:
> That's great, thanks! Sorry about nitpicking for a while. LGTM.

No problem at all! Thank you!

Issue #2846, opened 2024-09-04
[Unquoted Imports] co19 Tests


kallentu:
Write up co19 tests for the import shorthand feature.

Examples pulled from [the spec](https://github.com/dart-lang/language/blob/main/accepted/future-releases/unquoted-imports/feature-specification.md).

sgrekhov:
Completed. See https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Unquoted-imports

PR #2889, opened 2024-09-20
#2846. Add unquoted imports tests for path separator

sgrekhov:


LanguageFeatures/Unquoted-imports/path_separator_A01_t01.dart
```diff
@@ -0,0 +1,47 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Instead of `.`, this proposal uses `/`. This is obviously a
+/// natural path separator since it's the main path separator character used in
+/// most operating systems and in URLs on the web. Also, it is already the path
+/// separator character used inside Dart imports today.
+///
+/// @description Checks that it is a compile-time error to use `.` or `\` as a
+/// path separator.
```
eernstg, line 11: I don't think it is possible to maintain that any particular piece of text is "an unquoted Dart import directive that uses `X` as a path separator" where `X` is some character. So we can't really say that we're testing that it would be an error to do so.

The problem is that every character which is accepted by the lexical rules plays a particular role, and there's no way we can "use it as" anything other. In particular, `.` and `\` are never path separators.

So what we _can_ do is to test that the unquoted path `test.package1.dotted.name1` is transformed into the URI `package:test.package1.dotted.name1/name1.dart`. We would make sure that there is a package like this, and perhaps read a `libraryId` to confirm that we're importing the library that we expect to import.

We could also arrange for a package named `test` to exist, and for that package to have a `lib` that contains `package1/dotted/name1.dart`, and then we'd be able confirm that the import directive does _not_ import that library (that is, we don't get that value of `libraryId`), but that wouldn't really show anything if we already have confirmed that we're importing the right library (it's fair to assume that the same import directive won't import two or more libraries; besides, if we're checking `libraryId` then we _aren't_ importing both libraries because then `libraryId` would be a conflicted name ;-).

LanguageFeatures/Unquoted-imports/path_separator_A01_t01.dart
```diff
@@ -0,0 +1,47 @@
...
+//     ^^^^^^^^^^^^^^^^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+import test.package1/a\b/c;
+//     ^^^^^^^^^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+import test.package1/a/b.c;
```
eernstg, line 31: So we're assuming that there is no `a/b.c.dart` in the package `test.package1`? It does create some amount of confusion that the test depends on so many other packages and libraries to exist or not exist. I think it would be helpful to have some comments to explain the situation.

LanguageFeatures/Unquoted-imports/test.package1/dummy.dart
```diff
@@ -0,0 +1,11 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @description Library file for testing unquoted imports. Must be unavailable
+/// via `import 'package:...'` and its unquoted shortcut.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=unquoted-imports
+
+Never libraryId = "dummy.dart" as dynamic;
```
eernstg, line 11: How does it help in any test that this `libraryId` has type `Never`? If a particular declaration is unavailable then I'd expect every attempt to access it from another library to fail, and this would be just as true for an unsurprising declaration like `String libraryId = "Never used";`. Or is there something that makes a `libraryId` of type `Never` work better?

LanguageFeatures/Unquoted-imports/test.package1/lib/dotted.name1.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @description Library file for testing unquoted imports.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=unquoted-imports
+
+String libraryId = "dotted.name1.dart";
```
eernstg, line 10: Perhaps?:
```suggestion
String libraryId = "dotted.name11";
```

LanguageFeatures/Unquoted-imports/test.package1/lib/package1.dart
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @description Library file for testing unquoted imports.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=unquoted-imports
+
+String libraryId = "package1.dart";
```
eernstg, line 10: Perhaps?:
```suggestion
String libraryId = "package11";
```

LanguageFeatures/Unquoted-imports/test.package1/lib/test.package1.dart
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @description Library file for testing unquoted imports.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=unquoted-imports
+
+String libraryId = "test.package1.dart";
```
eernstg, line 10: Perhaps?:
```suggestion
String libraryId = "test.package11";
```

LanguageFeatures/Unquoted-imports/test_package2/dummy.dart
```diff
@@ -0,0 +1,11 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @description Library file for testing unquoted imports. Must be unavailable
+/// via `import 'package:...'` and its unquoted shortcut.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=unquoted-imports
+
+Never libraryId = "dummy.dart" as dynamic;
```
eernstg, line 11: Same comment as previous `Never` typed `libraryId`.

LanguageFeatures/Unquoted-imports/test_package2/lib/dotted.name2.dart
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @description Library file for testing unquoted imports.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=unquoted-imports
+
+String libraryId = "dotted.name2.dart";
```
eernstg, line 10: Perhaps?:
```suggestion
String libraryId = "dotted.name22";
```

LanguageFeatures/Unquoted-imports/test_package2/lib/test_package2.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @description Library file for testing unquoted imports.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=unquoted-imports
+
+String libraryId = "test_package2.dart";
```
eernstg, line 10: Perhaps?:
```suggestion
String libraryId = "test_package22";
```

eernstg (commented):
Lots of comments this time. ;-)

I think it's necessary to start with the specified semantics and test that we have that, and then basically drop the idea that we can "test that '.' isn't used as a path separator".

LanguageFeatures/Unquoted-imports/path_separator_A01_t01.dart
sgrekhov: Ok, let's remove `path_separator_*` tests then. I'll `static_semantics_*` tests in a separate PR that will proof that a path with `.` and `/` is transformed and points to a right library. This PR only adds necessary packages for the furter tests then.

LanguageFeatures/Unquoted-imports/path_separator_A01_t01.dart
sgrekhov: This test has been deleted

LanguageFeatures/Unquoted-imports/test.package1/dummy.dart
sgrekhov: Changed to a `String`.

LanguageFeatures/Unquoted-imports/test.package1/lib/dotted.name1.dart
sgrekhov: Why `dotted.name11`? Changed to `dotted.name1`.

LanguageFeatures/Unquoted-imports/test.package1/lib/package1.dart
sgrekhov: `package1`

LanguageFeatures/Unquoted-imports/test.package1/lib/test.package1.dart
sgrekhov: The same. `test.package1`.

LanguageFeatures/Unquoted-imports/test_package2/dummy.dart
sgrekhov: Fixed

LanguageFeatures/Unquoted-imports/test_package2/lib/dotted.name2.dart
sgrekhov: Same as the above

sgrekhov (commented):
Updated. `path_separator_*` tests removed at all. I'll add `static_semantics_*` tests in a separate PR later. PTAL

sgrekhov:
Merge conflict has been resolved. PTAL.

LanguageFeatures/Unquoted-imports/path_separator_A01_t01.dart
eernstg: Sounds good, thanks!

LanguageFeatures/Unquoted-imports/test.package1/lib/dotted.name1.dart
eernstg: I just looked at the other declarations playing a similar role: It looked like the `libraryId` would be created by removing `.dart` from the end of the file name and adding `1`, e.g., `c.d.dart` has `c.d1`.

eernstg (approved):
LGTM

PR #2890, opened 2024-09-25
#2846. Add unquoted imports tests for parts

sgrekhov:


LanguageFeatures/Unquoted-imports/lib/co19_unquoted_imports.dart
```diff
@@ -0,0 +1,21 @@
...
+/// library at `package:foo/bar.dart`, not part of the library with name
+/// `foo.bar`.
+///
+/// @description Checks that unquoted syntax can be used in `part` and `part of`
+/// directives.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=unquoted-imports,enhanced-parts
+
+part of co19_unquoted_imports/part_A01_t02;
```
eernstg, line 18: Is there a package named `co19_unquoted_imports`? I can't see it here, and I can't see it in https://github.com/dart-lang/co19/tree/master/LanguageFeatures/Unquoted-imports. If there is no such package (presumably, we aren't getting it from `dart pub get`) then this would be a compile-time error. Similarly for all the other unquoted imports that rely on a package with this name.

eernstg: I believe the expanded URI would be `'package:co19_unquoted_imports/part_A01_t02.dart'`, and I don't think that URI could be resolved successfully.

LanguageFeatures/Unquoted-imports/lib/co19_unquoted_imports_part.dart
```diff
@@ -0,0 +1,20 @@
...
+/// library at `package:foo/bar.dart`, not part of the library with name
+/// `foo.bar`.
+///
+/// @description Checks that unquoted syntax can be used in `part` and `part of`
+/// directives.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=unquoted-imports,enhanced-parts
+
+part of co19_unquoted_imports;
```
eernstg, line 18: Would be `'package:co19_unquoted_imports/co19_unquoted_imports.dart'`, which would also be an error as far as I can see.

eernstg (commented):
I think several of these unquoted imports will fail to resolve to anything. Perhaps this PR should have added a pubspec.yaml that introduces the package `co19_unquoted_imports`?

eernstg:
Oh, just found it after all: https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Unquoted-imports/pubspec.yaml.

LanguageFeatures/Unquoted-imports/lib/co19_unquoted_imports.dart
eernstg: OK, now that I've found the pubspec.yaml it looks fine!

LanguageFeatures/Unquoted-imports/lib/co19_unquoted_imports_part.dart
eernstg: Looks fine now!

LanguageFeatures/Unquoted-imports/of.parts.test/lib/part_A02_t01.dart
```diff
@@ -0,0 +1,25 @@
...
+/// lookahead past the of identifier to see if the next token is `;`, `.`, `/`,
+/// or another identifier.
+///
+/// @description Checks that `part of.parts.test` is parsed like a
+/// `part 'package:of.parts.test/test.dart';`.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=unquoted-imports
+
+part of.parts.test;
```
eernstg, line 19: This looks tricky, but it should be OK: Even though it might be a `part of` directive, this will fail to parse when `.` is encountered, and regular parser behavior will then ensure that `of` is never parsed as the keyword `of` in the sequence `part`, `of`, it's parsed as part of an unquoted URI.

Nothing surprising about this, that's just the standard ability of parsers to resolve ambiguities based on lookahead.

eernstg (approved):
LGTM, now that I've looked twice. ;-)

PR #2896, opened 2024-09-26
#2846. Add syntax and whitespaces tests

sgrekhov:


LanguageFeatures/Unquoted-imports/syntax_A01_t01.dart
```diff
@@ -0,0 +1,33 @@
...
+/// uri               ::= stringLiteral | packagePath
+/// packagePath       ::= pathSegment ( '/' pathSegment )*
+/// pathSegment       ::= segmentComponent ( '.' segmentComponent )*
+/// segmentComponent  ::= IDENTIFIER
+///                     | RESERVED_WORD
+///                     | BUILT_IN_IDENTIFIER
+///                     | OTHER_IDENTIFIER
+///
+/// @description Checks that it is a syntax error if a `pathSegment` starts or
+/// ends with a '/'.
```
eernstg, line 17: It's somewhat confusing to read "a `pathSegment` starts or ends with a `/`" because that is never true. How would we be able to test the handling of something that can't happen? The following is perhaps better:
```suggestion
/// @description Checks that it is a syntax error if a `/` is encountered when
/// a `packagePath` is expected, or a `;` is encountered when a `pathSegment`
/// is expected.
```
The point is that the grammar clearly indicates that a `packagePath` is expected when the token `import` has been accepted, and the next token isn't a string literal. So we're necessarily trying to recognize a `packagePath` when the first `/` is encountered, and that must be reported as a syntax error (or some other compile-time error, because a parser is always allowed to report syntax errors in that way).

For the `/` that occurs after a `pathSegment`, it is definitely not a syntax error in its own right: We could have another `pathSegment`, and the `/` would then fit in just fine. So it's the semicolon which is a syntax error. Again, I'm focusing on the non-terminal which is actually expected when the semicolon is encountered. We're doing `pathSegment ( '/' pathSegment )*` which means that we expect to see a `pathSegment` after each occurrence of a `/`, and a `pathSegment` cannot start with the token `;`, which must give rise to a syntax error.

LanguageFeatures/Unquoted-imports/syntax_A02_t01.dart
```diff
@@ -0,0 +1,38 @@
...
+/// uri               ::= stringLiteral | packagePath
+/// packagePath       ::= pathSegment ( '/' pathSegment )*
+/// pathSegment       ::= segmentComponent ( '.' segmentComponent )*
+/// segmentComponent  ::= IDENTIFIER
+///                     | RESERVED_WORD
+///                     | BUILT_IN_IDENTIFIER
+///                     | OTHER_IDENTIFIER
+///
+/// @description Checks that it is a syntax error if a `segmentComponent` is
+/// empty.
```
eernstg, line 17:
```suggestion
/// @description Checks that it is a syntax error if a `.` is encountered
/// when a `packagePath` is expected. Ditto if a `;` is encountered
/// when a `segmentComponent` is expected. Ditto if a `.` is encountered
/// when a `pathSegment` is expected.
```

eernstg (commented):
Looks good!

I'm a little worried about the wording in some cases (basically, we can't say something which would be like "check that `if` can't start with a `+`" in the extreme, because the answer is simply "but it doesn't!"), and I've tried to come up with something that doesn't have this kind of built-in contradiction.

LanguageFeatures/Unquoted-imports/syntax_A01_t01.dart
sgrekhov: There was a typo. There should be `packagePath` instead of `pathSegment`. Thank you for noticing, changed to the suggested wording.

LanguageFeatures/Unquoted-imports/syntax_A02_t01.dart
sgrekhov: Updated.

sgrekhov (commented):
Thank you. Updated, PTAL.

eernstg (approved):
LGTM

PR #2897, opened 2024-09-27
#2559. Add more tests for augmenting enum members and values.

sgrekhov:


eernstg (approved):
LGTM!

One thing could be interesting to try out, but that doesn't have to delay this PR: Copy LanguageFeatures/Augmentation-libraries/augmenting_enum_values_A05_t04.dart. Then modify the copy such that every `name` declaration is static. This should be accepted, and it should still _not_ be possible to invoke the extension getter `name` unless it is explicitly resolved as `EnumName(myEnum).name`. (That is: the static member with the same basename will suffice to shadow the extension member.)

PR #2899, opened 2024-09-27
#2559. Add test that `EnumName` extension is not augmentable

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmenting_enum_values_A05_t05.dart
```diff
@@ -0,0 +1,41 @@
...
+augment enum E2 {
+  augment e0;
+  augment static String get name => "name2";
+}
+
+main() {
+  Expect.equals("name1", E1.name);
+  Expect.equals("e0", E1.e0.name);
+  Expect.equals("name2", E2.e0.name);
+  Expect.equals("e0", E2.e0.name);
```
eernstg, line 40: Very good!

The other thing I mentioned should perhaps be added to a more general `enum` test: The `name` extension instance getter cannot be invoked when there is a static member in the on-type with basename `name`. 

In general: An extension instance member named `n` cannot be invoked when there is a static member in the on-type with the name `id` or `id=` where `id` is the basename of `n`.

We might have it, it just occurred to me that perhaps we don't. ;-)

eernstg (approved):
LGTM

LanguageFeatures/Augmentation-libraries/augmenting_enum_values_A05_t05.dart
sgrekhov: I can't find such tests. Which part of the spec contains this statement?

PR #2903, opened 2024-09-30
Fixes #2902. Add tests for an extension applicability

### LINKED ISSUE #2902: Add co19 for extension members invocation
Add co19 tests checking that an extension instance member named `n` cannot be invoked when there is a static member in the on-type with the name `id` or `id=` where `id` is the basename of `n`.

sgrekhov:


eernstg (commented):
LGTM.

The rule might well be adjusted such that it is not an error, but we can then change the test accordingly.

eernstg (approved):

Issue #2906, opened 2024-10-01
Add co19 tests for Digit separators


sgrekhov:
Specification: https://github.com/dart-lang/language/blob/main/accepted/future-releases/digit-separators/feature-specification.md
Implementation: https://github.com/dart-lang/sdk/issues/56188

PR #2910, opened 2024-10-02
#2906. Add digit separators tests

sgrekhov:


LanguageFeatures/Digit_separators/digit_separators_A04_t04.dart
```diff
@@ -0,0 +1,42 @@
...
+//          ^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+const pi2 = 3____.14;
+//          ^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+final e1 = 2._718;
```
eernstg, line 27: We could have an extra copy of this particular situation which should succeed:

```dart
final e3 = 2._719;

...

extension on int {
  int get _719 => this;
}
```

LanguageFeatures/Digit_separators/digit_separators_A04_t06.dart
```diff
@@ -0,0 +1,66 @@
...
+///
+/// That means only allowing `_`s between two `0-9` digits in `NUMBER` and
+/// between two `0-9`, `a-f`, `A-F` digits in `HEX_NUMBER`.
+///
+/// @description Check that it is a compile-time error if `_` follows or is
+/// followed by `+` or `-`.
+/// @author sgrekhov22@gmail.com
+
+const pi1 = 0.0314e_+2;
+//          ^^^^^^^^^^
```
eernstg, line 18: Is this error marking actually precise? I would expect `0.0314e_` to be an unrecognized token (a lexer error, no less), or a syntax error (a literalInteger followed by an identifier `e_`), and `+2` would in any case be a perfectly legit continuation of the expression.

eernstg (commented):
Looks good!

I think it would make sense to double check that a particular situation isn't a syntax error, it's just a failed lookup (`2._718` fails because `int` doesn't have a member named `_718`). I wrote a comment about this.

I looked at LanguageFeatures/Digit_separators/digit_separators_A04_t06.dart, thinking that there could be a similar situation there. I don't think so, after all, but the situation isn't quite clear. I wrote a comment on that, too.

LanguageFeatures/Digit_separators/digit_separators_A04_t06.dart
sgrekhov: Yes, it is presize. Both analyzer and VM produce here `Digit separators ('_') in a number literal can only be placed between two digits. Try removing the '_'.`  `0.0314` here cannot be a `literalInteger`, it contains `.`. But on the line 22 below `314` can be. But anyway both of the tools produce the same error on line 22 as well. Added
```dart
const e_ = 0;
const E_ = 0;
```
to make the test stronger.

sgrekhov (commented):
Thank you. Added new test to check the extension and comments regarding `digit_separators_A04_t06.dart`.
Also the directory renamed from `Digin_separators` to `Digit-separators` to follow naming stayle in `Language-features`. PTAL.

LanguageFeatures/Digit_separators/digit_separators_A04_t04.dart
eernstg: Thanks!

eernstg: (Innocent bypassers should not do that at home, of course ;-)

LanguageFeatures/Digit_separators/digit_separators_A04_t06.dart
eernstg: Very good, thanks!

eernstg (approved):
LGTM

sgrekhov:
Asked if we need it in the chat. The tests work on the bleeding edge without the flag.
```
sgrekhov@SGREKHOV-HP:~/Google/dart-sdk/sdk$ python3 tools/test.py -n cfe-strong-linux co19/LanguageFeatures/Digit-separators/
No build targets found.
Test configuration:
    cfe-strong-linux(architecture: x64, compiler: fasta, mode: release, runtime: none, system: linux)
Suites tested: co19
[00:48 | 100% | +   16 | -    0]

=== All 16 tests passed ===
sgrekhov@SGREKHOV-HP:~/Google/dart-sdk/sdk$ python3 tools/test.py -n analyzer-asserts-linux co19/LanguageFeatures/Digit-separators/
No build targets found.
Test configuration:
    analyzer-asserts-linux(architecture: x64, compiler: dart2analyzer, mode: release, runtime: none, system: linux, enable-asserts, use-sdk)
Suites tested: co19
[00:07 | 100% | +   16 | -    0]

=== All 16 tests passed ===
```

eernstg (approved):
LGTM

PR #2911, opened 2024-10-02
Fixes #2909. Fix extension applicability tests according to the updated spec

### LINKED ISSUE #2909: Static members do not interfere with extension instance member invocations
Adjust test to expect that a static member does not eliminate the applicability of an extension. See https://github.com/dart-lang/sdk/issues/56818 for background information.

sgrekhov:


LanguageFeatures/Extension-methods/applicability_A01_t01.dart
```diff
@@ -6,46 +6,44 @@
...
+///   Function types and the type `Function` are considered as having a `call`
+///   member. This ensure that if there is an applicable extension, the existing
+///   invocation would otherwise be a compile-time error. Members of `Object`
+///   exists on all types, so they can never be the target of implicit member
+///   invocations (they can also not be declared as extension members).
 ///
-/// @description Check that an extension member with the basename `m` is not
-/// applicable if on type has a static member with the same basename.
+/// @description Check that an extension member with the basename `m` is
+/// applicable if on type has a static member with the same name.
```
eernstg, line 20:
```suggestion
/// applicable even if the on type has a static member with the same basename.
```

eernstg (commented):
Looks good. Perhaps we should change a couple of static declarations or extension declarations such that the member name which is being invoked has the same basename as the static member, but exactly one of them has `=` at the end of the name? I think that's a case that might well be overlooked.

eernstg (approved):
LGTM!

PR #2922, opened 2024-10-07
#2825. Fix dotted.library.name syntax in `part of` directive

sgrekhov:


eernstg (commented):
Looks good!

I noticed one change that I think makes the text less natural for most readers: "An URI" looks funny (it seems to imply that the acronym is pronounced as a word, just like "NATO" or "UNICEF", but I've never encountered that pronunciation in real life...).

sgrekhov:
Thank you! I didn't know about this 'a', 'an' exception. Updated. PTAL.

eernstg (approved):
Thanks, LGTM!

Issue #2923, opened 2024-10-08
Add tests for import/export libraries with the same URI


sgrekhov:
According to the specification 

> All imports and exports of the same URI in a Dart program denotes the same library, and imports or exports of different URIs denote distinct libraries.

We have no tests checking this statement.

PR #2928, opened 2024-10-09
#2923. More tests for import and export library with the same URI.

sgrekhov:


Language/Libraries_and_Scripts/Imports/Semantics_of_Imports/same_uri_A01_t10.dart
```diff
@@ -0,0 +1,22 @@
...
+/// the same library, and imports or exports of different URIs denote distinct
+/// libraries.
+///
+/// @description Checks that imports of the same URI in a Dart program denotes
+/// the same library.
+/// @author sgrekhov22@gmail.com
+
+import '../../../../Utils/expect.dart';
+import 'same_uri_common_lib.dart' as l1;
+import '../Semantics_of_Imports/same_uri_common_lib.dart' as l2;
```
eernstg, line 15: I wouldn't expect these to be the same URI. Doesn't the test fail?

eernstg: OK, the discussion in https://github.com/dart-lang/sdk/issues/56869 revealed that at least some tools do already normalize the URIs to some extent.

The language specification does say that "different" URIs denote different libraries, and "equal" URIs denote the same library, so we should be able to test that.

eernstg (approved):
LGTM

Language/Libraries_and_Scripts/Imports/Semantics_of_Imports/same_uri_A01_t10.dart
sgrekhov: > I wouldn't expect these to be the same URI. Doesn't the test fail?
No, the test works well in both analyzer and vm on Win and Linux.

eernstg: Right, I found a comment to the effect that at least some tools choose to simplify paths where `..` occurs in some non-first location (so `SOMETHING/dir1/dir2/../dir3/foo.dart` is simplified to `SOMETHING/dir1/dir3/foo.dart`). The tools can in principle have any implementation specific behavior they want in almost every respect when it comes to URIs, so we should just test this in a minimal fashion and leave the details to the tool teams and their specialized test suite.

Issue #2925, opened 2024-10-08
co19/LanguageFeatures/Parts-with-imports/grammar_A05_t01


scheglov:
```
FAILED: dart2analyzer-none release_arm64 co19/LanguageFeatures/Parts-with-imports/grammar_A05_t01
Expected: Pass
Actual: MissingCompileTimeError

--- Command "dart2analyzer" (took 62ms):
DART_CONFIGURATION=ReleaseARM64 sdk/bin/dart --packages=.dart_tool/package_config.json pkg/analyzer_cli/bin/analyzer.dart --use-analysis-driver-memory-byte-store --dart-sdk=/Users/scheglov/Source/Dart/sdk.git/sdk/sdk -Dtest_runner.configuration=custom-configuration-1 --enable-experiment=macros --ignore-unrecognized-flags --packages=/Users/scheglov/Source/Dart/sdk.git/sdk/.dart_tool/package_config.json --format=json /Users/scheglov/Source/Dart/sdk.git/sdk/tests/co19/src/LanguageFeatures/Parts-with-imports/grammar_A05_t01.dart

static error failures:
- Missing expected unspecified error at tests/co19/src/LanguageFeatures/Parts-with-imports/grammar_A05_t01_lib.dart line 32, column 5, length 4.

--- Re-run this test:
python3 tools/test.py -m release -c dart2analyzer -a arm64 co19/LanguageFeatures/Parts-with-imports/grammar_A05_t01
```

`LanguageFeatures/Parts-with-imports/grammar_A05_t01_lib.dart` has
```
library;

/**/part of 'grammar_A05_t01.dart';
//  ^^^^
// [analyzer] unspecified
// [cfe] unspecified
```

Because it has `library;` it is not recognized by `Parts-with-imports/grammar_A05_t01.dart` as a part, and so is not analyzed, so no errors it it. But we do report an error in `Parts-with-imports/grammar_A05_t01.dart`.

eernstg:
And it isn't even reported that `LanguageFeatures/Parts-with-imports/grammar_A05_t01_lib.dart` has a syntax error?

scheglov:
We analyze `Parts-with-imports/grammar_A05_t01.dart`, which attempts to include `grammar_A05_t01_lib.dart` as a part. But because it has `library;` it is not considered to be a valid part, and skipped. So, it is not analyzed by the test.

In DAS it would be analyzed as a library on its own, and get an error reported.

eernstg:
OK, thanks!

PR #2931, opened 2024-10-10
#2925. Add comment explaining why the syntax error is not reported in grammar_A05_t01_lib.dart

sgrekhov:


eernstg (approved):
Thanks!

LGTM

Issue #2933, opened 2024-10-14
Add more `Stream` tests


sgrekhov:
There are no co19 tests for `Stream.error`, `Stream.multi` and `Stream.value`

PR #2935, opened 2024-10-15
#2933. Add tests for `Stream.multi` constructor.

sgrekhov:


LibTest/async/Stream/Stream.multi_A01_t01.dart
```diff
@@ -0,0 +1,64 @@
...
+    }
+    controller.close();
+  });
+  Expect.isFalse(stream.isBroadcast);
+
+  int i = 1;
+  stream.listen((v) {
+    Expect.equals(v, i++);
+  }, onDone: () {
+    asyncEnd();
```
lrhn, line 55: Maybe check here that `i` is 6. Ditto for `j` below.

LibTest/async/Stream/Stream.multi_A03_t01.dart
```diff
@@ -0,0 +1,62 @@
...
+    });
+    controller.close();
+    Expect.throws(() {
+      controller.stream;
+    });
+  });
+
+  stream.listen((v) {}, onDone: () {
+    asyncEnd();
+  });
```
lrhn, line 61: (Can be written as `onDone: asyncEnd`.)

LibTest/async/Stream/Stream.multi_A04_t01.dart
```diff
@@ -0,0 +1,58 @@
...
+/// will be true. This has no effect on the stream behavior, it is up to the
+/// `onListen` function to act like a broadcast stream if it claims to be one.
+///
+/// A multi-subscription stream can behave like any other stream. If the
+/// `onListen` callback throws on every call after the first, the stream behaves
+/// like a single-subscription stream. If the stream emits the same events to
+/// all current listeners, it behaves like a broadcast stream.
+///
+/// @description Checks that reading of [MultiStreamController.stream] is not
+/// supported.
```
lrhn, line 35: Description not updated for this test?
(Same as for A03_t01.)

LibTest/async/Stream/Stream.multi_A04_t01.dart
```diff
@@ -0,0 +1,58 @@
...
+/// @author sgrekhov22@gmail.com
+
+import "dart:async";
+import "../../../Utils/expect.dart";
+
+main() {
+  asyncStart();
+  var stream = Stream<int>.multi((MultiStreamController<int> controller) {
+    print(controller.hasListener);
```
lrhn, line 44: Test true or false? I think it's `true`. Also not paused, even though you can't send synchronous events before the call to `listen` has returned.

LibTest/async/Stream/Stream.multi_A05_t01.dart
```diff
@@ -0,0 +1,64 @@
...
+main() {
+  asyncStart(2);
+  var stream = Stream<int>.multi((controller) {
+    for (var v in [1, 2, 3, 4, 5]) {
+      controller.add(v);
+    }
+    controller.close();
+  }, isBroadcast: true);
+  Expect.isTrue(stream.isBroadcast);
+
```
lrhn, line 50: Nothing below here is needed for the test.

lrhn:
Are there tests for the `MultiStreamController` API? Which is the normal `StreamController` API plus `addSync`, `addErrorSync` and `closeSync`, which allows adding events synchronously (after `listen` has returned, not paused, and if there are no prior events still enqueued).

LibTest/async/Stream/Stream.multi_A02_t01.dart
```diff
@@ -0,0 +1,63 @@
...
+    }
+    for (var v in [1, 2, 3, 4, 5]) {
+      controller.add(v);
+    }
+    controller.close();
+  });
+
+  stream.listen((v) {}, onDone: () {
+    stream.listen((v) {}, onDone: () {
+      Expect.notEquals(c1, c2);
```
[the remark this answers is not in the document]
sgrekhov: Surprisingly but we had no `Expect.notIdentical()`. Added.

LibTest/async/Stream/Stream.multi_A03_t01.dart
sgrekhov: Updated

LibTest/async/Stream/Stream.multi_A04_t01.dart
sgrekhov: Yes, thank you for noticing.  Fixed.

LibTest/async/Stream/Stream.multi_A05_t01.dart
sgrekhov: Ok, removed.

sgrekhov (commented):
Updated. PTAL.

> Are there tests for the `MultiStreamController` API? Which is the normal `StreamController` API plus `addSync`, `addErrorSync` and `closeSync`, which allows adding events synchronously (after `listen` has returned, not paused, and if there are no prior events still enqueued).

I'm working on these tests now and will add them later in a separate PR. 

BTW could you, please, help me to understand if the following test is correct or not?
```dart
main() {
  asyncStart(2);
  int counter = 1;
  var stream = Stream<int>.multi((controller) {
    for (;;counter++) {
      controller.add(counter);
      if (counter % 5 == 0) {
        controller.close();
        counter++;
        return;
      }
    }
  });
  int i = 1;
  stream.listen((v) {
    print(v);
    Expect.equals(v, i++);
  }, onDone: () {
    Expect.equals(6, i);
    asyncEnd();
  });
  int j = 6;
  stream.listen((v) {
  print(v);
    Expect.equals(v, j++);
  }, onDone: () {
    Expect.equals(11, j);
    asyncEnd();
  });
}
```
The test works for me but it's unclear for me why. The output is
```
unittest-suite-wait-for-done
1
6
2
7
3
8
4
9
5
10
unittest-suite-success
```
Seems that the controller syncronously sends data to the first listener and then to the second. Is this expected? If yes, what's the difference between `add` and `addSync`?

lrhn (approved):

Utils/tests/Expect/notIdentical_A01_t02.dart
```diff
@@ -0,0 +1,34 @@
...
+  try {
+    Expect.notIdentical(arg1, arg2, reason);
+    throw new Exception("ExpectException expected");
+  } on ExpectException catch(_) {
+  }
+}
+
+main() {
+  int one = 1;
+  check("x", "x");
```
eernstg, line 30: Canonicalization of string literals is underspecified. I think it's safer to use two occurrences of a reference to the same string, to avoid depending on canonicalization.

For example, the following prints 'false' with `dart` and with `dart compile exe`:

```dart
void main() {
  var ab = "a" + "b";
  print(identical("ab", ab));
}
```

As usual, the underlying rationale is that we should not take the opportunity to test that string canonicalization will make the original version work, because this test is not about string literal canonicalization. The point is that we wish to pass two actual arguments that are _definitely_ the same object, and then we want to see that `notIdentical` throws an `ExpectException`.

```suggestion
  var x = "x";
  check(x, x);
```

We could make a similar argument about the different parts of the semantics of `identical`. For example, records get a very special treatment, and so do `double` objects. But it's up to some other tests to determine that all those different cases are treated correctly by `identical`.

LibTest/async/Stream/Stream.multi_A04_t01.dart
eernstg: Perhaps this test wasn't fully implemented?  Both `print` invocations should probably be `Expect.isTrue`/`Expect.isFalse`.

eernstg (commented):
Looks good, just a couple of comments.

LibTest/async/Stream/Stream.multi_A04_t01.dart
sgrekhov: It's just an accidental code. Removed.

Utils/tests/Expect/notIdentical_A01_t02.dart
sgrekhov: Thank you. Updated.

sgrekhov (commented):
Updated. PTAL.

lrhn:
> BTW could you, please, help me to understand if the following test is correct or not?


It's not wrong. I wouldn't *depend* on the interleaving always being what it is.

The `Stream.multi` callbacks here are both synchronous, which means they run immediately to completion when `listen` is called.
It doesn't matter whether they use `add` or `addSync` because no event can be emitted until `listen` has returned, so all five events are enqueued and a microtask is scheduled to deliver the the next enqueued value.
Both do this, the first one enqueues 1..5+done events and the second 6..10+done, and two microtasks are scheduled.
Then those trigger, each emitting one event and scheduling another microtasks to emit the next event.

It's guaranteed that the streams will emit those events, because all events are computed synchronously.

The scheduling gives the interleaving you see. I would not guarantee any specific interleaving of separate event sources. We can guarantee that the next enqueued event will be delivered in a *later* microtask, but not precisely *which* later microtask.

So, working as intended.

Issue #2930, opened 2024-10-10
Check Stream tests failures


sgrekhov:
See https://dart-current-results.web.app/#/filter=co19/LibTest/async/

A lot of tests fail on run time

sgrekhov:
See also https://github.com/dart-lang/sdk/issues/49264

PR #2936, opened 2024-10-16
#2930. Add MultiStreamController tests. Part 1.

sgrekhov:


LibTest/async/MultiStreamController/addStream_A03_t01.dart
```diff
@@ -0,0 +1,63 @@
...
+  var stream = Stream<int>.multi((controller) {
+    controller.addStream(source.stream, cancelOnError: false).then((_) {
+      controller.close();
+    });
+  });
+  listen(stream);
+}
+
+void listen(Stream<int> stream) {
+  int eventsCounter = 1;
```
eernstg, line 52: Wouldn't it still work if you used `eventsCounter = 0` and then `++eventsCounter` in line 55? Seems slightly simpler to understand.

(Also, just as a side remark, `++x` is simpler than `x++` because the former is a side-effect and a variable-read operation, but the latter is more like "cache the old value in a tmp variable", then the side effect, then "evaluate to the tmp variable". So I prefer `++x` unless there is a specific need for `x++`.)

LibTest/async/MultiStreamController/addStream_A03_t01.dart
```diff
@@ -0,0 +1,63 @@
...
+  int errorsCounter = 0;
+  stream.listen((v) {
+    Expect.equals(eventsCounter++, v);
+  }, onError: (e) {
+    Expect.equals(--errorsCounter, e);
+  }, onDone: () {
+    Expect.equals(4, eventsCounter);
```
eernstg, line 59: This would then be 3, I think.

LibTest/async/MultiStreamController/addStream_A03_t04.dart
```diff
@@ -0,0 +1,65 @@
...
+  listen(stream);
+}
+
+void listen(Stream<int> stream) {
+  int eventsCounter = 1;
+  int errorsCounter = 0;
+  stream.listen((v) {
+    Expect.equals(eventsCounter++, v);
+  }, onError: (e) {
+    errorsCounter++;
```
eernstg, line 58: Why not `--errorsCounter` as usual?

LibTest/async/MultiStreamController/addStream_A03_t05.dart
```diff
@@ -0,0 +1,68 @@
...
+  Completer completer = Completer();
+  completer.future.then((_) {
+    asyncEnd();
+  });
+  var source = StreamController<int>();
+  source.add(1);
+  source.add(2);
+  source.add(3);
+  source.addError(-1);
+  // No source.close(); here. The first error should complete the future
```
eernstg, line 45: Good catch!

LibTest/async/MultiStreamController/addStream_A04_t01.dart
```diff
@@ -0,0 +1,59 @@
...
+  });
+  listen(stream);
+}
+
+void listen(Stream<int> stream) {
+  int eventsCounter = 1;
+  stream.listen((v) {
+    Expect.equals(eventsCounter++, v);
+    if (v == 3) {
+      Future.delayed(Duration(milliseconds: 100), asyncEnd);
```
eernstg, line 54: I guess it's difficult to obtain an absolute guarantee that this test will never run on hardware which is tiny enough, and overloaded enough, such that the delay of 100 ms is too short?

However, I'd also assume that the test will actually succeed even in the case where `asyncEnd` is invoked too early. Right?

If so then it should be fine. Otherwise we would need to think about potential flaking.

eernstg (commented):
Looks good! Just one question and a couple of nits.

LibTest/async/MultiStreamController/addErrorSync_A02_t01.dart
```diff
@@ -0,0 +1,47 @@
...
+    i += 2;
+  }, onError: (e) {
+    delivered = true;
+    Expect.equals(2, e);
+  }, onDone: () {
+    Expect.isTrue(delivered);
+    asyncEnd();
+  });
+  ss.pause(Future<void>.delayed(Duration(milliseconds: 100), () {
+    ss.resume();
```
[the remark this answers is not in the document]
sgrekhov: I'm afraid that in this case the test may become racy. At the time of `ss.pause()` invocation we may already receive one, two, or even all events. I think it's better to leave it as is and `asyncEnd()` will tel us that everything was delivered as expected in spite of that listener was paused at some point.

LibTest/async/MultiStreamController/addStream_A03_t01.dart
sgrekhov: Makes sense, thank you! Updated.

LibTest/async/MultiStreamController/addStream_A03_t01.dart
sgrekhov: Of course

LibTest/async/MultiStreamController/addStream_A03_t04.dart
sgrekhov: Chenged to `--errorsCounter`

sgrekhov (commented):
Updated, except one place where, I believe, there is probability to make the test flaky. PTAL.

LibTest/async/MultiStreamController/addErrorSync_A02_t01.dart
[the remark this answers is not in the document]
eernstg: OK!

eernstg (approved):
LGTM

LibTest/async/MultiStreamController/addErrorSync_A02_t01.dart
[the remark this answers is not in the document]
lrhn: You should not be able to get any events synchronous while calling `listen`, so there is no race here.
Even without the pause, a call to `listen(stream)` would not receive any events and would return synchronously, so both `listen(stream)` calls must happen before any event can be delivered.

LibTest/async/MultiStreamController/addErrorSync_A02_t01.dart
```diff
@@ -0,0 +1,47 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion void addErrorSync( Object error, [ StackTrace? stackTrace ])
+/// Adds and delivers an error event.
+///
+/// Adds an error like [addError] and attempts to deliver it immediately.
+/// Delivery can be delayed if other previously added events are still pending
+/// delivery, if the subscription is paused, or if the subscription isn't
+/// listening yet.
```
lrhn, line 11: ... where "the subsciprion isn't listening yet" includes the entire synchronous run of `onListen`.
So even without the earlier `add(1)`, the `addErrorSync` would still be enqueued.

PR #2937, opened 2024-10-17
#2933. Add MultiStreamController tests. Part 2.

sgrekhov:


LibTest/async/MultiStreamController/closeSync_A02_t01.dart
```diff
@@ -0,0 +1,49 @@
...
+
+void listen(Stream<int> stream) {
+  int i = 0;
+  late StreamSubscription<int> ss;
+  ss = stream.listen((v) {
+    Expect.equals(++i, v);
+    if (v == 1) {
+      ss.pause(Future.delayed(Duration(milliseconds: 100), () {
+
+        ss.resume();
```
lrhn, line 42: This is a double-resume.

If you pass a future to `StreamSubscription.pause`, it resumes when that future completes.
Here you both pass a future *and* call `ss.resume()` just before that future completes. Either one is enough.

LibTest/async/MultiStreamController/closeSync_A03_t01.dart
```diff
@@ -0,0 +1,46 @@
...
+    bool done = false;
+    controller.done.then((_) {
+      done = true;
+      asyncEnd();
+    });
+    controller.add(1);
+    controller.add(2);
+    controller.add(3);
+    Expect.isFalse(done);
+    controller.closeSync();
```
lrhn, line 33: The `isFalse` check should be after the `closeSync()`?
(To check what the text above says that it checks.)

It's correct that the `closeSync` will be enqueued, both because there are earlier async events enqueued, and because it's called during the `listen` call.

LibTest/async/MultiStreamController/close_A01_t01.dart
```diff
@@ -0,0 +1,69 @@
...
+import "../../../Utils/expect.dart";
+
+main() {
+  asyncStart(2);
+  var stream = Stream<int>.multi((controller) {
+    controller.add(1);
+    controller.add(2);
+    controller.add(3);
+    controller.close().then((_) {
+      Expect.throws(() {
```
lrhn, line 40: Can also check that `hasListener` is false.

LibTest/async/MultiStreamController/close_A01_t01.dart
```diff
@@ -0,0 +1,69 @@
...
+  listen(stream);
+  listen(stream);
+}
+
+void listen(Stream<int> stream) {
+  int i = 0;
+  stream.listen((v) {
+    Expect.equals(++i, v);
+  }, onDone: () {
+    Expect.equals(3, i);
```
lrhn, line 67: Could add an `asyncEnd()` here too and up the `asyncStart` count to 4.

That should prove that the `done` event is delievered *and* the `done` future completes. (Without checking the order.)

LibTest/async/MultiStreamController/closeSync_A02_t01.dart
sgrekhov: Thank you! Removed.

LibTest/async/MultiStreamController/close_A03_t01.dart
```diff
@@ -0,0 +1,62 @@
...
+  asyncStart(4);
+  var stream = Stream<int>.multi((controller) {
+    controller.add(1);
+    controller.add(2);
+    controller.add(3);
+    controller.close().then((_) {
+      for (int i = 0; i < results.length; ++i) {
+        Expect.isTrue(
+            results[i], " Listener number $i is not done or cancelled yet");
+      }
```
lrhn, line 45: This does not look correct.

It checks that *all* listeners have received their first event when *any* listener's `cancel()` is registered and closed the `done` future for that listner's controller.

It might be true, but that's an implementation artifact.

What is promised is:
* When you call `listen` on the stream, the `onListen` callback is run synchronously.
* Here that enqueues a number of events and calls `close` to access the `done` future.
* That happens four times, synchronously.
* Then, in some later microtask, one of the stream subscriptions will deliever an event. Here that cancels the stream subscription, which calls the controller's `onCancel` callback and waits for that to complete. There isn't one installed so that does nothing. It then completes the `done` future "in some later microtask".
* At this point, it's allowed that the `done` future completes before the next subscription event is delivered. There is no *guarantee* of ordering between separate asynchronous event sources. (They'll likely be in order of scheduling, but if the enqueued delivery function recognizes that the stream has been cancelled after deliering the first event, it *could* just complete the `done` event right there.

So maybe, if possible, only check the one result that corresponds to the stream controller whose `done` future completed.

LibTest/async/MultiStreamController/close_A04_t01.dart
```diff
@@ -0,0 +1,46 @@
...
+/// is not sent and the future is never complete.
+/// @author sgrekhov22@gmail.com
+
+import "dart:async";
+import "../../../Utils/expect.dart";
+
+main() {
+  asyncStart();
+  Stream<int>.multi((controller) {
+    controller.add(1);
```
lrhn, line 36: You could just put
`Expect.fail("called onListen");`
as the body of this callback. 

If the stream is not listened to, the `onListen` callback is never called and no stream controller is created. (So it's trivial that there won't be a `done` future that completes, since there won't even be a controller to read `done` from.)

lrhn (approved):
Few comments, generally super fine!

LibTest/async/MultiStreamController/close_A03_t01.dart
sgrekhov: Thank you for the detailed explanation. Updated.

sgrekhov (commented):
Thank you. Updated. PTAL.

PR #2941, opened 2024-10-21
#2933. Add MultiStreamController tests. Part 4.

sgrekhov:


sgrekhov:
Converted to a draft. `onResume` test should be updated according to the https://github.com/dart-lang/sdk/issues/56927

sgrekhov:
Ready for review

LibTest/async/MultiStreamController/onResume_A02_t01.dart
```diff
@@ -0,0 +1,49 @@
...
+import "../../../Utils/expect.dart";
+
+main() {
+  asyncStart(2);
+  var stream = Stream<int>.multi((controller) {
+    Expect.isNull(controller.onResume);
+    controller.onResume = () {
+      Expect.fail("Unexpected onResume");
+    };
+    controller.onResume = null;
```
eernstg, line 27: Same consideration again around testing `myVar = nonNullValue; myVar = null;`.

eernstg (commented):
Added a few comments.

LibTest/async/MultiStreamController/isPaused_A01_t01.dart
```diff
@@ -0,0 +1,47 @@
...
+/// @description Checks that this getter returns `true` if the stream is paused
+/// and `false` otherwise.
+/// @author sgrekhov22@gmail.com
+
+import "dart:async";
+import "../../../Utils/expect.dart";
+
+main() {
+  asyncStart();
+  var controllers = <MultiStreamController<int>>[];
```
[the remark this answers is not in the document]
sgrekhov: Oh yes. Thank you. Fixed!

LibTest/async/MultiStreamController/isPaused_A01_t02.dart
```diff
@@ -0,0 +1,47 @@
...
+/// This is the case if the controller's stream has a listener and it is paused,
+/// or if it has not received a listener yet. In that case, the controller is
+/// considered paused as well.
+///
+/// A broadcast stream controller is never considered paused. It always forwards
+/// its events to all uncanceled subscriptions, if any, and let the
+/// subscriptions handle their own pausing and buffering.
+///
+/// @description Checks that this getter returns `true` if the stream is paused
+/// and `false` otherwise. Test `isBroadcast: true` case.
```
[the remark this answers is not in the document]
sgrekhov: Fixed.

LibTest/async/MultiStreamController/isPaused_A01_t02.dart
```diff
@@ -0,0 +1,47 @@
...
+/// @author sgrekhov22@gmail.com
+
+import "dart:async";
+import "../../../Utils/expect.dart";
+
+main() {
+  asyncStart();
+  var controllers = <MultiStreamController<int>>[];
```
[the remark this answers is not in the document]
sgrekhov: Fixed too

LibTest/async/MultiStreamController/onCancel_A01_t01.dart
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion FutureOr<void> Function()? onCancel
+/// The callback which is called when the stream is canceled.
+///
+/// May be set to `null`, in which case no callback will happen.
+///
+/// @description Checks that this callback is called when the stream is canceled
```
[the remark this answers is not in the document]
sgrekhov: But here we are testing the case when `onCancel` is not null.
```dart
  controller.onCancel = asyncEnd;
```

LibTest/async/MultiStreamController/onCancel_A02_t01.dart
```diff
@@ -0,0 +1,38 @@
...
+import "dart:async";
+import "../../../Utils/expect.dart";
+
+main() {
+  asyncStart(2);
+  var stream = Stream<int>.multi((controller) {
+    controller.onCancel = () {
+      Expect.fail("Unexpected onCancel");
+    };
+    controller.onCancel = null;
```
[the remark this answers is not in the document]
sgrekhov: Yes, but... Is there any other way to check that "May be set to `null`, in which case no callback will happen."?

LibTest/async/MultiStreamController/onPause_A01_t02.dart
```diff
@@ -0,0 +1,46 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion void Function()? onPause
+/// The callback which is called when the stream is paused.
+///
+/// May be set to `null`, in which case no callback will happen.
+///
+/// Pause related callbacks are not supported on broadcast stream controllers.
+///
+/// @description Checks that this callback is called when the stream is paused.
```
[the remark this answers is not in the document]
sgrekhov: Yes. Updated the description.

LibTest/async/MultiStreamController/onResume_A02_t01.dart
sgrekhov: If you insist I can remove these tests. Please confirm. I just don't see any other way to test the statement

sgrekhov (commented):
Thank you. Updated. Please confirm if we need to remove `.onEvent = ...; onEvent = null;` tests. I don't think that they do much harm, they simply tests an appropriate statement.

LibTest/async/MultiStreamController/onCancel_A02_t01.dart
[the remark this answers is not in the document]
eernstg: I'd simply think that if `x` is a variable then `x = null;` would be an action which can be described as "`x` is set to null", and we wouldn't have to worry about the history.

In other words, we wouldn't have to try things like "setting `x` to null in a situation where it has the value `v`" for a bunch of different `v`, for instance the cases where `v` is `() { /*something*/ }` and where `v` is null.

The daunting part is that we basically cannot introduce this rule in general (that is, "check different histories when we're checking the effect of setting a variable to a specific value"), because it might very well make the complexity of a huge number of tests grow substantially. So that's the reason why I think we should be really careful about taking the history into account here.

That said, it is true that anything that looks like an instance variable in Dart could be a getter and a setter, and "reading/writing the variable" could give rise to arbitrary computations, and there might actually be a difference between the behavior of `x` in the situation where it has a specific value `v`, but different histories.

So you could argue that we should _always_ test with different histories when we're supposed to test that setting a variable to a specific value has a specific effect. I just don't think it's doable, or worthwhile. So let's only test that kind of thing when the history is mentioned in the documentation/specification of the feature being tested.

LibTest/async/MultiStreamController/onResume_A02_t01.dart
eernstg: I'm not suggesting that we shouldn't test the given spec sentence, just that there's no need to deal with the history of values of `onResume`. Setting it to null should be enough, and it doesn't matter which value it had previously. So I'm actually just suggesting that we shouldn't have line 24-26.

Or did I miss something, is there a reason to expect that lines 24-26 will have an observable effect?

eernstg (commented):
Thanks for the fixes! I think we should take one more iteration, though, on the question about testing variables with or without history.

sgrekhov:
If we only lines 24-26 then how we will know that the test failed? Therefore removed these tests at all. PTAL.

sgrekhov:
Yes, I understand your concern, but with the suggested change how can we be sure that the tests don't fail? What if after setting `onSomething` to `null`  some callback is called anyway? We'll newer know about this and the tests will simply always pass. With
```dart
    controller.onCancel = () {
      Expect.fail("Unexpected onCancel");
    };
```
we at least make sure that setting `onCancel` to null clears the callback immediately, not after some microtask. Wihout it I'd prefer don't have a test which is always pass and only increases tryjobs load. Especiallly taking into account that `controller.onCancel = null;` doesn't change anything because it already has a `null` value and we check it in an earlier tests.

LibTest/async/MultiStreamController/isPaused_A01_t02.dart
```diff
@@ -0,0 +1,48 @@
...
+/// @assertion bool get isPaused
+/// Whether the subscription would need to buffer events.
+///
+/// This is the case if the controller's stream has a listener and it is paused,
+/// or if it has not received a listener yet. In that case, the controller is
+/// considered paused as well.
+///
+/// A broadcast stream controller is never considered paused. It always forwards
+/// its events to all uncanceled subscriptions, if any, and let the
+/// subscriptions handle their own pausing and buffering.
```
lrhn, line 14: Might want to say that a `MultiStreamController` is not a broadcast stream controller, even if its stream has `isBroadcast` as `true`. 
(It's just a controller for a single stream subscription on a stream which claims to be a broadcast stream, so it should cooperate with other controllers for other subscriptions on the same stream about acting like a broadcast stream.)

LibTest/async/MultiStreamController/onListen_A02_t01.dart
```diff
@@ -0,0 +1,34 @@
...
+    controller.add(3);
+    controller.close();
+  });
+  listen(stream);
+  listen(stream);
+}
+
+void listen(Stream<int> stream) {
+  stream.listen((v) {
+  }, onDone: asyncEnd);
```
lrhn, line 33: (You can use `null` as first argument to `listen` if you don't want to do anything for data events.)

LibTest/async/MultiStreamController/onPause_A01_t02.dart
```diff
@@ -0,0 +1,46 @@
...
+import "dart:async";
+import "../../../Utils/expect.dart";
+
+main() {
+  asyncStart(2);
+  var stream = Stream<int>.multi((controller) {
+    controller.onPause = () {
+      Expect.fail("Unexpected onPause");
+    };
+    controller.onPause = null;
```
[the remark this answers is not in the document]
lrhn: Agree. If you check first that `onPause` is `null`, like the test above, setting it to `null` should be expected to do nothing. (But it doesn't hurt either, and checks that you *can* call `onPause=` with `null`.)

LibTest/async/MultiStreamController/onResume_A01_t01.dart
```diff
@@ -0,0 +1,46 @@
...
+import "dart:async";
+import "../../../Utils/expect.dart";
+
+main() {
+  asyncStart(2);
+  var stream = Stream<int>.multi((controller) {
+    Expect.isNull(controller.onResume);
+    controller.onResume = () {
+      controller.close();
+      asyncEnd();
```
lrhn, line 24: No need for the `controller.close()` here. It will be delivered after `asyncEnd()` has run, so at least for the second stream, the test is already considered completed.

lrhn (approved):

LibTest/async/MultiStreamController/onResume_A01_t01.dart
sgrekhov: Thank you. Removed.

eernstg (commented):
Just a few extra comments.

LibTest/async/MultiStreamController/onResume_A01_t01.dart
lrhn: Correct on the behavior of `asyncStart`/`asyncEnd`.
The thing is that the test is considered *done* after the second `asyncEnd`. If nothing has failed before that, the test is *successful*, irrevocably.

Anything that happens after that does not affect the test result, so *preferably* nothing should happen at all. Something throwing after the test is complete is a bug in the test, not a test failure.

In this test, the behavior of `controller.close()` is not what is tested, so there is no need to call it at all.
If we do want to test it, then the `asyncEnd()` should be moved to after that test has been performed.

sgrekhov: `controller.close()` was removed from this test, so it should be Ok. `asyncEnd()` here checks that `onResume` was really called.

eernstg: > In a way, it does preempt the test result.

I don't think so, the test runner just checks that the subprocess output includes `unittest-suite-success`, but it is certainly still possible for the test to proceed, and to succeed or fail based on later actions.

```dart
import 'dart:async';
import "package:expect/async_helper.dart";

void main() {
  asyncStart();
  asyncEnd();
  throw "Failed!"; // Causes the test to fail, comment it out and it succeeds.
}
```

> Something throwing after the test is complete is a bug in the test, not a test failure

It isn't quite fair to say that this is what is happening in examples like the one above, or in LibTest/async/MultiStreamController/onResume_A01_t01.dart.

The test will succeed or fail according to the test runner (which is what counts), and there's no difference between failures happening before or after the final invocation of `asyncEnd()`.

What `asyncStart(_)`/`asyncEnd()` will do is simply to support behavior that is detected as a test run failure in the case where the stand/end invocation counts differ. This can be used to detect, in particular, that a given piece of code which was expected to run did not actually run.

> In this test, the behavior of `controller.close()` is not what is tested, so there is no need to call it at all.

OK, that's good to know! Perhaps `controller.close()` should then be deleted from all tests which aren't testing this, and then we should make sure there is one or more tests targeting the effects of invoking `controller.close()`.

> If we do want to test it, then the `asyncEnd()` should be moved to after that test has been performed.

That makes sense, too. It shouldn't matter where it occurs in a piece of code that does not include scheduling points (like `await`), but it's safer to put it at the end, indicating that _all_ of the given code has been executed when the test terminates.

LibTest/async/MultiStreamController/onListen_A02_t01.dart
eernstg: Do we have a test about passing `null` as the first argument to `listen`?

eernstg (commented):
Looks good, nearly done!

LibTest/async/MultiStreamController/isPaused_A01_t02.dart
sgrekhov: https://github.com/dart-lang/sdk/issues/56908#issuecomment-2421969826

LibTest/async/MultiStreamController/onListen_A02_t01.dart
sgrekhov: Yes, we have. https://github.com/dart-lang/co19/blob/master/LibTest/async/Stream/listen_A02_t01.test.dart. But I've updated this test too for the brevity.

LibTest/async/MultiStreamController/onPause_A01_t02.dart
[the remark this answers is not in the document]
sgrekhov: Updated.

sgrekhov (commented):
Updated. PTAL

LibTest/async/MultiStreamController/isPaused_A01_t02.dart
eernstg: Thanks!

LibTest/async/MultiStreamController/onPause_A01_t02.dart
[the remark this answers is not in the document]
eernstg: Thanks!

LibTest/async/MultiStreamController/onResume_A01_t01.dart
eernstg: Closing this thread. I don't think there's anything substantial to resolve here. ;-)

eernstg (approved):
LGTM

LibTest/async/MultiStreamController/onResume_A01_t01.dart
sgrekhov: The thread is closed but I can't find my comment here. Adding just in case.

There was a long discussion about it. See
https://github.com/dart-lang/co19/pull/2400#issuecomment-1831769630

Issue #2386, opened 2023-11-20
Add more tests for super-bounded types


sgrekhov:
We have no tests tor the following statement

> It is a compile-time error if a parameterized type T is super-bounded when
> it is used in any of the following ways:
> • T is an immediate subterm of a new expression (17.13.1) or a constant
> object expression (17.13.2).
> • T is an immediate subterm of a redirecting factory constructor signature
> (10.7.2).
> • T is an immediate subterm of an extends clause of a class (10.9), or it
> occurs as an element in the type list of an implements clause (10.10), or
> a with clause (10).

We also have no any test checking that `class C<T extends T> {}` is an error

PR #2942, opened 2024-10-22
#2386. Update super-bounded types tests

sgrekhov:
- Update descriptions
- Fix some tests by changing `[cfe] ...` to `[cfe] unspecified`
- Move some checks to appropriate tests


Language/Generics/Superbounded_types/class_A01_t05.dart
```diff
@@ -2,26 +2,35 @@
...
-///   [T] is an immediate subterm of a new expression (16.15.1) or a constant
-///   object expression
-/// @description Checks that compile error is thrown when not well-bounded
-/// parametrized type is used in the constant object expression with [is]
-/// constructions.
+/// - `T` is an immediate subterm of a new expression or a constant object
+///   expression.
+///
+/// @description Checks that it is a compile-time error if a parametrized
+/// super-bounded type is used in a constant `is` expression and not a top type.
```
eernstg, line 11: What we're checking here is that a type which is not well-bounded is an error as the right hand operand of `is`. Of course, it would be an error in every other location as well.

There's no great need to mention 'super-bounded types', because there aren't any of them here. Perhaps the point is that "this is not a super-bounded type, and we want the compiler to say so"? Not sure how to say that in a way that helps the reader more than a plain "this type is an error because it isn't well-bounded".

Finally, it doesn't make much sense to talk about a _parameterized_ super-bounded type because super-bounded types are _always_ parameterized.

Perhaps something like this could be used:

Checks that it is a compile-time error if a type which is not well-bounded
is used as the right hand operand in an `is` expression.

eernstg (commented):
I think it would be very helpful if all the `@description`s would be revised based on the comments about class_A01_t04.dart.

sgrekhov:
- Tests, checking (not-) well-bounded types in `is` and `as` expressions renamed and assertion updated
- Descriptions updated is some other tests
PTAL

Language/Generics/Superbounded_types/class_A03_t01.dart
```diff
@@ -7,9 +7,8 @@
 ///   ...
 /// - `T` is an immediate subterm of a redirecting factory constructor signature
 ///
-/// @description Checks that compile error is thrown when a parameterized
-/// super-bounded type `T` is used in a redirecting factory constructor
-/// signature
+/// @description Checks that it is a compile-time error if a super-bounded type
+/// is an immediate subterm of a redirecting factory constructor signature
```
eernstg, line 11: We have two errors here: (1) The redirection uses a super-bounded type (this is the one that we're trying to test), and (2) the redirection corresponds to a type which isn't assignable to the return type of the constructor, `A<X>`.

So we'd need to use `class A {...}` rather than `class A<X> {...}` in order to make the redirection type correct, such that the _only_ error is the fact that the redirection type is super-bounded.

Something like this should work:

```dart
class A {
  ...
}

class C<X extends C<X>> extends A {}
```

This implies that `C<S>` is a subtype of `A` for all `S`.

Language/Generics/Superbounded_types/class_A03_t02.dart
```diff
@@ -7,8 +7,8 @@
 ///   ...
 /// - `T` is an immediate subterm of a redirecting factory constructor signature
 ///
-/// @description Checks that it is a compile-time error if a type `T` is used in
-/// the redirecting factory constructor signature
+/// @description Checks that it is a compile-time error if a super-bounded type
+/// is an immediate subterm of a redirecting factory constructor signature
```
eernstg, line 11: This test needs the same fix as Language/Generics/Superbounded_types/class_A03_t01.dart.

Language/Generics/Superbounded_types/class_A03_t03.dart
```diff
@@ -7,9 +7,8 @@
 ///   ...
 /// - `T` is an immediate subterm of a redirecting factory constructor signature
 ///
-/// @description Checks that it is a compile-time error if a parameterized
-/// super-bounded type `T` is used in a redirecting factory constructor
-/// signature
+/// @description Checks that it is a compile-time error if a super-bounded type
+/// is an immediate subterm of a redirecting factory constructor signature
```
eernstg, line 11: Perhaps just delete this test? I don't see any super-bounded types anywhere, we just have the (irrelevant) lack of subtyping.

Language/Generics/Superbounded_types/class_A03_t04.dart
```diff
@@ -8,7 +8,7 @@
 /// - `T` is an immediate subterm of a redirecting factory constructor signature
 ///
 /// @description Checks that it is a compile-time error if a super-bounded type
-/// `T` is used in a redirecting factory constructor signature
+/// is an immediate subterm of a redirecting factory constructor signature
```
eernstg, line 11: This test seems to need the same fix as Language/Generics/Superbounded_types/class_A03_t01.dart. However, it might also be redundant, is there anything new in here?

Language/Generics/Superbounded_types/well_bounded_A06_t02.dart
```diff
@@ -2,13 +2,11 @@
...
-/// @assertion It is a compile-time error if a parameterized type `T` is
-/// super-bounded when it is used in any of the following ways:
-/// - `T` is an immediate subterm of a new expression or a constant object
-///   expression.
+/// @assertion Any use of a type `T` which is not well-bounded is a
+/// compile-time error.
 ///
-/// @description Checks that it is a compile-time error if a parametrized
-/// super-bounded type is used in a constant `is` expression and not a top type.
+/// @description Checks that it is a compile-time error if not a well-bounded
```
eernstg, line 8: I think we could use this:
```suggestion
/// @description Checks that it is a compile-time error if a malbounded
```

No, I don't know why it is 'malbounded' without a hyphen, but 'well-bounded'. ;-)

Language/Generics/Superbounded_types/well_bounded_A06_t02.dart
```diff
@@ -2,13 +2,11 @@
...
+/// type is used as a right hand operand in an `is` expression.
```
eernstg, line 9: Nit.
```suggestion
/// type is used as the right hand operand in an `is` expression.
```
We always know _which_ right hand operand it is, because there is defi-nit-ely only one. ;-)

Language/Generics/Superbounded_types/well_bounded_A06_t03.dart
```diff
@@ -2,13 +2,11 @@
...
-/// super-bounded when it is used in any of the following ways:
-/// - `T` is an immediate subterm of a new expression or a constant object
-///   expression.
+/// @assertion Any use of a type `T` which is not well-bounded is a
+/// compile-time error.
 ///
-/// @description Checks that it is not an error if a parametrized super-bounded
-/// type is used in a constant `as` expression and a top type.
+/// @description Checks that it is not an error if a well-bounded type is used
+/// as a right hand operand in an `as` expression.
```
eernstg, line 9:
```suggestion
/// as the right hand operand in an `as` expression.
```

Language/Generics/Superbounded_types/well_bounded_A06_t04.dart
```diff
@@ -2,13 +2,11 @@
...
-/// @assertion It is a compile-time error if a parameterized type `T` is
-/// super-bounded when it is used in any of the following ways:
-/// - `T` is an immediate subterm of a new expression or a constant object
-///   expression.
+/// @assertion Any use of a type `T` which is not well-bounded is a
+/// compile-time error.
 ///
-/// @description Checks that it is a compile-time error if a parametrized
-/// super-bounded type is used in a constant `as` expression and not a top type.
+/// @description Checks that it is a compile-time error if not a well-bounded
```
eernstg, line 8:
```suggestion
/// @description Checks that it is a compile-time error if a malbounded
```

Language/Generics/Superbounded_types/well_bounded_A06_t04.dart
```diff
@@ -2,13 +2,11 @@
...
+/// type is used as a right hand operand in an `as` expression.
```
eernstg, line 9:
```suggestion
/// type is used as the right hand operand in an `as` expression.
```

Language/Generics/Superbounded_types/well_bounded_A06_t05.dart
```diff
@@ -2,13 +2,11 @@
...
-/// super-bounded when it is used in any of the following ways:
-/// - `T` is an immediate subterm of a new expression or a constant object
-///   expression.
+/// @assertion Any use of a type `T` which is not well-bounded is a
+/// compile-time error.
 ///
-/// @description Checks that it is not an error if a parametrized super-bounded
-/// type is used in a constant `is` expression.
+/// @description Checks that it is not an error if a well-bounded  type is used
+/// as a right hand operand in an constant `is` expression.
```
eernstg, line 9: Nits.
```suggestion
/// @description Checks that it is not an error if a well-bounded type is used
/// as the right hand operand in a constant `is` expression.
```

Language/Generics/Superbounded_types/well_bounded_A06_t06.dart
```diff
@@ -2,13 +2,11 @@
...
-/// super-bounded when it is used in any of the following ways:
-/// - `T` is an immediate subterm of a new expression or a constant object
-///   expression.
+/// @assertion Any use of a type `T` which is not well-bounded is a
+/// compile-time error.
 ///
-/// @description Checks that it is not an error if a parametrized super-bounded
-/// type is used in a constant `as` expression.
+/// @description Checks that it is not an error if a well-bounded type is used
+/// as a right hand operand in a constant `as` expression.
```
eernstg, line 9:
```suggestion
/// @description Checks that it is not an error if a well-bounded type is used
/// as the right hand operand in a constant `as` expression.
```

eernstg (commented):
Looks good! I commented on a handful of typos.

Language/Generics/Superbounded_types/class_A03_t01.dart
sgrekhov: Thank you. Fixed

Language/Generics/Superbounded_types/class_A03_t02.dart
sgrekhov: Fixed.

Language/Generics/Superbounded_types/class_A03_t03.dart
sgrekhov: Deleted.

Language/Generics/Superbounded_types/class_A03_t04.dart
sgrekhov: If to make the same changes then this test will become identical to `Language/Generics/Superbounded_types/class_A03_t01.dart`. Therefore deleted.

sgrekhov (commented):
Thank you! Updated. PTAL.

Language/Generics/Superbounded_types/class_A03_t01.dart
eernstg: Thanks!

Language/Generics/Superbounded_types/class_A03_t04.dart
eernstg: OK, thanks for checking!

eernstg (approved):
LGTM!

PR #2949, opened 2024-10-25
Fixes #2386. Add more super-bounded types tests

### LINKED ISSUE #2386: Add more tests for super-bounded types

sgrekhov:


Language/Generics/Superbounded_types/well_bounded_A01_t04.dart
```diff
@@ -16,10 +16,17 @@ class B1<X extends A<Object?>> {}
 class B2<X extends A<dynamic>> {}
 class B3<X extends A<void>> {}
 class B4<X extends A<Never>> {}
```
eernstg, line 18: This one is actually not super-bounded.

Language/Generics/Superbounded_types/well_bounded_A01_t04.dart
```diff
@@ -16,10 +16,17 @@ class B1<X extends A<Object?>> {}
...
+class B5<X extends A<X>> {}
```
eernstg, line 19: Ditto.

Language/Generics/Superbounded_types/well_bounded_A01_t04.dart
```diff
@@ -16,10 +16,17 @@ class B1<X extends A<Object?>> {}
...
+
+mixin M<X extends A<X>> on A<X> {}
```
eernstg, line 21: Same here, but in order to cover mixins we could do this:
```suggestion
mixin M<X extends A<Object?>> on A<Never> {}
```

Language/Generics/Superbounded_types/well_bounded_A01_t04.dart
```diff
@@ -16,10 +16,17 @@ class B1<X extends A<Object?>> {}
...
+extension type ET<X extends A<X>>(A<X> _) implements A<X> {}
```
eernstg, line 22: Something like this should suffice:
```dart
extension type E<X extends A<Object?>>(String _) implements Object {}
```
We are just interested in the bound of `X`, so there's no need to fight with the representation type or the implemented types, just use something that works.

eernstg (commented):
Looks good! Just one test has a couple of cases that aren't super-bounded as claimed in the description.

sgrekhov:
Thank you! Updated. PTAL.

eernstg (approved):
LGTM

PR #2952, opened 2024-10-28
Fixes #2947. Expect AUGMENTATION_OF_DIFFERENT_DECLARATION_KIND in augmenting_types_A02_t14_lib.dart

### LINKED ISSUE #2947: co19/LanguageFeatures/Augmentation-libraries/augmenting_types_A02_t14
```dart
part 'augmenting_types_A02_t14_lib.dart';

mixin C {}
```

and in the part there is currently:
```dart
augment class C {} // errors

<cut>

augment mixin C {} // Ok
```

I'm currently making changes to the analyzer that break this `Ok`. The scenario that leads to the error is following: initially we give `C` the meaning `mixin C`, but when we see `augment class C {}` we change its meaning to `class C`, so when we get to `augment mixin C {}` we see that the meaning is not `mixin C` and report an error.

sgrekhov:


eernstg (approved):
I don't quite understand the explanation in https://github.com/dart-lang/co19/issues/2947, but this is about the _placement_ of error messages, not whether the situation is an error or not, so we'll just accept whatever the analyzer wants to do. ;-)

LGTM

PR #2955, opened 2024-10-29
Fixes #2954. Don't expect null-aware warnings in CFE

### LINKED ISSUE #2954: Expectations of warnings should be removed for the CFE in the tests for Null-aware elements
The following test expects warnings to be emitted by both the Analyzer and the CFE. However, reporting warnings is optional, and the CFE doesn't emit them.

co19/LanguageFeatures/Control-flow-collections/Null-aware-elements/constants_A01_t02

sgrekhov:


chloestefantsova (approved):
LGTM. Thanks!

PR #2966, opened 2024-11-05
#2559. Add extension types to `augmented` expression tests. Part 7.

sgrekhov:


LanguageFeatures/Augmentation-libraries/augmented_expression_A10_t02_lib.dart
```diff
@@ -362,3 +362,84 @@ augment extension Ext {
...
   }
 }
+
+augment extension type ET {
+  static var staticVariable = () {
+    int augmented = 42;
+//      ^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    for (int augmented = 0; augmented < 0; augmented++) {}
```
eernstg, line 372: Maybe avoid using the declaration? We could have a bug whereby the declaration is accepted, but the usages of `augmented` as an expression give rise to an error, and then we'd have a false positive on this error. Perhaps the following would suffice:
```suggestion
    for (int augmented = 0;;) {}
```

eernstg: (I guess this question wasn't raised when the non-new text was created, e.g., line 359)

eernstg (commented):
Looks good!

I think just one change would be useful (but it does seem to affect a bit of existing code as well): It is probably safer to avoid using `augmented` as an expression in a line which is supposed to give rise to an error because of the fact that `augmented` is also the name of a _declaration_ in the same line.

The point is (as usual) that we shouldn't test derived errors, and it's not easy to say what's the correct handling (silently skip over, or report error for) a location where an ill-defined variable is used.

sgrekhov:
Makes sense! Updated. PTAL.

eernstg (approved):
Thanks! LGTM

PR #2973, opened 2024-11-11
Fixes #2943. Use name `dartaotruntime` instead of `dart_precompiled_runtime`

### LINKED ISSUE #2943: The file co19/src/Utils/test_mode_check.dart uses the name 'dart_precompiled_runtime' to determine mode which could be incorrect
The CL https://dart-review.googlesource.com/c/sdk/+/389760 is making the name of the AOT executable consistent by using the
name '`dartaotruntime`'  (which is what the shipping Dart sdk uses). The executable '`dart_precompiled_runtime`' in the build directories is being renamed to '`dartaotruntime`'.

The file co19/src/Utils/test_mode_check.dart  uses the name '`dart_precompiled_runtime`' to determine if the test is running in AOT mode, this breaks with the above change. The check needs to be modified to use '`dartaotruntime`'

sgrekhov:


eernstg (approved):
LGTM

eernstg:
@a-siva, do you have further comments?

Issue #2956, opened 2024-10-29
Add more co19 tests for type `void`


sgrekhov:
Add more tests for "20.9 Type Void"

PR #2977, opened 2024-11-13
#2956. Add more tests for the type `void`

sgrekhov:


Language/Types/Type_Void/permitted_use_A12_t01.dart
```diff
@@ -0,0 +1,49 @@
...
+///   and `e1` is of the form `⟨expressionWithoutCascade⟩`.
+///   If `s` is an `⟨assignableSelector⟩` of the form `.id` or `?.id` where the
+///   static type of the identifier `id` is `void`, `e1` may have type `void`.
+///   Otherwise, if `s` is an `⟨assignableSelector⟩` of the form `[ e0 ]` where
+///   the static type of the first formal parameter in the statically known
+///   declaration of operator `[]=` is `void`, `e0` may have type `void`. Also,
+///   if the static type of the second formal parameter is `void`, `e1` may have
+///   type `void`.
+///
+/// @description Checks that if in an expression of the form `.. S s = e1` where
```
eernstg, line 23: We probably have to keep the words "ending in" somewhere here, because `.. S s = e1` isn't an expression.

Language/Types/Type_Void/permitted_use_A12_t02.dart
```diff
@@ -0,0 +1,42 @@
...
+///   and `e1` is of the form `⟨expressionWithoutCascade⟩`.
+///   If `s` is an `⟨assignableSelector⟩` of the form `.id` or `?.id` where the
+///   static type of the identifier `id` is `void`, `e1` may have type `void`.
+///   Otherwise, if `s` is an `⟨assignableSelector⟩` of the form `[ e0 ]` where
+///   the static type of the first formal parameter in the statically known
+///   declaration of operator `[]=` is `void`, `e0` may have type `void`. Also,
+///   if the static type of the second formal parameter is `void`, `e1` may have
+///   type `void`.
+///
+/// @description Checks that if in an expression of the form `.. S s = e1` where
```
eernstg, line 23: Needs "ending in" again.

eernstg (commented):
Looks good! There are a couple of cases where I believe we should delete a case (because it tries to assign a `void` expression result to a `Null` variable, which causes some noise because we have both `void` rules and assignability rules). A couple of other tests have a `@description` which is slightly confusing.

Language/Types/Type_Void/permitted_use_A12_t01.dart
sgrekhov: Updated

Language/Types/Type_Void/permitted_use_A12_t02.dart
sgrekhov: And here

sgrekhov (commented):
Updated. PTAL

eernstg (approved):
LGTM

Issue #2979, opened 2024-11-14
co19 tests for NNBD static errors needs update


sgrekhov:
[Errors and Warnings](https://github.com/dart-lang/language/blob/main/accepted/2.12/nnbd/feature-specification.md#errors-and-warnings) section of NNBD specifications has updates that were not reflected in appropriate co19 tests. Check the tests and the spec text and update the tests when needed.

PR #2984, opened 2024-11-19
#2979. Update existing and add new tests checking nnbd static errors

sgrekhov:


LanguageFeatures/nnbd/static_errors_A07_t01.dart
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is an error to derive a mixin from a class declaration which
+/// contains an instance variable with a potentially non-nullable type and no
+/// initializer expression unless the variable is marked with the `late`
+/// modifier.
+///
+/// @description Check that it is an error to derive a mixin from a class
```
eernstg, line 10: It's a little bit funny because we can't "derive a mixin from a class declaration" in Dart 3.0 or higher. So the NNBD spec should really be updated to talk about a `mixin class` rather than deriving anything. But we aren't going to update the NNBD spec because it's being integrated into the language specification (yes, something _is_ happening there ;-), and we don't generally update feature specifications with later features.

So we basically need to treat the NNBD spec as if it _had already_ been updated, such that newly written tests will test the current language.

So the description could perhaps be something like this:

"Check that it is an error for a mixin class to contain an instance variable ..."

This shows the reader that this test was written for a language that has mixin classes, and the reader will then know which kind of thinking gave rise to this particular snippet of code.

LanguageFeatures/nnbd/static_errors_A07_t01.dart
```diff
@@ -0,0 +1,30 @@
...
+mixin class C {
+  Object instance;
+//       ^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+mixin M1 on C {
+}
+
```
eernstg, line 24: I think `M1` could just as well be deleted because there is no need to use `C` as a class, the test is exclusively about an error that may occur in a mixin (and hence also in a mixin class).
```suggestion
```

LanguageFeatures/nnbd/static_errors_A07_t02.dart
```diff
@@ -0,0 +1,28 @@
...
+/// `late`.
+/// @author sgrekhov22@gmail.com
+
+mixin class C {
+  late Object instance;
+}
+
+mixin M1 on C {
+}
+
```
eernstg, line 22: Similar:
```suggestion
```

LanguageFeatures/nnbd/static_errors_A07_t02.dart
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It is an error to derive a mixin from a class declaration which
+/// contains an instance variable with a potentially non-nullable type and no
+/// initializer expression unless the variable is marked with the `late`
+/// modifier.
+///
+/// @description Check that it is not an error to derive a mixin from a class
```
eernstg, line 10: This description could be adjusted similarly.

LanguageFeatures/nnbd/static_errors_A20_t04.dart
```diff
@@ -0,0 +1,31 @@
...
+class A {
+  int one = 1;
+}
+
+class C extends A {
+  int two = 2;
+  late Function p1 = () => this.one;
+  late final p2 = () {
+    return this.two;
+  };
```
eernstg, line 25: I think we should have a case that shows a direct reference to `this`, not wrapped by a function literal:

```dart 
class A {
  late final x = this.y; // OK!
  final y;
  A(this.y);
}
```

eernstg (commented):

LanguageFeatures/nnbd/static_errors_A07_t01.dart
sgrekhov: Description updated

LanguageFeatures/nnbd/static_errors_A07_t01.dart
sgrekhov: Removed

LanguageFeatures/nnbd/static_errors_A07_t02.dart
sgrekhov: Yes, updated similarly

LanguageFeatures/nnbd/static_errors_A07_t02.dart
sgrekhov: Deleted

LanguageFeatures/nnbd/static_errors_A20_t04.dart
sgrekhov: The test `static_errors_A20_t03.dart` has this. It contains https://github.com/dart-lang/co19/blob/31124f678d4a8691c52b3bec29c7719317acd5e9/LanguageFeatures/nnbd/static_errors_A20_t03.dart#L14-L20

sgrekhov (commented):
Descriptions updated. PTAL

LanguageFeatures/nnbd/static_errors_A07_t01.dart
eernstg: Thanks!

LanguageFeatures/nnbd/static_errors_A20_t04.dart
eernstg: OK, great!

eernstg (approved):
LGTM

PR #2989, opened 2024-11-21
#2559. Update assertions for some Augmenting libraries tests

sgrekhov:
This makes updated assertions wisible for spec coverage viewer tool


eernstg (approved):
LGTM

PR #2997, opened 2024-11-25
#2956. Add more type `void` tests

sgrekhov:


Language/Functions/function_body_short_syntax_A03_t04.dart
```diff
@@ -0,0 +1,34 @@
...
+/// - of the form `=> e` or the form `async => e`, which both return the value
+///   of the expression `e` as if by a `return e`. ...  Let `T` be the declared
+///   return type of the function that has this body. It is a compile-time error
+///   if one of the following conditions hold:
+///   – The function is synchronous, `T` is not `void`, and it would have been a
+///     compile-time error to declare the function with the body `{ return e; }`
+///     rather than `=> e`.
+///
+/// @description Checks that it is not an error to declare a synchronous
+/// function with a short syntax and the return type `FutureOr<void>`.
```
eernstg, line 16: That's true. The type `FutureOr<void>` (which is a [bad type](https://dart.dev/tools/linter-rules/avoid_futureor_void) anyway, but we still need to test it) does not get special treatment. In particular, `FutureOr<void>` isn't "a kind of void". (Similarly, `List<void>` doesn't get special treatment, so there is no error if we do things like `List<Object?> xs = <void>['secret!'];`.)

`FutureOr<void>` is a top type, though, so we can return anything we want when the return type of a non-asynchronous function is `FutureOr<void> (except that expressions of type `void` can't be used unless the situation is on the allowlist, and the allowlist doesn't say anything about `FutureOr<void>`).

It looks like the analyzer emits warnings, but that's not part of the language definition, so we shouldn't interfere with that.

eernstg (approved):
LGTM. I left a comment, but it just confirms the choices made in the test.

Issue #2976, opened 2024-11-13
[Enum Shorthands] Add co19 tests for "static access shorthand" feature


sgrekhov:
Specification https://github.com/dart-lang/language/blob/main/working/3616%20-%20enum%20value%20shorthand/proposal-simple-lrhn.md

kallentu:
Added this to the enum shorthands project.
Thanks for working on this @sgrekhov .

sgrekhov:
Closing as completed

PR #2998, opened 2024-11-26
#2976. Add semantics tests

sgrekhov:


LanguageFeatures/Static-access-shorthand/semantics_A01_t01.dart
```diff
@@ -0,0 +1,121 @@
...
+/// `.id(args)`, `.id<typeArgs>(args)`, `.new` or `.new(args)`.
+/// ...
+/// The general rule is that any of the expression forms above, starting with
+/// `.id`, are treated exactly as if they were prefixed by a fresh identifier
+/// `X` which denotes an accessible type alias for the greatest closure of the
+/// context type scheme of the following primary and selector chain.
+///
+/// @description Checks that expressions of the form `.id`, `.id(args)` and
+/// `.id<typeArgs>(args)` are treated as if they are prefixed by a fresh
+/// identifier `X` which denotes an accessible type alias.
```
eernstg, line 25: This is _really_ hard to test! ;-)

The point could be that a type alias can be non-generic and still denote a parameterized type (`typedef X = SomeType<With, Args>;`), and this allows for the constructor invocations to have the actual type arguments, and it also allows for a static method invocation to _ignore_ the same type arguments (because that's how we treat type aliases when used to access static members).

Apart from this, there would be no distinction between this description and a shorter one that simply says "the semantics of `.id` is `T.id`, where `T` is the greatest closure of the context type.

However, that works for invocations of static methods, but it is _not_ the actual semantics of constructor invocations: For them we strip off the actual type arguments (so the greatest closure operation is ignored ... for both static member invocations and constructor invocations!), and then type inference is performed as if no actual type arguments were passed.

This means that it isn't hugely meaningful to test "as if prefixed by a type alias" at all.

What we need to test is that it works to call a static method even in the case where the context type has actual type arguments (where those actual type arguments must be ignored).

Next, we should test that it works to invoke a constructor in a case where the context type has actual type arguments (e.g., `List<num>` as context type should make `.filled(10, 10)` mean `List<num>.filled(10, 10)`, and the type argument should not be `int`, as it would be if the context type were `_`).

Finally, it should work to invoke a constructor with a context type schema (say, `List<Iterable<_>>`), and we should then see that the actual arguments to the constructor invocation plays a role:

```dart
Object foo<X>(Map<int, X> map) => map;
var map = foo(.fromIterables([1], ['Hello']));
```
In this case we'd get `int` from the context type and `String` from the actual argument `['Hello']`, and the result would then be a `Map<int, String>`.

LanguageFeatures/Static-access-shorthand/semantics_A01_t01.dart
```diff
@@ -0,0 +1,121 @@
...
+  Expect.equals(1, et1.t);
+
+  ET<String> et2 = .id2("et2");
+  Expect.equals("et2", et2.t);
+
+  ETAlias<int> et3 = .id1;
+  Expect.equals(1, et3.t);
+
+  ETInt et4 = .id2<int>(4);
+  Expect.equals(4, et4.t);
```
eernstg, line 120: I think this test is fine, but it doesn't test the location in the feature specification which is mentioned in the `@assertion`, it just tests that the processing of the context type includes a type alias expansion, as it should.

So the `@description` should be adjusted to match this situation.

The remaining tests are fine in the same sense as well.

LanguageFeatures/Static-access-shorthand/semantics_A01_t04.dart
```diff
@@ -0,0 +1,114 @@
...
+
+extension type ET(int _) implements Object {
+  static X id<X>(X x) => x;
+}
+typedef ETAlias = ET;
+
+main() {
+  Object o = C();
+  if (o is C) {
+    o = .id<int>;
```
eernstg, line 57: Nice!

eernstg (commented):
Looks good, with a couple of comments!

The feature specification talks about "like an accessible type alias" which is a somewhat Byzantine way to say that we can ignore type arguments as needed (namely, when we're invoking a static member of the context type).

This means that we basically can't test whether it is actually "like an accessible type alias", because there's no way we could do anything _other_ than ignoring any actual type arguments in cases like `.myStaticMethod()` with context type `C<String>`.

However, it gets worse: Constructor invocations _will_ ignore the actual type arguments in the context type, and use normal inference to find them. Often, this means to "re-find" the actual type arguments because we will frequently infer exactly the same type arguments that we just erased. However, it actually matters in some cases, especially with non-trivial selector chains.

So I'm suggesting that the `@description` is adjusted to say that these tests are just testing that when we're computing a static namespace from a given context type we will also expand all type aliases.

eernstg (approved):
LGTM

sgrekhov:
Thank you for the detailed review and explanation! I've updated tests descriptions to say that we are testing that processing of a context type includes type aliases (is there a better place in the spec to test it?). In the `main()` function the only checks for the type aliases left (there are and there will be plenty of checks for non-type aliases).  In the tests
 ```dart
  if (o is C) {
    o = .id<int>;
...
  }
```
I'm testing not type aliases only because there are no such tests yet.

Tests that check that type arguments should be ignored will be added in a separate PR as a type inference tests.

File: LanguageFeatures/Static-access-shorthand/semantics_A01_t01.dart (hunk 1 above)
Comment (lrhn) [reply LanguageFeatures/Static-access-shorthand/semantics_A01_t01.dart]: > This is really hard to test! ;-)

And the spec also doesn't use the "like a type alias" phrasing any more. Since yesterday.

What's important to test is that it:
- Accesses the correct static member.
- Does so, even if the name of that type is shadowed, or not even in scope.

so `.filled` with context type `List<int>` denotes (the constructor) `List.filled`. Even if `List` is not in scope.

The current spec also doesn't use "the greatest closure of the context type" any more. An occurrence of `_` doesn't matter, it's either in a position where it's irrelevant, or the scheme is `_` in which case it doesn't denote a static namespace.

To find the static namespace, it just looks at the context type scheme, and if it's of the form `S?` or `FutureOr<S>`, it recurses on the `S`, and otherwise it only recognizes *types* of the form `C`/`C<typeArgs>` where `C` is a type introduced by a `class`, `mixin`, `enum` or `extension type` declaration.

(It's written declaratively, "a type scheme denotes a static namespace *N* if and only if either the type scheme is a type introduced by a type declaration and *N* is the static namespace of that declaration, or the type scheme is either `S?` or `FutureOr<S>` and *S* denotes the static namespace *N*.".)

So if the context type scheme is a non-generic type or instantiated generic type scheme whose type is introduced by the declaration *D*, then `.id` denotes the static member named `id` or constructor with base name `id` in *D*.
And if it's a constructor, then it's a *raw* reference to that constructor, where type arguments to the class can be inferred from arguments or *actual* context type (not shorthand context, in the cases where those differ.)

Comment (sgrekhov) [reply LanguageFeatures/Static-access-shorthand/semantics_A01_t01.dart]: >-  Does so, even if the name of that type is shadowed, or not even in scope.
> so `.filled` with context type `List<int>` denotes (the constructor) `List.filled`. Even if `List` is not in scope.

How can we have a context type `List<int>`  if `List` is not in scope or shadowed? It's just a compile-time error.
```dart
class C {
  int List() => 42;
  
  test() {
    List<int> l = List.filled(1, 1);
//  ^^^^  Error: List isn't a type
  }
}
```
```dart
class List<T> {}

main() {
  List<int> l = List<int>.filled(1, 1);
  //                      ^^^^^^ Error: Couldn't find constructor 'List.filled'.
}
```
Please, clarify.

eernstg: > How can we have a context type `List<int>` if `List` is not in scope or shadowed?

You can declare a function like `void foo<X>(List<X> xs) {}` in another library 'lib.dart', and then call `foo` from a library that imports 'lib.dart' and imports 'dart:core' with a prefix such that `List` isn't in scope without the prefix.

sgrekhov: Thank you!

PR #3000, opened 2024-11-28
#2976. Update assertion texts according to the new spec

sgrekhov:


eernstg:
Wow, 3000 -- We should have cake! ;-)

eernstg (approved):
LGTM

PR #3017, opened 2024-12-16
#2559. Add metadata tests. Part 1.

sgrekhov:


sgrekhov:
Sorry, not ready for review yet

sgrekhov:
Ready for review now.

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t28.dart
```diff
@@ -0,0 +1,172 @@
...
+/// @assertion The exact result of an `augmented` expression depends on what is
+/// being augmented, but it generally follows the same rules as any normal
+/// identifier:
+/// - Augmenting getters: Within an augmenting getter `augmented` invokes the
+///   augmented getter and evaluates to its return value. If augmenting a
+///   variable with a getter, this will invoke the implicitly induced getter
+///   from the augmented variable declaration.
+///
+/// @description Checks that it is a compile-time error if an augmenting getter
+/// has a metadata named `augmented`.
```
eernstg, line 14: I think this is a little bit to strong, it should only be an error for the metadata `@augmented()` to occur _inside_ a declaration which is augmenting (that is: which has the modifier `augment`). It is not an error (see line 27) for an augmenting declaration to have such metadata, as long as it's not inside a construct that has the modifier `augment`.

The reason why it's fine to have "the same thing in a different place" is that the problem is the confusion which can be caused by the _word_ `augmented` occurring in a scope where `augmented`/`augmented()` can be an invocation of an augmented method, getter, setter, or initializing expression. There is nothing wrong with the _meaning_ of `augmented` in these cases, they would have been perfectly OK if their name had been different.

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t28.dart
```diff
@@ -0,0 +1,172 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+  @augmented() String get instanceGetter2 => "Augmented";
+// ^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+extension type ET(int _) {
+  @augmented()
```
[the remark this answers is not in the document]
eernstg: And several more in the following test libraries.

LanguageFeatures/Augmentation-libraries/augmented_expression_A02_t26.dart
```diff
@@ -0,0 +1,171 @@
...
+/// being augmented, but it generally follows the same rules as any normal
+/// identifier:
+/// ...
+/// - Augmenting setters: Within an augmenting setter `augmented` must be
+///   followed by an `=` and will directly invoke the augmented setter. If
+///   augmenting a variable with a setter, this will invoke the implicitly
+///   induced setter from the augmented variable declaration.
+///
+/// @description Checks that it is a compile-time error if an augmenting setter
+/// has a metadata named `augmented`.
```
eernstg, line 15: Same kind of adjustment as with the getter. Similarly for other kinds of members in the following test libraries.

eernstg (commented):
Looks good! I'm suggesting that one extra case is covered (`@augmented()` is OK on an augmenting declaration as long as it isn't _inside_ that declaration). This gives rise to a somewhat large number of changes, but they are very systematic.

sgrekhov (commented):
Thank you! Updated. PTAL.

LanguageFeatures/Augmentation-libraries/augmented_expression_A01_t28.dart
eernstg: Works - thank you!

PR #3018, opened 2024-12-17
#2559. Fix metadata tests

sgrekhov:


LanguageFeatures/Augmentation-libraries/metadata_A01_t02.dart
```diff
@@ -68,13 +71,14 @@ extension type ET(int id) {
...
-  var symbol = MirrorSystem .getSymbol("topLevelVariable");
+  var symbol = MirrorSystem.getSymbol("topLevelVariable");
   DeclarationMirror varMirror =
-    libraryMirror.declarations[symbol] as DeclarationMirror;
-  Expect.equals('.Meta1',
-      MirrorSystem.getName(varMirror.metadata[0].type.qualifiedName));
-  Expect.equals('.Meta2',
-      MirrorSystem.getName(varMirror.metadata[1].type.qualifiedName));
+      libraryMirror.declarations[symbol] as DeclarationMirror;
+  print(MirrorSystem.getName(varMirror.metadata[0].type.qualifiedName));
```
eernstg, line 77: Leftover debugging code?

eernstg (commented):
Looks very good, thanks! I think one line might be a remnant from debugging?

This means that we won't complain if the order of metadata objects differs from what we'd expect by looking at the syntax, and that's fine: There is no language specification rule to prevent that. (Also, it seems harmless in practice.)

LanguageFeatures/Augmentation-libraries/metadata_A01_t02.dart
sgrekhov: Yes, thank you! Removed.

eernstg (approved):
LGTM

PR #3020, opened 2024-12-18
Fix Finalizer_A01_t04.dart test

sgrekhov:
See https://github.com/dart-lang/sdk/issues/59752


eernstg (approved):
LGTM.
@dcharkes, it would be great if you could take a look as well.

PR #3021, opened 2024-12-19
#2976. Add equality tests. Part 1.

sgrekhov:


LanguageFeatures/Static-access-shorthand/equality_A01_t01.dart
```diff
@@ -0,0 +1,81 @@
...
+
+/// @assertion If an expression has the form `e1 == e2` or `e1 != e2`, or a
+/// pattern has the form `== e2` or `!= e2`, where the static type of `e1`, or
+/// the matched value type of the pattern, is S1, and e2 is precisely a
+/// `<staticMemberShorthand>` expression, then assign the type S1 as the
+/// shorthand context of the `<staticMemberShorthandHead>` of e2 before
+/// inferring its static type the same way as above.
+///
+/// @description Checks that if an expression has the form `e1 == e2` or
+/// `e1 != e2` and `e2` is a shorthand expression, then it has context type from
```
eernstg, line 13: I think it would be helpful to use the terminology from the feature spec:
```suggestion
/// `e1 != e2` and `e2` is a shorthand expression then it has shorthand context from
```
The point is that it isn't actually the context type for `e2`. The context type for `e2` is determined by the statically known declaration of operator `==` and static access shorthands won't change anything about that. This means that equality expressions is _the_ topic where we need to make it explicit that we have a context type, and we have a shorthand context, and they are not the same (even though they are the same everywhere else).

LanguageFeatures/Static-access-shorthand/equality_A02_t01.dart
```diff
@@ -0,0 +1,73 @@
...
+/// This special-casing is only against an immediate static member shorthand. It
+/// does not change the context type of the second operand, so it would not work
+/// with, for example, `Endian.host == wantBig ? .big : .little`. Here the
+/// second operand is not a `<staticMemberShorthand>`, so it won't have a
+/// shorthand context set, and the context type of the second operand of `==` is
+/// the empty context `_`. (It's neither the static type of the first operand,
+/// nor the parameter type of the first operand's `operator==`.)
+///
+/// @description Checks that it is a compile-time error if the right operand of
+/// `==` operator isn not a precisely static member shorthand but a ternary
```
eernstg, line 17: There are a few occurrences of this one:
```suggestion
/// `==` operator is not a precisely static member shorthand but a ternary
```

eernstg (commented):
Looks good, noted one typo (in a few sentences that are similar).

Another thing came to mind, for a different PR if that's more convenient:

It may look like an oversight that the patterns `== e` and `!= e` allow `e` to be a `<staticMemberShorthand>` because this allows for `<selector>*` at the end, and we almost never have a constant expression when this sequence of `<selector>` is non-empty.

However, we _can_ do a few things: `== 'Hello'.length` should be supported. So we should now be able to do `== .aString.length` as well. Similarly, `== .myGenericStaticMethod<int>` should be possible. It is probably a tricky exercise to find a suitable context type such that the comparison will be non-trivial, but it shouldn't be a big problem even if `unrelated_type_equality_checks` will complain.

There may be more, but I can't immediately come up with any other examples. Perhaps there would be some extra ones popping up by looking at the language specification section about constant expressions.

sgrekhov:
Thank you! Descriptions are updated. I'll add constants with non-empty selectors in a separate PR

eernstg (approved):
LGTM, thanks!

PR #3024, opened 2024-12-24
#2976. Add constant expressions tests

sgrekhov:


LanguageFeatures/Static-access-shorthand/constant_expression_A07_t01.dart
```diff
@@ -0,0 +1,54 @@
...
+  const C();
+  const C.id();
+}
+
+extension type const ET<X, T>(int _) {
+  const ET.id(this._);
+}
+
+void foo<T>() {
+  C<int, T> c1 = const .new();
```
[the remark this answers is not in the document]
eernstg: Seems like `const C<int, void Function<Y extends T>()> c1 = const .new();` will suffice for now. I created https://github.com/dart-lang/language/issues/4223 in order to clarify the topic.

LanguageFeatures/Static-access-shorthand/constant_expression_A08_t02.dart
```diff
@@ -0,0 +1,43 @@
...
+  const C c = .id<int>();
+//            ^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  const ET et = .id<int>();
+//              ^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
```
eernstg, line 43: Same thing here, it would be useful to verify that the error is about constancy and not something else.

LanguageFeatures/Static-access-shorthand/constant_expression_A09_t01.dart
```diff
@@ -0,0 +1,84 @@
...
+
+import '../../Utils/expect.dart';
+
+class C {
+  final String value;
+  const C(this.value);
+  const C.id(this.value);
+  const factory C.f(String value) = C;
+
+  static const C instance = const C("C instance");
```
eernstg, line 25: Could we use the type `C?` to avoid some `unnecessary_non_null_assertion`s below?

LanguageFeatures/Static-access-shorthand/constant_expression_A09_t01.dart
```diff
@@ -0,0 +1,84 @@
...
+}
+
+mixin M on C {
+  static const M instance = const MC("M instance");
```
eernstg, line 29: Type `M?`?

LanguageFeatures/Static-access-shorthand/constant_expression_A09_t01.dart
```diff
@@ -0,0 +1,84 @@
...
+  const E(this.value);
+
+  static const E instance = E.e0;
+}
+
+extension type const ET(int value) {
+  const ET.id(this.value);
+  const factory ET.f(int value) = ET;
+
+  static const zero = const ET(0);
```
eernstg, line 48: Type `ET?`?

LanguageFeatures/Static-access-shorthand/constant_expression_A09_t01.dart
sgrekhov: Yes. Updated here and below.

LanguageFeatures/Static-access-shorthand/constant_expression_A08_t01.dart
```diff
@@ -34,31 +34,37 @@ extension type const ET(int _) {
 }
 
 main() {
+  C nc1 = .id1; // Ok
   const C c1 = .id1;
-//             ^^^^
+//              ^^^^
```
eernstg, line 39: Unintended?

LanguageFeatures/Static-access-shorthand/constant_expression_A09_t01.dart
```diff
@@ -58,16 +58,16 @@ main() {
   const C c3 = const .f("three")!; // ignore: unnecessary_non_null_assertion
   Expect.equals("three", c3.value);
 
-  const C c4 = .instance!; // ignore: unnecessary_non_null_assertion
+  const C c4 = .instance!;
```
eernstg, line 61: Oh, but that wasn't a constant expression after all.

LanguageFeatures/Static-access-shorthand/constant_expression_A07_t01.dart
```diff
@@ -31,20 +31,20 @@ extension type const ET<X, T>(int _) {
 }
 
 void foo<T>() {
-  C<int, T> c1 = const .new();
-//                     ^^^^
+  C<int, void Function<Y extends T>()> c1 = const .new();
```
eernstg, line 34: This will stop working if we decide that `const .new()` should be inferred as `const C<int, Never>.new()`, which would not have any errors.

I guess we can just proceed for now (keeping this test as it is with this PR), and handle the missing compile-time error later if constant-expression inference is changed in this manner.

eernstg: Similarly for the other cases.

LanguageFeatures/Static-access-shorthand/constant_expression_A02_t01.dart
```diff
@@ -0,0 +1,47 @@
...
+/// @assertion Given an expression that is a prefix of
+/// `<staticMemberShorthandHead> <selector>*`, whose assigned shorthand context
+/// denotes a declaration D, and where the identifier or `new` of the
+/// `<staticMemberShorthandHead>` denotes a static declaration or constructor
+/// declaration S when looked up on D.
+/// ...
+/// - An expression of the form `.<identifier>` is a constant expression if S
+///   declares a constant getter.
+///
+/// @description Checks that expressions of the form `<identifier>` is a
```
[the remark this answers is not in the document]
eernstg: It is probably more readable for readers who are used to reading grammar related information if we keep the tokens separated by whitespace, that is, `'.' <identifier>`. I think syntax people usually do this simply because they think it looks better, but it might also make sense to have the spaces such that it is (almost certainly) possible to use the space to disambiguate quotes: " '" is a single quote begin and "' " is a single quote end. If we drop the spaces then we can have `'things''like'<demonstrativePronoun>` where `''` might be confusing to read, especially if the non-terminals aren't marked by angle brackets: `'things''like'demonstrativePronoun`

LanguageFeatures/Static-access-shorthand/constant_expression_A08_t01.dart
```diff
@@ -0,0 +1,66 @@
...
+  const ET et2 = .id1();
+//               ^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  const ET et3 = .id2();
+//               ^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
```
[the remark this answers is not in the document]
eernstg: Thanks!

sgrekhov:
`e!` test is removed, descriptions are updated. PTAL

eernstg (approved):
LGTM, thanks!

PR #3025, opened 2024-12-25
#2976. Add constant expressions tests.Part 2.

sgrekhov:


LanguageFeatures/Static-access-shorthand/constant_expression_A10_t01.dart
```diff
@@ -0,0 +1,32 @@
...
+///
+/// @description Checks that a qualified reference to a static constant variable
+/// that is not qualified by a deferred prefix, is a potentially constant and
+/// constant expression
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=enum-shorthands
+
+import '../../Utils/expect.dart';
+import 'shorthand_lib.dart'as p;
```
eernstg: Typo:
```suggestion
import 'shorthand_lib.dart' as p;
```

eernstg (approved):
LGTM. Just one missing space character, and I added that.

I noticed that there are many occurrences of `static const v = const ...` where the `const` after `=` can be omitted. Let's just keep them. Tests aren't supposed to follow all the style rules, they are supposed to cover a lot of cases.

PR #3027, opened 2024-12-27
#2976. Add grammar tests for shorthand with different operators.

sgrekhov:


LanguageFeatures/Static-access-shorthand/grammar_A04_t01.dart
```diff
@@ -0,0 +1,79 @@
...
+/// <staticMemberShorthandHead> ::=
+///       <staticMemberShorthandValue>
+///     | '.' 'new'                                       -- shorthand unnamed constructor
+///
+/// <staticMemberShorthandValue> ::=                      -- something that can potentially create a value.
+///     | '.' <identifier>                                -- shorthand for qualified name
+///     | 'const' '.' (<identifier> | 'new') <arguments>  -- shorthand for constant object creation
+///  ```
+///
+/// @description Checks that the static access short syntax can be used with
```
eernstg, line 24: Let's just stick to the standard terminology:
```suggestion
/// @description Check that static access shorthand syntax can be used with
```

eernstg (commented):
Looks good!

Perhaps we should have a test on `>>>`, `[]=`, and `[]`, now that we have all the others?

We might also consider `<compoundAssignmentOperator>` (I guess they can be tested along with their suboperator, so the `+` test could have some `+=` cases as well).

LanguageFeatures/Static-access-shorthand/grammar_A04_t01.dart
eernstg: We could also use 'Checks that static access shorthands can be used with', just to avoid changing the next line if this line gets too long.

eernstg (commented):
(Looks like I have to press this button one more time in order to send the latest comment)

sgrekhov:
- Descriptions updated
- Compound assignment operators added
- Test for `>>>` operator added

Test for `[]` and `[]=` operators will be added in a separate PR. Please take another look.

PR #3032, opened 2024-12-31
Fixes #3028. Add tests for mixin application constant constructor

### LINKED ISSUE #3028: Add tests for constant constructor and mixin application
Add co19 test(s) covering https://github.com/dart-lang/sdk/issues/59796

sgrekhov:


eernstg (approved):
LGTM.

I was just wondering about one thing: The tests starting with `...A03_t06...` seem to differ by having different kinds of parameters (optional named, required named). I couldn't see an obvious reason why this would help testing anything which is concerned with constant constructors?

If there is a need to change anything then we can do it in a new PR, I'll land this one now.

sgrekhov:
> I was just wondering about one thing: The tests starting with `...A03_t06...` seem to differ by having different kinds of parameters (optional named, required named). I couldn't see an obvious reason why this would help testing anything which is concerned with constant constructors?

There are 3 statements in the specification, about constructor with no optional parameters, with optional positional parameters and with named parameters. `...A03...` tests checking mixin application constructor with named parameters.
`...t06`: optional named parameter, mixin has a static variable,  mixin application of the form `class MA = A with M;`
`...t07`: required named parameter, mixin has a static variable,  mixin application of the form `class MA = A with M;`
`...t08`: optional named parameter, mixin has a static variable,  mixin application of the form `class MA extends A with M...`
`...t09`: required named parameter, mixin has a static variable,  mixin application of the form `class MA extends A with M...`
and 4 more test but with the instance variable in a mixin.

I believe it makes sense to check all combinations. What if mixin application constructor looses its constantness, say, in case of required named parameter only? See, for example, https://github.com/dart-lang/sdk/issues/59796.

eernstg:
Makes sense, thank you!

PR #3036, opened 2025-01-07
Fixes #3030. Add additional test checking non const in a const constructor initializer list

### LINKED ISSUE #3030: Add tests for non-constant values in a constant constructor initializer list
See https://github.com/dart-lang/sdk/issues/59804

sgrekhov:
Github shows a wrong diff as if `..._02` was renamed to `...t03` and `...t03` to `...t02`. Hope that it won't be a problem for the review.


eernstg (approved):
LGTM

PR #3037, opened 2025-01-08
#2976. Add constant expressions tests. Part 4.

sgrekhov:


LanguageFeatures/Static-access-shorthand/constant_expression_A03_t03.dart
```diff
@@ -0,0 +1,63 @@
...
+    }
+    return true;
+  }
+}
+class MA = Object with M;
+
+test<T>() {
+  Expect.isFalse(C() == .foo<T>);
+  Expect.isFalse((MA() as M) == .foo<T>);
+}
```
eernstg, line 59: I'm a little bit worried about this because it _requires_ certain torn-off functions to be non-canonicalized. I don't think we want to prevent it if a compiler will actually detect that it can apply this optimization in some cases.

We could test that it doesn't crash, and ignore the returned boolean value.

I think it would be nice to have a comment, as well, explaining why we do not require the result to be false.

LanguageFeatures/Static-access-shorthand/constant_expression_A04_t02.dart
```diff
@@ -0,0 +1,38 @@
...
+  C(this.t);
+
+  @override
+  bool operator ==(Object other) {
+    return identical(C<int>.new, other);
+  }
+}
+
+main() {
+  Expect.isTrue(C(1) == .new);
```
eernstg, line 37: Wouldn't `.new` expand as something like `C<Object?>.new` or `C<dynamic>.new`? In that case we'd get false.

LanguageFeatures/Static-access-shorthand/constant_expression_A04_t03.dart
```diff
@@ -0,0 +1,42 @@
...
+  C(this.t);
+
+  @override
+  bool operator ==(Object other) {
+    return identical(C<int>.new, other);
+  }
+}
+
+void test<T>(T t) {
+  Expect.isFalse(C<T>(t) == .new);
```
eernstg, line 37: This would need to be adjusted as well, such that we don't succeed just because the type arguments are different. Also, this would be one more case where we don't want to _prevent_ canonicalization as an optimization.

eernstg (commented):
Looks good! Tricky stuff!

I think it would be useful to consider extension types one more time: They can obtain an implementation of operator `==` from a non-extension type superinterface. There are several tests where an extension type case could probably be added, I commented on the first couple of cases I noticed.

LanguageFeatures/Static-access-shorthand/constant_expression_A04_t02.dart
sgrekhov: Yes, really. Removed this test. Short syntax assumes that `.new` always means `C.new` and in case of tear-off we don't have actual paramenters to trigger an inference. So, I don't think that we can test this case (when type inference has inferred constant type arguments for the target class).

LanguageFeatures/Static-access-shorthand/constant_expression_A04_t03.dart
sgrekhov: The same as the above. Removed this test.

sgrekhov (commented):
Thanx! Updated, extension type cases added, some tests removed. PTAL

LanguageFeatures/Static-access-shorthand/constant_expression_A03_t01.dart
```diff
@@ -0,0 +1,68 @@
...
+/// declaration S when looked up on D.
+/// ...
+/// - An expression of the form `.<identifier>` that is not followed by an
+///   `<argumentPart>`, is a constant expression if S declares a static method
+///   or constructor with base name `<identifier>`, and either type inference
+///   has not added type arguments as a generic function instantiation coercion
+///   to the method, or to the target class for a constructor, or the added type
+///   arguments are constant types.
+///
+/// @description Checks that an expression of the form `'.'<identifier>` that is
```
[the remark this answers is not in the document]
eernstg: Thanks!

LanguageFeatures/Static-access-shorthand/constant_expression_A03_t01.dart
```diff
@@ -0,0 +1,68 @@
...
+  bool operator ==(Object other) {
+    if (other is Function) {
+      Expect.equals("M", other());
+      return identical(M.foo, other);
+    }
+    return false;
+  }
+}
+class MA = Object with M;
+
```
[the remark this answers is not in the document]
eernstg: Great, thanks!

LanguageFeatures/Static-access-shorthand/constant_expression_A03_t03.dart
eernstg: Thanks! I really want to avoid preventing those optimizations, so it's a good thing to keep an eye on things that would outlaw them.

LanguageFeatures/Static-access-shorthand/constant_expression_A04_t02.dart
eernstg: OK!

eernstg (approved):
LGTM

PR #3047, opened 2025-01-15
#2976. Add more tests for records, factories and parenthesized expressions

sgrekhov:


LanguageFeatures/Static-access-shorthand/constant_expression_A01_t04.dart
```diff
@@ -0,0 +1,55 @@
...
+/// declaration S when looked up on D.
+/// - An expression of the form `const .id(arguments)` or
+///   `const .new(arguments)` is a constant expression. It's a compile-time
+///   error if S does not declare a constant constructor, and if any expression
+///   in `arguments`, which are all in a constant context, is not a constant
+///   expression.
+///
+/// @description Checks that expressions of the form
+/// `const '(' '.' id(arguments) ',' ')'` and
+/// `const '(' '.' new(arguments) ',' ')'` are constant expressions.
```
eernstg, line 18: I can't really see the connection to the assertion: The assertion is about an expression `e` of the form `<staticMemberShorthandHead> <selector>*` or `'.' <identifierOrNew>?` or `'const'` or `'const' '.'` or `'const' '.' <identifierOrNew>` or `'const' '.' <identifierOrNew> <arguments>` (or a bunch of other forms which are prefxies of `'const' '.' <identifierOrNew> <arguments>` that I haven't already mentioned) ...

(I'm not 100% convinced we want to allow for all those forms, but anyway)

then consider the case where `e` also has the form `'const' '.' <identifierOrNew> <arguments>`. In this case `e` is a constant expression, except that it is an error if it doesn't denote a constant constructor, or any argument is non-constant.

There's no reference to record literals anywhere, so where do they come from?

I think the natural thing to test would be that `const .name(14)` is a constant expression with context type `C` when `C.name` is a constant constructor that accepts this actual argument list.

Testing constant records is at least one step further away and doesn't actually test the use of `const .name(14)`. It actually tests the use of `.name(14)` as a constant expression because `.name(14)` occurs in a constant context when it occurs as `const (.name(14),)`.

So the test is OK in its own right, if the purpose is to test that constant record literals can contain static access shorthands denoting a constructor invocation, but it doesn't quite match the assertion.

eernstg: If we don't already have a test where a constant record literal contains a static access shorthand then we could just take a copy of this one and change the assertion, and then adjust this one to work on `const .name(14)` rather than `const (.name(14),)`.

eernstg (commented):
Looks good!

I just don't understand the connection between the assertion and the test in the first one.

LanguageFeatures/Static-access-shorthand/constant_expression_A01_t04.dart
sgrekhov: The idea here was to test parenthesized expressions. To check that `(.id)` is treated as `.id`. But in case of constants `const (...)` is always parsed as a constant record. I think it makes sense to check that `const (.id,)` is still treated as a record and `const (.id)` is not a parenthesized constant but a syntax error. Moved to grammar tests with the apropriate assertion (can't find the better one). PTAL.

LanguageFeatures/Static-access-shorthand/grammar_A01_t07.dart
```diff
@@ -1,21 +1,30 @@
...
+/// <staticMemberShorthandValue> ::=                      -- something that can potentially create a value.
+///     | '.' <identifier>                                -- shorthand for qualified name
+///     | 'const' '.' (<identifier> | 'new') <arguments>  -- shorthand for constant object creation
+///  ```
 ///
 /// @description Checks that expressions of the form
 /// `const '(' '.' id(arguments) ',' ')'` and
-/// `const '(' '.' new(arguments) ',' ')'` are constant expressions.
+/// `const '(' '.' new(arguments) ',' ')'` are records containing constant
+/// shorthand expressions.
```
eernstg, line 27: Sounds good!

LanguageFeatures/Static-access-shorthand/grammar_A01_t08.dart
```diff
@@ -0,0 +1,70 @@
...
+///     | '.' 'new'                                       -- shorthand unnamed constructor
+///
+/// <staticMemberShorthandValue> ::=                      -- something that can potentially create a value.
+///     | '.' <identifier>                                -- shorthand for qualified name
+///     | 'const' '.' (<identifier> | 'new') <arguments>  -- shorthand for constant object creation
+///  ```
+///
+/// @description Checks that expressions of the form
+/// `const '(' '.' id(arguments) ')'` and
+/// `const '(' '.' new(arguments) ')'` are still compile-time errors.
```
eernstg, line 26: Sounds good!

LanguageFeatures/Static-access-shorthand/constant_expression_A01_t04.dart
eernstg: Looks good, thanks!

eernstg (approved):
LGTM

PR #3052, opened 2025-01-21
#2145. Update existing "Evaluation of Implicit Variable Getters" tests

sgrekhov:
- Update assertions
- Rename
- Remove duplicated tests
- Add test cases for mixins, enums, extesions and extension types


sgrekhov:
Updated according to the https://github.com/dart-lang/sdk/issues/59945#issuecomment-2609853912. Please review.

Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A01_t03.dart
```diff
@@ -0,0 +1,101 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Case ⟨Static or library variable⟩. If `d` declares a static or
+/// library variable, the implicitly induced getter of `id` executes as follows:
+/// - Non-constant variable with an initializer. In the case where `d` has an
+///   initializing expression and is not constant, the implicitly induced getter
+///   of `id` is a late-initialized getter. This determines the semantics of an
+///   invocation.
+///
+/// @description Checks that it is not an error if during an evaluation of the
```
eernstg, line 12: Typo:
```suggestion
/// @description Checks that it is an error if during an evaluation of the
```

This is actually a case which is covered by dart-lang/sdk#50383, and the rule is even stronger than the description here: It is enough that, for example, the declaration of `sVar` has an initializing expression that contains a reference to `sVar`, the evaluation of the initializing expression doesn't have to need that value (which is in fact the case here, because `func` doesn't invoke its argument).

eernstg: So the error expectations below are fine, but the description could be strengthened to say that a circular dependency is an error even when it is only syntactic, and there's no actual infinite loop during evaluation.

eernstg (commented):
Looks good! One description may not have been updated?

eernstg (approved):
LGTM

Issue #2138, opened 2023-07-17
Add more `Type of a Function` tests


sgrekhov:
Add more tests for `9.3 Type of a Function`

sgrekhov:
There are no tests for the following statement

• of the form => e or the form async => e, which both return the value
of the expression e as if by a return e. The other modifiers do not apply
here, because they apply only to generators, discussed below. Generators are
not allowed to explicitly return anything, objects are added to the generated
stream or iterable using yield or yield*. Let T be the declared return type
of the function that has this body. It is a compile-time error if one of the
following conditions hold:
– The function is synchronous, T is not void, and it would have been a
compile-time error to declare the function with the body { return e;
} rather than => e

PR #3053, opened 2025-01-23
#2138. Add more tests for `call` method/getter

sgrekhov:
Based on the [test](https://dartpad.dev/?id=20c5ad78d018c515d829ff4e7191822f) mentioned in https://github.com/dart-lang/sdk/issues/55803#issuecomment-2124182059

### DIFF
--- a/Language/Expressions/Function_Invocation/Function_Expression_Invocation/call_A04_t01.dart
+++ b/Language/Expressions/Function_Invocation/Function_Expression_Invocation/call_A04_t01.dart
@@ -1,0 +1,182 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A function expression invocation i has the form
+/// ef <A1, . . . , Ar>(a1, . . . , an, xn+1: an+1, . . . , xn+k: an+k),
+/// where ef is an expression.
+/// ...
+/// Let F be the static type of ef . If F is an interface type that has a method
+/// named call, i is treated as the ordinary invocation
+/// ef .call<A1, . . . , Ar>(a1, . . . , an, xn+1: an+1, . . . , xn+k: an+k)
+///
+/// @description Checks that an implicit invocation of a `.call()` invokes an
+/// appropriate function.
+/// @author sgrekhov22@gmail.com
+/// @author lrhn
+/// @issue 59952
+
+// See https://github.com/dart-lang/sdk/issues/55803#issuecomment-2124182059
+
+import '../../../../Utils/expect.dart';
+
+void main() {
+  // Real functions, invoked typed or dynamic.
+  check = expr<int Function(int)>(good, "function")(42);
+  check = expr<Function>(good, "function as Function")(42);
+  check = expr<dynamic>(good, "function as dynamic")(42);
+
+  // These implicit `.call` invocations should work:
+
+  // Callable types.
+  check = expr<C>(C())(42);
+  check = expr<M>(MC())(42);
+  check = expr<E>(E.instance)(42);
+  check = expr<ET>(ET(null), "ET")(42);
+  check = expr<NSMC>(NSMC())(42); // NSM-forwarder as call method.
+  // Extension types inheriting those call methods.
+  check = expr<ETC>(C())(42);
+  check = expr<ETM>(MC())(42);
+  check = expr<ETE>(E.instance)(42);
+  check = expr<ETET>(ET(null), "ET")(42);
+  check = expr<ETNSMC>(NSMC())(42); // NSM-forwarder as call method.
+  // Callable extension methods.
+  check = expr<EC>(EC())(42);
+  check = expr<EM>(EMC())(42);
+  check = expr<EE>(EE.instance)(42);
+  check = expr<EET>(EET(null), "EET")(42);
+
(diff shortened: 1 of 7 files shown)


Language/Expressions/Function_Invocation/Function_Expression_Invocation/call_A04_t02.dart
```diff
@@ -0,0 +1,59 @@
...
+  }, null, "BM as dynamic did not throw");
+  Expect.throws(() {
+    (BE.instance as dynamic)(42);
+  }, null, "BC as dynamic did not throw");
+  Expect.throws(() {
+    (rec as dynamic)(42);
+  }, null, "Record as dynamic did not throw");
+}
+
+int bad(int x) => 100 + x;
```
eernstg, line 40: I don't think any of these tests are actually relying on this function having any bad properties. It's just a function like any other function of that type, and the point is that `call` is a getter, not that it returns something which is "bad" in some sense.

So perhaps the test would be slightly more self-explanatory if the function had a neutral name like `fun` or `foo`?

eernstg: Similarly for a few other tests. (Perhaps the name `bad` came from Language/Expressions/Function_Invocation/Function_Expression_Invocation/call_A04_t01.dart, where we actually have a `good` and a `bad` function and an invocation of the bad one implies that the semantics of some other thing was wrong?)

Language/Expressions/Function_Invocation/Function_Expression_Invocation/call_A04_t03.dart
```diff
@@ -0,0 +1,118 @@
...
+  int Function(int) get call => bad;
+}
+
+// Representation type getter is a getter.
+extension type BERT(int Function(int) call) {}
+
+class BNSMC {
+  // Should be NSM-forwarder.
+  int Function(int) get call;
+  Object? noSuchMethod(Invocation i) {
```
eernstg, line 93: Perhaps `Object? noSuchMethod(Invocation i) => null;` would suffice? The code won't run (unless the tool chain is highly broken ;-). The point is simply that the implementation seems to make the test less readable and less to-the-point.

eernstg (commented):
Looks good!

I've commented on the first test because I think it's a good idea to stick to a small number of language features in a language test (to remain focused). In the remaining tests there might be some elements which have taken inspiration from the first test and I'm suggesting that a couple of things a simplified slightly.

Language/Expressions/Function_Invocation/Function_Expression_Invocation/call_A04_t02.dart
```diff
@@ -0,0 +1,59 @@
...
+  // (dynamic invocation with `call` getter).
+  Expect.throws(() {
+    (BC() as dynamic)(42);
+  }, null, "BC as dynamic did not throw");
+  Expect.throws(() {
+    (BMC() as dynamic)(42);
+  }, null, "BM as dynamic did not throw");
+  Expect.throws(() {
+    (BE.instance as dynamic)(42);
+  }, null, "BC as dynamic did not throw");
```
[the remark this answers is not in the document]
sgrekhov: Thank you! Fixed.

Language/Expressions/Function_Invocation/Function_Expression_Invocation/call_A04_t02.dart
sgrekhov: Yes, makes sense. Tanx! Simplified and renamed to `foo` here and it the all other tests.

Language/Expressions/Function_Invocation/Function_Expression_Invocation/call_A04_t01.dart
```diff
@@ -0,0 +1,182 @@
...
+void main() {
+  // Real functions, invoked typed or dynamic.
+  check = expr<int Function(int)>(good, "function")(42);
+  check = expr<Function>(good, "function as Function")(42);
+  check = expr<dynamic>(good, "function as dynamic")(42);
+
+  // These implicit `.call` invocations should work:
+
+  // Callable types.
+  check = expr<C>(C())(42);
```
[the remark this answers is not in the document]
sgrekhov: First,  it was my decision to simplify the original test https://dartpad.dev/?id=20c5ad78d018c515d829ff4e7191822f but then found that in some cases it won't work. So, I decided to leave it as in the original test.

Language/Expressions/Function_Invocation/Function_Expression_Invocation/call_A04_t03.dart
sgrekhov: Indeed. Simplified.

sgrekhov (commented):
Thanx! Updated. PTAL.

Language/Expressions/Function_Invocation/Function_Expression_Invocation/call_A04_t01.dart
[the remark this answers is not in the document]
eernstg: Sounds good, thanks!

Language/Expressions/Function_Invocation/Function_Expression_Invocation/call_A04_t02.dart
eernstg: Thanks!

eernstg (approved):
LGTM

Issue #3054, opened 2025-01-24
Add more tests for Property extraction


sgrekhov:
Check the existing tests and add missing ones. In particular we have no tests for "17.22.3 Instance Method Closurization".

PR #3060, opened 2025-01-29
#3054. Add more method clozurization tests

sgrekhov:


Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_named_params_A04_t01.dart
```diff
@@ -0,0 +1,70 @@
...
+///  The parameter types `Tj,j ∈ 1..n + k`, are determined as follows: Let the
+///  method declaration `D` be the implementation of `m` which is invoked by the
+///  expression in the body. Let `T` be the class that contains D.
+///  ...
+///  If `T` is a non-generic class then for `j ∈ 1..n+k, Tj` is a type
+///  annotation that denotes the same type as that which is denoted by the type
+///  annotation on the corresponding parameter declaration in `D`. If that
+///  parameter declaration has no type annotation then `Tj` is `dynamic`.
+///
+/// @description Check that if `T` is a non-generic class then for
```
eernstg, line 28: But `C` is a generic class. We need a test for both cases, so I'd suggest that the type parameter declaration is removed from the declaration of `C` and the parameter `r1` gets some other type (`String` is fine, because we don't have any other positional parameters whose type is a plain constant type).

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_named_params_A05_t01.dart
```diff
@@ -0,0 +1,79 @@
...
+/// @author sgrekhov22@gmail.com
+
+import '../../../../Utils/expect.dart';
+import '../../../../Utils/static_type_helper.dart';
+
+class C<Z extends String> {
+  List<Y> m<X, Y extends num>(
+    List<Z> r1,
+    List<Y> r2,
+    List r3, {
```
eernstg, line 44: OK, this is not really testing the description, it is testing instantiation to bound. (So if instantiation to bound is changed and the `@assertion` is not then it breaks, and also if instantiation to bound is kept the same but the `@assertion` changes).

I think the easy solution is just to keep it as a plain `r3`, not `T r3` for any `T`, even if this causes the type `dynamic` to pop up.

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_named_params_A05_t01.dart
```diff
@@ -0,0 +1,79 @@
...
+    List<X>? p1,
+    List<int> p2 = const [],
+    List p3 = const [],
```
eernstg, line 47: This one should also continue to work if the type annotation is removed.

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_positional_params_A04_t01.dart
```diff
@@ -0,0 +1,71 @@
...
+/// `j ∈ 1..n+k, Tj` is a type annotation that denotes the same type as that
+/// which is denoted by the type annotation on the corresponding parameter
+/// declaration in `D` or `dynamic` if the parameter declaration has no type
+/// annotation.
+/// @author sgrekhov22@gmail.com
+
+import '../../../../Utils/expect.dart';
+import '../../../../Utils/static_type_helper.dart';
+
+class C<Z extends String> {
```
eernstg, line 39: Again, `C` should be non-generic in order to match the description.

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_positional_params_A05_t01.dart
```diff
@@ -0,0 +1,80 @@
...
+/// @author sgrekhov22@gmail.com
+
+import '../../../../Utils/expect.dart';
+import '../../../../Utils/static_type_helper.dart';
+
+class C<Z extends String> {
+  List<Y> m<X, Y extends num>(
+    List<Z> r1,
+    List<Y> r2,
+    List r3, [
```
eernstg, line 45: Same issue: This is testing instantiation to bound, not the rule in the description.

eernstg (commented):

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_named_params_A04_t01.dart
sgrekhov: Ah, now I see. I thought that `T` is a type argument but missed that `Let T be the class that contains D.`. Then, of course, `C` should be a non-generic to follow the description. Fixed. Thank you for pointing.

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_named_params_A05_t01.dart
sgrekhov: Thanx. Fixed.

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_named_params_A05_t01.dart
sgrekhov: And here

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_positional_params_A04_t01.dart
sgrekhov: Fixed

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_positional_params_A05_t01.dart
sgrekhov: Fixed here too.

sgrekhov (commented):
Thank you. Updated. PTAL

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_named_params_A04_t01.dart
eernstg: Thanks!

eernstg (approved):
LGTM!

PR #3062, opened 2025-01-31
#3054. Add generic method instantiation tests

sgrekhov:


Language/Expressions/Property_Extraction/Generic_Method_Instantiation/generic_method_A01_t01.dart
```diff
@@ -0,0 +1,49 @@
...
+  String Function(int) f1 = c1.foo;
+//                          ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  int Function(num) f2 = c1.foo;
+//                       ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+  C2 c2 = C2();
```
eernstg, line 40: Perhaps
```suggestion
  var c2 = C2<int>();
```
such that the actual type argument is provided explicitly, the test doesn't depend on instantiation to bound, and the type argument isn't the bound.

eernstg (commented):

Language/Expressions/Property_Extraction/Generic_Method_Instantiation/generic_method_A03_t02.dart
```diff
@@ -0,0 +1,45 @@
...
+  X foo<X extends T>(X x) => x + v as X;
+}
+
+void main() {
+  var c1 = C1(1);
+  int Function(int) f1 = c1.foo;
+  int Function(int) f2 = c1.foo;
+  Expect.isTrue(f1 == f2);
+
+  var c2 = C2(2);
```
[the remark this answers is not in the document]
sgrekhov: > (We could have a test for throwing, too ;-)

`generic_method_A01_t02.dart` tests the throwing.

sgrekhov (commented):
Updated. PTAL.

Language/Expressions/Property_Extraction/Generic_Method_Instantiation/generic_method_A01_t02.dart
```diff
@@ -20,22 +20,13 @@
...
-  C1 c1 = C1();
-  Expect.throws(() {
-    String Function(int) f1 = c1.foo as dynamic;
-  });
-
-  C2 c2 = C2();
+  var c = C<num>();
   Expect.throws(() {
-    int Function(num) f2 = c2.foo as dynamic;
+    int Function(num) f = c.foo as dynamic;
```
eernstg, line 30: These tests are about generic function instantiation, so there should be a part of the code where this operation occurs. However, I can't see any such location in this test. In particular, `c.foo as dynamic` gives `c.foo` the context type `_`, which means that there will not be a generic function instantiation.

If we want to test the case where generic function instantiation occurs because of the failing caller-side check that I mentioned this morning then we'd need to satisfy two requirements:

- `c` must have a covariant type (as in `C<num> c = C<int>();`.
- An attempt is made to access `c.foo`.

This should be enough to make the program throw.

However, the thing that we should test is that this dynamic type check (that fails) is actually performed also in the case where the context calls for a generic function instantiation (which should be flawless according to the static types, such that we don't get any errors because of the function instantiation itself, it must throw simply because we're accessing `c.foo`).

Something like this:

```suggestion
    num Function(num) f = c.foo;
```

eernstg (commented):
Looks good! I did have an extra comment on the throwing test that we briefly talked about this morning.

eernstg (approved):
LGTM!

Language/Expressions/Property_Extraction/Generic_Method_Instantiation/generic_method_A01_t02.dart
eernstg: Thanks!

PR #3068, opened 2025-02-06
Fixes #3067. Rename the static access shorthand feature experiment flag

### LINKED ISSUE #3067: [Dot Shorthands] Rename the static access shorthand feature experiment flag
@sgrekhov 
https://dart-review.googlesource.com/c/sdk/+/407603 fails currently because it's hanging from all the co19 tests using `enum-shorthands` which is an unknown experiment flag after we rename it to `dot-shorthands`. 

The solution is to make sure the tests use a known experiment. So I'm creating an issue to ask if we could rename all flags in the `co19/LanguageFeatures/Static-access-shorthand` tests to `dot-shorthands` and I'll update the DEPS in [that PR](https://dart-review.googlesource.com/c/sdk/+/407603) so we can sync the new flag with the tests using it. 

Thanks!

sgrekhov:


eernstg (approved):
LGTM

PR #3069, opened 2025-02-06
#3054. Add more test cases to closurization tests. Part 2.

sgrekhov:


Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_named_params_A06_t02.dart
```diff
@@ -0,0 +1,55 @@
...
+
+extension type ET(int _) {
+  num m(int r1, {p1 = ""}) => r1;
+}
+
+main() {
+  A a = A();
+  final fa1 = a.m;
+  final fa2 = a.m;
+  Expect.notEquals(fa1, fa2);
```
eernstg, line 49: Note that the spec wording says that the two tear-offs are equal if and only if they are identical. But it doesn't say that they can't be identical.

In this particular case we actually need to preserve the permission to optimize the extension methods: If an extension method doesn't actually depend on the receiver (as in this example: it just returns an argument and never uses `this`) then it must be OK for an implementation to compile this extension method to, for example, a simple top-level function like `num compiled_m(int r1, {String p1 = ""}) => r1;` and then return that (simple and constant) function object when the corresponding extension method is torn off.

In this case we'll have `identical(fa1, fa2)` (and hence also `fa1 == fa2`).

So we have to weaken the test to say that it's OK for `fa1` and `fa2` to be unequal, and it's also OK for them to be equal (and even identical, if anyone asks ;-).

As usual, evaluation of `fa1 == fa2` shouldn't throw, so we shouldn't just skip the test entirely.

So perhaps this would do?:

```suggestion
  Expect.isTrue(fa1 == fa2 || fa1 != fa2);
```

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_named_params_A06_t02.dart
```diff
@@ -0,0 +1,55 @@
...
+
+  ET et = ET(0);
+  final fet1 = et.m;
+  final fet2 = et.m;
+  Expect.notEquals(fet1, fet2);
```
eernstg, line 54: Similar.

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_positional_params_A03_t03.dart
```diff
@@ -37,13 +37,41 @@ class C<X extends num> {
...
-  var o = C<num>();
-  final f = o.m;
-  f.expectStaticType<
+  var c = C<num>();
+  final fc = c.m;
+  fc.expectStaticType<
         Exactly<num Function<Y extends num>(num r1, Y r2, [int? p1])>
       >();
+  Expect.isTrue(
+    fc is num Function<Y extends num>(Object? r1, Object? r2, [Object? p1]));
```
eernstg, line 61: Why wouldn't `r2` have type `Y`? Note that the relaxation of a parameter type to `Object?` which is based on covariance doesn't apply to `Y` because it is a type parameter of the method, not a type parameter of an enclosing class/mixin/etc.

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_positional_params_A03_t03.dart
```diff
@@ -37,13 +37,41 @@ class C<X extends num> {
...
+
+  M<num> m = MO<num>();
+  final fm = m.m;
+  fm.expectStaticType<
+      Exactly<num Function<Y extends num>(num r1, Y r2, [int? p1])>
+  >();
+  Expect.isTrue(
+      fm is num Function<Y extends num>(Object? r1, Object? r2, [Object? p1]));
```
eernstg, line 69: Ditto.

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_positional_params_A03_t03.dart
```diff
@@ -37,13 +37,41 @@ class C<X extends num> {
...
 
+  final fe = E.e0.m;
+  fe.expectStaticType<
+      Exactly<num Function<Y extends num>(num r1, Y r2, [int? p1])>
+  >();
   Expect.isTrue(
-    f is num Function<Y extends num>(Object? r1, Object? r2, [Object? p1]));
+      fe is num Function<Y extends num>(Object? r1, Object? r2, [Object? p1]));
```
eernstg, line 76: Ditto again.

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_positional_params_A06_t02.dart
```diff
@@ -0,0 +1,56 @@
...
+
+extension type ET(int _) {
+  num m(int r1, [p1 = ""]) => r1;
+}
+
+main() {
+  A a = A();
+  final fa1 = a.m;
+  final fa2 = a.m;
+  Expect.notEquals(fa1, fa2);
```
eernstg, line 50: Again, should allow this, but it should also allow optimization that causes `fa1 == fa2`.

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_positional_params_A06_t02.dart
```diff
@@ -0,0 +1,56 @@
...
+
+  ET et = ET(0);
+  final fet1 = et.m;
+  final fet2 = et.m;
+  Expect.notEquals(fet1, fet2);
```
eernstg, line 55: Ditto.

eernstg (commented):
Looks good!

We only have a couple of cases where I'm worried: We shouldn't use tests to prevent optimizations that are reasonable, and that we could otherwise have (now or later).

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_named_params_A06_t02.dart
sgrekhov: Thank you. Makes sense. I simply deleted this test. Expect.isTrue(fa1 == fa2 || fa1 != fa2); will hardly ever fail and will always be true. Therefore, this test does nothing except confirm that extension and extension type methods can be torn off. We have lots of other tests (including ones in this PR) that check this.

sgrekhov (commented):
Updated. PTAL.

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_positional_params_A03_t03.dart
eernstg: Thanks!

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_named_params_A06_t02.dart
eernstg: Do we have a test that verifies that the operator `==` on these function objects isn't buggy and throws?

eernstg (commented):
Just one little thing to think about...

Language/Expressions/Property_Extraction/Instance_Method_Closurization/method_closurization_named_params_A06_t02.dart
sgrekhov: Please clarify, how `==` may throw on such function objects?

eernstg: That would be a bug, but bugs is the main reason why we have tests.

eernstg (approved):
LGTM, thanks a lot!

Issue #3057, opened 2025-01-27
Add more flow analysis tests


sgrekhov:
In particular, we have no tests for promotion via type casts. Add them and any other missing ones, if applicable.

sgrekhov:
Closing as completed

PR #3071, opened 2025-02-10
#3057. Add more flow analysis tests for type Never. Part 2.

sgrekhov:


TypeSystem/flow-analysis/reachability_A01_t18.dart
```diff
@@ -0,0 +1,34 @@
...
+/// @assertion
+/// - Variable or getter: If `N` is an expression of the form `x` where the
+///   type of `x` is `T` then:
+/// If `T <: Never` then:
+///   - Let `after(N) = unreachable(before(N))`.
+/// Otherwise:
+///   - Let `after(N) = before(N)`.
+///
+/// @description Checks that the code is unreachable after a variable of type
+/// `Never`. Test a map literal.
```
eernstg, line 14:
```suggestion
/// `Never`. Test a set literal.
```

TypeSystem/flow-analysis/reachability_A01_t20.dart
```diff
@@ -0,0 +1,34 @@
...
+/// @assertion
+/// - Variable or getter: If `N` is an expression of the form `x` where the
+///   type of `x` is `T` then:
+/// If `T <: Never` then:
+///   - Let `after(N) = unreachable(before(N))`.
+/// Otherwise:
+///   - Let `after(N) = before(N)`.
+///
+/// @description Checks that the code is unreachable after a variable of type
+/// `Never`. Test an `if` statement in a list literal.
```
eernstg, line 14:
```suggestion
/// `Never`. Test an `if` element in a list literal.
```

TypeSystem/flow-analysis/reachability_A01_t20.dart
```diff
@@ -0,0 +1,34 @@
...
+/// @author sgrekhov22@gmail.com
+
+void test(Never n) {
+  late int i;
+  bool b = (() => true)();
+  if (b) {
+    [
+      if (n is String)        // The code after this point is unreachable
```
eernstg, line 22: Ignore unnecessary cast lint?

TypeSystem/flow-analysis/reachability_A01_t21.dart
```diff
@@ -0,0 +1,34 @@
...
+/// @assertion
+/// - Variable or getter: If `N` is an expression of the form `x` where the
+///   type of `x` is `T` then:
+/// If `T <: Never` then:
+///   - Let `after(N) = unreachable(before(N))`.
+/// Otherwise:
+///   - Let `after(N) = before(N)`.
+///
+/// @description Checks that the code is unreachable after a variable of type
+/// `Never`. Test an `if` statement in a set literal.
```
eernstg, line 14:
```suggestion
/// `Never`. Test an `if` element in a set literal.
```

TypeSystem/flow-analysis/reachability_A01_t22.dart
```diff
@@ -0,0 +1,34 @@
...
+/// @assertion
+/// - Variable or getter: If `N` is an expression of the form `x` where the
+///   type of `x` is `T` then:
+/// If `T <: Never` then:
+///   - Let `after(N) = unreachable(before(N))`.
+/// Otherwise:
+///   - Let `after(N) = before(N)`.
+///
+/// @description Checks that the code is unreachable after a variable of type
+/// `Never`. Test an `if` statement in a map literal.
```
eernstg, line 14:
```suggestion
/// `Never`. Test an `if` element in a map literal.
```

TypeSystem/flow-analysis/reachability_A01_t23.dart
```diff
@@ -0,0 +1,83 @@
...
+/// @assertion
+/// - Variable or getter: If `N` is an expression of the form `x` where the
+///   type of `x` is `T` then:
+/// If `T <: Never` then:
+///   - Let `after(N) = unreachable(before(N))`.
+/// Otherwise:
+///   - Let `after(N) = before(N)`.
+///
+/// @description Checks that the code is unreachable after a variable of type
+/// `Never`. Test a `for(;;)` statement in a list literal.
```
eernstg, line 14:
```suggestion
/// `Never`. Test a `for(;;)` element in a list literal.
```

TypeSystem/flow-analysis/reachability_A01_t23.dart
```diff
@@ -0,0 +1,83 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+void test3(Never n) {
+  late int i;
+  bool b = (() => true)();
+  if (b) {
+    [
+      for (;; n)  // The code after this for statement is unreachable
```
eernstg, line 55: Interesting&mdash;I thought the flow analysis wouldn't catch this! It needs to recognize that the condition is implicitly turned into `true`; then it needs to notice that the expression list which is the third element is _guaranteed_ to be executed, and then it will conclude that this for statement will definitely not execute the body.

I hadn't expected the flow analysis to have a notion of "will _definitely_ evaluate to `true`". Anyway, it could be a special treatment of the case where the condition is empty.

However, if this test fails then it seems reasonable to double check the spec and, most likely, adjust the expectation.

TypeSystem/flow-analysis/reachability_A01_t23.dart
```diff
@@ -0,0 +1,83 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+void test4(Never n) {
+  late int i;
+  bool b = (() => true)();
+  if (b) {
+    [
+      for (var j = 0; j < 0; n)  // Ok, not a dead code
```
eernstg, line 71: This makes sense: The flow analysis won't try to reason about the value of `j`, it is just going to say "who knows?", and conclude that `i` may or may not have been initialized in line 75.

TypeSystem/flow-analysis/reachability_A01_t24.dart
```diff
@@ -0,0 +1,83 @@
...
+/// If `T <: Never` then:
+///   - Let `after(N) = unreachable(before(N))`.
+/// Otherwise:
+///   - Let `after(N) = before(N)`.
+///
+/// @description Checks that the code is unreachable after a variable of type
+/// `Never`. Test a `for(;;)` statement in a set literal.
+/// @author sgrekhov22@gmail.com
+/// @issue 60089
+
```
eernstg, line 17: Similar to the previous test.

TypeSystem/flow-analysis/reachability_A01_t24.dart
```diff
@@ -0,0 +1,83 @@
...
+/// @assertion
+/// - Variable or getter: If `N` is an expression of the form `x` where the
+///   type of `x` is `T` then:
+/// If `T <: Never` then:
+///   - Let `after(N) = unreachable(before(N))`.
+/// Otherwise:
+///   - Let `after(N) = before(N)`.
+///
+/// @description Checks that the code is unreachable after a variable of type
+/// `Never`. Test a `for(;;)` statement in a set literal.
```
eernstg, line 14:
```suggestion
/// `Never`. Test a `for(;;)` element in a set literal.
```

TypeSystem/flow-analysis/reachability_A01_t25.dart
```diff
@@ -0,0 +1,83 @@
...
+/// @assertion
+/// - Variable or getter: If `N` is an expression of the form `x` where the
+///   type of `x` is `T` then:
+/// If `T <: Never` then:
+///   - Let `after(N) = unreachable(before(N))`.
+/// Otherwise:
+///   - Let `after(N) = before(N)`.
+///
+/// @description Checks that the code is unreachable after a variable of type
+/// `Never`. Test a `for(;;)` statement in a map literal.
```
eernstg, line 14:
```suggestion
/// `Never`. Test a `for(;;)` element in a map literal.
```

TypeSystem/flow-analysis/reachability_A01_t25.dart
```diff
@@ -0,0 +1,83 @@
...
+/// @author sgrekhov22@gmail.com
+/// @issue 60089
+
```
eernstg, line 17: Again similar.

eernstg (commented):
Looks good! Commented on a few typos.

I also commented on a couple of cases where I'm surprised that the flow analysis would be able to reach a particular conclusion. If it does work, and the spec actually says it should, then that's fine. If it doesn't work, and the spec agrees, then the expectation should be adjusted.

Otherwise we may need to revisit the spec.

TypeSystem/flow-analysis/reachability_A01_t18.dart
sgrekhov: Fixed

TypeSystem/flow-analysis/reachability_A01_t20.dart
sgrekhov: Thanx, fixed here and below.

TypeSystem/flow-analysis/reachability_A01_t20.dart
sgrekhov: Added. Now the warning is ignored.

TypeSystem/flow-analysis/reachability_A01_t23.dart
sgrekhov: The test doesn't fail, it works in both the analyzer and the CFE

sgrekhov (commented):
Updated. PTAL

TypeSystem/flow-analysis/reachability_A01_t23.dart
eernstg: OK, I can see that the `true` literal does indeed get this kind of treatment. Thanks!

eernstg (approved):
LGTM

PR #3072, opened 2025-02-11
#3057. Add more reachability tests for `true` and `false` literals

sgrekhov:
Also added for-in tests for the type `Never`.

### DIFF
--- a/TypeSystem/flow-analysis/reachability_A02_t05.dart
+++ b/TypeSystem/flow-analysis/reachability_A02_t05.dart
@@ -1,0 +1,43 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion True literal: If `N` is the literal `true`, then:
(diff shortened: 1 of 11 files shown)


TypeSystem/flow-analysis/reachability_A02_t05.dart
```diff
@@ -0,0 +1,43 @@
...
+
+/// @assertion True literal: If `N` is the literal `true`, then:
+///  Let `true(N) = before(N)`.
+///  Let `false(N) = unreachable(before(N))`.
+///
+/// @description Checks reachability after `true` literal in collection literals
+/// @author sgrekhov22@gmail.com
+
+void test1() {
+  late int i;
```
eernstg, line 13: I think this one could be non-late. Same situation below.

eernstg (commented):
Looks good! Just one test could be stronger if an `int` were non-late.

TypeSystem/flow-analysis/reachability_A02_t05.dart
sgrekhov: Of course. Thank you for noticing! Fixed.

eernstg (approved):
LGTM

PR #3076, opened 2025-02-13
#3057. Add flow analysis tests for throw expression

sgrekhov:


TypeSystem/flow-analysis/reachability_A10_t01.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion throw: If `N` is a throw expression of the form `throw E1`, then:
+/// - Let `before(E1) = before(N)`
+/// - Let `after(N) = unreachable(after(E1))`
+///
+/// @description Checks that a code after throw expression is unreachable.
```
eernstg, line 9: I think 'code' is a mass noun (that is, we can't have "one code, two codes" when it refers to source code, that would be more likely if the "codes" were the individual ways to encrypt a secret message; similar rules apply for 'milk', 'sand' and similar words denoting something that we can't count). In that case the most natural wording would be to omit an indefinite article ("the milk" is fine, but "a milk" doesn't occur). On the other hand, 'one expression, two expressions' works just fine, so we should include the indefinite article there.
```suggestion
/// @description Checks that code after a throw expression is unreachable.
```

eernstg: Similarly for a few more tests below.

TypeSystem/flow-analysis/reachability_A10_t03.dart
```diff
@@ -0,0 +1,82 @@
...
+/// @description Checks that a code after throw expression is unreachable. Test
+/// collection literals.
+/// @author sgrekhov22@gmail.com
+
+test1() {
+  late int i;
+  if ((() => false)()) {
+    [
+      0,
+      if (throw "Dead code after this point")
```
eernstg, line 18: I think this is kind of a branching point in the design of which cases to cover: We already have a test where a list literal contains an `if` element that contains a condition whose static type is `Never` (that was `[.... if (n) e ....]`), and here's another one.

I think it's going to explode if we are consistently covering all kinds of `Never`-typed expressions in a setting that already has structure ('list literal with `if` element'), so it's probably better to say that "we already have this case" in the sense that we already have another test that covers the two top levels of the structure.

So, in order to avoid drowning in tests, it's probably better to drop A10_t03 .. A10_t05.

t06 and t07 are probably new because they are testing a for-in element.

eernstg (commented):
Looks good, couple of typos.

A slightly deeper thing is that some tests look like they are covering cases that are a bit too deep (if we start doing that consistently then the number of tests will explode).

eernstg (approved):
LGTM!

PR #3078, opened 2025-02-14
#3057. Add instance check tests

sgrekhov:
We have `promotion_via_type_test_A*.dart` test that check promotion. So, only missing cases added here.

### DIFF
--- a/TypeSystem/flow-analysis/reachability_A12_t01.dart
+++ b/TypeSystem/flow-analysis/reachability_A12_t01.dart
@@ -14,13 +14,13 @@
 /// the case `T` is `R?` and `Null <: S`.
 /// @author sgrekhov22@gmail.com
 
+import '../../Utils/static_type_helper.dart';
+
 main() {
   int? i = 2 > 1 ? null : 42;
   if (i is Null) {
-    Null n = i; // Ok
-    int j = i;
-//          ^
-// [analyzer] unspecified
-// [cfe] unspecified
+    Null n = i;
+  } else {
+    i.expectStaticType<Exactly<int>>();
   }
 }
--- a/TypeSystem/flow-analysis/reachability_A12_t01.dart
+++ b/TypeSystem/flow-analysis/reachability_A12_t01.dart
@@ -1,0 +1,26 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion instance check If `N` is an expression of the form `E1 is S`
+/// where the static type of `E1` is `T` then:
+/// - Let `before(E1) = before(N)`
+/// - Let `true(N) = promote(E1, S, after(E1))`
+/// - Let `false(N) = promote(E1, factor(T, S), after(E1))`
+///
+/// @description Checks that for expression of the form `E1 is S`
+/// `true(N) = promote(E1, S, after(E1))` and
+/// `false(N) = promote(E1, factor(T, S), after(E1))`. Test `factor(T, S)` for
+/// the case `T` is `R?` and `Null <: S`.
+/// @author sgrekhov22@gmail.com
+
+main() {
+  int? i = 2 > 1 ? null : 42;
+  if (i is Null) {
+    Null n = i; // Ok
+    int j = i;
+//          ^
+// [analyzer] unspecified
+// [cfe] unspecified
+  }
+}
--- a/TypeSystem/flow-analysis/reachability_A12_t02.dart
+++ b/TypeSystem/flow-analysis/reachability_A12_t02.dart
@@ -11,16 +11,16 @@
 /// @description Checks that for expression of the form `E1 is S`
 /// `true(N) = promote(E1, S, after(E1))` and
 /// `false(N) = promote(E1, factor(T, S), after(E1))`. Test `factor(T, S)` for
-/// the case `T` is `R?` and `Null <: S`.
+/// the case `T` is `R?` and `S` is not subtype of `Null`.
 /// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
 
 main() {
   int? i = 2 > 1 ? null : 42;
   if (i is int) {
     int j = i;
-    Null n = i;
-//           ^
-// [analyzer] unspecified
-// [cfe] unspecified
+  } else {
+    i.expectStaticType<Exactly<Null>>();
   }
 }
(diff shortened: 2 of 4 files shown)


TypeSystem/flow-analysis/reachability_A12_t01.dart
```diff
@@ -0,0 +1,26 @@
...
+/// `true(N) = promote(E1, S, after(E1))` and
+/// `false(N) = promote(E1, factor(T, S), after(E1))`. Test `factor(T, S)` for
+/// the case `T` is `R?` and `Null <: S`.
+/// @author sgrekhov22@gmail.com
+
+main() {
+  int? i = 2 > 1 ? null : 42;
+  if (i is Null) {
+    Null n = i; // Ok
+    int j = i;
```
eernstg, line 21: I think this one isn't placed correctly. The `factor` rule applies to the false continuation, that is, the `else` branch. In that branch, the type of `i` should be `int`, because null was `factor`ed out.

So perhaps the best test of the treatment would be `i.expectStaticType<Exactly<int>>()` rather than `int j = i;`?

TypeSystem/flow-analysis/reachability_A12_t02.dart
```diff
@@ -0,0 +1,26 @@
...
+/// @assertion instance check If `N` is an expression of the form `E1 is S`
+/// where the static type of `E1` is `T` then:
+/// - Let `before(E1) = before(N)`
+/// - Let `true(N) = promote(E1, S, after(E1))`
+/// - Let `false(N) = promote(E1, factor(T, S), after(E1))`
+///
+/// @description Checks that for expression of the form `E1 is S`
+/// `true(N) = promote(E1, S, after(E1))` and
+/// `false(N) = promote(E1, factor(T, S), after(E1))`. Test `factor(T, S)` for
+/// the case `T` is `R?` and `Null <: S`.
```
eernstg, line 14: But that's not true in this case: `S` is `int`. Also, the test should still be in the false continuation.

TypeSystem/flow-analysis/reachability_A12_t03.dart
```diff
@@ -0,0 +1,34 @@
...
+
+import 'dart:async';
+
+class C {
+  void foo() {}
+}
+
+test(FutureOr<C> foc) async {
+  if (foc is Future<C>) {
+    foc.foo();
```
eernstg, line 25: Should be in the false continuation again, and then it should not be an error.

TypeSystem/flow-analysis/reachability_A12_t04.dart
```diff
@@ -0,0 +1,32 @@
...
+
+import 'dart:async';
+
+class C {
+  void foo() {}
+}
+
+test(FutureOr<C> foc) async {
+  if (foc is C) {
+    foc.foo(); // Ok
```
eernstg, line 25: This should again be in the false continuation, and then it should be an error (because the "rest type" that remains when `C` has been "removed" is `Future<C>`).

eernstg (commented):
This one needs to be adjusted.

TypeSystem/flow-analysis/reachability_A12_t01.dart
sgrekhov: Thanx! Fixed.

TypeSystem/flow-analysis/reachability_A12_t02.dart
sgrekhov: Fixed

TypeSystem/flow-analysis/reachability_A12_t03.dart
sgrekhov: Fixed

TypeSystem/flow-analysis/reachability_A12_t04.dart
sgrekhov: Added `expectStaticType` in the `else` branch.

sgrekhov (commented):
Updated. PTAL.

TypeSystem/flow-analysis/reachability_A12_t02.dart
```diff
@@ -11,16 +11,16 @@
...
-/// the case `T` is `R?` and `Null <: S`.
+/// the case `T` is `R?` and `S` is not subtype of `Null`.
 /// @author sgrekhov22@gmail.com
 
+import '../../Utils/static_type_helper.dart';
+
 main() {
   int? i = 2 > 1 ? null : 42;
   if (i is int) {
     int j = i;
```
eernstg, line 22: I'm afraid I was a little bit careless here. The current approach will succeed even if the static type of `i` after the test `i is int` happens to be `dynamic` or `Never`. So `i.expectStaticType<Exactly<int>>()` will be a stronger test than `int j = i;`.

TypeSystem/flow-analysis/reachability_A12_t03.dart
```diff
@@ -22,13 +22,12 @@ class C {
 
 test(FutureOr<C> foc) async {
   if (foc is Future<C>) {
```
eernstg, line 24: I think we're checking the entire rule about `E1 is S`, in the particular case where `factor(T, S)` gets a certain kind of input. This means that it makes sense to verify the treatment of the branch that relies on `factor` as well as the other branch.

As before, the safest test would probably be `expectStaticType` expecting `Future<C>` in the "then" branch.

eernstg: Alternatively, maybe these tests are exclusively about `factor`, and we just want the bare minimum that reveals the behavior of `factor`? In that case a couple of other tests could have one empty branch, and some non-trivial code in the branch that uses `factor`.

TypeSystem/flow-analysis/reachability_A12_t04.dart
```diff
@@ -15,14 +15,16 @@
...
 
 import 'dart:async';
+import '../../Utils/static_type_helper.dart';
 
 class C {
   void foo() {}
 }
 
 test(FutureOr<C> foc) async {
   if (foc is C) {
```
eernstg, line 25: Similar.

eernstg (commented):
Looks good!

One thing looks slightly inconsistent: We may be testing a complete rule (for example, about `E1 is S`), in which case a couple of tests should populate both branches. Alternatively, we're strictly aiming at `factor`, in which case the non-`factor` branch should be left empty.

TypeSystem/flow-analysis/reachability_A12_t03.dart
eernstg: Thanks!

TypeSystem/flow-analysis/reachability_A12_t04.dart
eernstg: Thanks!

TypeSystem/flow-analysis/reachability_A12_t02.dart
eernstg: Thanks!

eernstg (approved):
LGTM

PR #3082, opened 2025-02-19
#3057. Add tests for logical and, or and null-check operators

sgrekhov:


TypeSystem/flow-analysis/reachability_A18_t02.dart
```diff
@@ -0,0 +1,20 @@
...
+/// - Let `false(N) = merge(split(false(E1)), false(E2))`.
+///
+/// @description Checks that in an expression of the form `E1 && E2` if `E1` is
+/// `true` then `E2` is always executed.
+/// @author sgrekhov22@gmail.com
+
+main() {
+  int i;
+  true && (i = 42) != 42;
+  i; // Definitely assigned
```
eernstg, line 19: I can see (in DartPad) that the flow analysis does conclude that `i = 42` is definitely being executed. It is not so obvious how this would be a necessary consequence of the rules in the `@assertion`.

But `false(E1)` is unreachable, so `split(false(E1))` is unreachable, so the `merge(...)` would be the same as `false(E2)`. This implies that `E2` has been evaluated if we're in the false continuation of the expression `E1 && E2`.

Similarly, `before(E2)` is `before(N)` (because `true(E1)` is `before(E1)`), so if `E1 && E2` is reached then `E2` is also reached.

OK!

TypeSystem/flow-analysis/reachability_A19_t03.dart
```diff
@@ -0,0 +1,23 @@
...
+/// - Let `true(N) = merge(split(true(E1)), true(E2))`.
+///
+/// @description Checks that in an expression of the form `E1 || E2` if `E1` is
+/// `true` then `E2` is never executed.
+/// @author sgrekhov22@gmail.com
+
+main() {
+  late int i;
+  true || (i = 42) == 42;
+  i; // Definitely unassigned
```
eernstg, line 19: Right! Tricky, too, but `before(E2)` is unreachable because `false(E1)` is unreachable.

eernstg (approved):
LGTM

PR #3083, opened 2025-02-19
#3057. Add tests for method invocation

sgrekhov:
Please review. In the [spec](https://github.com/dart-lang/language/blob/main/resources/type-system/flow-analysis.md#expressions) we also have:

> Binary operator: All binary operators other than `==`, `&&`, `||`, and `??` are handled as calls to the appropriate operator method.

So, if these tests are Ok, then I'm going to add the similar tests for `E1 + E2` and all other operators with two operands.

### DIFF
--- a/TypeSystem/flow-analysis/reachability_A22_t01.dart
+++ b/TypeSystem/flow-analysis/reachability_A22_t01.dart
@@ -1,0 +1,33 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion - Method invocation: If `N` is an expression of the form
+/// `E1.m1(E2)`, then:
+/// - Let `before(E1) = before(N)`
+/// - Let `before(E2) = after(E1)`
+/// - Let `T` be the static return type of the invocation
+/// - If `T <: Never` then:
+///   - Let `after(N) = unreachable(after(E2))`.
+/// - Otherwise:
+///   - Let `after(N) = after(E2)`.
+///
+/// @description Checks that for an expression of the form `E1.m1(E2)`
+/// `before(E2) = after(E1)`.
+/// @author sgrekhov22@gmail.com
+
+void test<T extends Never>(T n) {
+  late int i;
+  if (2 > 1) {
+    n.foo(i = 42);
+  }
+  i; // Definitely unassigned
+//^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+
+main() {
+  print(test);
+}
--- a/TypeSystem/flow-analysis/reachability_A22_t01.dart
+++ b/TypeSystem/flow-analysis/reachability_A22_t01.dart
@@ -13,7 +13,8 @@
 ///   - Let `after(N) = after(E2)`.
 ///
 /// @description Checks that for an expression of the form `E1.m1(E2)`
-/// `before(E2) = after(E1)`.
+/// `before(E2) = after(E1)`. Test that if `after(E1)` is unreachable then
+/// `before(E2)` is also unreachable.
 /// @author sgrekhov22@gmail.com
 
 void test<T extends Never>(T n) {
--- a/TypeSystem/flow-analysis/reachability_A22_t01.dart
+++ b/TypeSystem/flow-analysis/reachability_A22_t01.dart
@@ -27,7 +27,6 @@ void test<T extends Never>(T n) {
 // [cfe] unspecified
 }
 
-
 main() {
   print(test);
 }
--- a/TypeSystem/flow-analysis/reachability_A22_t03.dart
+++ b/TypeSystem/flow-analysis/reachability_A22_t03.dart
@@ -1,0 +1,37 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion - Method invocation: If `N` is an expression of the form
+/// `E1.m1(E2)`, then:
+/// - Let `before(E1) = before(N)`
+/// - Let `before(E2) = after(E1)`
+/// - Let `T` be the static return type of the invocation
+/// - If `T <: Never` then:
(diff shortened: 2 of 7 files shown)


TypeSystem/flow-analysis/reachability_A22_t01.dart
```diff
@@ -0,0 +1,32 @@
...
+/// - Let `before(E1) = before(N)`
+/// - Let `before(E2) = after(E1)`
+/// - Let `T` be the static return type of the invocation
+/// - If `T <: Never` then:
+///   - Let `after(N) = unreachable(after(E2))`.
+/// - Otherwise:
+///   - Let `after(N) = after(E2)`.
+///
+/// @description Checks that for an expression of the form `E1.m1(E2)`
+/// `before(E2) = after(E1)`.
```
eernstg, line 16: At this location in the feature spec we have a couple of 'TODO' comments. For example, there's no mention of the case where `m1` is a getter whose return type is a function type or a interface type that includes a method named `call` ... or `Never`, which would allow the invocation `E1.m1(E2)` as well.

I think it would make sense to have those cases as well. When the feature spec gets expanded to cover those cases we'll just need to revisit these tests and check that they aren't expecting anything that contradicts the newly added spec cases.

For this PR, I'll just assume that we're calling a method.

eernstg: Regarding "checks that .. `before(E2) = after(E1)`", that's a big ask. Perhaps just say that we're checking one consequence of that property (namely that if `after(E1)` is unreachable then `before(E2)` is also unreachable)?

TypeSystem/flow-analysis/reachability_A22_t03.dart
```diff
@@ -0,0 +1,63 @@
...
+  i; // Not definitely unassigned
+}
+
+void test3() async {
+  late int i;
+  if (2 > 1) {
+    await C().bar();
+    i = 42;
+  }
+  i; // Not definitely unassigned
```
eernstg, line 53: Looks like a typo:
```suggestion
  i; // Definitely unassigned
```

TypeSystem/flow-analysis/reachability_A22_t05.dart
```diff
@@ -0,0 +1,55 @@
...
+/// - Let `before(E1) = before(N)`
+/// - Let `before(E2) = after(E1)`
+/// - Let `T` be the static return type of the invocation
+/// - If `T <: Never` then:
+///   - Let `after(N) = unreachable(after(E2))`.
+/// - Otherwise:
+///   - Let `after(N) = after(E2)`.
+///
+/// @description Checks that if the static type of the expression of the form
+/// `E1.m1(E2)` static type of `E1`is not `Never` then `after(N) = after(E2)`.
```
eernstg, line 16: Again, that's a big claim. There could be all kinds of detailed differences between `after(N)` and `after(E2)` that we'd have to write many tests in order to cover. (For example, some innocent bystander of a local variable could suddenly have been promoted, because of a bug, so `after(N) != after(E2)` with respect to that particular variable; and perhaps this only happens when the name of the variable is `xyzzy31` ;-). Perhaps:
```suggestion
/// `E1.m1(E2)` static type of `E1`is not `Never` then `after(N) = after(E2)`,
/// which is tested by detecting that `i = 42` is considered to be guaranteed
/// to have been executed when `i;` is executed.
```
It is probably possible to cut that down to fewer words and still inform the reader about the situation, but the point is just that we can't claim to have shown that `after(N) == after(E2)`, and we don't even want to get anywhere near to being able to claim that.

TypeSystem/flow-analysis/reachability_A22_t01.dart
eernstg: I had a couple of other comments, so it might make sense to add some extra tests to this PR&mdash;whatever is more convenient.

eernstg (commented):

sgrekhov:
Updated. Added cases for getter returning callable object and type `Never`. PTAL.

TypeSystem/flow-analysis/reachability_A22_t05.dart
```diff
@@ -13,7 +13,9 @@
 ///   - Let `after(N) = after(E2)`.
 ///
 /// @description Checks that if the static type of the expression of the form
-/// `E1.m1(E2)` static type of `E1`is not `Never` then `after(N) = after(E2)`.
+/// `E1.m1(E2)` static type of `E1`is not `Never` then `after(N) = after(E2)`,
+/// which is tested by detecting that `i = 42` is considered to be guaranteed
+/// to have been executed when `i;` is executed.
```
eernstg, line 18: This one needs an adjustment. Perhaps it is intended to say something like the following?:
```suggestion
/// @description Checks that for an expression of the form `E1.m1(E2)`, if the
/// static type of `E1`is not `Never` then `after(N) = after(E2)`. This is
/// tested by detecting that `i = 42` is considered to be guaranteed to have
/// been executed when `i;` is executed.
```

TypeSystem/flow-analysis/reachability_A22_t06.dart
```diff
@@ -0,0 +1,58 @@
...
+/// `E1.m1(E2)`, then:
+/// - Let `before(E1) = before(N)`
+/// - Let `before(E2) = after(E1)`
+/// - Let `T` be the static return type of the invocation
+/// - If `T <: Never` then:
+///   - Let `after(N) = unreachable(after(E2))`.
+/// - Otherwise:
+///   - Let `after(N) = after(E2)`.
+///
+/// @description Checks that if the static type of the expression of the form
```
eernstg, line 15: This one would need a similar adjustment.

TypeSystem/flow-analysis/reachability_A22_t07.dart
```diff
@@ -0,0 +1,50 @@
...
+  i; // Definitely unassigned
+//^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+main() {
+  print(test1);
+  print(test2);
+}
```
eernstg, line 50: Nice! ;-)

eernstg (commented):
Looks good, but I think a couple of descriptions got slightly garbled at some point.

eernstg (approved):
LGTM

PR #3087, opened 2025-03-03
Fixes #3086. Update context type for expected value

### LINKED ISSUE #3086: Static-access-shorthand/non_ambiguity_A01_t02 - Unexpected error in test
I ran into this while testing the dot shorthands CFE implementation for static gets.

It seems that the `non_ambiguity_A01_t02` has the error: `Error: A value of type 'Object' can't be assigned to a variable of type 'C'.` which I believe is due to `e2`.

I tried testing the same code on without a shorthand and I get the same error.

```dart
// No shorthand, still errors.
class C {
  int value;
  C(this.value);
  static C id = C(0);
}

extension on bool {
  int get id => 42;
}

main() {
  bool e1 = 2 > 1; // true
  var e2 = "value";
  Object o = <C>{C(1)};
  if (o is Set<C>) {
    o = {e1? C.id: e2};  // Error: A value of type 'Object' can't be assigned to a variable of type 'C'.
  }
}
```

Could we fix the test so that we don't have that error? Thanks!

cc. @sgrekhov @eernstg

sgrekhov:
I believe that with this change the context type of `e1? .id: e2` expression should be `C`.

### DIFF
--- a/LanguageFeatures/Static-access-shorthand/non_ambiguity_A01_t02.dart
+++ b/LanguageFeatures/Static-access-shorthand/non_ambiguity_A01_t02.dart
@@ -28,7 +28,7 @@ extension on bool {
 
 main() {
   bool e1 = 2 > 1; // true
-  var e2 = "value";
+  var e2 = C(2);
   Object o = <C>{C(1)};
   if (o is Set<C>) {
     o = {e1? .id: e2};


eernstg:
Ah, @kallentu, I didn't see that you are a reviewer as well, so I just landed it. If you have further comments then I'd suggest that we handle them in a separate PR.

kallentu:
@eernstg No worries. It looks good. Thank you @sgrekhov.

Issue #3088, opened 2025-03-03
Remove weak mode


johnniwinther:
After [[cfe] Remove nnbd mode](https://dart-review.googlesource.com/c/sdk/+/412881) the CFE no longer supports weak mode, so the tests `co19/LanguageFeatures/nnbd/static/weak/*` can be deleted.

PR #3092, opened 2025-03-04
#3088. Remove weak mode from LanguageFeatures/nnbd. Part 1.

sgrekhov:
Also some excessive empty lines removed.


eernstg (approved):
LGTM!

I can see that there's a bit of cleanup going on as well (moving a couple of `@issue` lines, removing some empty lines), thanks for the hint!

PR #3097, opened 2025-03-05
#3057. Add reachability tests for operator +

sgrekhov:
If these tests are ok, then I'll add the similar tests for other operators with two operands.

### DIFF
--- a/TypeSystem/flow-analysis/reachability_A20_t03.dart
+++ b/TypeSystem/flow-analysis/reachability_A20_t03.dart
@@ -1,0 +1,57 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion - Binary operator: All binary operators other than `==`, `&&`,
+/// `||`, and `??` are handled as calls to the appropriate `operator` method.
+///
+/// @description Checks that if the static type of the expression of the form
+/// `E1 + E2` is `Never` `after(N) = unreachable(after(E2))`.
+/// @author sgrekhov22@gmail.com
+
+class C<T extends Never> {
+  T operator +(int other) => throw "";
+}
+
+class C2<T extends Never> {
+  Future<T> operator +(int other) async => throw "";
+}
+
+void test1() {
+  late int i;
+  if (2 > 1) {
+    C() + 0;
+    i = 42;
+  }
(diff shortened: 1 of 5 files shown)


TypeSystem/flow-analysis/reachability_A20_t03.dart
```diff
@@ -0,0 +1,57 @@
...
+  i; // Definitely unassigned
+//^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+void test2() {
+  late int i;
+  if (2 > 1) {
+    C2() + 0;
```
eernstg, line 35: Wouldn't this make the program throw after `main` has returned? So we're relying really heavily on the fact that this program won't run (because it has errors).

Perhaps use some other type than `Future<Never>` when the point is simply that the type isn't `Never`.

TypeSystem/flow-analysis/reachability_A20_t03.dart
```diff
@@ -0,0 +1,57 @@
...
+    i = 42;
+  }
+  i; // Not definitely unassigned
+}
+
+void test3() async {
+  late int i;
+  if (2 > 1) {
+    await (C2() + 0);
```
eernstg, line 44: OK, this case is actually using the fact that it is a `Future<Never>`.

eernstg (commented):
Looks good!

One case is somewhat unstable: TypeSystem/flow-analysis/reachability_A20_t03.dart uses a construct `C2() + 0` which will evaluate to a `Future<Never>`, and discards it. This means that it will give rise to an exception after `main` has returned.

However, there is one line in the program that gives rise to a compile-time error, so we will never see the run-time error after all.

So this work, but if someone changes the test, e.g., by moving that error case to a different library, then the test will start to fail at run time because of that `throw ""` in `C2`.

It would be safer to use some other type than `Future<Never>` in this case where no part of the test of `C2() + 0` actually needs the result to be a `Future<Never>`.

TypeSystem/flow-analysis/reachability_A20_t03.dart
eernstg: I'm just (slightly) worried because it is a source of confusion: Comment out a line that has a compile-time error and run the test. In this case the test will get the static checking, and it may also be executed (if there are no other compile-time errors left). However, it will now throw, because of a `Future<Never>` which will be completed after `main` returns.

eernstg (approved):
LGTM, ignoring the worries about the `Future<Never>` after `main` for now.

PR #3099, opened 2025-03-06
#3057. Add reachability tests for operators. Part 2.

sgrekhov:
Like the previous ones but for other operators. BTW they use type parameter of the type `<T extends Never>`.


eernstg (approved):
LGTM

PR #3101, opened 2025-03-07
#3057. Update assertions according to the spec

sgrekhov:


eernstg (approved):
LGTM

This is kind of messy because these changes are being introduced into the feature specification in response to the fact (which has been so for years) that flow analysis takes weak mode execution into account, so we can't assume that an expression whose type is non-nullable will _actually_ not evaluate to null.

So we're finally going to have the correct specification of the behavior that we've had for years.

At the same time, we're going to change the spec and the static analysis such that it relies on the actual situation today, and that's going to happen very, very soon: There is no weak mode execution, so we _can_ trust non-nullable expressions to yield a non-null result.

So we'll change this one more time very soon.

PR #3104, opened 2025-03-11
#3057. Add more reachability tests for && and || operators

sgrekhov:


TypeSystem/flow-analysis/reachability_A18_t04.dart
```diff
@@ -0,0 +1,25 @@
...
+
+Never foo() => throw "Never";
+
+main() {
+  late int i;
+  if (2 > 1) {
+    false && foo();
+    i = 42;
+  }
+  i; // Not definitely unassigned
```
eernstg, line 24: This would be true also if `false && foo()` _might_ not execute `foo()`, which is too weak because we really want to conclude that `false && foo()` will _definitely_ not execute `foo()`.

Perhaps `int i; try { false && foo(); i = 42; } catch (_) {} i;` would work.

TypeSystem/flow-analysis/reachability_A19_t04.dart
```diff
@@ -0,0 +1,25 @@
...
+
+Never foo() => throw "Never";
+
+main() {
+  late int i;
+  if (2 > 1) {
+    true || foo();
+    i = 42;
+  }
+  i; // Not definitely unassigned
```
eernstg, line 24: This one is similar to TypeSystem/flow-analysis/reachability_A18_t04.dart.

eernstg (commented):
I think two of these tests can be strengthened a bit.

TypeSystem/flow-analysis/reachability_A18_t04.dart
sgrekhov: Hm... We have here.
```dart
Never foo() => throw "Never";

main() {
  int i;
  false && foo();
  i = 42;
  i; // Definitely assigned
}
```
The above works, but if flow analysis wrongly assumes that the code after `false && foo();` is dead code, then `i;` will also become dead code. Let's not write tests with important checks in a potentially dead code.

Now let's try your suggestion.
```dart
Never foo() => throw "Never";

main() {
  int i;
  try {
    false && foo();
    i = 42;
  } catch (_) {}
  i; // Error: Non-nullable variable 'i' must be assigned before it can be used.
}
```
According to the [spec](https://github.com/dart-lang/language/blob/main/resources/type-system/flow-analysis.md#statements)
> - try catch: If `N` is a try/catch statement of the form `try B alternatives` then:
>   - Let `before(B) = before(N)`
>   - For each catch block on `Ti Si` in `alternatives`:
>     - Let `before(Si) = conservativeJoin(before(N), assignedIn(B), capturedIn(B))`
>   - Let `after(N) = join(after(B), after(C0), ..., after(Ck))`

As far as I understand after try-catch we should heve `join(after(B), after(""))`. The flow analysis should detect that `i` is definitely assigned in `B`, shouldn't it? A bug? Please confirm.

eernstg: I'm thinking about the situation where `foo()` in `false && foo()` is (wrongly) considered to be non-dead such that we may or may not execute `foo()`. This means that perhaps we don't execute `foo()` in which case `i` will be initialized, that is, `i` is 'not definitely unassigned'.

The point I'm making is that we wouldn't catch this bug. But perhaps there is some other test which _would_ catch that bug? Also, perhaps this test will catch some other bug that my proposed variant wouldn't?

In any case, I'm still sticking to the point that the current test seems to let a bug slip through, and I thought I had come up with a way to catch it.

Next, the variant you mention where there is no try/catch statement relies on properties of dead code, I agree that this is a less reliable situation, let's not do that.

> Now let's try your suggestion.

It sounds like you are saying that this version gives rise to the stated error with the current implementation of the tool chain.

In other words, the flow analysis does _not_ conclude that `foo()` is unreachable, so `i = 42` will be executed, so `i` is guaranteed to be initialized when we evaluate it.

But that's correct: Any expression or statement in the body of the try part could throw (say, out of memory could be assumed to be possible at every single step of the execution, even if that isn't quite true). This implies that we wouldn't reach `i = 42`, and hence `i` isn't 'definitely assigned' after the try/catch statement.

So that's working as it should (and the spec looks correct, as far as I can see: the part about `before(Si)` can only assume `before(B)` because `B` is assumed to be able to throw at any time whatsoever).

So this is all good. But then let's try another variant:

```dart
Never foo() => throw "Never";

main() {
  late int i;
  false && (i = 42) == 42;
  i; // Error, definitely unassigned.
}
```

In this variant, if `false && e` is considered to "possibly evaluate `e`" then `i` will no longer be definitely unassigned. I think that should do it.

Returning to an earlier question, is there some bug that this test doesn't catch that the original version would catch?

eernstg (commented):
Thinking about a tricky case ...

TypeSystem/flow-analysis/reachability_A18_t04.dart
eernstg: OK!

TypeSystem/flow-analysis/reachability_A19_t04.dart
eernstg: OK!

eernstg (approved):
LGTM!

PR #3106, opened 2025-03-12
#3057. Add conditional statements tests

sgrekhov:


TypeSystem/flow-analysis/reachability_conditional_A02_t01.dart
```diff
@@ -0,0 +1,27 @@
...
+/// @assertion Conditional statement: If `N` is a conditional statement of the
+/// form `if (E) S1 else S2` then:
+/// - Let `before(E) = before(N)`.
+/// - Let `before(S1) = split(true(E))`.
+/// - Let `before(S2) = split(false(E))`.
+/// - Let `after(N) = merge(after(S1), after(S2))`.
+///
+/// @description Checks that if `E` has the form `if-case` then even if all
+/// conditions are `true` then assignment in `then` branch is not treated as
+/// definite assignment.
```
eernstg, line 14: This library and the following 2 whose description says "even if" are basically focused on the fact that if-case (and pattern matching in general) makes no attempt to conclude at compile time that the outcome will be a specific boolean value, even in the case where the scrutinee is a boolean literal and the pattern is `== true` or `== false`. 

In other words, flow analysis considers the scrutinee to be "some boolean" even though it's specifically one of the literals `true` or `false`, and hence it doesn't conclude that "we will choose one specific branch, and the other one is unreachable". Instead, it considers both branches to be reachable.

So how do we say that? (I think there must be a better way to say it than "all conditions are `true`").

```suggestion
/// @description Checks that if `E` has the form `if-case` then the flow
/// analysis will not conclude that a specific branch will definitely be taken,
/// even in the case where the scrutinee is a boolean literal and the pattern
/// is of the form `== b` where `b` is also a boolean literal. This is tested
/// via the definite assignment status of a variable.
```

TypeSystem/flow-analysis/reachability_conditional_A03_t02.dart
```diff
@@ -0,0 +1,23 @@
...
+/// - Let `after(N) = merge(after(S1), after(S2))`.
+///
+/// @description Checks that if `E` has the form `if-case` then initialization
+/// in the `when` part is visible for the flow analysis.
+/// @author sgrekhov22@gmail.com
+/// @issue 60269
+
+main() {
+  late int i;
+  if (false case == false when (i = 42) > 0) {
```
eernstg, line 19: Perhaps:
```suggestion
  if (false case == true when (i = 42) > 0) {
```

... just because the test seems stronger if the flow analysis doesn't "know" that we will definitely not execute `i = 42` (which is known because `i` is not definitely unassigned further down).

As long as we're testing `false case == false`, a flow analysis that (wrongly) "knows" the outcome will also know that `i = 42` will definitely be evaluated, which means that it is also 'not definitely unassigned' below.

TypeSystem/flow-analysis/reachability_conditional_A04_t01.dart
```diff
@@ -0,0 +1,28 @@
...
+main() {
+  int i;
+  bool b = true;
+  print(b); // no assignment to `b`
+  if (b) {
+    i = 42;
+  } else {
+    foo();
+  }
+  i; // Definitely assigned
```
eernstg, line 27: Wouldn't this test be slightly stronger if the flow analysis does not "know" that `b` has the value `true`?

Yes, again, it shouldn't rely on this knowledge, but if the flow analysis (wrongly) relies on this knowledge then we'd still conclude that `i` is definitely assigned below.

If we could instead have a situation where the flow analysis surely does not know the value of `b` then the test that `i` is definitely assigned would be known to take the return type `Never` of `foo` into account.

I think it would be enough to use `bool b = 2 > 1;`. Another way could be to make `b` a top-level variable.

TypeSystem/flow-analysis/reachability_conditional_A04_t02.dart
```diff
@@ -0,0 +1,28 @@
...
+///
+/// @description Checks that `before(S1) = split(true(E))` and
+/// `before(S2) = split(false(E))`.
+/// @author sgrekhov22@gmail.com
+
+Never foo() => throw "Never";
+
+main() {
+  int i;
+  bool b = false;
```
eernstg, line 20: Similar.

eernstg (commented):

TypeSystem/flow-analysis/reachability_conditional_A02_t01.dart
sgrekhov: Thank you. Updated here and below.

TypeSystem/flow-analysis/reachability_conditional_A03_t02.dart
sgrekhov: Updated.

TypeSystem/flow-analysis/reachability_conditional_A04_t01.dart
sgrekhov: Ah, yes! Of course! Thank you. Updated.

sgrekhov (commented):
Thank you! Updated. PTAL.

TypeSystem/flow-analysis/reachability_conditional_A01_t01.dart
```diff
@@ -0,0 +1,35 @@
...
+/// unreachable.
+/// @author sgrekhov22@gmail.com
+
+Never foo() => throw "Never";
+
+void test() {
+  late int i;
+  if (2 > 1) {
+    if (foo()) {
+      i = 42; // Variable is initialized in a dead code. This leaves it definitely unassigned
```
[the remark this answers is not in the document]
eernstg: Thanks!

eernstg (approved):
LGTM

PR #3108, opened 2025-03-17
#3057. Add `for(;;)` statements tests

sgrekhov:


TypeSystem/flow-analysis/reachability_for_A01_t01.dart
```diff
@@ -0,0 +1,59 @@
...
+/// that if the static type of `D` is `Never` then `C` is dead code and an
+/// assignment in it will never happen.
+/// @author sgrekhov22@gmail.com
+/// @issue 60089
+
+Never foo() => throw "Never";
+
+test1() {
+  late int i;
+  for (foo(); (i = 42) < 0;) {
```
eernstg, line 24: How about wrapping the `for` statement in an `if (2 > 1) {}` such that there is a path which will not throw? This means that the code in line 26 will not be dead. Also, this won't weaken the test because 'definitely unassigned' in line 26 (line 28 after the wrap) still says that we won't reach the condition part of the forLoopParts.

The current version of this test might wrongfully succeed if line 26 is reported as dead code, and 'definitely unassigned' is never detected.

TypeSystem/flow-analysis/reachability_for_A01_t01.dart
```diff
@@ -0,0 +1,59 @@
...
+  }
+  i; // Definitely unassigned.
+//^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+test2() {
+  late int i;
```
eernstg, line 33: The same kind of wrapping should work here, and in the corresponding location below.

TypeSystem/flow-analysis/reachability_for_A02_t01.dart
```diff
@@ -0,0 +1,71 @@
...
+/// @description Checks that `before(S) = split(true(C))`. Test that if the
+/// static type of `C` is `Never`, then an assignment in `S` will never happen.
+/// @author sgrekhov22@gmail.com
+/// @issue 60089
+
+Never foo() => throw "Never";
+
+test1() {
+  late int i;
+  for (; foo();) {
```
eernstg, line 22: This is one of those tests where the `if (2 > 1) {}` wrapper would make line 25 non-dead.

eernstg: Same for the other cases in this test.

TypeSystem/flow-analysis/reachability_for_A03_t04.dart
```diff
@@ -0,0 +1,64 @@
...
+/// @author sgrekhov22@gmail.com
+
+Never foo() => throw "Never";
+
+test1() {
+  late int i;
+  if (2 > 1) {
+    for (;; i = 42) {
+      foo();
+      break;
```
eernstg, line 24: Does it help to have `break;` here? It should be dead code.

TypeSystem/flow-analysis/reachability_for_A03_t05.dart
```diff
@@ -0,0 +1,47 @@
...
+/// `for (D; C; U) S,` then:
+/// - Let `before(D) = before(N)`.
+/// - Let `before(C) = conservativeJoin(after(D), assignedIn(N), capturedIn(N))`
+/// - Let `before(S) = split(true(C))`.
+/// - Let `before(U) = merge(after(S), continue(S))`.
+/// - Let `after(N) = inheritTested(join(false(C), unsplit(break(S))), after(U))`
+///
+/// @description Checks that `before(U) = merge(after(S), continue(S))`. Test
+/// that if the static type of `D` is `Never`, then an assignment in `U` will
+/// never happen.
```
eernstg, line 15: OK, let's just keep this one where we are testing the treatment of dead code: It has comments to remind a reader that the code is dead, and it seems pragmatic/useful for the analysis to avoid reporting any other diagnostics than 'dead code' for line 23, 31, 40.

However, I don't see any cases where `D` has type `Never`. Missing update in `@description`?

TypeSystem/flow-analysis/reachability_for_A01_t01.dart
sgrekhov: Makes sense. Added, thank you!

TypeSystem/flow-analysis/reachability_for_A02_t01.dart
sgrekhov: Added.

TypeSystem/flow-analysis/reachability_for_A03_t04.dart
sgrekhov: Hm. Indeed. Removed, thanx.

TypeSystem/flow-analysis/reachability_for_A03_t05.dart
sgrekhov: Yes, thank you.  Description update was missing. Fixed.

sgrekhov (commented):
Thank you! Updated. PTAL.

TypeSystem/flow-analysis/reachability_for_A01_t01.dart
eernstg: Thanks!

eernstg (approved):
LGTM

PR #3110, opened 2025-03-18
#3057. Add more for-statement tests

sgrekhov:
It turns out that flow analysis after `break` statement can be not intuitive :)

### DIFF
--- a/TypeSystem/flow-analysis/reachability_for_A01_t10.dart
+++ b/TypeSystem/flow-analysis/reachability_for_A01_t10.dart
@@ -1,0 +1,84 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion for statement: If `N` is a for statement of the form
+/// `for (D; C; U) S,` then:
+/// - Let `before(D) = before(N)`.
+/// - Let `before(C) = conservativeJoin(after(D), assignedIn(N), capturedIn(N))`
+/// - Let `before(S) = split(true(C))`.
+/// - Let `before(U) = merge(after(S), continue(S))`.
+/// - Let `after(N) = inheritTested(join(false(C), unsplit(break(S))), after(U))`
+///
+/// @description Checks that
+/// `before(C) = conservativeJoin(after(D), assignedIn(N), capturedIn(N))`. Test
+/// that `capturedIn(C)` is detected by flow analysis.
+/// @author sgrekhov22@gmail.com
+
+int? x = (2 > 1) ? 1 : null;
+
+test1(int? n) {
+  if (n != null) { // n promoted to `int`
+    for (; () {
+        late int? i = (n = x); // n demoted to `int?`
+        return false;
+      }();) {
+    }
+    n.isEven;
+//    ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  }
+}
+
+test2(int? n) {
+  if (n != null) {
+    [
+      for (; () {
+        late int? i = (n = x);
+        return false;
+      }();) 0,
+      n.isEven
+//      ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    ];
+  }
+}
+
+test3(int? n) {
+  if (n != null) {
+    <Object, int>{
+      for (; () {
+        late int? i = (n = x);
+        return false;
+      }();) 0: 0,
+      n.isEven: 0
+//      ^^^^^^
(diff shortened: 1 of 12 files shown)


eernstg:
Oh, I thought the flow analysis was always clear as day! ;-D

TypeSystem/flow-analysis/reachability_for_A01_t10.dart
```diff
@@ -0,0 +1,84 @@
...
+/// `before(C) = conservativeJoin(after(D), assignedIn(N), capturedIn(N))`. Test
+/// that `capturedIn(C)` is detected by flow analysis.
+/// @author sgrekhov22@gmail.com
+
+int? x = (2 > 1) ? 1 : null;
+
+test1(int? n) {
+  if (n != null) { // n promoted to `int`
+    for (; () {
+        late int? i = (n = x); // n demoted to `int?`
```
eernstg, line 23: I don't think write-captured variables care about which type we're assigning to them: They can't be promoted at any location where the write-capture has occurred (that is: the function object exists, no matter whether we "know" that it has or hasn't been called). So `x` could be a plain `int`, and it would still prevent all promotion of `n`.

I think this implies that the test would be slightly stronger if `x` were actually of type `int`.

eernstg: Also, the flow analysis makes no attempt to prove that `n = x` is dead code because we never read the value of `i`. All we need is the fact that `n = x` _occurs_ in the function literal, no matter where.

TypeSystem/flow-analysis/reachability_for_A01_t12.dart
```diff
@@ -0,0 +1,74 @@
...
+test2(int? n) {
+  if (n != null) {
+    [
+      for (;n > 0 ;)
+//            ^
+// [analyzer] unspecified
+// [cfe] unspecified
+        () {
+          late int? v = (n = x);
+        }
```
eernstg, line 40: Yes, this puts `n` into `capturedIn(N)`.

eernstg: Similarly for the following cases.

TypeSystem/flow-analysis/reachability_for_A03_t06.dart
```diff
@@ -0,0 +1,26 @@
...
+
+/// @assertion for statement: If `N` is a for statement of the form
+/// `for (D; C; U) S,` then:
+/// - Let `before(D) = before(N)`.
+/// - Let `before(C) = conservativeJoin(after(D), assignedIn(N), capturedIn(N))`
+/// - Let `before(S) = split(true(C))`.
+/// - Let `before(U) = merge(after(S), continue(S))`.
+/// - Let `after(N) = inheritTested(join(false(C), unsplit(break(S))), after(U))`
+///
+/// @description Checks that `before(U) = merge(after(S), continue(S))`. Test
```
eernstg, line 13: This is going to merge the flow models which is `after(S)` plus all the flow models from locations where `S` contains a statement of the form `continue;` or `continue L;` which is targeting this `for` statement. But then we should probably have at least one test that contains `continue;` or `continue L;`.

One thing which could be tested is that a variable which may be assigned before a `continue;` in the body is "not definitely unassigned" when it is used in `U`.

This test seems to be testing something which is quite different, and I'm actually not quite sure how I would describe it...

eernstg (commented):
This is tricky! ;-)

I've tried to comment on a number of cases where a test has a description which claims to be about a specific formula in the spec (in particular, the conservative join), but it doesn't really seem to test that formula. It's more natural to say that it is testing some other thing (e.g., that `after(B)` contains a specific setting of 'assigned' for a variable, or 'captured', etc.)

TypeSystem/flow-analysis/reachability_for_A03_t07.dart
```diff
@@ -0,0 +1,26 @@
...
+/// that if `D`, `C` are empty and variable is assigned in `S` after `continue`,
+/// then this variable is treated in `U` as 'possibly assigned`.
+/// @author sgrekhov22@gmail.com
+
+main() {
+  late int i;
+  try {
+    for (;; i) { // Possibly assigned. https://github.com/dart-lang/sdk/issues/42232#issuecomment-690681385
+      continue;
+      i = 42;
```
eernstg, line 23: This is a good one because it shows that the analysis is somewhat lossy: We will actually never execute line 23, but the analysis still thinks that it might have been executed when we reach `U`.

Of course, developers would be expected to modify their code such that there is no dead code, which means that it isn't a big problem that the analysis can be made less useful by dead code.

eernstg (approved):
LGTM

PR #3115, opened 2025-03-24
#3057. Add flow analysis tests for the `for-in` statement

sgrekhov:


TypeSystem/flow-analysis/reachability_for_in_A03_t03.dart
```diff
@@ -0,0 +1,25 @@
...
+/// an assignment in `S` before the `continue` statement is treated by flow
+/// analysis as possibly assigned.
+/// @author sgrekhov22@gmail.com
+
+main() {
+  late int i;
+  if (2 > 1) {
+    for (var j in [42]) {
+      i = 42;
+      continue;
```
eernstg, line 21: I think the 'conservative' part of this step in the flow analysis would make the test succeed even in the case where the `continue;` statement occurs _before_ the assignment.

The usual rule could then say "don't test that, because we don't want to depend on the treatment of dead code", but in this particular case it _is_ actually specified that `assignedIn` can't see whether the code is live or dead.

eernstg (commented):
Looks good!

I've suggested adding an `if (2 > 1) {...}` wrapper around various pieces of code, in order to ensure that we test the same thing, but do not depend on the treatment of dead code, when possible.

TypeSystem/flow-analysis/reachability_for_in_A03_t03.dart
sgrekhov: There is `reachability_continue_A01_t02.dart` that tests an assignment after `continue` statement. But there were no tests checking an assignment before the `continue`.  Therefore added this one.

sgrekhov (commented):
Updated. PTAL. Adding an `if (2 > 1) {...}` wrapper sometimes hides an issue (which, I believe, is https://github.com/dart-lang/sdk/issues/60265). Therefore, I added the wrapper only partially. However, there is another possible approach. We could add the wrapper to all of these tests and include a couple of new ones without it. These would demonstrate sdk#60265 while still containing important checks in dead code. Please let me know which option is better for you.

TypeSystem/flow-analysis/reachability_for_in_A03_t03.dart
eernstg: Thanks!

eernstg (commented):
Please take one more look at the situation: I hope it will be possible to de-emphasize testing of the properties of dead code (that's what all the `if (2 >1) ...` thingies are for), and I hope there will be at least one test which catches the bug that you mention (and, apparently, that one _must_ do some testing of the properties of dead code).

TypeSystem/flow-analysis/reachability_for_in_A01_t01.dart
```diff
@@ -0,0 +1,49 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion for each statement: If `N` is a for statement of the form
+/// `for (T X in E) S`, `for (var X in E) S`, or `for (X in E) S`, then:
+/// - Let `before(E) = before(N)`
+/// - Let `before(S) = conservativeJoin(after(E), assignedIn(N), capturedIn(N))`
+/// - Let `after(N) = join(before(S), break(S))`
+///
+/// @description Checks that if `E` has type `Never` then an assignment in `X`
+/// is unreachable.
```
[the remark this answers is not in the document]
sgrekhov: > So what's the situation? Do we have a broader "local variable has type Never, flow analysis forgets" bug? Do we have a test (another one) that catches this bug?

No we don't have one general issue for that. Flow analysis mostly correctly process type `Never` but forgets about it sometimes. I'm going to have a separate small isuue for each of these cases.

Filled https://github.com/dart-lang/sdk/issues/60395 for dead code and https://github.com/dart-lang/sdk/issues/60394 and https://github.com/dart-lang/sdk/issues/60393 for live code. Added wrappers and `reachability_for_in_A01_t10.dart` with no wrappers.

sgrekhov (commented):
Added wrappers and one test with no wrappers. PTAL.

TypeSystem/flow-analysis/reachability_for_in_A01_t01.dart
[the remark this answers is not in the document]
eernstg: >  I'm going to have a separate small isuue for each of these cases.

Sounds good!

> Added wrappers and `reachability_for_in_A01_t10.dart` with no wrappers.

Thanks! I was worried that we could end up with lots of tests relying on a specific treatment of dead code, and then the whole topic of "how to deal with dead code" is changed, and we don't really have a record of which tests depend on it. It might be OK to deal with this because "everything that just broke must depend on the treatment of dead code", but it seems cleaner if we keep the dependency on dead code analysis to a minimum, with some kind of indicator (the description or a comment should say 'dead code' somewhere, and add a few words about what's going on).

TypeSystem/flow-analysis/reachability_for_in_A01_t04.dart
```diff
@@ -0,0 +1,59 @@
...
+/// assignment in `X` is unreachable.
+/// @author sgrekhov22@gmail.com
+/// @issue 60265
+
+Never foo() => throw "Never";
+
+test1() {
+  late int i;
+  for (i in [foo()]) {}
+  i; // Definitely unassigned.
```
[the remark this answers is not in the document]
eernstg: Thanks!

TypeSystem/flow-analysis/reachability_for_in_A01_t10.dart
```diff
@@ -0,0 +1,52 @@
...
+
+/// @assertion for each statement: If `N` is a for statement of the form
+/// `for (T X in E) S`, `for (var X in E) S`, or `for (X in E) S`, then:
+/// - Let `before(E) = before(N)`
+/// - Let `before(S) = conservativeJoin(after(E), assignedIn(N), capturedIn(N))`
+/// - Let `after(N) = join(before(S), break(S))`
+///
+/// @description Checks that if `E` has type `Never` or `List<Never>` then an
+/// assignment in `X` or `S` is unreachable. Test use of a definitely unassigned
+/// variable in dead code.
```
eernstg, line 13: Yes, thanks for mentioning the dead code!

eernstg (approved):
LGTM

PR #3116, opened 2025-03-25
#3057. Fix tests for literals and the type-check

sgrekhov:


TypeSystem/flow-analysis/reachability_A05_t01.dart
```diff
@@ -2,26 +2,21 @@
...
-  }
-  try {
-    i;
-    Expect.fail("Error expected");
-  } on Error catch(e) {
-    // Ok
+  if ("Lily was here" == null) { // ignore: unnecessary_null_comparison
+    i = 42;
   }
+  i; // Definitely unassigned
```
eernstg, line 18: OK, so this test assumes that we have already introduced the changes to the flow analysis that are discussed in dart-lang/language#3100.

@stereotype441, how do we coordinate this with a spec update? The [flow-analysis feature spec](https://github.com/dart-lang/language/blob/main/resources/type-system/flow-analysis.md) already refers to issue 3100 in one location, but there might be additional adjustments coming up as this is implemented, right?

TypeSystem/flow-analysis/reachability_A05_t02.dart
```diff
@@ -2,22 +2,18 @@
...
 /// @author sgrekhov@unipro.ru
 /// @issue 41985
+/// @issue 60114
 
 main() {
   int i;
-  if ("Lily was here" != null) {
-    i = 42;   // `i` is not definitely assigned because in a weak mode the
-              // condition may be false
+  if ("Lily was here" != null) { // ignore: unnecessary_null_comparison
```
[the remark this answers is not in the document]
eernstg: All the tests in this PR rely on the same spec update.

stereotype441:
@eernstg said:
> @stereotype441, these changes would presumably be valid when the spec and implementation changes discussed in [dart-lang/language#3100](https://github.com/dart-lang/language/issues/3100) have been performed. What's the right ordering/scheduling of these changes and the associated tests?

I think the changes are significant enough that they should be language-versioned, which means that we should handle them the same way we handle other language versioned features. That means something like:
- Create an experimental flag for the feature. I can do this later today; if there are no objections, let's call it `sound-flow-analysis`.
- Include the line `// SharedOptions=--enable-experiment=sound-flow-analysis` in any co19 tests that exercise the new behavior.
- Include the line `// @dart=3.7` in any co19 tests that exercise the old behavior (3.7 is the latest stable release version, so it unambiguously selects the old behavior).

I'll be including similar comments in tests that I write in `tests/language` for the new feature.

Note that the new tests will initially fail, because I haven't implemented the new functionality yet. That's fine. We can ignore the failures for now, and when I get around to implementing the new functionality, I'll search for pre-existing failures to make sure they all start passing.

stereotype441:
> I think the changes are significant enough that they should be language-versioned, which means that we should handle them the same way we handle other language versioned features. That means something like:
> 
> * Create an experimental flag for the feature. I can do this later today; if there are no objections, let's call it `sound-flow-analysis`.

FYI, the new experimental flag has landed: https://github.com/dart-lang/sdk/commit/82420ba66cc7acf631d067e00d40c325919bac44

sgrekhov:
Thank you!

sgrekhov:
Experimental flag and the language version added. Please review.

TypeSystem/flow-analysis/reachability_A05_t01.dart
```diff
@@ -10,6 +10,8 @@
 /// @author sgrekhov@unipro.ru
 /// @issue 41985
 
+// @dart = 3.7.0
+
 import "../../Utils/expect.dart";
 
 main() {
```
eernstg, line 17: I can't comment on line 21 directly, but 'condition may be true' seems to imply that a string literal could actually evaluate to null during unsound execution, which was never true. It would be safer to say that the 'value of a non-nullable expression is considered to be possibly null'.

This reflects the actual level of detail that the analysis is going into when this decision is made: It doesn't make any attempt to distinguish between `"A string literal"` and a `function()` even though it's only the latter which could _actually_ be null even when `function()` has a non-nullable return type.

eernstg (commented):
Looks good!

I'm suggesting an adjustment of some comments (because the current wording gives readers too much justification for saying "that's not true!").

sgrekhov:
Updated. PTAL

eernstg (approved):
LGTM, thanks!

PR #3121, opened 2025-03-31
#3057. Add switch statement tests

sgrekhov:


TypeSystem/flow-analysis/reachability_switch_A04_t02.dart
```diff
@@ -0,0 +1,51 @@
...
+
+void test1() {
+  late int i;
+  switch (42) {
+    case 1:
+      break;
+      i = 42;
+    default:
+  }
+  i; // Definitely assigned
```
eernstg, line 28: Typo:
```suggestion
  i; // Definitely unassigned
```

eernstg: One more below.

TypeSystem/flow-analysis/reachability_switch_A02_t06.dart
```diff
@@ -0,0 +1,36 @@
...
+
+void test(int? n) {
+  if (n != null) {  // n promoted to int
+    switch ((){n = 42;}) { // n demoted to int?
+      label: case 1: // ignore: constant_pattern_never_matches_value_type
+        n.isEven;
+//        ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+      default:
```
eernstg, line 29: I think the demotion should occur here as well because it is a property in `after(E)`, the conservative join just preserves this property.

TypeSystem/flow-analysis/reachability_switch_A02_t07.dart
```diff
@@ -0,0 +1,69 @@
...
+  if (n != null) {  // n promoted to int
+    switch (42) {
+      case 1:
+        () {n = 42;}; // n demoted to int?
+      label: case 2:
+        n.isEven;
+//        ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+      default:
```
eernstg, line 31: But I think this one should allow `n.isEven`.

TypeSystem/flow-analysis/reachability_switch_A02_t07.dart
```diff
@@ -0,0 +1,69 @@
...
+
+void test2(int? n) {
+  if (n != null) {
+    switch (42) {
+      label: case 1:
+        n.isEven;
+//        ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+      case 2:
```
eernstg, line 44: This would should allow `n.isEven` if it occurs before the function literal. `default` should allow it, too.

TypeSystem/flow-analysis/reachability_switch_A02_t07.dart
```diff
@@ -0,0 +1,69 @@
...
+
+void test3(int? n) {
+  if (n != null) {
+    switch (42) {
+      label: case 1:
+        n.isEven;
+//        ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+      default:
```
eernstg, line 59: Again, `n.isEven` should be OK before the function literal.

eernstg (commented):

I think a few tests can be slightly stronger: The conservative join applies to each case that has a label, but not the others, and this seems to allow certain expressions to occur in those non-labeled cases.

TypeSystem/flow-analysis/reachability_switch_A02_t06.dart
sgrekhov: Added to default as well.

TypeSystem/flow-analysis/reachability_switch_A02_t07.dart
sgrekhov: Added here and below

TypeSystem/flow-analysis/reachability_switch_A04_t02.dart
sgrekhov: Thank you. Fixed.

sgrekhov (commented):
Thank you! Fixed. PTAL.

TypeSystem/flow-analysis/reachability_switch_A02_t06.dart
```diff
@@ -14,8 +14,8 @@
 ///
 /// @description Checks that if `C` is labelled then
 /// `before(S) = conservativeJoin(after(E), assignedIn(N), capturedIn(N))`. Test
-/// that if `C` is labeled and a promoted variable was captured in `E` then it
-/// is demoted in `S`.
+/// that if `C` is labeled or `default` and a promoted variable was captured in
```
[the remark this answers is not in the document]
sgrekhov: Thank you! Updated. PTAL.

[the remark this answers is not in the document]
eernstg: Thanks!

eernstg (approved):
LGTM

PR #3124, opened 2025-04-01
#3057. Add pattern assignment cases to do-while tests

sgrekhov:


TypeSystem/flow-analysis/reachability_do_while_A01_t02.dart
```diff
@@ -12,9 +12,42 @@
...
+  late int i;
+  do {
+    (x: i) = (x: 42);
+  } while (i < 0);
+}
+
+test4() {
+  late int i;
+  do {
+    (i,) = (42,);
```
eernstg, line 44: Should this be an object pattern based pattern assignment?

eernstg (commented):
Looks good! I think there's one typo where `(i,)` should have been `C(v: i)`.

TypeSystem/flow-analysis/reachability_do_while_A01_t02.dart
sgrekhov: Of course! Thank you for noticing! Fixed.

eernstg (approved):
LGTM

Issue #3128, opened 2025-04-02
Create CLI tool showing coverage of the Dart specification by co19 test


sgrekhov:
Create CLI tool that will show the coverage of the dart specification by co19 test. The tool should search assertion text in the spec for the each test and print if it found or not.

PR #3129, opened 2025-04-03
#3128. Initial commit

sgrekhov:
Please note `Tools/spec_coverage/resources/readme.txt`


Tools/spec_coverage/README.md
```diff
@@ -0,0 +1,5 @@
+Command-line application that shows Dart specification coverage by co19 tests.
+
+Usage:
+`$ cd Tools/spec_coverage`
```
athomas, line 4: Can we use `tools` instead? co19 is the only place we use uppercase in directories, and in the Dart SDK we have a `tools` directory as well. Preferable we'd change `Utils` as well. But we should leave the test directories alone because we don't want the churn of renaming tests.

Tools/spec_coverage/resources/dartLangSpec.tex
athomas, line 1: Do we have to check this in?

Tools/spec_coverage/resources/spec.txt
athomas, line 1: Do we have to check this in?

Tools/spec_coverage/resources/dartLangSpec.tex
sgrekhov: It simplifies the tool for now. Without it I need:
- download spec
- save it in some temp directory
- perform `latex2text dartLangSpec.tex > spec.txt` (need to add system requirements about `latex2text`)
- do the work
- perform the cleanup

I'd like to have it for now as a temp solution to complete the tool itself first. Then to add the above.

Tools/spec_coverage/README.md
sgrekhov: Renamed.

Tools/spec_coverage/resources/dartLangSpec.tex
athomas: As a temporary solution, would it be ok to add it to .gitignore, so you would have it locally in your checkout and other people would have to create the files manually?

Once it's in the git history, it will be there forever.

sgrekhov: > As a temporary solution, would it be ok to add it to .gitignore, so you would have it locally in your checkout and other people would have to create the files manually?

Sure. Done. Updated README.md as well.

athomas (approved):
LGTM, will not merge and let Erik take a look as well.

eernstg:
(Just FYI: I'm working on this. ;-)

tools/spec_coverage/lib/co19.dart
```diff
@@ -0,0 +1,68 @@
...
+
+class Co19 {
+  static const languageDirName = "Language";
+  late final TestDir root;
+  late final TestDir language;
+
+  Co19(String path) {
+    Directory dir = Directory(path);
+    if (!dir.existsSync()) {
+      throw Exception("Directory '$path' does not exists");
```
eernstg, line 11:
```suggestion
      throw Exception("Directory '$path' does not exist");
```

tools/spec_coverage/lib/co19.dart
```diff
@@ -0,0 +1,68 @@
...
+    List<FileSystemEntity> entities = root.listSync();
+    for (FileSystemEntity fse in entities) {
+      if (skip(fse)) {
+        continue;
+      }
+      FileStat fs = fse.statSync();
+      if (fs.type == FileSystemEntityType.directory) {
+        subDirs.add(TestDir(Directory(fse.path)));
+      } else if (fs.type == FileSystemEntityType.file) {
+        tests.add(TestFile(fse.path));
```
eernstg, line 38: Would it be useful to specify a regular expression matching actual test file names in `Co19` and passing it around such that we can include only test files here? Or perhaps other files (imported libraries and whatnot) could also have an `@assertion` and/or a `@description`?

tools/spec_coverage/lib/co19.dart
```diff
@@ -0,0 +1,68 @@
...
+      }
+    }
+  }
+
+  static String resolvePath(String relativePath) {
+    String basePath = Directory.current.path;
```
eernstg, line 44: So this utility _must_ be executed from the root of the package? OK, I think that's a rather common restriction (but it might be nice to have an error message if the current directory at startup doesn't look like a package root).

tools/spec_coverage/lib/co19.dart
```diff
@@ -0,0 +1,68 @@
...
+    Uri resolvedUri = baseUri.resolve(relativePath);
+    return resolvedUri.toFilePath();
+  }
+
+  String get name => _entityName(path);
+
+  String _entityName(String p) =>
+      p.substring(p.lastIndexOf(Platform.pathSeparator) + 1);
+
+  bool skip(FileSystemEntity fse) {
```
eernstg, line 55: Private?

tools/spec_coverage/lib/spec.dart
```diff
@@ -0,0 +1,37 @@
+import 'dart:io';
+
+import 'package:spec_coverage/spec_parser.dart';
+
+class Spec {
+  late final List<Chapter> chapters;
+
+  Spec.fromTxt(String path) {
```
eernstg, line 8: Perhaps this could be a static method returning `Spec` (possibly renamed as `Specification`). The class could then have a private constructor accepting the list, and `chapters` wouldn't have to be late.

tools/spec_coverage/lib/spec_parser.dart
```diff
@@ -0,0 +1,61 @@
+import 'spec.dart';
+
+class SpecParser {
```
eernstg, line 3: OK, so this class depends in detail on the format of `spec.txt`. It's a good thing that this dependency is encapsulated to this class alone.

eernstg (commented):
Many good things here! I'm not quite sure about the status&mdash;some parts of the code look like it is not nearly finished. So we need to find a good approach to check in this code such that it is well-understood what the status is, and what the code is expected to be able to do. We should probably talk about these things in an online meeting because it's a new situation.

tools/spec_coverage/bin/main.dart
```diff
@@ -0,0 +1,42 @@
...
+  Spec spec = Spec.fromTxt(config.specPath);
+  Co19 co19 = Co19(config.co19Dir);
+  findSpecChapters(co19.language.subDirs, spec.chapters);
+}
+
+void findSpecChapters(List<TestDir> testDirs, List<Chapter> chapters) {
+  for (TestDir td in testDirs) {
+    bool found = false;
+    for(Chapter ch in chapters) {
+      if (td.name.toLowerCase() == ch.co19DirName.toLowerCase()) {
```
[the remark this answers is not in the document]
sgrekhov: Thank you. Fixed! Please review.

sgrekhov (commented):
Thank you for the very detailed and useful review! Updated, PTAL. For now the only that this tool does is comparing spec chapters and test directories names and print the output like:
```
Found spec for co19\Language\Generics
Not found spec for co19\Language\Generics\Superbounded_types. Chapters are:
instantiation_to_bound
super_bounded_types
variance
```
And we immediately realyse why the appropriate chapter was not found for the directory.
The next step is to parse spec text and find appropriate assertions there.

PTAL.

PR #3130, opened 2025-04-03
#3057. Add pattern assignment cases to for-in tests

sgrekhov:


TypeSystem/flow-analysis/reachability_for_in_A03_t04.dart
```diff
@@ -25,17 +30,89 @@ test1() {
 }
```
eernstg, line 30: Need to wrap `i;` in a try-catch? It looks like this test would run, and then it would fail when `i` is evaluated here, and it hasn't been initialized (even though flow analysis thinks that it _might_ have been initialized, it hasn't).

eernstg: Similar issue with other cases in this test.

eernstg (commented):
Looks good!

It does look like a test will run and then incur a run time failure, though.

TypeSystem/flow-analysis/reachability_for_in_A03_t04.dart
sgrekhov: That's why `main` prints these functions instead of running them. I can make the update, but do we need it? Flow analysis is not about runtime behavior, so I don't think we need to run them

eernstg (approved):
LGTM

PR #3131, opened 2025-04-04
#3057. Update instance check tests and add negated instance check ones

sgrekhov:


TypeSystem/flow-analysis/reachability_A23_t01.dart
```diff
@@ -0,0 +1,30 @@
...
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+main() {
+  int? i = 2 > 1 ? null : 42;
+  if (i is! Null) {
+    i.expectStaticType<Exactly<int>>();
+  } else {
+    Null n = i;
```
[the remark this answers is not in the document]
eernstg: OK, `v is Null` and `v is! Null` _does_ promote `v` to have type `Null` in the appropriate branch, it's only `== null` and `!= null` that doesn't promote anything to `Null`. Testing against the type `Null` is linted, so that's probably the reason why I haven't noticed this behavior before. I think it's in line with the specification (couldn't find evidence to the contrary).

TypeSystem/flow-analysis/reachability_A23_t05.dart
```diff
@@ -0,0 +1,28 @@
...
+/// `E1 is! S` where the static type of `E1` is `T` then:
+/// - Let `before(E1) = before(N)`
+/// - If `T` is a bottom type, then:
+///   - Let `true(N) = after(E1)`.
+///   - Let `false(N) = unreachable(after(E1))`.
+/// - Otherwise:
+///   - Let `true(N) = promote(E1, factor(T, S), after(E1))`
+///   - Let `false(N) = promote(E1, S, after(E1))`
+///
+/// @description Checks that `T is a bottom type then `true(N) = after(E1)`.
```
eernstg, line 15: Typo:
```suggestion
/// @description Checks that if `T` is a bottom type then `true(N) = after(E1)`.
```

eernstg: About 4 other locations have the same text.

eernstg (commented):
Looks good, couple of typos!

sgrekhov:
Missed this one, sorry. Updated. PTAL.

eernstg (approved):
LGTM (added a few missing backquotes).

PR #3135, opened 2025-04-07
#3057. Add switch-expression tests

sgrekhov:


TypeSystem/flow-analysis/reachability_switch_expression_A01_t01.dart
```diff
@@ -0,0 +1,69 @@
...
+/// `switch (E) {alternatives}` then:
+/// - Let `before(E) = before(N)`.
+/// - For each `C` in `alternatives` with statement body `S`:
+///   - If `C` is labelled let
+///     `before(S) = conservativeJoin(after(E), assignedIn(N), capturedIn(N))`
+///     otherwise let `before(S) = after(E)`.
+///  - If the cases are exhaustive, then let `after(N) = break(N)` otherwise let
+///    `after(N) = join(after(E), break(N))`.
+/// TODO (sgrekhov): there is no switch expressions in the flow analysis spec
+/// yet (April 2025). Update the assertion after spec update.
```
eernstg, line 15: Thanks for a useful comment!

This is tricky because switch statements and switch expressions differ in several ways. In particular, switch expressions can't have labels (which would be easy to handle, we just take into account that no usages of a rule involving labels can occur), and they can't have `break` or `continue` statements (so we will need to consider whether we have any conservative joins which are not needed for a switch expression). Switch expressions are also guaranteed to be exhaustive, so we'll need to only use the "branches" of the ruleset where exhaustiveness is required or allowed.

All in all, I think we'll have a rule along the following lines (and surely there's a `split` or `unsplit` which is missing or should be placed differently):

```dart
/// @assertion switch expression: If `N` is a switch expression of the form
/// `switch (E) {alternatives}` then:
/// - Let `before(E) = split(before(N))`.
/// - For each case `Cj` in `alternatives` with expression `Ej`, for `j` in `1 .. k`:
///   `before(Ej) = after(E)`.
/// - `after(N) = join(after(E1), after(E2), .. after(Ek))`.
```

Does it make sense to handle switch expressions based on the switch statement rule if the switch expression rule would be so different?

Maybe. ;-)

One thing which is definitely still missing is that these rules are pre-pattern, which means that they do not take into account that pattern matching has control flow, and it can have side effects (including throwing an exception, e.g., because they use an expression of type `Never`).

In any case, I think we should be prepared to go ahead cautiously: Testing things that are unavoidable (no matter which rules we'll have, this particular test _must_ see the expected behavior, for soundness reasons or because it's an obvious case).

Then we can aim for a specification where switch expressions are included and focus on tricky corners later on.

I created https://github.com/dart-lang/language/issues/4320 in order to clarify which way ahead is better.

eernstg: Right now I'll review this series of test based "apparently obvious outcomes", hoping that this will be sufficient. ;-)

TypeSystem/flow-analysis/reachability_switch_expression_A02_t03.dart
```diff
@@ -0,0 +1,80 @@
...
+/// @author sgrekhov22@gmail.com
+
+class C {
+  int v;
+  C(this.v);
+}
+
+void test1(int? n) {
+  if (n != null) {  // n promoted to int
+    var x = switch ((){n = 42;}) { // n demoted to int?
```
eernstg, line 29: It seems reasonable to avoid the "never_matches" diagnostic because this makes the code somewhat artificial (I don't have a really good reason, but it seems reasonable to focus mostly on code that could actually run and reach the pieces of code which are in focus for the test). So:
```suggestion
    var x = switch ((){n = 42; return 1;}) { // `n` demoted to `int?`.
```
Similarly for the other cases.

eernstg (commented):

One issue should be considered, though: We don't have a specification of switch expressions, but (worse) we also don't have a specification of the flow analysis of pattern matching (which would make a huge difference for switch expression as well as switch statements).

I think we might well have covered much of this based on the ordering considerations that were part of the discussions about pattern tests, but there could still be some lack of coverage. I created an issue (mentioned in a comment).

So I think we can land this PR when the typos have been fixed, but then we probably need to think about further testing of similar constructs (involving pattern matching and/or other recent features).

TypeSystem/flow-analysis/reachability_switch_expression_A01_t01.dart
```diff
@@ -0,0 +1,69 @@
...
+/// @assertion switch statement: If `N` is a switch statement of the form
+/// `switch (E) {alternatives}` then:
+/// - Let `before(E) = before(N)`.
+/// - For each `C` in `alternatives` with statement body `S`:
+///   - If `C` is labelled let
+///     `before(S) = conservativeJoin(after(E), assignedIn(N), capturedIn(N))`
+///     otherwise let `before(S) = after(E)`.
+///  - If the cases are exhaustive, then let `after(N) = break(N)` otherwise let
+///    `after(N) = join(after(E), break(N))`.
+/// TODO (sgrekhov): there is no switch expressions in the flow analysis spec
```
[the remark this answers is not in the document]
sgrekhov: Updated.

TypeSystem/flow-analysis/reachability_switch_expression_A01_t01.dart
sgrekhov: Thank you! Updated the assertion text. I'll update the text and the tests when the spec will be changed.

TypeSystem/flow-analysis/reachability_switch_expression_A01_t01.dart
```diff
@@ -0,0 +1,69 @@
...
+
+void test1() {
+  late int i;
+  if (2 > 1) {
+    var x = switch (foo()) {
+      1 => (i = 42),
+      _ => (i = 42)
+    };
+  }
+  i; // Definitely unassigned
```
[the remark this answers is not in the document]
sgrekhov: There is an `if` condition above so it is not a dead code.

TypeSystem/flow-analysis/reachability_switch_expression_A02_t03.dart
sgrekhov: A very good advice! Thank you! Updated.

sgrekhov (commented):
Thank you. Updated. PTAL.

TypeSystem/flow-analysis/reachability_switch_expression_A01_t01.dart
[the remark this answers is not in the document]
eernstg: Ah, of course! Thanks.

TypeSystem/flow-analysis/reachability_switch_expression_A01_t01.dart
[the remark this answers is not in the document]
eernstg: Thanks!

TypeSystem/flow-analysis/reachability_switch_expression_A01_t01.dart
eernstg: Sounds good! I'm not sure about the placement of `split` operations, but it probably doesn't create any problems here even in the case where it isn't 100% right.

TypeSystem/flow-analysis/reachability_switch_expression_A02_t03.dart
eernstg: Thanks!

eernstg (approved):
LGTM

PR #3137, opened 2025-04-08
#3057. Add flow analysis try-catch tests

sgrekhov:


TypeSystem/flow-analysis/reachability_try_catch_A02_t05.dart
```diff
@@ -0,0 +1,73 @@
...
+test4(int? n) {
+  if (n != null) {
+    try {
+    } on Exception catch (_) {
+      () {C(v: n) = C(42);};
+    } catch (_) {
+      n.isEven;
+    }
+  }
+}
```
eernstg, line 66: Perhaps swap the order (e.g., in `test2` and `test4`) such that we also have the case where `n.isEven` comes in a textually earlier position than the code that captures `n` (it should be equally bad even though it might look like a case that is less likely to be buggy because the elements have a "natural" ordering).

TypeSystem/flow-analysis/reachability_try_catch_A01_t04.dart
```diff
@@ -0,0 +1,80 @@
...
+  try {
+  } on Exception catch (_) {
+    C(v: i) = C(42);
+  } catch (_) {
+    i;
+//  ^
+// [analyzer] unspecified
+// [cfe] unspecified
+  }
+}
```
eernstg, line 73: Perhaps reorder a couple of cases; see the comment on TypeSystem/flow-analysis/reachability_try_catch_A02_t05.dart for details.

eernstg (commented):
Looks good! Just one tiny thing: We might want to try both orderings in a couple of cases where two constructs play a similar role and the requirement is just that "one kind of thing most be in one construct, another kind in the other construct".

sgrekhov:
Thanks. Reordered. PTAL

TypeSystem/flow-analysis/reachability_try_catch_A02_t05.dart
eernstg: Thanks!

eernstg (approved):
LGTM

PR #3144, opened 2025-04-14
Fixes #3143. Update assertions, add language version, add new tests

### LINKED ISSUE #3143: Some additional tests need version comments because they depend on unsound flow analysis
I believe I have now completed the implementation work for the `sound-flow-analysis` language feature. The feature isn't enabled yet (because I still have to do more testing, write the spec, and prepare the internal google3 codebase for the new functionality). But I have an experimental CL that I've been using to see if any unexpected test failures will occur when the feature is turned on: https://dart-review.googlesource.com/c/sdk/+/420240. That CL is currently failing the following co19 tests, because they make assumptions about the behavior of flow analysis that will no longer be true when `sound-flow-analysis` is enabled:

```
TypeSystem/flow-analysis/reachability_A07_t07
TypeSystem/flow-analysis/reachability_A07_t08
TypeSystem/flow-analysis/reachability_A07_t09
TypeSystem/flow-analysis/reachability_A07_t10
```

I recommend adding the comment `// @dart = 3.8` to these tests, so that once `sound-flow-analysis` is turned on, they will continue to exercise the behavior of older language versions.

(Note that in the past I've recommended using `// @dart = 3.7` for a similar purpose; that was because I didn't know if I would get a chance to finish the implementation of `sound-flow-analysis` before Dart version 3.8 shipped. It turns out that I didn't (the branch cutoff for Dart 3.8 has now passed), so it's now safe to use `// @dart = 3.8` for this purpose. It's also fine to continue using `// @dart = 3.7` if you prefer.)

sgrekhov:


TypeSystem/flow-analysis/reachability_A07_t07.dart
```diff
@@ -2,29 +2,23 @@
...
-///    Let notNull(N) = promoteToNonNull(x, before(N))
-///
-/// @description Checks reachability after method call. Test non-nullable return
-/// type
+/// @description Checks reachability after method invocation. Tests a method
+/// with a non-nullable return type.
 /// @author sgrekhov@unipro.ru
 /// @issue 41985
 
+// @dart = 3.8
```
eernstg, line 20: Just to make sure: Dart 3.8 does _not_ have flow analysis that relies on sound null safety, which means that there's no need to update the comment below in line 32-33. I think it would be nice to have a comment to this effect.

TypeSystem/flow-analysis/reachability_A07_t08.dart
```diff
@@ -2,29 +2,23 @@
...
-///    Let notNull(N) = promoteToNonNull(x, before(N))
-///
-/// @description Checks reachability after method call. Test static method and
-/// non-nullable return type
+/// @description Checks reachability after method invocation. Tests a static
+/// method with a non-nullable return type.
 /// @author sgrekhov@unipro.ru
 /// @issue 41985
 
+// @dart = 3.8
```
eernstg, line 20: Again, a comment here reminding the reader that we need `@dart = 3.8` in order to ensure that there is no sound-flow-analysis.

eernstg: Same in a couple of other libraries.

TypeSystem/flow-analysis/reachability_A07_t11.dart
```diff
@@ -0,0 +1,33 @@
...
+///  - Let `T` be the static return type of the invocation
+///  - If `T <: Never` then:
+///   - Let `after(N) = unreachable(after(E2))`.
+///  - Otherwise:
+///    - Let `after(N) = after(E2)`
+///
+/// @description Checks reachability after method invocation. Tests a method
+/// with a non-nullable return type.
+/// @author sgrekhov22@gmail.com
+
```
eernstg, line 18: Shouldn't we enable the experiment `sound-flow-analysis` here?

eernstg (commented):
Looks good! I think it would be helpful to deal with language versions more explicitly, though.

TypeSystem/flow-analysis/reachability_A07_t07.dart
sgrekhov: Updated here and below

TypeSystem/flow-analysis/reachability_A07_t11.dart
sgrekhov: Right. Thank you! Added here and below.

sgrekhov (commented):
Thank you. Updated. PTAL

TypeSystem/flow-analysis/reachability_A07_t07.dart
eernstg: Thanks!

eernstg (approved):
LGTM

PR #3146, opened 2025-04-14
#3057. Add more try-catch-finally tests

sgrekhov:


TypeSystem/flow-analysis/reachability_try_catch_A03_t02.dart
```diff
@@ -0,0 +1,63 @@
...
+  }
+}
+
+test2() {
+  S s = S();
+  try {
+    if (s is T) {}
+  } catch (_) {
+    (s,) = (T(),);
+    s.answer();
```
eernstg, line 45: I would expect this one to fail because `before(S0) = conservativeJoin(before(N), ...)`, where `T` is not a type of interest for `s`.

TypeSystem/flow-analysis/reachability_try_catch_A03_t03.dart
```diff
@@ -0,0 +1,63 @@
...
+}
+
+test2() {
+  S s = S();
+  try {
+  } on Exception catch (_) {
+    if (s is T) {}
+  } catch (_) {
+    (s,) = (T(),);
+    s.answer();
```
eernstg, line 45: I'd expect this to fail because each alternative starts with `before(N)` (plus some `assigned` and some `captured` bits, but they won't change anything about types of interest).

TypeSystem/flow-analysis/reachability_try_finally_A04_t01.dart
```diff
@@ -0,0 +1,75 @@
...
+test2() {
+  S s = S();
+  if (s is T) {}
+  try {
+  } catch (_) {
+    (s,) = (T(),);
+    s.answer();
+  } finally {
+  }
+}
```
eernstg, line 46: Right, we should certainly test try/catch/finally even though it's not yet in the spec. Best effort, as usual. ;-)

But I guess they'd fit better in a series of tests named `reachability_try_catch_finally_A*_t*.dart`.

TypeSystem/flow-analysis/reachability_try_finally_A04_t02.dart
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+
+test2() {
+  S s = S();
+  try {
+    if (s is T) {}
+  } catch (_) {
+    (s,) = (T(),);
+    s.answer();
```
eernstg, line 44: This one should again fail, considering that the catch blocks have no more than `before(N)` when it comes to types of interest.

TypeSystem/flow-analysis/reachability_try_finally_A04_t02.dart
```diff
@@ -0,0 +1,76 @@
...
+}
+
+test3() {
+  S s = S();
+  try {
+    if (s is T) {}
+  } catch (_) {
+  } finally {
+    (x: s) = (x: T());
+    s.answer();
```
eernstg, line 56: Ditto.

TypeSystem/flow-analysis/reachability_try_finally_A04_t03.dart
```diff
@@ -0,0 +1,78 @@
...
+
+/// @assertion try finally: If `N` is a try/finally statement of the form
+/// `try B1 finally B2` then:
+/// - Let `before(B1) = split(before(N))`
+/// - Let `before(B2) = split(join(drop(after(B1)),
+///     conservativeJoin(before(N), assignedIn(B1), capturedIn(B1))))`
+/// - Let `after(N) = restrict(after(B1), after(B2), assignedIn(B2))`
+///
+/// @description Checks that if a type `T` is made a type of interest in a catch
+/// part of `B1` then it can be promoted in `B1`, `B2` and `after(N)`.
```
eernstg, line 13: I don't think `B1` can include a catch part. I think we have rules about `try .. catch .. catch` and about `try .. finally`, but the rule about `try .. catch .. catch .. finally` is missing.

So we're testing `try`-`catch`-`finally` statements as a separate kind of case, noting that these tests must be revisited when the spec rule is in place.

TypeSystem/flow-analysis/reachability_try_finally_A04_t03.dart
```diff
@@ -0,0 +1,78 @@
...
+}
+
+test2() {
+  S s = S();
+  try {
+  } on Exception catch (_) {
+    if (s is T) {}
+  } catch (_) {
+    (s,) = (T(),);
+    s.answer();
```
eernstg, line 46: This should fail, `S0` is not a predecessor of `S1` in the flow graph.

TypeSystem/flow-analysis/reachability_try_finally_A04_t03.dart
```diff
@@ -0,0 +1,78 @@
...
+}
+
+test3() {
+  S s = S();
+  try {
+  } on Exception catch (_) {
+    if (s is T) {}
+  } finally {
+    (x: s) = (x: T());
+    s.answer();
```
eernstg, line 58: Similarly, the `catch` block is not a predecessor of the `finally` block.

eernstg (commented):
Looks good! I think, however, that a few cases are expected to pass, but must be errors.

TypeSystem/flow-analysis/reachability_try_catch_A03_t02.dart
sgrekhov: But according to the https://github.com/dart-lang/sdk/issues/60519 this should be an error.

eernstg: Exactly, it should be an error because `T` is not a type of interest for `s` at the location of the pattern assignment.

eernstg (commented):
(Clarifications)

TypeSystem/flow-analysis/reachability_try_catch_A03_t02.dart
eernstg: However, I just created https://github.com/dart-lang/language/issues/4328 in order to clarify the situation: The implemented behavior appears to be different from the specified behavior.

eernstg (commented):
I think this one might need to be blocked on a spec clarification, too: https://github.com/dart-lang/language/issues/4328.

sgrekhov:
Partially tests are duplicates of existing ones, partially replaced by https://github.com/dart-lang/co19/pull/3148. Closing this one.

PR #3147, opened 2025-04-15
#3057. Add switch expression/statement tests for promotion

sgrekhov:


TypeSystem/flow-analysis/reachability_switch_A06_t01.dart
```diff
@@ -0,0 +1,50 @@
...
+/// - Let `before(E) = before(N)`.
+/// - For each `C` in `alternatives` with statement body `S`:
+///   - If `C` is labelled let
+///     `before(S) = conservativeJoin(after(E), assignedIn(N), capturedIn(N))`
+///     otherwise let `before(S) = after(E)`.
+///  - If the cases are exhaustive, then let `after(N) = break(N)` otherwise let
+///    `after(N) = join(after(E), break(N))`.
+///
+/// @description Checks that if a type `T` is made a type of interest in `E`
+/// then it can be promoted in `alternatives` and `after(N)`
```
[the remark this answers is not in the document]
eernstg: Similarly for a few other locations in other test libraries below.

TypeSystem/flow-analysis/reachability_switch_A06_t02.dart
```diff
@@ -0,0 +1,51 @@
...
+/// - Let `before(E) = before(N)`.
+/// - For each `C` in `alternatives` with statement body `S`:
+///   - If `C` is labelled let
+///     `before(S) = conservativeJoin(after(E), assignedIn(N), capturedIn(N))`
+///     otherwise let `before(S) = after(E)`.
+///  - If the cases are exhaustive, then let `after(N) = break(N)` otherwise let
+///    `after(N) = join(after(E), break(N))`.
+///
+/// @description Checks that if a type `T` is made a type of interest in
+/// `alternatives` then it can be promoted in `alternatives` and `after(N)`
```
eernstg, line 16: I don't think that's true, neither in the situation where the case is labelled or in the case where it is not. First, `before(S) = after(E)` certainly doesn't introduce anything about types of interest in other cases. Next, in the labelled case, the `conservativeJoin` doesn't transfer information about types of interest anywhere, it only transfers information about being assigned and being write-captured.

In short, an unlabelled case can't see any other cases. A labelled case can see all other cases, but only with respect to 'assigned' and 'captured'.

TypeSystem/flow-analysis/reachability_switch_A06_t02.dart
```diff
@@ -0,0 +1,51 @@
...
+}
+
+void test2() {
+  S s = S();
+  switch (s is T) {
+    case true:
+    case false:
+  }
+  s = T();
+  s.answer();
```
eernstg, line 45: This works!

TypeSystem/flow-analysis/reachability_switch_expression_A04_t01.dart
```diff
@@ -0,0 +1,48 @@
...
+}
+
+void test2() {
+  S s = S();
+  var x = switch (s is T) {
+    true => 1,
+    _ =>  0
+  };
+  s = T();
+  s.answer();
```
eernstg, line 42: This might fail, but it would be a bug, so the test is fine.

TypeSystem/flow-analysis/reachability_switch_expression_A04_t03.dart
```diff
@@ -0,0 +1,50 @@
...
+}
+
+test1() {
+  S s = S();
+  var x = switch (42) {
+    1 => [s = T(), s.answer()],
+//                   ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    _ =>  [s is T ? 1: 2] // Make `T` a type of interest
```
eernstg, line 32: Typo:
```suggestion
    _ => [s is T ? 1: 2] // Make `T` a type of interest
```

eernstg, line 32: Typo:
```suggestion
    _ => [s is T ? 1: 2] // Make `T` a type of interest
```

eernstg: Same below.

TypeSystem/flow-analysis/reachability_switch_expression_A04_t03.dart
```diff
@@ -0,0 +1,50 @@
...
+/// - Let `before(E) = split(before(N))`.
+/// - For each case `Cj` in `alternatives` with expression `Ej`, for `j` in
+///   `1 .. k`:
+///   - Let `before(Ej) = after(E)`.
+/// - Let `after(N) = join(after(E1), after(E2), .. after(Ek))`.
+/// TODO (sgrekhov): there is no switch expression in the flow analysis spec yet
+/// (April 2025). Update the assertion after spec update.
+///
+/// @description Checks that if a type `T` is made a type of interest in
+/// `alternatives` then it cannot be promoted in other `alternatives`.
```
eernstg, line 16: How can this be phrased? It is not possible for one alternative to _prevent_ promotion of a variable `s` in other alternatives unless it is because it makes `s` non-promotable entirely (`() { s = something; })`). What we _can_ say is that making `T` a type of interest for a given variable in one alternative does not make `T` a type of interest for that variable in any other cases.

Perhaps:
```suggestion
/// @description Checks that if a type `T` is made a type of interest for `s`
/// in an alternative then there is no guarantee that `T` is of interest for `s`
/// in any other alternative.
```

eernstg (commented):
Looks good! But we definitely need some tests where one or more cases are labelled, because the labelled cases are treated differently.

TypeSystem/flow-analysis/reachability_switch_expression_A04_t03.dart
sgrekhov: Fixed.

sgrekhov: Fixed.

TypeSystem/flow-analysis/reachability_switch_expression_A04_t03.dart
sgrekhov: Thank you. Updated.

sgrekhov (commented):
Thank you! Updated. Labelled cases are added. PTAL.

PR #3154, opened 2025-04-21
#3057. Add promotion tests for for-in loop

sgrekhov:


TypeSystem/flow-analysis/reachability_for_in_A04_t06.dart
```diff
@@ -0,0 +1,53 @@
...
+  ];
+}
+
+test3() {
+  S s = S();
+  [
+    for (var j in [s is T ? 1 : 2]) 0,
+    s = T(),
+    s.answer()
+  ];
```
eernstg, line 46: Perhaps intended to be like this?:
```suggestion
  ];
  s = T();
  s.answer();
```
Just because this would correspond more closely to the last test of other libraries in this PR.

eernstg: Well, the opposite might just as well be argued: "It's the other tests that should be like this one".

The question arises "should we test the situation where the last promotion occurs "a bit later" (there general approach is otherwise to put the code immediately after `N` when testing anything about `after(N)`), but that's an endless tarpit. So it's probably best to keep the promotion testing code immediately after `N` in order to test something about `after(N)`.

TypeSystem/flow-analysis/reachability_for_in_A04_t09.dart
```diff
@@ -0,0 +1,104 @@
...
+    for (var j in [])
+      [s is T ? 1 : 2, s = T(), s.answer()]
+  };
+}
+
+test4() {
+  S s = S();
+  <int, List<Object>>{
+    for (var j in [])
+      s is T ? 1 : 2 : [s = T(), s.answer()]
```
eernstg, line 53: Nice colons! ;-)

TypeSystem/flow-analysis/reachability_for_in_A04_t04.dart
```diff
@@ -0,0 +1,65 @@
...
+}
+
+test4() {
+  S s = S();
+  if (s is T) {}
+  <int, int>{
+    for (var j in []) 0: 0
+  };
+  s = T();
+  s.answer();
```
eernstg, line 57: This is an example of the placement that I mentioned in the comment on TypeSystem/flow-analysis/reachability_for_in_A04_t06.dart: The promotion testing expressions could be inside the map literal such that they occur immediately after the given `N`.

eernstg (commented):
Looks good! Just one tiny inconsistency: I think a couple of tests would be more strictly consistent if a couple of promotion testing expressions are moved into a collection literal.

TypeSystem/flow-analysis/reachability_for_in_A04_t06.dart
sgrekhov: Moved outside of the `List` literal as it is done for `Set` and `Map` literals. PTAL.

eernstg: Oh, I'm afraid I've moved in the direction of the opposite choice:

We can't make it a principle that we're testing the specific static analysis information changes along a path in the control flow graph (such as promoting a variable or making the location unreachable) _at some point one or more steps after the change is specified to occur_. This would explode, and it is very likely to be useless (we aren't going to catch any extra bugs).

Conversely, the principle that we _can_ have is that such tests should be located in the tightest possible location relative to the specified change. In this case it means testing that the promotion can be performed in code that follows immediately after the element (inside the list literal, rather than after the list literal and in the next statement).

So, I'm sorry about pushing on this multiple times, but I think it's useful to look for principles like this one ("test flow analysis effects as soon as they are specified to have occurred, not one or more steps later") and applying them consistently. (Even though we will surely get it wrong now and then, I think it's a good habit to aim for a tight style in this sense).

eernstg (commented):
Just being stubborn. ;-)

sgrekhov:
No problem! :) Updated.

eernstg (approved):
Thanks! LGTM

PR #3156, opened 2025-04-22
#3057. Add promotion tests for `if` and `break` statements

sgrekhov:


TypeSystem/flow-analysis/reachability_break_A05_t01.dart
```diff
@@ -0,0 +1,85 @@
...
+    s.answer();
+  }
+}
+
+test2() {
+  S s = S();
+  for (var j in []) {
+    if (s is T) {}
+    break;
+    s = T();
```
eernstg, line 39: As usual, I think it would be nice to keep track of test cases that are concerned with the properties of dead code. Perhaps:

```suggestion
    // Unreachable, but does support promotion.
    s = T();
```

It should be fine if there's one such comment in each of the tests that are dealing with unreachable code.

TypeSystem/flow-analysis/reachability_conditional_A05_t03.dart
```diff
@@ -0,0 +1,165 @@
...
+  } else {
+    s is T ? 1 : 2;
+  }
+  s = T();
+  s.answer();
+}
+
+test14() {
+  S s = S();
+  [if (1 > 2) 0 else s is T ? 1 : 2];
```
eernstg, line 129: Again, for consistency (and in order to get the tightest possible test), it would be good to have the promotion related expressions inside the collection literal, immediately after the `if` element.

eernstg (commented):
Looks good, with a few comments!

As usual, I'm worried that tests where significant code is dead are too confusing for future readers (they might very well think "OK, this is wrong, that's dead code!" and then proceed to delete or change the test). So it would be nice to have just one comment in the first of several cases in each of the tests that are testing the treatment of dead code.

sgrekhov:
Thank you. Updated. PTAL.

TypeSystem/flow-analysis/reachability_break_A05_t01.dart
eernstg: Thanks!

TypeSystem/flow-analysis/reachability_conditional_A05_t03.dart
eernstg: Thanks!

eernstg (approved):
LGTM, thanks!

PR #3158, opened 2025-04-24
#3057. Add return statement in loops tests

sgrekhov:


TypeSystem/flow-analysis/reachability_while_A03_t01.dart
```diff
@@ -8,11 +8,11 @@
 /// - Let `before(S) = split(true(E))`.
 /// - Let `after(N) = inheritTested(join(false(E), unsplit(break(S))), after(S))`
 ///
-/// @description Checks that if `E` is not a boolean literal flow analysis
-/// detects assignment in `S`.
+/// @description Checks that if `E` is a boolean `true` literal and there is no
+/// reachable `break` or `return` in `S` then `after(N) is unreachable.
```
eernstg, line 12: I don't think a `return` statement in `S` could make `after(N)` reachable. How would that happen?

TypeSystem/flow-analysis/reachability_while_A03_t02.dart
```diff
@@ -0,0 +1,27 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion while statement: If `N` is a while statement of the form
+/// `while (E) S` then:
+/// - Let `before(E) = conservativeJoin(before(N), assignedIn(N), capturedIn(N))`
+/// - Let `before(S) = split(true(E))`.
+/// - Let `after(N) = inheritTested(join(false(E), unsplit(break(S))), after(S))`
+///
+/// @description Checks that if `E` is a boolean `true` literal and there is no
+/// reachable `break` or `return` in `S` then `after(N)` is unreachable.
```
eernstg, line 12: Same consideration here.

eernstg (commented):
Looks good! Just a couple of comments about the ability (or not) of a `return` statement to make something reachable.

sgrekhov:
Right. Thank you! Removed mentions of `return` statement from there. You may ask why then `return` is left in the code? Let's check that FA doesn't treat it as a `break`. PTAL.

eernstg (commented):
LGTM

eernstg (approved):

Issue #3122, opened 2025-04-01
[Dot Shorthands] Bugs in static-access-shorthand tests


kallentu:
I was going through a bunch of the co19 tests while doing CFE implementation and came across these.
I believe we need an update on these tests.

cc. @eernstg to confirm for some of the spec related changes. I might also be wrong, so it'd be nice to get a double-check.

https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Static-access-shorthand/type_inference_A01_t02.dart
**Problem**: This errors even without a shorthand:
```
tests/co19/src/LanguageFeatures/Static-access-shorthand/type_inference_A01_t02.dart:162:40: Error: The getter 'value' isn't defined for the class 'ET<dynamic>'.
Try correcting the name to the name of an existing getter, or defining a getter or field named 'value'.
  Expect.equals(3, (await (await et9)).value);
```

https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Static-access-shorthand/type_inference_A01_t01.dart
**Problem**: `e1` doesn't exist for `E`.
```
tests/co19/src/LanguageFeatures/Static-access-shorthand/type_inference_A01_t01.dart:85:19: Error: Member not found: 'e1'.
  Expect.equals(E.e1, e1);
```

https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Static-access-shorthand/semantics_A05_t01.dart
**Problem**: The LHS context is now `C` for example and `C` doesn’t have a `.new`. I think only `Object` as the context would work here for constructor tearoffs.
```
tests/co19/src/LanguageFeatures/Static-access-shorthand/semantics_A05_t01.dart:44:10: Error: The static getter or field 'new' isn't defined for the type 'C<int>'.
 - 'C' is from 'tests/co19/src/LanguageFeatures/Static-access-shorthand/semantics_A05_t01.dart'.
Try correcting the name to the name of an existing static getter or field, or defining a getter or field named 'new'.
    o = .new;
         ^^^
tests/co19/src/LanguageFeatures/Static-access-shorthand/semantics_A05_t01.dart:51:10: Error: The static getter or field 'id' isn't defined for the type 'C<int>'.
 - 'C' is from 'tests/co19/src/LanguageFeatures/Static-access-shorthand/semantics_A05_t01.dart'.
Try correcting the name to the name of an existing static getter or field, or defining a getter or field named 'id'.
    o = .id;
```

[...]

PR #3159, opened 2025-04-25
#3122. Expect an error in case of non-constant in a constant expression

sgrekhov:
The previous change in this test was wrong.  `static M one = const MC(1);` is not a potentially constant and therefore cannot be used in a constant expression.


eernstg (approved):
LGTM

PR #3160, opened 2025-04-25
#3057. Update promotion via assignment tests

sgrekhov:


TypeSystem/flow-analysis/promotion_via_assignment_A03_t02.dart
```diff
@@ -0,0 +1,26 @@
...
+/// - and `T` is a type of interest for `x` in `tested`
+///
+/// @description Checks that if `T` is not subtype of `S` then promotion via
+/// assignment cannot be performed
+/// @author sgrekhov22@gmail.com
+
+main() {
+  String s = "Some string";
+  if (s is int) {} // Make `int` a type of interest for `s`
+  s = 42 as dynamic;
```
eernstg, line 21: I think we aren't testing precisely what the description says. Here, `s` has type `String` (and it hasn't been promoted in any way), and we're assigning a new value of type `dynamic`. So there will be a dynamic type check at run time (if we were to get that far), checking that `42` is indeed a `String` (and then the program incurs a run-time failure).

Statically, we just declare that the assignment is fine, because we know that it will maintain soundness at run time, either by succeeding or by throwing.

However, the next statement is a compile-time error because `String` has no member named `isEven`.

So we're never even close to a promotion, which makes the description somewhat misleading.

How about letting `s` start as `Object?`, initializing it with the same value as now, and then promoting `s` to `String` using `s as String;` or something like that.

Next, make `int` a type of interest, such that assignment can promote at all.

Next, assign `s` using `s = 42;` (which would be OK and demote `s`).

Now we can say that promotion _could_ have taken place (yielding an `s` whose type is `int`), but it _doesn't_ (because of the requirement `T <: S` that this is all about). Instead, it demotes `s` to an earlier type (namely the declared type, because `String` isn't a supertype of `int`).

eernstg (commented):
Looks good! Just one test seems to test something which isn't quite what the description says.

TypeSystem/flow-analysis/promotion_via_assignment_A03_t02.dart
sgrekhov: Thank you for the a very good advice! I rewrote the tests as yuo've suggested. But.
```dart
main() {
  Object? s = "Some string";
  s as String; // `s` promoted to `String`
  if (s is int) {} // Make `int` a type of interest for `s`.
  s = 42; // `s` demoted to `Object?`
  s.isEven; // No this error!
//  ^^^^^^
// [analyzer] unspecified
// [cfe] unspecified
  s.substring(0); // Ok
//  ^^^^^^^^^
// [analyzer] unspecified
// [cfe] unspecified
}
```
Seems that `s = 42` didn't demote `s` but promoted it to `int` (or demoted to `Object?` and then promoted to `int`). Looks like an isse. Please confirm.

eernstg: Thanks!

I think 'No this error!' means that no error is reported at that line, which would match what I can see in DartPad. I do think this means that `s` is "demoted from `String` to `int`" by the assignment in the previous line.

The definition of `assign(x, E, M)` [here](https://github.com/dart-lang/language/blob/main/resources/type-system/flow-analysis.md#promotion) would then have to proceed as follows: 

- Reaching the case "`x` is demotable".
- `promoted` is `[String, Object?]`, `S` is `Object?`, `previous` is `[Object?]` (because `int <: Object?` but `int <: String` is not true). Hence, `Q` is `Object`.
- So `demoted` is `[Object, Object?]` and the new type of the variable is `Object`.

With that, `s.isEven` and `s.substring(0)` should both be errors.

So that's an issue.

sgrekhov:
Thank you! Updated. Also updated `promotion_via_assignment_A03_t01.dart` to follow the assertion. Please review.

TypeSystem/flow-analysis/promotion_via_assignment_A03_t01.dart
```diff
@@ -11,23 +11,31 @@
...
 import '../../Utils/static_type_helper.dart';
 
-class S {
-  num answer() => 42.42;
+test1(FutureOr<Object> v) {
+  if (v is Object) {} // ignore: unnecessary_type_check
+  v = Object();
+  // `Object` <: `FutureOr<Object>` and `FutureOr<Object>` <: `Object`.
+  // Therefore `v` is not promoted to `Object`.
+  v.expectStaticType<Exactly<FutureOr<Object>>>();
```
eernstg, line 26: OK, nice!

eernstg (approved):
LGTM

PR #3161, opened 2025-04-25
#3057. Add type of interest tests.

sgrekhov:


TypeSystem/flow-analysis/type_of_interest_A01_t03.dart
```diff
@@ -0,0 +1,40 @@
...
+/// @author sgrekhov22@gmail.com
+
+class S {}
+
+class T extends S {
+  int foo() => 42;
+}
+
+test1(S? s) {
+  if (s is T?) {}  // make `T` a type of interest
```
eernstg, line 19: I think this might help a future reader:
```suggestion
  if (s is T?) {}  // make `T?` and `T` types of interest
```

TypeSystem/flow-analysis/type_of_interest_A01_t04.dart
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We say that at type `T` is a type of interest for a variable `x`
+/// in a set of tested types `tested` if `tested` contains a type `S` such that
+/// `T` is `S`, or `T` is `NonNull(S)`.
+///
+/// @description Checks that if `T` is a type of interest for a variable `x`,
+/// then it doesn't make `T?` a type of interest for `x`.
```
eernstg, line 10: Yep, we just had that case recently. Good to have a test about this here!

TypeSystem/flow-analysis/type_of_interest_A03_t01.dart
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We say that at type `T` is a type of interest for a variable `x`
+/// in a set of tested types `tested` if `tested` contains a type `S` such that
+/// `T` is `S`, or `T` is `NonNull(S)`.
+///
+/// @description Checks that a type `void` cannot be made a type of interest for
+/// a variable.
```
eernstg, line 10: I think the variable has to have declared type `Object?`, or some other top type. Otherwise it could be allowed to make `void` a type of interest, but it should not make a difference because types of interest that aren't subtypes of the declared type of the variable can never have an effect on anything.

We shouldn't be able to promote the variable to have type `void`, because that isn't a _proper_ subtype of the type of the variable (promoted or not, it can't be a proper supertype of a top type).

Nevertheless, it does make sense to see that `if (s is Void) { /*..'s' could be `void` here..*/ }`, and the result should show that it does not have the type `void`, so (e.g.) `s.substring(0)` is OK.

TypeSystem/flow-analysis/type_of_interest_A03_t02.dart
```diff
@@ -0,0 +1,36 @@
...
+/// for a variable.
+/// @author sgrekhov22@gmail.com
+/// @issue 60620
+
+import 'dart:async';
+
+test1() {
+  String s = "a";
+  if (s is dynamic) {}  // Try to make `dynamic` a type of interest
+  s = "b";
```
eernstg, line 19: This shouldn't do anything to the type of `s`, it already has type `String`, and assigning an expression of static type `String` could never change that.

eernstg: I think the situation is similar: The variable must have a declared type which is a top type, such that `dynamic` is at least a subtype of the declared type (though not a proper subtype). We can then have `if (s is dynamic) {/*..'s' could then have type `dynamic` here..*/}`, but `s` should still have the declared type.

eernstg: This change might destroy the purpose of the test (because it's not about a type of interest that enables a promotion via assignment any more). If it gets too awkward then perhaps the test should just be deleted, or perhaps it fits into some other series of tests (and isn't a duplicate), in which case it should probably be renamed.

TypeSystem/flow-analysis/type_of_interest_A03_t03.dart
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We say that at type `T` is a type of interest for a variable `x`
+/// in a set of tested types `tested` if `tested` contains a type `S` such that
+/// `T` is `S`, or `T` is `NonNull(S)`.
+///
+/// @description Checks that a type `Object` cannot be made a type of interest
+/// for a variable.
```
eernstg, line 10: I don't see a reason why this should not be possible. There is nothing special about `Object` in this respect.

Of course, the declared type must be a supertype of `Object` (that is, a top type) in order to allow the promotion to `Object` to take place.

TypeSystem/flow-analysis/type_of_interest_A03_t04.dart
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We say that at type `T` is a type of interest for a variable `x`
+/// in a set of tested types `tested` if `tested` contains a type `S` such that
+/// `T` is `S`, or `T` is `NonNull(S)`.
+///
+/// @description Checks that a type `Null` cannot be made a type of interest
+/// for a variable.
```
eernstg, line 10: This one is more tricky. I actually think it should be possible to promote to `Null` (e.g., from `T?` for some `T`) when the construct that does it uses the type `Null` explicitly (so `s == null` won't do it).

When the variable has been promoted, we could use `expectStaticType<..>()` to verify that this  promotion has indeed occurred.

TypeSystem/flow-analysis/type_of_interest_A03_t04.dart
```diff
@@ -0,0 +1,39 @@
...
+
+import 'dart:async';
+
+extension on Null {
+  void foo() {}
+}
+
+test1() {
+  String s = "a";
+  if (s is Null) {}  // Try to make `Null` a type of interest
```
eernstg, line 21: This won't work because `Null` is not a subtype of the declared type of `s`.

eernstg: It should work to change the declared type of `s` to `String?` (and perhaps cheat the static analysis to think that it can't be promoted to `String` immediately at the declaration).

TypeSystem/flow-analysis/type_of_interest_A03_t04.dart
```diff
@@ -0,0 +1,39 @@
...
+  s = "b";
```
eernstg, line 22: This would in any case eliminate all promotions, if any promotions did occur. I think that would prevent us from learning anything about said promotions.

TypeSystem/flow-analysis/type_of_interest_A03_t04.dart
```diff
@@ -0,0 +1,39 @@
...
+  s.foo();
```
eernstg, line 23: This is the place where a member invocation won't suffice (because `Null` does not have any members that aren't shared by _all_ objects).

But `expectStaticType` should work.

TypeSystem/flow-analysis/type_of_interest_A03_t05.dart
```diff
@@ -0,0 +1,32 @@
...
+/// @description Checks that a type `Never` cannot be made a type of interest
+/// for a variable.
+/// @author sgrekhov22@gmail.com
+
+import 'dart:async';
+
+test1() {
+  String s = "a";
+  if (s is Never) {}  // Try to make `Never` a type of interest
+  s = "b";
```
eernstg, line 18: Could use `s = throw 0;`, I think this should promote `s` to have type `Never` in subsequent code, which will be unreachable but still subject to flow analysis as if we could reach it.

TypeSystem/flow-analysis/type_of_interest_A03_t05.dart
```diff
@@ -0,0 +1,32 @@
...
+  s.substring(0);
+}
+
+test2() {
+  String s = "a";
+  if (s is FutureOr<Never>) {}
```
eernstg, line 24: This would again require that `s` has a much more general declared type. `Object` should work.

eernstg (commented):
Looks good. Added several comments.

TypeSystem/flow-analysis/type_of_interest_A03_t01.dart
sgrekhov: Removed top-types tests from this PR. I'll add them as promotion via assigment and promotion via type tests. And they will test type of interest as well.

sgrekhov (commented):
Updated. Removed top-types tests from this PR. I'll add them later as promotion tests.

eernstg (commented):
Here's a review which is only submitted because that action might finally take the PR out of the "Processing updates" state.

TypeSystem/flow-analysis/type_of_interest_A03_t01.dart
eernstg: OK.

eernstg (approved):
LGTM

PR #3162, opened 2025-04-28
#3057. Add null-aware method invocation tests

sgrekhov:


TypeSystem/flow-analysis/reachability_A24_t04.dart
```diff
@@ -0,0 +1,47 @@
...
+}
+
+test1(C? c) {
+  try {
+    late int i;
+    if (c == null) {  // This doesn't promote `c` to `Null`
+      c?.m1();
+      i = 42;
+    }
+    i; // Possibly assigned
```
eernstg, line 25: Perhaps this test is a little bit loose?

Intuitively, `i` can be 'possibly assigned' at `i;` because `c` could be null, and then `c?.m1()` would be a no-op, and `i = 42` would be executed.

We'd be analyzing `c?.m1()` under the assumption that `c` has type `C?`, so it might be a no-op and it might throw (as opposed to: `c` has type `Null` and `c?.m1()` is definitely a no-op). So the `if` statement body will either throw or it will definitely initialize `i`.

Doesn't this mean that we can put `i = 42` into the else branch and achieve a situation where `i` will be definitely assigned after the `if` statement, and we can remove `late`?

Or perhaps this just turns the test into a duplicate of another test?

eernstg (commented):
Looks good! I think it's worth taking one more look at one test which might be a little too permissive.

TypeSystem/flow-analysis/reachability_A24_t04.dart
sgrekhov: Thank you! No, the change won't make this test a duplicate of another one. Updated. PTAL.

TypeSystem/flow-analysis/reachability_A24_t04.dart
```diff
@@ -0,0 +1,45 @@
...
+
+test1(C? c) {
+  int i;
+  if (c == null) {  // This doesn't promote `c` to `Null`
+    c?.m1();
+    i = 42;
+  } else {
+    i = 42;
+  }
+  i; // Definitely assigned
```
eernstg, line 26: Oh, I'm afraid my cunning plan isn't working. ;-)

First, I'm assuming that the test `c == null` is the core point for this test (because of other tests having various other tests, including `c is Null`). We want to see that the decisions made by the flow analysis based on this test are appropriate, and the special point is that we do _not_ use the fact that `c` is null to conclude anything, it is as if we hadn't tested `c == null` in the first place.

Isn't this true?

If so, we're supposed to verify that `c?.m1()` is considered to have two possible outcomes: It is a no-op (because `c` is null), or it throws (because `c` is non-null), because `c` hasn't been promoted by the null test (we only narrow the type when the test shows that it can be `C` rather than `C?`, not when it shows that the type could be `Null` rather than `C?`). It shouldn't have just one outcome, saying that it is a no-op.

But the current version of this test can't detect this, I think.

The test could perfectly well be run by an implementation whose analysis says "`c?.m1()` is definitely a no-op because `c` is `Null`, so `i = 42` is definitely executed if we reach `i;` from the "then" branch, and similarly for the `else` branch, so `i` is definitely assigned in line 26.

I'm almost thinking "But we have that test already, it's TypeSystem/flow-analysis/reachability_A24_t01.dart". ;-)

I can't really find a way to adjust this test such that it actually shows anything significant beyond what is already verified in A24_t01 (except for turning it into a duplicate of that test). Perhaps I just misread it all the time?

eernstg (commented):
Sorry about pushing this one more time, I have difficulties seeing how we can conclude anything new from TypeSystem/flow-analysis/reachability_A24_t04.dart. Perhaps I just need to understand better what it is intended to show.

TypeSystem/flow-analysis/reachability_A24_t04.dart
sgrekhov: > Oh, I'm afraid my cunning plan isn't working. ;-)
> 
> First, I'm assuming that the test `c == null` is the core point for this test (because of other tests having various other tests, including `c is Null`). We want to see that the decisions made by the flow analysis based on this test are appropriate, and the special point is that we do _not_ use the fact that `c` is null to conclude anything, it is as if we hadn't tested `c == null` in the first place.
> 
> Isn't this true?

Yes! In `reachability_A24_t03.dart` we are testing the case `if (c != null) { c?.m();}` and in `reachability_A24_t04.dart` check the case `if (c == null) {c?.m();}`.

> 
> If so, we're supposed to verify that `c?.m1()` is considered to have two possible outcomes: It is a no-op (because `c` is null), or it throws (because `c` is non-null), because `c` hasn't been promoted by the null test (we only narrow the type when the test shows that it can be `C` rather than `C?`, not when it shows that the type could be `Null` rather than `C?`). It shouldn't have just one outcome, saying that it is a no-op.
> 
> But the current version of this test can't detect this, I think.
> 
> The test could perfectly well be run by an implementation whose analysis says "`c?.m1()` is definitely a no-op because `c` is `Null`, so `i = 42` is definitely executed if we reach `i;` from the "then" branch, and similarly for the `else` branch, so `i` is definitely assigned in line 26.
> 
> I'm almost thinking "But we have that test already, it's TypeSystem/flow-analysis/reachability_A24_t01.dart". ;-)
> 
> I can't really find a way to adjust this test such that it actually shows anything significant beyond what is already verified in A24_t01 (except for turning it into a duplicate of that test). Perhaps I just misread it all the time?

What I had in mind is that in `reachability_A24_t01.dart` and `reachability_A24_t04.dart` we have different type of expression `c?.m()`. In the first case we have expression of type `void` or no-op. If there are any code after this expression it will be definitely executed.

In the second case we have expression of type `Never` or no-op, and there is no guarantee that `i = 42;` after this expression will be executed. But `if (c == null)...` and `?.` operator should give us this guarantee. In other words we are checking that FA doesn't theat code after `c?.m()` as unreachable (as would be if we didn't use `?.` operator).

eernstg: OK, very good! The current version of the test does not make a distinction between the case where `c?.m1()` has one outcome (e.g., "it is guaranteed to throw", or "it is guaranteed to be a no-op") and the case where it has two outcomes (that is, "it might throw or it might be a no-op").

We want the latter and not the former. However, I can't see a way to actually make that distinction.

So let's land it!

eernstg (approved):
LGTM

PR #3163, opened 2025-04-29
#3057. Remove obsolete `promotion_via_initialization_*` tests

sgrekhov:


TypeSystem/flow-analysis/promotion_via_assignment_A05_t01.dart
```diff
@@ -0,0 +1,36 @@
...
+/// @assertion We say that a variable `x` is promotable via assignment of an
+/// expression of type `T` given variable model `VM` if
+/// - `VM = VariableModel(declared, promoted, tested, assigned, unassigned,
+///     captured)`
+/// - and captured is false
+/// - and `S` is the current type of `x` in `VM`
+/// - and `T <: S` and not `S <: T`
+/// - and `T` is a type of interest for `x` in `tested`
+///
+/// @description Checks that if a type of variable is not specified and there is
```
eernstg, line 14: Typo:
```suggestion
/// @description Checks that if the type of a variable is not specified and there is
```

TypeSystem/flow-analysis/promotion_via_assignment_A05_t01.dart
```diff
@@ -0,0 +1,36 @@
...
+/// no initializer then its type is `dynamic` in spite of assignment.
```
eernstg, line 15: I don't think there is anything special about `dynamic` here. We used to have a rule saying that a variable of type `dynamic` is never promoted (the idea is that this would _take away_ member invocations that you'd be allowed to perform, and promotion should never eliminate any allowed constructs), but we stopped having that rule.

This matches the behavior that I can see, too: A variable of type `dynamic` can be promoted by a type test, just like a variable of any other type (`dynamic d = 1; if (d is int) ...`).

TypeSystem/flow-analysis/promotion_via_assignment_A05_t01.dart
```diff
@@ -0,0 +1,36 @@
...
+}
+
+main() {
+  var x;
+  x = S();
+  if (x is T) {} // Make `T` a type of interest
+  x = T();
+  x.foo();
+  Expect.throws(
+    () {x.bar();},  // Check that `x` is dynamic
```
eernstg, line 33: I think `x` has been promoted to type `T` in the body of `main`. However, this occurrence of `x` is in a function literal, which means that it cannot be promoted (because flow analysis doesn't make any attempts to understand exactly at which locations in code this function literal could be executed). So `x` isn't promoted to type `T` in the body of the function literal, it's still of type `dynamic`.

This means that the test in the function literal says nothing which can be connected to "in spite of assignment".

Perhaps this is an updated version of an old test, written way back when a `dynamic` variable was actually never promoted?

TypeSystem/flow-analysis/promotion_via_assignment_A05_t02.dart
```diff
@@ -0,0 +1,33 @@
...
+/// expression of type `T` given variable model `VM` if
+/// - `VM = VariableModel(declared, promoted, tested, assigned, unassigned,
+///     captured)`
+/// - and captured is false
+/// - and `S` is the current type of `x` in `VM`
+/// - and `T <: S` and not `S <: T`
+/// - and `T` is a type of interest for `x` in `tested`
+///
+/// @description Checks that if a type of variable is not specified and there is
+/// an initializer then its type is defined by the initializer.
```
eernstg, line 15: Perhaps more direct:
```suggestion
/// an initializer then its type is inferred from the initializing expression.
```

eernstg (commented):
Looks good! One test needs to be discussed a bit more, though: TypeSystem/flow-analysis/promotion_via_assignment_A05_t01.dart.

TypeSystem/flow-analysis/promotion_via_assignment_A05_t01.dart
sgrekhov: Yes, exactly! This is an updated version of the old test. Fixed. PTAL.

TypeSystem/flow-analysis/promotion_via_assignment_A05_t01.dart
eernstg: Thanks!

TypeSystem/flow-analysis/promotion_via_assignment_A05_t01.dart
eernstg: Thanks!

eernstg (commented):
LGTM

eernstg (approved):

PR #3165, opened 2025-05-01
#3057. Add promotion via type test tests. Check top and bottom types and Null

sgrekhov:


TypeSystem/flow-analysis/promotion_via_type_test_A03_t07.dart
```diff
@@ -0,0 +1,26 @@
...
+/// @description Checks that a variable is promoted if `T` is a subtype of `S`.
+/// Test type `Object?` as `T`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+main() {
+  dynamic d = "a";
+  if (d is Object?) { // ignore: unnecessary_type_check
+    d.expectStaticType<Exactly<Object?>>();
```
eernstg, line 24: We can't promote from `dynamic` to `Object?`, but this test won't tell us that `d` still has the type `dynamic` _before run time_ (it will throw because there is no such method).

Perhaps use `if (1 > 2) d.testDynamic;` instead?

TypeSystem/flow-analysis/promotion_via_type_test_A03_t09.dart
```diff
@@ -0,0 +1,26 @@
...
+///
+/// @description Checks that a variable is promoted if `T` is a subtype of `S`.
+/// Test type `dynamic` as `T`.
+/// @author sgrekhov22@gmail.com
+
+main() {
+  Object? o = 1 > 2 ? null : "a";
+  if (o is dynamic) { // ignore: unnecessary_type_check
+    try {
+      o.whatever;
```
eernstg, line 23: I think we can just expect an error here: There are no other tests in this library, so it's not a problem if we change it from being a "compiles and runs" test into an "encounters compile-time errors" test.

(Or would that imply that the test should have a different name because error tests and non-error tests can somehow be recognized by their names? In that case I guess it could just be renamed).

TypeSystem/flow-analysis/promotion_via_type_test_A03_t10.dart
```diff
@@ -0,0 +1,26 @@
...
+/// @description Checks that a variable is not promotable if `T` is not subtype
+/// of `S`. Test type `Null` as `T`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+main() {
+  String s = "a";
+  if (s is Null) { // ignore: type_check_with_null
+    s.expectStaticType<Exactly<String>>();
```
eernstg, line 24: Yes, that's nice!

This is nowhere near as tricky as the promotion-via-assignment tests.

TypeSystem/flow-analysis/promotion_via_type_test_A03_t12.dart
```diff
@@ -0,0 +1,27 @@
...
+/// - and `captured` is `false`
+/// - and `S` is the current type of `x` in `VM`
+/// - and not `S <: T`
+/// - and `T <: S` or (`S` is `X extends R` and `T <: R`) or (`S` is `X & R` and
+///   `T <: R`)
+///
+/// @description Checks that a variable is not promoted if the type `T` is
+/// `Never`.
+/// @author sgrekhov22@gmail.com
+/// @issue 60657
```
eernstg, line 18: It is _very_ helpful to have this comment!

TypeSystem/flow-analysis/promotion_via_type_test_A04_t01.dart
```diff
@@ -0,0 +1,55 @@
...
+/// `T` given variable model `VM` if
+/// - `VM = VariableModel(declared, promoted, tested, assigned, unassigned,
+///     captured)`
+/// - and `captured` is `false`
+/// - and `S` is the current type of `x` in `VM`
+/// - and not `S <: T`
+/// - and `T <: S` or (`S` is `X extends R` and `T <: R`) or (`S` is `X & R` and
+///   `T <: R`)
+///
+/// @description Checks that a variable is promoted after the type test in one
```
eernstg, line 15: Typo:
```suggestion
/// @description Checks that if a variable is promoted after the type test in one
```

eernstg: I think the same phrase occurs in all the tests following this one.

TypeSystem/flow-analysis/promotion_via_type_test_A04_t02.dart
```diff
@@ -0,0 +1,47 @@
...
+///     captured)`
+/// - and `captured` is `false`
+/// - and `S` is the current type of `x` in `VM`
+/// - and not `S <: T`
+/// - and `T <: S` or (`S` is `X extends R` and `T <: R`) or (`S` is `X & R` and
+///   `T <: R`)
+///
+/// @description Checks that a variable is promoted after the type test in one
+/// of the branches, then is is not promoted in another branch and after it.
+/// Test type `void`.
```
eernstg, line 17: I'm afraid I'd recommend that this particular test is deleted.

The point is that a variable can never be promoted to have type `void` (because it would then have to have at least a declared type which is a proper supertype of `void`, and there is no such type). So we can't even get started investigating how a promotion to `void` would work, because they don't exist.

TypeSystem/flow-analysis/promotion_via_type_test_A04_t03.dart
```diff
@@ -0,0 +1,44 @@
...
+///     captured)`
+/// - and `captured` is `false`
+/// - and `S` is the current type of `x` in `VM`
+/// - and not `S <: T`
+/// - and `T <: S` or (`S` is `X extends R` and `T <: R`) or (`S` is `X & R` and
+///   `T <: R`)
+///
+/// @description Checks that a variable is promoted after the type test in one
+/// of the branches, then is is not promoted in another branch and after it.
+/// Test type `dynamic`.
```
eernstg, line 17: Same here.

TypeSystem/flow-analysis/promotion_via_type_test_A04_t04.dart
```diff
@@ -0,0 +1,54 @@
...
+///     captured)`
+/// - and `captured` is `false`
+/// - and `S` is the current type of `x` in `VM`
+/// - and not `S <: T`
+/// - and `T <: S` or (`S` is `X extends R` and `T <: R`) or (`S` is `X & R` and
+///   `T <: R`)
+///
+/// @description Checks that a variable is promoted after the type test in one
+/// of the branches, then is is not promoted in another branch and after it.
+/// Test type `Object?`.
```
eernstg, line 17: And again same here.

TypeSystem/flow-analysis/promotion_via_type_test_A04_t06.dart
```diff
@@ -0,0 +1,44 @@
...
+/// of the branches, then is is not promoted in another branch and after it.
+/// Test type `Never`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+test1() {
+  String? s = 1 > 2 ? null : "a";
+  if (s is Never) {
+  } else {
```
eernstg, line 25: We do have another test (or several) about the outcome of testing for `Never`, but it might still be good for the overall consistency and readability of this test to add this to the "then" part:

  s.expectStaticType<Exactly<String?>>(); // Dead code and no promotion.

eernstg (commented):
Looks good! I'm suggesting that a couple of tests should be deleted, because they unfold some cases where the test is basically not possible to perform ("check how this promotion works, except that this promotion never occurs").

Also, a couple of typos.

TypeSystem/flow-analysis/promotion_via_type_test_A03_t04.dart
```diff
@@ -0,0 +1,57 @@
...
+  Object? s = "a"; // Promoted to `Object`. `void` is not subtype of `Object`
+  if (s is Void) { // ignore: unnecessary_type_check
+    s.expectStaticType<Exactly<Object>>();
+  }
+}
+
+test4() {
+  Object? s = 1 > 2 ? null : "a";
+  if (s is FutureOr<void>) { // ignore: unnecessary_type_check
+    s.expectStaticType<Exactly<Object?>>();
```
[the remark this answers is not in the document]
eernstg: The point was actually that line 48 won't be able to show whether the program analysis behaves as specified or not: `s` has a top type (`Object?`) from the declaration and is then tested against another top type, namely `FutureOr<void>`. This shouldn't change anything (in the true branch, or in the false branch) because we're testing a type which is not a proper subtype, it's just a subtype.

But even in the case where there is a bug and `s` is "promoted" to `FutureOr<void>`, we won't be able to detect that anything is wrong. The reason for this is that `expectStaticType` is blind to distinctions between types that are mutual subtypes (e.g., all top types), it just considers them all to be the same type (which is perfectly fine from a subtyping perspective).

So I was trying to say that line 48 would need to be adjusted in order to detect that something is wrong in the case where `s` gets the type `FutureOr<void>` (as I understand it, the purpose of this test is to show that this doesn't happen).

Here's the technique that I had in mind, which is used to get an error when the variable has type `FutureOr<void>` and _not_ getting an error when it has type `Object?`:

```dart
import 'dart:async';

void main() async {
  FutureOr<void> s = 1; // No error below if the type is `Object?`.
  if (s is Future<void>) {
    // `s` has type `Future<void>` here.
  } else {
    // `s` has type `void` here, based on factoring.
    s.hashCode; // Compile-time error.
  }
}
```

I was thinking that this `if` statement could basically replace the invocation of `expectStaticType`. So that's not an extra test case, but a fix which could be applied to the existing test case (because it needs a fix).

> But what about testing of the type `FutureOr<void>` against `Void`?

Right, you can do that, too. And I agree that the type of `s` in the `else` branch should have been `Future<void>`.

However, that isn't very useful because the `else` branch is dead code when the "then" branch is testing against a top type.

You could report it as an issue and then we'll see whether anyone thinks it is a bug. ;-)

eernstg (commented):
Just one location in TypeSystem/flow-analysis/promotion_via_type_test_A03_t04.dart has a comment.

TypeSystem/flow-analysis/promotion_via_type_test_A03_t09.dart
eernstg: Thanks!

TypeSystem/flow-analysis/promotion_via_type_test_A04_t06.dart
eernstg: Thanks!

TypeSystem/flow-analysis/promotion_via_type_test_A03_t04.dart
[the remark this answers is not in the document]
sgrekhov: Thank you for the detailed explanation! I got it wrong at first. Removed `test4` from this test and moved it to a separate tests (`promotion_via_type_test_A03_t13.dart`) which expects compile-time error on `s.hashCode`. Please take another look.

eernstg (commented):
Test t13 remains tricky because `factor` works in ways that may seem really confusing in special cases like top types.

TypeSystem/flow-analysis/promotion_via_type_test_A03_t13.dart
```diff
@@ -12,36 +12,39 @@
 /// - and `T <: S` or (`S` is `X extends R` and `T <: R`) or (`S` is `X & R` and
 ///   `T <: R`)
 ///
-/// @description Checks that a variable is promoted after the type test in one
-/// of the branches, then is is not promoted in another branch and after it.
-/// Test type `void`.
+/// @description Checks that a variable is not promotable if `T` is not subtype
+/// of `S`. Test type `FutureOr<void>`.
```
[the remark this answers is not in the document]
sgrekhov: Thank you! Updated. PTAL.

eernstg:
Perhaps the update was not pushed?

eernstg (approved):
LGTM, thanks!

eernstg:
No. Problem. ;-D

PR #3167, opened 2025-05-02
#3057. Add promotion via assignment tests. Check top types.

sgrekhov:


eernstg:
First (before I take a look at the actual tests), note this comment and the enclosing issue: https://github.com/dart-lang/sdk/issues/60609#issuecomment-2845954988.

This confirms that the treatment of top types as "the same type" vs. "different types" (that is, essentially, saying "No" or "Yes" to the use of MORETOP to make some top types different from each other) is a non-settled issue.

The comment from @stereotype441 indicates that we might end up with the approach where top types are just considered to be the same type, and hence there's no support for promoting from one top type to another.

I'd suggest that we proceed to make that assumption ("we can't promote to a top type"), and then we can make adjustments if the clarification (of the spec and the implementation) breaks something.

TypeSystem/flow-analysis/promotion_via_assignment_A03_t03.dart
```diff
@@ -0,0 +1,65 @@
...
+// [cfe] unspecified
+}
+
+test2() {
+  Object? v = 1 > 2 ? null : Object();
+  if (v is Void) {} // ignore: unnecessary_type_check
+  v = 1 > 2 ? null : Object(); // can't assign `void`, let's assign subtype of `void`
+  // `Object?` <: `void` and `void` <: `Object?`.
+  // Therefore `v` is not promoted to `void`.
+  v.expectStaticType<Exactly<Object?>>();
```
eernstg, line 41: That's a weird way to verify that the static type is `Object?` rather than `void`. ;-)

The point is that they are mutual subtypes, so the `expectStaticType` test would just say yes ("`v` has type `void` so we can confirm that it has type `Object?`!").

However, if `v` actually has type `void` then it is an error to invoke `expectStaticType` on it. So we _will_ get an error if `v` has type `void`, but for a reason which is not readable.

Perhaps just do `v.hashCode` in order to get the error if the type is `void`?

TypeSystem/flow-analysis/promotion_via_assignment_A03_t03.dart
```diff
@@ -0,0 +1,65 @@
...
+  dynamic d = 42;
+  if (d is Object?) {} // ignore: unnecessary_type_check
+  d = 1 > 2 ? null: Object();
+  d.testDynamic; // `d` should be `dynamic`
+}
+
+test4() {
+  dynamic d = 42;
+  if (d is Void) {} // ignore: unnecessary_type_check
+  d = Object();     // can't assign `void`, let's assign subtype of `void`
```
eernstg, line 54: I don't think this can work. We would want to perform an assignment to `v` in a situation where it does not have type `void` (or `Void`, shouldn't matter), and the assigned expression would have static type `void`. I can't see how that expression could be a non-error.

Assigning something which is a subtype should not change the type of `v` as long as it is not a type of interest.

This means that this test depends on the outcome of the discussion in https://github.com/dart-lang/sdk/issues/60622. So it would actually be blocked.

TypeSystem/flow-analysis/promotion_via_assignment_A03_t05.dart
```diff
@@ -0,0 +1,28 @@
...
+/// @issue 60620
+
+import '../../Utils/static_type_helper.dart';
+
+typedef Void = void;
+
+main() {
+  String s = "x";
+  if (s is Void) {} // ignore: unnecessary_type_check
+  s = "y"; // Cannot assign `void`, let's assign `String <: void`
```
eernstg, line 26: Same issue about being blocked.

TypeSystem/flow-analysis/promotion_via_assignment_A03_t06.dart
```diff
@@ -0,0 +1,26 @@
...
+/// assignment cannot be performed. Test type `Object?`;
+/// @author sgrekhov22@gmail.com
+/// @issue 60620
+
+import '../../Utils/static_type_helper.dart';
+
+main() {
+  String s = "x";
+  if (s is Object?) {} // ignore: unnecessary_type_check
+  s = "y"; // Cannot assign `Object?`, let's assign `String <: Object?`
```
eernstg, line 24: This is again not really doing what we need. `s = "y"` is just an assignment of an expression whose type is the declared type of the variable which hasn't been promoted. There's simply nothing going on about the type of `s` here, and there is certainly no reason why we would try to "promote" `s` to have type `Object?`.

Perhaps this test is just impossible? ;-)

TypeSystem/flow-analysis/promotion_via_assignment_A03_t07.dart
```diff
@@ -0,0 +1,25 @@
...
+/// @description Checks that if `T` is not subtype of `S` then promotion via
+/// assignment cannot be performed. Test type `Null`;
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+main() {
+  String s = "x";
+  if (s is Null) {} // ignore: unnecessary_type_check
+  s = "y";
```
eernstg, line 23: Again, I suspect that this test just can't be made using an assignment, we'd need to rely on a type query (that is, putting code into the body of the `if` statement) in order to determine whether or not the promotion to `Null` has occurred -- expecting that it hasn't.

TypeSystem/flow-analysis/promotion_via_assignment_A03_t08.dart
```diff
@@ -0,0 +1,25 @@
...
+/// @description Checks that if `T` is not subtype of `S` then promotion via
+/// assignment cannot be performed. Test type `Never`;
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+main() {
+  String s = "x";
+  if (s is Never) {} // ignore: unnecessary_type_check
+  s = "y";
```
eernstg, line 23: Same difficulty here, we really can't learn anything about "promotion via assignment" to `Never` by assigning a value of type `String`. ;-)

TypeSystem/flow-analysis/promotion_via_assignment_A06_t01.dart
```diff
@@ -0,0 +1,26 @@
...
+/// supertype of the declared type as type of interest.
+/// @author sgrekhov22@gmail.com
+/// @issue 60620, 60622
+
+import '../../Utils/static_type_helper.dart';
+
+main() {
+  String s = "a";
+  if (s is Pattern) {}
+  s = "";
```
eernstg, line 24: Same difficulty again.

eernstg (commented):
This one is tricky. I'm tempted to say that these tests just can't be written, because it's inherently not possible to obtain a "promotion via assignment" to a type which is not a subtype of the declared type.

It is possible to do something when the declared type is a top type, but even in those cases there are difficulties (like assigning `void` to a variable whose type is not currently `void`).

So what do you think? Do we need to simply delete some of these tests?

eernstg:
"Check top types" seems to imply that the types being tested are top types. However, in some cases the type is `String` and that seems to create a number of difficulties that may not have a reasonable solution.

TypeSystem/flow-analysis/promotion_via_assignment_A03_t07.dart
sgrekhov: It's again test for sdk#60620, sdk#60622. We have 5 special types, 'Object?`, `void`, `dynamic`, `Null` and `Never` (probably `FutureOr` is also can be added to this list). The first 3 are top types and if one of them is a type of interes for  a variable then any assignment to this variable demotes the type. The goal of this test was to check that if `Null` is among the types of interests for a variable, then an assignment doesn't demote its type to `Null`.  What if to rewrite it as 
```dart
/// @description Checks that if a variable `s` has a nullable type and `Null` is among its types of interest
/// then assertion to `s` doeesn't demote it to `Null`.

main() {
  String? s = 1 > 2 ? null : "x";
  if (s is Null) {} // ignore: unnecessary_type_check
  s = 1 > 2 ? null : "y";
  s.expectStaticType<Exactly<String?>>();
}
```

TypeSystem/flow-analysis/promotion_via_assignment_A03_t08.dart
sgrekhov: Let's delete this one :). Thank you!

TypeSystem/flow-analysis/promotion_via_assignment_A06_t01.dart
sgrekhov: And this is a test for sdk#60620, sdk#60622. May be to change the description to "Checks that if a variable `x` of type `T` such that `T <: S` has `S` as a type of interest, then assignment type `T` to `x` doesn't demote its type  `S`"?

sgrekhov (commented):
Let's mark this PR as blocked but first decide which tests should be deleted, which one updated (including the description). Please take another look.

TypeSystem/flow-analysis/promotion_via_assignment_A03_t07.dart
eernstg: Same as before, the promotion to `Null` will not occur because that's not even correct with respect to the declared type. This is also the reason why `s = null` can't be used because that's just another compile-time error. But then we're assigning something to `s` (namely a `String`) which is completely irrelevant to `Null`.

So if we want to test that something fails (here: `s` isn't promoted to have type `Null` via assignment) then it should fail for the reason that we're testing, not for one or more other reasons.

The current type of `s` can be `String`, that's fine (because it satisfies the requirement that `Null` isn't a subtype of `String`), but we need the declared type of `s` to allow for the assignment `s = null` (with `Null`, we can't assign an actual value whose type is a subtype, so it must be `s = e` where `e` has static type `Null`, so it might as well be `s = null`).

Then we have something like this:

```dart
main() {
  Object? s = "x";
  s as String;
  if (s is Null) {} // ignore: unnecessary_type_check
  s = null;
  s.expectStaticType<Exactly<Null>>();
}
```

This might look like a failure (because we _do_ confirm that the type of `s` is now `Null`), but the fact is that this is not because the 'promote' rule has been implemented incorrectly, it is because the situation is handled using the 'demote' rule, and this is the outcome.

It is still not a 'promotion' because the type changes from `String` to `Null`, and the latter is not a subtype of the former.

I think it makes sense to do it in this way, and adding a few words to the description saying that it is not a promotion, so we aren't testing that line 11 is implemented correctly, but we are testing demotion, and it works OK (again: assuming the desired spec updates).

TypeSystem/flow-analysis/promotion_via_assignment_A03_t08.dart
eernstg: OK, let's delete this test!

TypeSystem/flow-analysis/promotion_via_assignment_A06_t01.dart
eernstg: With the desired spec updates, this will promote `s` to `Pattern`. Let's just expect that.

eernstg (commented):
These tests would still be blocked on the outcome of discussions about the promotion and demotion rules, but I think we can resolve all the outstanding comment threads, based on the assumption that the rules will change as desired (that is, promotion can only promote to a proper subtype, and promotion/demotion will promote to a supertype which is of interest on assignment of a type which isn't of interest).

TypeSystem/flow-analysis/promotion_via_assignment_A03_t05.dart
sgrekhov: Hm... Befor going on, what about https://github.com/dart-lang/sdk/issues/60620? The current behavior is:
```dart
typedef Void = void;

test1() {
  String s = "Some string";
  if (s is Void) {}
  s = "";
  s.substring(0); // Compile error: This expression has type 'void' and can't be used.
}
```
What should we expect in this case according to the updated rules? `String` is proper subtype of `void` (I was unable to find a definition of "being proper subtype of", but according to ChatGPT and Geminy it's  a type which is a subtype but not equal to the initial type), `void` is a type of interest, so it's ok to demote `s` to `void`? Again, what about the fact that `String` is also type of interest for `s`?

sgrekhov (commented):
@eernstg, thank you for the detailed review! But before proceeding, I need a better understanding of the intended rules. So please take a look at the questions below.

TypeSystem/flow-analysis/promotion_via_assignment_A03_t05.dart
eernstg: First, promotion should never be able to give a variable `s` with declared type `String` any type which isn't a subtype of `String`. In particular, promotion/demotion can never give `s` the type `void`. If this does happen then it's a bug.

We could make it `Object? s = "Some string";` in order to enable `Void` as a type of interest, but let's keep `String` as the declared type for this discussion.

So the expected behavior of the example above is that `s is Void` may or may not add `Void` to the set of types of interest (we shouldn't be able to see the difference, but I think the spec says that the type is added to `tested` no matter what, but then it should be ignored during promotion/demotion; it might be nicer to just avoid adding it to `tested` in the first place, so that might be the outcome of the discussion; still, it shouldn't matter).

So `s is Void` does nothing.

Next, `s = ""` assigns an expression whose static type is the declared type of `s`, which hasn't been promoted in the meantime. Again, nothing happens, `s` has type `String`.

Finally, `s.substring(0)` is accepted with no errors (compile time and run time) because the receiver has type `String`.

eernstg (commented):
Added a few more comments.

sgrekhov (commented):
Updated. PTAL. Mainly add two tests. One with `Object?` as a declared type and the second test with the declared type `String` (and this test check sdk#60620).

TypeSystem/flow-analysis/promotion_via_assignment_A03_t03.dart
eernstg: Thx! ;-)

TypeSystem/flow-analysis/promotion_via_assignment_A03_t08.dart
eernstg: Thanks!

TypeSystem/flow-analysis/promotion_via_assignment_A06_t01.dart
eernstg: The problem with `test2` is that it is not a test which is based on language related expectations, it's specifically a #60620 regression test. From a language perspective there is absolutely no reason to expect that `s = ""` should equip `s` with a new type: It already has type `String` based on the declaration (and no promotions), and the assigned expression has static type `String`.

So it's kind of crazy to expect that it would get another type (such as `Pattern`) based on this situation.

It makes sense in a regression test (because we've actually seen the crazy behavior in real life!), but it does not really make sense as a language test.

Anyway, let's keep it just because this PR has been lingering forever. ;-)

eernstg (commented):
Converging! Couple of comments.

TypeSystem/flow-analysis/promotion_via_assignment_A03_t05.dart
```diff
@@ -0,0 +1,41 @@
...
+
+import '../../Utils/static_type_helper.dart';
+
+typedef Void = void;
+
+test1() {
+  Object? s = "x";
+  s as String;
+  if (s is Void) {} // ignore: unnecessary_type_check
+  s = 1; // Cannot assign `void`, let's assign `String <: void`
```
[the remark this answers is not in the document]
sgrekhov: Thank you! Fixed.

TypeSystem/flow-analysis/promotion_via_assignment_A03_t05.dart
sgrekhov: > Do we have a special naming scheme for regression tests?
There is https://github.com/dart-lang/co19/tree/master/LanguageFeatures/regression keeping the set of tests for NNBD and LUB. But Ok,  let's drop `test2()`. The issue is reported, and there should be a regression test for it when fixed.

TypeSystem/flow-analysis/promotion_via_assignment_A03_t06.dart
sgrekhov: Updated. `test2` removed, `test1()` renamed to `main()`

sgrekhov (commented):
Updated and removed `test2()` as suggested. PTAL

TypeSystem/flow-analysis/promotion_via_assignment_A03_t05.dart
eernstg: Thanks!

eernstg (approved):
LGTM!

PR #3168, opened 2025-05-05
#3057. Update promotion_via_assignment_A03_t02.dart according to the implementation

sgrekhov:
According to the https://github.com/dart-lang/sdk/issues/60646#issuecomment-2848198769


eernstg (approved):
I'm still double checking whether it is demoted to `Object?` or `Object`, but we can adjust that in a separate PR if needed. It doesn't change the end result.

PR #3171, opened 2025-05-07
Fixes #3170. Fix spread collections tests

### LINKED ISSUE #3170: [Dot Shorthands] Spreads have `Iterable<..>` context
co19/LanguageFeatures/Static-access-shorthand/non_ambiguity_A02_t06
co19/LanguageFeatures/Static-access-shorthand/non_ambiguity_A02_t05
co19/LanguageFeatures/Static-access-shorthand/non_ambiguity_A02_t04
co19/LanguageFeatures/Static-access-shorthand/non_ambiguity_A02_t03

For the tests above, we use spreads on dot shorthands, but they should produce an error because the context type of the operand of a spread element is `Iterable<C>` where `C` is the element type of the collection.
-- Meaning that when we're looking up the namespace for finding which static member to resolve to, we're looking in `Iterable` and these tests will produce an error.

Essentially, we'd need to change these to expect an error. The tests are still valuable to have to make sure we don't crash while parsing.

cc. @eernstg

sgrekhov:


LanguageFeatures/Static-access-shorthand/non_ambiguity_A02_t07.dart
```diff
@@ -0,0 +1,40 @@
...
+/// `... .id`
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=dot-shorthands
+
+import '../../Utils/expect.dart';
+
+main() {
+  var list = <num>[
+    ....castFrom<num, int>([1, 2, 3])
```
eernstg, line 22: Cool! So we can test the actual behavior in a non-error way, too. I hadn't thought of this way to do it.

LanguageFeatures/Static-access-shorthand/non_ambiguity_A02_t03.dart
```diff
@@ -59,30 +57,62 @@ extension type ET(int value) {
 
 main() {
   var l1 = <C>[
-  ....one,
+  ....one, // `...` has type `Iterable<C>` but it has no static member `one`
```
eernstg, line 60: Perhaps:
```suggestion
  ....one, // `...` yields the context `Iterable` which has no static member `one`
```
Elements (including spread elements) don't have a type (they are not expressions), but they contribute to the determination of an element type for a list or set literal, and to a key or value type for a map literal, so it's somewhat confusing to talk about a spread element having a type. Also, `...` in itself is even further from having a type. The operand of a spread element does have a context type, and it does have a static namespace which is used for dot shorthand lookups.

eernstg (commented):
Looks good! Just one comment (occurring in about 4 locations) is a little bit confusing, and I suggested a way to make it clearer.

sgrekhov:
Thank you! Updated. PTAL.

eernstg (commented):
LGTM, thanks!

eernstg (approved):

PR #3173, opened 2025-05-08
#3057. Add more reachability tests for `switch`

sgrekhov:


TypeSystem/flow-analysis/reachability_switch_A08_t01.dart
```diff
@@ -0,0 +1,38 @@
...
+
+/// @assertion switch statement: If `N` is a switch statement of the form
+/// `switch (E) {alternatives}` then:
+/// - Let `before(E) = before(N)`.
+/// - For each `C` in `alternatives` with statement body `S`:
+///   - If `C` is labelled let
+///     `before(S) = conservativeJoin(after(E), assignedIn(N), capturedIn(N))`
+///     otherwise let `before(S) = after(E)`.
+///  - If the cases are exhaustive, then let `after(N) = break(N)` otherwise let
+///    `after(N) = join(after(E), break(N))`.
```
eernstg, line 13: I think it would be useful to have an explicit statement that this spec paragraph targets the old version of switch statements, and as of May 2025 there is no corresponding spec paragraph about switch expressions or switch statements using patterns.

TypeSystem/flow-analysis/reachability_switch_A08_t02.dart
```diff
@@ -0,0 +1,28 @@
...
+/// - Let `before(E) = before(N)`.
+/// - For each `C` in `alternatives` with statement body `S`:
+///   - If `C` is labelled let
+///     `before(S) = conservativeJoin(after(E), assignedIn(N), capturedIn(N))`
+///     otherwise let `before(S) = after(E)`.
+///  - If the cases are exhaustive, then let `after(N) = break(N)` otherwise let
+///    `after(N) = join(after(E), break(N))`.
+///
+/// @description Checks that if a variable is initialized in a `when` part of a
+/// `case` statement then this variable is not definitely assigned.
```
eernstg, line 16: If you say `int _` then it seems likely that the flow analysis _could_ know that the case will definitely match and hence the `when` expressions will definitely be evaluated (and it may be true or false, but that doesn't change the fact that the `when` expression has been evaluated).

So maybe this test will fail, and we should expect `i` to be definitely assigned after all? Let's rely on the implementation behavior in order to see which way it goes.

eernstg: Ok, the next test already does this. Closing.

TypeSystem/flow-analysis/reachability_switch_A08_t04.dart
```diff
@@ -0,0 +1,48 @@
...
+
+void test2() {
+  int i;
+  switch (42) {
+    label:
+    case String _:
+      break;
+    case _ when (i = 42) > 0:
+  }
+  i;
```
eernstg, line 39: If we remove the label and assume that the flow analysis does not know that the intersection between `String` and `int` is the empty type (`Never`) then `i` should still be not-definitely assigned at this point. In other words, it's not about the label (unless flow analysis knows that the labeled case is unreachable). Is there a "dead code" warning?

Otherwise I think it isn't really justified to say that we're testing anything about labels in this test case.

TypeSystem/flow-analysis/reachability_switch_expression_A06_t01.dart
```diff
@@ -0,0 +1,30 @@
...
+/// - Let `before(E) = split(before(N))`.
+/// - For each case `Cj` in `alternatives` with expression `Ej`, for `j` in
+///   `1 .. k`:
+///   - Let `before(Ej) = after(E)`.
+/// - Let `after(N) = join(after(E1), after(E2), .. after(Ek))`.
+/// TODO (sgrekhov): there is no switch expression in the flow analysis spec yet
+/// (May 2025). Update the assertion after spec update.
+///
+/// @description Checks that if a variable is initialized in a `when` part of a
+/// `case` statement then this variable is possibly assigned.
```
eernstg, line 16: Perhaps we should say that it is a "switch expression case" (not `case`) because the actual syntax doesn't have that keyword.

TypeSystem/flow-analysis/reachability_switch_expression_A06_t01.dart
```diff
@@ -0,0 +1,30 @@
...
+/// @author sgrekhov22@gmail.com
+
+void test() {
+  late int i;
+  var x = switch (42) {
+    String _ when (i = 42) > 0 => 42,
+    _ => 0
+  };
+  i; // Possibly assigned
```
eernstg, line 25: If it is possibly assigned then we know that `String` and `int` having the empty intersection is _not_ something that the flow analysis knows.

TypeSystem/flow-analysis/reachability_switch_A08_t04.dart
eernstg: Perhaps test some unrelated types where the flow analysis cannot know that the intersection is empty (`class A {}` and `class B {}` could have a shared subtype)? That seems to be a more readable test case.

TypeSystem/flow-analysis/reachability_switch_expression_A06_t01.dart
eernstg: This would be relevant for the comment in TypeSystem/flow-analysis/reachability_switch_A08_t04.dart.

TypeSystem/flow-analysis/reachability_switch_expression_A06_t02.dart
```diff
@@ -0,0 +1,33 @@
...
+/// - Let `before(E) = split(before(N))`.
+/// - For each case `Cj` in `alternatives` with expression `Ej`, for `j` in
+///   `1 .. k`:
+///   - Let `before(Ej) = after(E)`.
+/// - Let `after(N) = join(after(E1), after(E2), .. after(Ek))`.
+/// TODO (sgrekhov): there is no switch expression in the flow analysis spec yet
+/// (May 2025). Update the assertion after spec update.
+///
+/// @description Checks that if a variable is initialized in a `when` part of a
+/// `case` statement then this variable is not definitely assigned.
```
eernstg, line 16: Again: expression, not statement.

eernstg (commented):

TypeSystem/flow-analysis/reachability_switch_expression_A06_t01.dart
sgrekhov: > If it is possibly assigned then we know that String and int having the empty intersection is not something that the flow analysis knows.

Why? Flow analysis doesn't know that `int` and `String` have no intersection and therefore assumes that `when` part _may_ be executed. And that means that `i` _may_ be initialized. I don't see what's wrong here.

TypeSystem/flow-analysis/reachability_switch_expression_A06_t01.dart
sgrekhov: Updated, thank you.

TypeSystem/flow-analysis/reachability_switch_expression_A06_t02.dart
sgrekhov: Updated.

TypeSystem/flow-analysis/reachability_switch_A08_t04.dart
sgrekhov: >  Is there a "dead code" warning?

No, there isn't. Removed `test2()` at all. `test1()` fails without the label (no expected error) and succeeds with it (there is an expected error).

sgrekhov (commented):
Updated. PTAL. One test left unchanged because I belive it is ok.

TypeSystem/flow-analysis/reachability_switch_A08_t04.dart
eernstg: OK, sounds good!

PR #3174, opened 2025-05-09
#1401. Add additional tests for shared case scoping

sgrekhov:


LanguageFeatures/Patterns/shared_case_scope_A01_t05.dart
```diff
@@ -0,0 +1,88 @@
...
+//          ^
+// [analyzer] unspecified
+// [cfe] unspecified
+    default:
+      print(x);
+  }
+
+  switch (n) {
+    case int x when 1 > 2:
+    case num when 1 > 2:
```
eernstg, line 67: Same potential confusion.

eernstg (commented):
I think it doesn't matter technically, but `case num` looks confusing, so I'm suggesting to write something which is more straightforward.

sgrekhov:
Updated. PTAL.

eernstg (approved):
LGTM

PR #3184, opened 2025-05-14
#3057. Add more promotion tests for extension types

sgrekhov:


TypeSystem/flow-analysis/reachability_A08_t09.dart
```diff
@@ -0,0 +1,66 @@
...
+///   - Let after(N) = after(E2).
+/// Note that it is tempting to generalize the two `null` literal cases to apply
+/// to any expression whose type is `Null`, but this would be unsound in cases
+/// where `E2` assigns to `x`. (Consider, for example,
+/// `(int? x) => x == (x = null) ? true : x.isEven`, which tries to call
+/// `null.isEven` in the event of a non-null input).
+///
+/// @description Checks that if `equivalentToNull(T1)` and `T2` is not
+/// non-nullable, or `equivalentToNull(T2)` and `T1` is not non-nullable, then
+/// `after(N) = after(E2)`.
```
eernstg, line 31: This is the before-sound-flow-analysis spec language. With the update considered in dart-lang/language#3100, we'd have `true(N) = unreachable(after(E2))` and `false(N) = after(E2)`. I suppose we have other tests using the same spec text, so they'll all need to be updated when the dart-lang/language#3100 effort has been completed. So that's fine.

Next, `after(N) = after(E2)` cannot be tested because this is a data structure (a flow model) that may contain data about any number of variables, and about many different properties of those variables.

So let's do the same thing as previously: Add a much more narrow goal to the description which can actually be tested, e.g., assigned/unassigned-ness. This seems to be what the test is already doing anyway.

Also, it would be nice to know which part of the spec language we're testing. When `T1` is equivalent to null, but `T2` is not non-nullable (or the converse) then the case in lines 12-14 is _not_ applicable. The assumptions do not allow us to conclude whether the `stripParens` cases are applicable, but the concrete terms do not have parentheses. So we must be in the fall-back case in line 22.

When `after(N)` is specified but `true(N)` and `false(N)` are not, the general rule is that all three of them are the same.

In particular, the assignedness below should be the same for the "then" and the `else` branch.

eernstg (commented):
OK, it took some time to clarify the situation, but the tests are fine! I suggested a couple of changes to the '@description`s.

sgrekhov:
Thank you. Descriptions updated, new tests added. PTAL

eernstg (approved):
LGTM

PR #3185, opened 2025-05-15
#3057. Add more tests for promotion in switch

sgrekhov:


TypeSystem/flow-analysis/promotion_switch_A04_t02.dart
```diff
@@ -0,0 +1,65 @@
...
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+test1(Object? x) {
+  switch (x) {
+    case int _:
+      print("");
+    case _:
+      x = 42;
```
eernstg, line 20: At this point, `x` has version `x#2` because of the assignment, and we have

> Since the version of `e` is no longer `o#1`, `o` is not promoted.

But that would be the type tests implied by pattern matching (because they are still applied to the value of `x#1`, so we can't allow a pattern match on `int _` to say anything about the new value of `x`).

It should still be the case that `int` is a type of interest for `x`, and `x` should still be promoted via assignment of an expression whose static type is of interest (or it has a unique best supertype that is). So line 20 _should_ promote `x` to `int`.

OK!

TypeSystem/flow-analysis/promotion_switch_A05_t01.dart
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion For now (May, 2025) promotion in switch statement and
+/// expressions are not specified yet.
+/// TODO (sgrekhov): update when specified
+///
+/// @description Checks that pattern match doesn't promote the variable when a
+/// case scope is shared abd cases have different types.
```
eernstg, line 10: Typo:
```suggestion
/// case scope is shared and cases have different types.
```

TypeSystem/flow-analysis/promotion_switch_A05_t01.dart
```diff
@@ -0,0 +1,52 @@
...
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+test1(Object? x) {
+  switch (x) {
+    case int v:
+    case num _:
```
eernstg, line 18: I'd think there should be two control flow paths to line 19: Matching the pattern in line 17, or matching the pattern in line 18. At the join, the promotion chains should be combined; `joinV` does this by taking the intersection of the two promotion chains (that is, include exactly the types that occur in both promotion chains, preserving the order), so that would be the empty list here.

OK, so it's not because the cases have different types, it's because these promotion chains happen to have the empty intersection.

So it would be interesting to have a case where the intersection is non-empty (like `[Object, num]` and `[Object, int]`, or `[Object, int]` and `[num, int]`).

eernstg: Perhaps the logic doesn't actually treat this situation as a full-fledged join, but in that case it would be nice to know it. ;)

eernstg (commented):

TypeSystem/flow-analysis/promotion_switch_A04_t01.dart
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion For now (May, 2025) promotion in switch statement and
+/// expressions are not specified yet.
+/// TODO (sgrekhov): update when specified
```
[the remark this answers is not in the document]
eernstg: Thanks!

eernstg (commented):
Just a clarification.

TypeSystem/flow-analysis/promotion_switch_A05_t04.dart
```diff
@@ -0,0 +1,55 @@
...
+      x.expectStaticType<Exactly<int>>();
+    case _:
+      x.expectStaticType<Exactly<Object?>>();
+  }
+}
+
+test3(Object? x) {
+  switch (x) {
+    case num _ && int v:
+    case Object v && num _: // ignore: unreachable_switch_case
```
eernstg, line 41: Ah, yes, that's an interesting case, too!

eernstg (approved):
LGTM

Issue #3182, opened 2025-05-14
[augmentations] Update co19 tests for augmentations


sgrekhov:
Update co19 tests according to https://github.com/dart-lang/language/commit/ebd9e1882484380edc6bc3c5cccad23ea1ea845b

PR #3188, opened 2025-05-19
#3182. Remove `augmented` expression tests

sgrekhov:


eernstg (approved):
Wow! That's a lot of tests about this topic --- kind of sad to go back and forth on a topic like this, but I do think it's helpful to simplify the feature.

LGTM.

PR #3190, opened 2025-05-20
#3182. Update scoping tests

sgrekhov:


LanguageFeatures/Augmentation-libraries/scoping_A01_t05.dart
```diff
@@ -2,18 +2,17 @@
...
-/// `name` is not in the lexical scope at all, in which case it’s interpreted as
-/// `this.name` if it occurs inside a scope where a `this` is available.
+/// @assertion The static and instance member namespaces for an augmented type
+/// or extension declaration include the declarations of all members in the
+/// introductory and augmenting declarations. Identifiers in the bodies of
+/// members are resolved against that complete merged namespace. In other words,
+/// augmentations are applied before identifiers inside members are resolved.
 ///
 /// @description Checks that `name` is interpreted as `this.name` if it occurs
 /// inside a scope where a `this` is available. Test a variable.
```
eernstg, line 12: This should be updated: The scope of an augmented class is treated the same as the top-level scope: It's all gathered into a single namespace and then lookups proceed to look up names in there.

In particular, each of `class C`, `mixin M`, etc has a declaration of all the names `foo`, `bar`, `baz`, `qux`, and hence references to those names in the body of such a declaration must be resolved to the class body member (instance or static).

As far as I can see, this means that the top-level declarations of the same names are never used.

eernstg: `name` is still interpreted as `this.name` when it occurs in a location where there is access to `this`. That's crucial for the ability to use inherited members (concrete or abstract ones from the superclass chain as well as abstract ones from the `implements` or `mixin`-`on` clause).

LanguageFeatures/Augmentation-libraries/scoping_A01_t06.dart
```diff
@@ -2,18 +2,17 @@
...
-/// `name` is not in the lexical scope at all, in which case it’s interpreted as
-/// `this.name` if it occurs inside a scope where a `this` is available.
+/// @assertion The static and instance member namespaces for an augmented type
+/// or extension declaration include the declarations of all members in the
+/// introductory and augmenting declarations. Identifiers in the bodies of
+/// members are resolved against that complete merged namespace. In other words,
+/// augmentations are applied before identifiers inside members are resolved.
 ///
 /// @description Checks that `name` is interpreted as `this.name` if it occurs
 /// inside a scope where a `this` is available. Test a getter.
```
eernstg, line 12: Similar.

I don't understand how `foo` in line 28 could ever have been the global `foo`, with the old scoping or the new one... not sure how that slipped through our discussions without being noticed. ;-)

LanguageFeatures/Augmentation-libraries/scoping_A01_t07.dart
```diff
@@ -2,18 +2,17 @@
...
-/// `name` is not in the lexical scope at all, in which case it’s interpreted as
-/// `this.name` if it occurs inside a scope where a `this` is available.
+/// @assertion The static and instance member namespaces for an augmented type
+/// or extension declaration include the declarations of all members in the
+/// introductory and augmenting declarations. Identifiers in the bodies of
+/// members are resolved against that complete merged namespace. In other words,
+/// augmentations are applied before identifiers inside members are resolved.
 ///
 /// @description Checks that `name` is interpreted as `this.name` if it occurs
 /// inside a scope where a `this` is available. Test a setter.
```
eernstg, line 12: This looks similar to the others, so it should presumably never invoke any of the global declarations.

LanguageFeatures/Augmentation-libraries/scoping_A01_t08.dart
```diff
@@ -2,18 +2,17 @@
...
-/// `name` is not in the lexical scope at all, in which case it’s interpreted as
-/// `this.name` if it occurs inside a scope where a `this` is available.
+/// @assertion The static and instance member namespaces for an augmented type
+/// or extension declaration include the declarations of all members in the
+/// introductory and augmenting declarations. Identifiers in the bodies of
+/// members are resolved against that complete merged namespace. In other words,
+/// augmentations are applied before identifiers inside members are resolved.
 ///
 /// @description Checks that `name` is interpreted as `this.name` if it occurs
 /// inside a scope where a `this` is available. Test a method.
```
eernstg, line 12: Similar again.

eernstg (commented):
There are a bunch of tests that are affected by the new scope rule.

LanguageFeatures/Augmentation-libraries/scoping_A01_t05.dart
sgrekhov: Great, thank you! overlooked this one! Fixed.

LanguageFeatures/Augmentation-libraries/scoping_A01_t06.dart
sgrekhov: Yes, indeed. Thank you for noticing! Fixed.

LanguageFeatures/Augmentation-libraries/scoping_A01_t07.dart
sgrekhov: Fixed.

sgrekhov (commented):
Thank you for being so attentive! Fixed. PTAL.

eernstg (approved):
LGTM

PR #3193, opened 2025-05-21
#3182. Update type_inheritance_A02/3_t*.dart

sgrekhov:
The A02 tests are updated such that they satisfy the constraint that member implementations can be provided by an augmentation, but not overridden. The A03 tests are updated to test abstract instance variables (and several cases handling augmentation of other variables have been deleted because concrete variables cannot be augmented according to the new specification of augmentations).

### DIFF
--- a/LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t01.dart
+++ b/LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t01.dart
@@ -10,8 +10,9 @@
 /// compile-time error if the type denoted by the augmenting declaration is not
 /// the same type as the type in the corresponding declaration being augmented.
 ///
-/// @description Check that it is a compile-time error if augmenting declaration
-/// specifies a different return type than the introductory declaration.
+/// @description Check that it is a compile-time error if an augmenting
+/// declaration specifies a different return type than the introductory
+/// declaration.
 /// @author sgrekhov22@gmail.com
 
 // SharedOptions=--enable-experiment=augmentations,enhanced-parts
--- a/LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t01.dart
+++ b/LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t01.dart
@@ -2,23 +2,19 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-/// @assertion An augmenting declaration may have no type annotations for a
-/// return type, variable type, parameter type, or type parameter bound type. In
-/// the last case, that includes omitting the extends keyword. For a variable or
-/// parameter, a var keyword may replace the type.
+/// @assertion If the type annotation or type parameter bound is omitted in the
+/// augmenting declaration, it is inferred to be the same as the corresponding
+/// type annotation or type parameter bound in the declaration being augmented.
 ///
-/// When applying an augmenting declaration that contains a type annotation at
-/// one of these positions, to a definition to be augmented, it's a compile-time
-/// error if the type denoted by the augmenting declaration is not the same type
-/// as the type that the augmented definition has at the corresponding position.
-/// An augmenting declaration can omit type annotations, but if it doesn't, it
-/// must repeat the type from the augmented definition.
+/// If the type annotation or type parameter bound is not omitted, then it's a
+/// compile-time error if the type denoted by the augmenting declaration is not
+/// the same type as the type in the corresponding declaration being augmented.
 ///
 /// @description Check that it is a compile-time error if augmenting declaration
-/// specifies a different return type than the augmented declaration.
+/// specifies a different return type than the introductory declaration.
 /// @author sgrekhov22@gmail.com
 
-// SharedOptions=--enable-experiment=macros
+// SharedOptions=--enable-experiment=augmentations,enhanced-parts
 
 part 'type_inheritance_A02_t01_lib.dart';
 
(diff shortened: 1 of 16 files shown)


LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t03.dart
```diff
@@ -2,94 +2,90 @@
...
+  Expect.equals(0, E.e0.instanceMethod());
+  Expect.equals(0, Ext.staticGetter);
+  Expect.equals(1, Ext.staticMethod());
+  Expect.equals(1, A().instanceGetter);
+  Expect.equals(0, A().instanceMethod());
+  Expect.equals(0, ET.staticGetter);
+  Expect.equals(1, ET.staticMethod());
+  Expect.equals(1, ET(0).instanceGetter);
+  Expect.equals(0, ET(0).instanceMethod());
+  Expect.equals(42, ET(42).id);
```
eernstg, line 90: Considering the comment on LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t02.dart, perhaps there shouldn't be an `id` case here, either?

eernstg: Or perhaps it's the other way around and an `id` case should be added to LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t02.dart after all?

LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t04_lib.dart
```diff
@@ -2,32 +2,28 @@
...
 /// wrong top type.
 /// @author sgrekhov22@gmail.com
 
-// SharedOptions=--enable-experiment=macros
+// SharedOptions=--enable-experiment=augmentations,enhanced-parts
 
 part of 'type_inheritance_A02_t04.dart';
 
 augment Object? get topLevelGetter => 0;
 //      ^^^^^^
```
eernstg, line 23: Perhaps?:
```suggestion
//      ^^^^^^^
```

LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t01.dart
```diff
@@ -2,23 +2,19 @@
...
-/// one of these positions, to a definition to be augmented, it's a compile-time
-/// error if the type denoted by the augmenting declaration is not the same type
-/// as the type that the augmented definition has at the corresponding position.
-/// An augmenting declaration can omit type annotations, but if it doesn't, it
-/// must repeat the type from the augmented definition.
+/// If the type annotation or type parameter bound is not omitted, then it's a
+/// compile-time error if the type denoted by the augmenting declaration is not
+/// the same type as the type in the corresponding declaration being augmented.
 ///
 /// @description Check that it is a compile-time error if augmenting declaration
```
eernstg, line 13: Typo:
```suggestion
/// @description Check that it is a compile-time error if an augmenting declaration
```
Seems to occur in every test, certainly in several.

eernstg (commented):
Looks good, with a couple of typos!

I was wondering why so many cases had been removed from the last 6 test libraries, but the reason is surely that they are concerned with the correct treatment of an augmentation with some wrong parts (namely: the type annotation), and this makes no sense in cases where the augmentation is an error in the first place (because variables can't be augmented any more). Added a note about this to the description of the PR.

LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t01.dart
sgrekhov: Thanks! Fixed here and below.

LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t02.dart
```diff
@@ -2,91 +2,87 @@
...
+  Expect.equals(1, E.e0.instanceGetter);
+  Expect.equals(0, E.e0.instanceMethod());
+  Expect.equals(0, Ext.staticGetter);
+  Expect.equals(1, Ext.staticMethod());
+  Expect.equals(1, A().instanceGetter);
+  Expect.equals(0, A().instanceMethod());
+  Expect.equals(0, ET.staticGetter);
+  Expect.equals(1, ET.staticMethod());
+  Expect.equals(1, ET(0).instanceGetter);
+  Expect.equals(0, ET(0).instanceMethod());
```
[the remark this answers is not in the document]
sgrekhov: Forgot about it. Added. Representation variable is a getter which can be augmented (but with no body). Thank you.

LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t03.dart
sgrekhov: Yes, has been added already :)

LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t04_lib.dart
sgrekhov: Updated.

sgrekhov (commented):
Updated. PTAL.

LanguageFeatures/Augmentation-libraries/type_inheritance_A02_t02.dart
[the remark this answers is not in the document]
eernstg: Sounds good! This confirms that it is still possible to add metadata and DartDoc comments to a declaration via an augmenting declaration, even for the representation variable (no matter how much it is claimed that "it is not a variable" ;-).

eernstg (approved):
LGTM

Issue #3180, opened 2025-05-13
Dart web interop testing: Adding 'dart:js_interop' tests


biggs0125:
Today we only have co19 tests for the now deprecated `dart:html`. However, we are in the process of migrating Dart users from `dart:html` to `package:web` and `dart:js_interop`. 

We should consider adding tests specifically for `dart:js_interop` to the `co19` test suite to ensure conformance of this new interop standard. Public documentation for new interop is here:
https://dart.dev/interop/js-interop/usage

Though `package:web` is essentially the replacement for `dart:html`, it isn't worth adding tests to co19 for that package for 3 reasons:
1) Unlike 'dart:html' there's no special code in the web compilers to support it directly.
2) It's effectively a bunch of static JS interop declarations. Unlike 'dart:html' we don't any wrappers for the underlying JS entities so there is minimal logic to test.
3) It isn't actually part of the Dart SDK.

For now, we want to retain the tests for `dart:html` as the Dart ecosystem still contains enough old interop code that we want it maintain the coverage those tests provide.

cc @athomas @srujzs @sigmundch @eernstg

srujzs:
> We should consider adding tests specifically for dart:js_interop to the co19 test suite to ensure conformance of this new interop standard. Public documentation for new interop is here:
https://dart.dev/interop/js-interop/usage

Some of the API surface in `dart:js_interop`/`dart:js_interop_unsafe` is special (like the `toJS`/`toDart` conversions for example) where there's an actual implementation involved in the patch files. In other cases like `JSUint8Array.new` for example, there's nothing special - they're just interop declarations. Just like with `package:web`, testing the latter isn't that useful.

sgrekhov:
@srujzs Is there anything in `js_interop` that isn’t covered by the tests yet, or anything you’d like to have tests for? Otherwise, I’m going to close this issue as completed.

I was thinking about adding tests for all good and bad cases listened [here](https://dart.dev/tools/linter-rules/invalid_runtime_check_with_js_interop_types) but according to the https://github.com/dart-lang/sdk/issues/61265#issuecomment-3198788202 the expected result here is compiler-specific.

sgrekhov:
The current status of this issue is that I'm waiting for the landing of https://dart-review.googlesource.com/c/sdk/+/450140 to update test comments and after that I'm going to close this as completed. @srujzs is there anything that I missed?

sgrekhov:
Closing as completed. Please feel free to create new co19 issues related with js_interop.

PR #3205, opened 2025-05-27
#3180. Add tests for NullableUndefineableJSAnyExtension.

sgrekhov:


LibTest/js_interop/NullableUndefineableJSAnyExtension/isNull_t01.dart
```diff
@@ -0,0 +1,28 @@
...
+main() {
+  eval(r'''
+    var n = null;
+    var u = undefined;
+    var s = "s";
+  ''');
+
+  Expect.isTrue(globalContext["n"].isNull);
+  Expect.isFalse(42.toJS.isNull);
+  Expect.isFalse(globalContext["u"].isNull);
```
eernstg, line 25: This is the most interesting case. I don't know whether or not `isNull` should see any difference between `null` and `undefined`.

The documentation says

> Currently, there is no way to distinguish between JavaScript `undefined`
> and JavaScript `null` when compiling to Wasm. Therefore, this getter
> should only be used for code that compiles to JavaScript and will throw
> when compiling to Wasm.

.. so there may be a need to treat compilation to Wasm and compilation to JS differently.

eernstg (approved):
LGTM.

@srujzs, it would be great if you could take a look at this as well!

LibTest/js_interop/NullableUndefineableJSAnyExtension/isNull_t01.dart
srujzs: Thanks for the cc, Erik! And thanks for adding tests, Sergey!

Yes, when compiling to Wasm, we're forced to convert to Wasm's null ref regardless of whether the JS value is `undefined` or `null` in order to maintain the same nullability semantics for interop across all compilers, so there's no way of distinguishing.

In some future, we may provide an opt-in annotation users to preserve `undefined` separately, but that doesn't exist yet.

So, both `isNull` and `isUndefined` need to be handled differently based on the platforms.

sgrekhov:
Yes, we use status files for these purposes. But in this particular case I think that it's simpler to use `if(isJS) {/* Code, not supported by Wasm*/}`

srujzs:
> if this test is executed on any platform where the generated code isn't JavaScript, presumably the eval thing wouldn't work

`eval` is an interop call: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval, so it should work even when compiling to Wasm. We use it often in the SDK tests for a quick way to set up the JS environment.

srujzs:
It's a runtime-error (you specifically get an `UnimplementedError`) when compiling to Wasm.

eernstg (approved):
LGTM, based on expected changes to status files.

PR #3211, opened 2025-05-30
Fixes #3210. Don't expect promotion from `FutureOr<void>` to `Future<void>` or `void`

### LINKED ISSUE #3210: One test will need updating when we stop promoting to mutual subtypes
https://dart-review.googlesource.com/c/sdk/+/429200 changes flow analysis so that starting in Dart 3.9 (when the `sound-flow-analysis` feature is switched on), an `is` test or `as` expression no longer promotes from one type to a mutual subtype.

When this change lands, the following co19 test will need to be updated: `co19/TypeSystem/flow-analysis/promotion_via_type_test_A03_t13`

What's happening is that the test is trying to check that a variable can be promoted from `FutureOr<void>` to `Future<void>`, yielding `void` in the `false` continuation. With the change, this promotion no longer occurs, since `FutureOr<void>` and `Future<void>` are mutual subtypes.

See https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8713598040047675201/+/u/test_results/new_test_failures__logs_ for the failure provoked by the change.

sgrekhov:


TypeSystem/flow-analysis/promotion_via_type_test_A03_t13.dart
```diff
@@ -12,11 +12,8 @@
 /// - and `T <: S` or (`S` is `X extends R` and `T <: R`) or (`S` is `X & R` and
 ///   `T <: R`)
 ///
-/// @description Checks that a variable can be promoted from `FutureOr<void>` to
-/// `Future<void>` yielding `void` in the `false` continuation. In the next
-/// test, `is Void` does not promote the variable in the `true` continuation, of
-/// `S`, but the `false` continuation promotes the variable to `void`. In both
-/// cases, `factor` determines the outcome.
+/// @description Checks that a variable cannot be promoted from `FutureOr<void>`
+/// to `Future<void>` or `void`.
```
eernstg, line 16: But the test for `Future<void>` definitely does promote the variable in the true continuation, and this description also contradicts the actual expectation in line 27.

It's the behavior of `factor(_, _)` that prevents certain kinds of promotion, and that's actually just because we have a lot of subtype relationships because the types in question are all top types.

So, because `factor(FutureOr<void>, Future<void>) == factor(void, Future<void>) == void`, we're trying to promote `s` to `void` in the `else` branch line 29, but that's not a proper subtype of the current type of `s`, so nothing happens.

In line 35, no promotion has taken place because `Void` isn't a proper subtype of the current type of `s`. In the `else` branch we consider `factor(FutureOr<void>, Void) == Never` (so we should expect a warning about dead code in the `else` branch, but that's the fix which is now being handled in #60718).

Finally, I'd expect line 37 to show that the type of `s` is `Never` at that point.

@stereotype441, wouldn't your changes to the flow analysis (possibly the ones associated with #60718) cause `s` to have type `Never` in line 37?

eernstg (commented):
I think the description in A03_t13 needs to be adjusted. I'd suggest that line 37 expects `s` to have type `Never`, and then we'll wait for input from @stereotype441.

TypeSystem/flow-analysis/promotion_via_type_test_A03_t13.dart
stereotype441: > Finally, I'd expect line 37 to show that the type of `s` is `Never` at that point.
> 
> @stereotype441, wouldn't your changes to the flow analysis (possibly the ones associated with #60718) cause `s` to have type `Never` in line 37?

@eernstg and I discussed this a bit more at our meeting on Monday, and we decided that the long term plan is to change flow analysis so that it never promotes to the types `Never` or `Null`. It's going to take a while to get to an implementation that's fully consistent in this regard, but I've already started moving in that direction.

As far as this particular case is concirned, the changes I've made for https://github.com/dart-lang/sdk/issues/60718 have made it so that when the result of applying the `factor` algorithm is `Never`, the code path is marked as unreachable, but the variable is not promoted to `Never`. So on line 37, expecting the static type to be `FutureOr<void>` is correct.

eernstg: Thanks!

eernstg (approved):
LGTM

PR #3216, opened 2025-06-05
#3180. Add JsAnyOperation.and tests

sgrekhov:


eernstg (approved):
LGTM.

Just wondering about the semantics of repeated executions of `eval('var v = e; ...')`: Is it safe to assume that nothing changes observably when it is executed for the second, third, etc time?

@srujzs, perhaps you could suggest another interop expert reviewer, just so we don't put too much of a burden on you? Otherwise, it would be great if you could take a look!

LibTest/js_interop/JSAnyOperatorExtension/and_t01.dart
```diff
@@ -0,0 +1,57 @@
...
+    var b1 = object && true;       // true
+    var b2 = object && false;      // false
+    var nan = object && (0 / 0);   // NaN
+    var f = object && foo;         // function foo() {}
+  ''');
+
+  Expect.equals(globalContext["one"], JSObject().and(1.toJS));
+  Expect.equals(globalContext["string"], JSObject().and("s".toJS));
+  Expect.listEquals(
+    globalContext["a1"].dartify(),
```
srujzs: Like the comments on the `/` PR, there's no reason to do `dartify()`.

LibTest/js_interop/JSAnyOperatorExtension/and_t01.dart
```diff
@@ -0,0 +1,57 @@
...
+    JSObject().and([].jsify()).dartify(),
```
srujzs: Also like the comments on the `/` PR, prefer fetching the array/object from the global context here, on line 41, and on line 55. Same comments everywhere else.

LibTest/js_interop/JSAnyOperatorExtension/and_t01.dart
```diff
@@ -0,0 +1,57 @@
...
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+import '../../../Utils/expect.dart';
+import '../js_utils.dart';
+
+main() {
+  eval(r'''
+    function foo() {}
+    
+    var object = {};
```
srujzs: Same meta-level comment as the `/` PR - instead of replicating this test file we can just use eval to have a different definition for the LHS JS value.

srujzs:
> Just wondering about the semantics of repeated executions of eval('var v = e; ...'): Is it safe to assume that nothing changes observably when it is executed for the second, third, etc time?

I'd avoid eval with `var underTest = ...` calls. The test relies on this evaluation occurring in a location such that the use of `underTest` can access that value. The compiler does not guarantee that. I'd instead just place it in the global context e.g. `globalThis.underTest =` so it doesn't matter where that eval call occurs.

----

At a meta level, I wonder if these tests are a bit overkill. There are tens of these operators, the implementation is quite direct, and there are a *lot* of values that can be used with these operators (any JS value, in fact!). Maybe it makes sense to test them all as thoroughly as we are, but I'm worried we'll be spending too many test cycles on something that will break in an obvious way if it ever changes. Thoughts, @eernstg?

----

> @srujzs, perhaps you could suggest another interop expert reviewer, just so we don't put too much of a burden on you? Otherwise, it would be great if you could take a look!

I'm happy to keep taking a look at all the PRs! @nshahan has expressed interest in taking a look as well.

sgrekhov:

PR #3221, opened 2025-06-09
#3180. Add tests for `JsAnyOperation.divide()`

sgrekhov:


eernstg (approved):
LGTM.

@srujzs, once again it would be great if you could take a look!

LibTest/js_interop/JSAnyOperatorExtension/divide_t02.dart
```diff
@@ -0,0 +1,94 @@
...
+    globalContext["divZero"].dartify(),
+    underTest.divide(0.toJS).dartify(),
+  );
+  Expect.equals(
+    globalContext["divNaN"].dartify(),
+    underTest.divide((0 / 0).toJS).dartify(),
+  );
+  Expect.equals(
+    globalContext["divArrayEmpty"].dartify(),
+    underTest.divide([].jsify()).dartify(),
```
srujzs, line 67: Prefer to just fetch the array you created in `eval` through the global context here and below. Same comment in other files.

LibTest/js_interop/JSAnyOperatorExtension/divide_t02.dart
```diff
@@ -0,0 +1,94 @@
...
+    underTest.divide(JSObject()).dartify(),
+  );
+  Expect.equals(
+    globalContext["divFunction"].dartify(),
+    underTest.divide(globalContext["foo"]).dartify(),
+  );
+  var dartObj = {"id": "I am an object in Dart"};
+  Expect.equals(
+    globalContext["divObject"].dartify(),
+    underTest.divide(dartObj.jsify()).dartify(),
```
srujzs, line 92: Prefer to fetch the `JSObject` you created or just call `JSObject()`. Same comment in other files.

LibTest/js_interop/JSAnyOperatorExtension/divide_t01.dart
```diff
@@ -0,0 +1,95 @@
...
+
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+import '../../../Utils/expect.dart';
+import '../js_utils.dart';
+
+void main() {
+  eval(r'''
+    function foo() {}
+    var underTest = {};
```
srujzs, line 20: It looks like we have 12 copies of this test because this keeps changing. Instead, we could just eval another time to define `underTest` for each test and then call out to common testing code e.g.

```
eval(r'''
  globalThis.underTest = {};
''');
runTest();
```

and then `runTest` would do the full eval, fetch `globalContext['underTest']`, and do all the divide expectations. It's up to you if you want to keep the 12 tests (perhaps that's better), but another alternative would be to just do this in one singular test:

```
eval(r'''
  globalThis.underTest = {};
''');
runTest();

eval(r'''
  globalThis.underTest = 42;
''');
runTest();
...

sgrekhov:

PR #3222, opened 2025-06-10
#3180. Add tests for `JsAnyOperation.equals()`

sgrekhov:
If this PR approach is OK, I'll rewrite the earlier tests this way.

### DIFF
--- a/LibTest/js_interop/JSAnyOperatorExtension/equals_lib.dart
+++ b/LibTest/js_interop/JSAnyOperatorExtension/equals_lib.dart
@@ -1,0 +1,56 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion JSBoolean equals( JSAny? any )
+/// The result of `this == other` in JavaScript.
+///
+/// @description Check that `equals` returns result of `this == other` in
+/// JavaScript. Test an empty JS object as `this`.
+/// @author sgrekhov22@gmail.com
+
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+import '../../../Utils/expect.dart';
+import '../js_utils.dart';
+
+void testEquals(JSAny? underTest, String strObj) {
+  eval("var underTest = $strObj;");
+  eval(r'''
+    function foo() {}
+    
+    var eqZero = underTest == 0;
+    var eqNum = underTest == 2; 
+    var eqFloat = underTest == 3.14;
+    var eqStringNum = underTest == "2";
+    var eqStringEmpty = underTest == "";
+    var eqStringText = underTest == "text";
+    var eqNull = underTest == null;
+    var eqNaN = underTest == (0 / 0);
+    var eqArrayEmpty = underTest == [];
+    var eqArrayNum = underTest == [1, 2];
+    var eqTrue = underTest == true;
+    var eqFalse = underTest == false;
+    var eqObject = underTest == {};
+    var eqFunction = underTest == foo;
+  ''');
+
+  Expect.equals(globalContext["eqZero"], underTest.equals(0.toJS));
+  Expect.equals(globalContext["eqNum"], underTest.equals(2.toJS));
+  Expect.equals(globalContext["eqFloat"], underTest.equals(3.14.toJS));
+  Expect.equals(globalContext["eqStringNum"], underTest.equals("2".toJS));
+  Expect.equals(globalContext["eqStringEmpty"], underTest.equals("".toJS));
+  Expect.equals(globalContext["eqStringText"], underTest.equals("text".toJS));
+  Expect.equals(globalContext["eqNull"], underTest.equals(null.jsify()));
+  Expect.equals(globalContext["eqNull"], underTest.equals(null));
+  Expect.equals(globalContext["eqNaN"], underTest.equals((0 / 0).jsify()));
+  Expect.equals(globalContext["eqArrayEmpty"], underTest.equals([].jsify()));
+  Expect.equals(globalContext["eqArrayNum"], underTest.equals([1, 2].jsify()));
+  Expect.equals(globalContext["eqTrue"], underTest.equals(true.toJS));
+  Expect.equals(globalContext["eqFalse"], underTest.equals(false.toJS));
+  Expect.equals(globalContext["eqObject"], underTest.equals(JSObject()));
+  Expect.equals(
+    globalContext["eqFunction"],
+    underTest.equals(globalContext["foo"]),
+  );
+}
--- a/LibTest/js_interop/JSAnyOperatorExtension/equals_lib.dart
+++ b/LibTest/js_interop/JSAnyOperatorExtension/equals_lib.dart
@@ -6,7 +6,7 @@
 /// The result of `this == other` in JavaScript.
 ///
 /// @description Check that `equals` returns result of `this == other` in
-/// JavaScript. Test an empty JS object as `this`.
+/// JavaScript
 /// @author sgrekhov22@gmail.com
 
 import 'dart:js_interop';
--- a/LibTest/js_interop/JSAnyOperatorExtension/equals_t01.dart
+++ b/LibTest/js_interop/JSAnyOperatorExtension/equals_t01.dart
@@ -1,0 +1,20 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion JSBoolean equals( JSAny? any )
+/// The result of `this == other` in JavaScript.
+///
+/// @description Check that `equals` returns result of `this == other` in
+/// JavaScript. Test an empty JS object as `this`.
+/// @author sgrekhov22@gmail.com
+
+import 'dart:js_interop';
+import '../../../Utils/expect.dart';
+import 'equals_lib.dart';
+
+void main() {
+  var underTest = JSObject();
+  testEquals(underTest, "{}");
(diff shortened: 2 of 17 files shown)


eernstg:
I can see that this PR uses the same approach as https://github.com/dart-lang/co19/pull/3216 where I asked why some tests had two `eval` invocations and executed initialization of the global environment multiple times. So the questions there would be relevant for this PR as well.

eernstg (approved):
LGTM, with a couple of questions about the fact that certain declarations are evaluated more than once.

@srujzs, here's another one. Do tell us if we need to find other interop expert reviewers!

osa1:
> It looks like the WASM behavior differs from the expected behavior

@eernstg which lines/tests do you mean? I run all of the tests here and they pass with dart2wasm.

osa1 (approved):
The tests cases LGTM, but similar @srujzs in https://github.com/dart-lang/co19/pull/3216#issuecomment-2960768440, I also wonder if this might be overkill. `JSAny` is just a reference to a JS object, and `operator ==` simply call JS `===`: https://github.com/dart-lang/sdk/blob/b67482d6bc21fd7462c98122086eae56450d7d11/sdk/lib/_internal/wasm/lib/js_helper.dart#L158-L159

I suspect it should be the same in dart2js as well.

So if Dart to JS and JS to Dart conversions are correct, then `operator ==` cannot be wrong.

sgrekhov:
> > It looks like the WASM behavior differs from the expected behavior
> 
> @eernstg which lines/tests do you mean? I run all of the tests here and they pass with dart2wasm.

I think it's just a wrong PR. @osa1 please, take a look at https://github.com/dart-lang/co19/pull/3221#discussion_r2138806570
Let's do what @srujzs suggests in a file `LibTest/js_interop/JSAnyOperatorExtension/divide_t01.dart` on the strings 39-40. Let's remove `dartify()` from them. Then the test starts passing with JS and fail with WASM. Is this expected? 

For me it's strange that it passes on JS. `{} / 2` produces `NaN` in JS and without `dartify()` we are comparing two JS objects but `NaN == NaN` is `false` in JS. So, the test should fail. Is this an issue or I'm missing anything?

Another example.
```dart
main() {
  Expect.isTrue(true.toJS.isTruthy);
}
```
This works with JS but fails with WASM. Adding `.dartify()` fixes it. But is this as designed and expected?

eernstg:
@osa1 wrote:
> I think the output is as expected ...

Thanks, @osa1, that's very helpful!

sgrekhov:
Hmm... According to @srujzs `JSBoolean` is extension type on `bool` and this explains why `Expect.isTrue(true.toJS.isTruthy);` works on JS without additional `.dartify()`. Probably it makes sense to update documentation. Please see https://api.dart.dev/dart-js_interop/JSBoolean-extension-type.html

> JSBoolean extension type
> A JavaScript boolean.
> 
> on
> JSBooleanRepType
> Implemented types
> [JSAny](https://api.dart.dev/dart-js_interop/JSAny-extension-type.html)

And there's no any information about `JSBooleanRepType` which is extension type on Dart `bool` I guess (probably not directly).

osa1:
@sgrekhov do you have the link to @srujzs's comment on this?

This behavior is the same in `JSNumber` as well, i.e. if you do `x == y` where `x` is `JSNumber` and `y` is `double`, they will never be equal in dart2wasm but they will be in dart2js. Example:

```dart
import 'dart:js_interop';
import 'expect.dart';

@JS('eval')
external void eval(String s);

@JS('foo')
external JSNumber foo();

main() {
  eval('''
    globalThis.foo = function() {
      return 123;
    };
  ''');

  print(foo());
  print(foo() == 123.0); // true in dart2js, false in dart2wasm
  print(foo() == 123);  // true in dart2js, false in dart2wasm
}
```

The reason is because in dart2wasm we have to box the JS references returned from JS calls, and once you box the references you can implement your own `operator ==` where you check that the types are compatible.

In dart2js you don't have to box, and boxing would cost you performance. So if we make dart2js work like dart2wasm here, `js_interop` in dart2js will get slower, and we probably cannot afford that.

So we have three options:

1. Accept this difference in `js_interop` semantics.

2. Make dart2js work like dart2wasm: this would cost dart2js performance as it would have to box returned JS references to be able to implement a `operator ==` that checks types and distinguishes `JSBoolean` from `bool` (and `JSNumber` from `double`).

   This would cost dart2js performance, which would be unacceptable.

3. Make dart2wasm work like dart2js: this would require dartifying the reference when comparing it with a non-js-interop type.

   This would cost dart2wasm some performance, but I suspect it may be acceptable. Depending on the RHS type, we would call different `dartify` functions that can fail, so that if you compare with a `bool` you do `dartifyBool`, and if the receiver isn't a bool it fails and you return false. Same for other types.

I don't see any other options.

@srujzs have you considered this case already? Do you have any opinions on the options above?

sgrekhov:
@osa1 is there a way to detect `wasm` somehow? Currently we detect JS as `final bool isJS = identical(1.0, 1);`. Probably, depending on the above we'll need test like:
```dart
Expect.isTrue(true.toJS.isTruthy.dartify());
if (isJS) {
  Expect.isTrue(true.toJS.isTruthy);
}
if (isWasm) {
  Expect.isFalse(true.toJS.isTruthy);
}
```

osa1:
@sgrekhov `identical(1.0, 1)` is the only way to do this in runtime that I'm aware of.

srujzs:
> And there's no any information about JSBooleanRepType which is extension type on Dart bool I guess (probably not directly).

It is an extension type on Dart bool directly in the JS compilers. Re: documentation, I agree it might be helpful to be clearer about the representation type differences, but I don't want to specify that the representation type is a specific type, because that may and should be allow to change. https://api.dart.dev/dart-js_interop/ specifies the runtime type differences and that `identical` checks won't work, but perhaps it should also mention that one shouldn't do `==` checks (or anything that calls that like `Expect.equals`) between a JS value and a Dart value.

My comment [here](https://github.com/dart-lang/co19/pull/3221/files/c865ca7f44635a6fd444b7d453fff3c07b57b841#r2138806570) about not using `dartify` is meant for the case where the two values are JS values, in which case `==` is okay and will do the sensible thing of checking the underlying values and not the boxes.

> Do you have any opinions on the options above?

We could lint as a fourth option. We already lint for `is` and `as` in cases we think may result in platform differences. We could consider linting `==` and `Expect.equals` when the static types are different. With some of the more dynamic expectation testing in `package:test` (like passing a JS value as a "matcher" argument), it may be harder to catch though.

> const bool isDart2Wasm = bool.fromEnvironment('dart.tool.dart2wasm');

I think this would be the preferable way to check for the compiler.

osa1:
Thanks. So we accept this difference in semantics as expected.

Do we want to test `bool` vs. `JSBoolean` comparisons (and for other types) in this repo?

Given that this is not a specified behavior of `js_interop` (backends are allowed to do different things), I wonder if this should be tested in the SDK, some directory specific to dart2js, rather than in co19 as a language spec test.

sgrekhov:
> const bool isDart2Wasm = bool.fromEnvironment('dart.tool.dart2wasm');

It works, thank you! 

>  https://api.dart.dev/dart-js_interop/ specifies the runtime type differences and that identical checks won't work, but perhaps it should also mention that one shouldn't do == checks (or anything that calls that like Expect.equals) between a JS value and a Dart value.

Now there is a note that one _should_ do `==`.

> The types defined in this library only provide static guarantees. The runtime types differ based on the backend, so it is important to rely on static functionality like the conversion functions, for example toJS and not runtime mechanisms like type checks (is) and casts (as). Similarly, identical may return different results for the same JS value depending on the compiler. **Use == to check for equality of two JS types instead.**

I still don't have a clear understanding of what kind of tests we should write.
1) **Equals and `dartify`.**
  If the fourth option is preferrable (make a lint about `==`) then how should we test that, say, `globalContext` returns a right value? Currently, the test below (without `dartify()`) works in `dart2js` but fails in `dart2wasm`.
```dart
main() {
  eval(r'''
    var jsVar = "jsVar";
  ''');
  Expect.equals("jsVar", globalContext["jsVar"]);
}
```
We could instead write something like:
```dart
main() {
  eval(r'''
    var jsVar = "jsVar";
  ''');
  Expect.equals("jsVar", globalContext["jsVar"].dartify());
  if (isJS) {
    Expect.equals("jsVar", globalContext["jsVar"]);
  }
  if (isWasm) {
    Expect.notEquals("jsVar", globalContext["jsVar"]); // If one day this fail it'd mean that the lint also should be removed
  }
}
```
@srujzs @osa1 @eernstg do we need the tests like the above? Or this, probably, makes sense for `==` operator only? And for all other cases just let's compare dartified values?

Honestly, I'd vote for option 3: to make `dart2wasm` and `dart2js` behave the same way. The code below looks very strange to me and requires a strong understanding of the internal workings of `dart2wasm`:
```dart
  // On dart2wasm
  print(globalContext["jsVar"]); // prints 'jsVar'
  print(globalContext["jsVar"] == "jsVar"); // false
```
Just saw @osa1's note that we’ve accepted the fact that the semantics of `==` differ between `dart2js` and `dart2wasm`. Ok, thank you!

2. **Tests for `JSAnyOperatorExtension`.**
If all operations defined in `JSAnyOperatorExtension` are straightforward and simple, then I’ll simplify the tests. Just one check for each primitive type, array, function, and object. Don't check for specific values like `42`, `0`, `-1`, `3.13`, `-0.0`; just verify that `42 + ""` in JS produces the same result as `42.toJS.add("".toJS)` in Dart. Use `dartify()` to compare the results (otherwise, we'll get different results on different compilers). For `==` check the results with `dartify()` and without it.

Does that work for you?

sgrekhov:
> So the tests would then be simplified considerably. Presumably that would be done by adding more commits to this PR, as well as a few others?

No. I created https://github.com/dart-lang/co19/pull/3223 as a replacement for this and the others. Please review it. If that PR is landed, I'll close this and the other PRs.

sgrekhov:

PR #3229, opened 2025-06-16
#3180. Add `JSArray` tests.

sgrekhov:
`from_t0*` also check `[]=` and `[]` operators. I don't think that we need any separate tests for them.

### DIFF
--- a/LibTest/js_interop/JSArray/from_t02.dart
+++ b/LibTest/js_interop/JSArray/from_t02.dart
@@ -1,0 +1,25 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion JSArray<T> from<T extends JSAny>( JSObject arrayLike )
+/// Creates a new, shallow-copied JavaScript `Array` instance from a JavaScript
+/// iterable or array-like object.
+///
+/// @description Checks that this constructor creates a new, shallow-copied
+/// JavaScript `Array` instance from a JavaScript array object.
+/// @author sgrekhov22@gmail.com
+/// @issue 60934
+
+import 'dart:js_interop';
+import '../../../Utils/expect.dart';
+
+main() {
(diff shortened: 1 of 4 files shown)


LibTest/js_interop/JSArray/from_t02.dart
```diff
@@ -0,0 +1,25 @@
...
+import '../../../Utils/expect.dart';
+
+main() {
+  JSArray<JSArray<JSNumber>> ar = JSArray<JSArray<JSNumber>>.withLength(1);
+  JSArray<JSNumber> val = JSArray<JSNumber>.withLength(1);
+  val[0] = 42.toJS;
+  ar[0] = val;
+  JSArray<JSArray<JSNumber>> copy = JSArray.from(ar);
+  Expect.equals(val, copy[0]);
+  Expect.identical(val, copy[0]);
```
srujzs, line 24: Does this `identical` check work in dart2wasm? I would assume you would get an error because you get a new box. At any rate, I'd avoid doing an `identical` check as we warn users against doing that.

LibTest/js_interop/JSArray/withLength_t01.dart
```diff
@@ -0,0 +1,35 @@
...
+
+  JSArray<JSString> a1 = JSArray<JSString>.withLength(1);
+  Expect.equals(1, a1.length);
+  globalContext["a1"] = a1;
+  eval(r'''
+    var b1 = a1.length === 1;
+    var v1 = a1[0];
+    ''');
+  Expect.isTrue((globalContext["b1"] as JSBoolean).toDart);
+  Expect.isNull(globalContext["v1"].dartify());
```
srujzs, line 34: Ooh, this one is interesting. JS specifies that you actually get "empty slots" when you create an array, and neither `null` nor `undefined`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Indexed_collections#sparse_arrays. I believe the access with `[0]` is what coerces it to `undefined`.

You can remove the `dartify` call here as `null`-ish values are always converted/coerced whenever they cross the interop boundary.

srujzs (approved):
A few comments, but generally LGTM!

eernstg (approved):
LGTM

PR #3236, opened 2025-06-25
#3180. Add more JS arrays tests

sgrekhov:


srujzs (approved):
LGTM!

eernstg (approved):
LGTM

It occurred to me that `byteOffset` could be ignored (so the actual JavaScript "typed arrays" could all use a byte offset of 0, or some other fixed value), and the tests would still succeed. Is that true? Is there a need to test this, or is it a JavaScript-internal property where we'd just trust JavaScript because this is only concerned with testing the glue between the Dart program and the underlying JavaScript?

I'll land this in any case. If there is something extra to test then it can be done in a different PR.

PR #3237, opened 2025-06-26
#3180. Add JSUintXArray tests

sgrekhov:


sgrekhov:
As far as I understand `byteOffset` controls where exactly in the `buffer` data should be stored. It's offset in the buffer, not in the array. If we have buffer `[1, 2]` and create `Int8Array(buffer, 1, 1)`, then after `a[0] = 42;` buffer will became `[1, 42]`. I think it makes sense to add an appropriate test.

eernstg:
Right, the point would be that we can trust the `byteOffset` to be used properly if we create a buffer and use that with two different typed arrays with different offsets (say 8 and 16), and then write to one of them from index 1 and read it using the other typed array at index 0, and then we'd expect to get the value we just wrote.

PR #3247, opened 2025-07-03
#3180. Add `NullRejectionException` tests

sgrekhov:
There is also a public constructor `NullRejectionException.new` but I didn't find a way how it can be used. @srujzs do you know it? Does it makes sense to make it public?


srujzs (approved):

srujzs:
I don't see a reason to make it public. Maybe if users want to throw the exception themselves in the Dart code that gets converted to a `Promise` and eventually back to a `Future`? I doubt there's any useful case there, so let's mark it private. https://github.com/dart-lang/sdk/issues/61054

PR #3249, opened 2025-07-04
#3180. Add additional `NullRejectionException` test

sgrekhov:


LibTest/js_interop/NullRejectionException/nullRejectionException_t03.dart
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Exception for when the promise is rejected with a `null` or
+/// `undefined` value.
+///
+/// @description Checks that this exception doesn't throw occurs when the
```
eernstg, line 8: Perhaps:
```suggestion
/// @description Checks that this exception isn't thrown when the
```

LibTest/js_interop/NullRejectionException/nullRejectionException_t03.dart
```diff
@@ -0,0 +1,43 @@
...
+/// promise is rejected with not a `null` or `undefined` value.
```
eernstg, line 9: Perhaps more readable:
```suggestion
/// promise is rejected with not a value which is not `null` or `undefined`.
```

eernstg (approved):

LibTest/js_interop/NullRejectionException/nullRejectionException_t02.dart
```diff
@@ -27,7 +27,9 @@ main() {
     Expect.fail("NullRejectionException expected");
   }).onError((e, st) {
     Expect.isTrue(e is NullRejectionException);
-    Expect.isTrue((e as NullRejectionException).isUndefined);
+    if (!isWasm) {
```
osa1, line 30: Do we have a `isJs`? That would make this easier to read without the negation: `if (isJs) { ... }`.

LibTest/js_interop/NullRejectionException/nullRejectionException_t02.dart
```diff
@@ -27,7 +27,9 @@ main() {
...
+      Expect.isTrue((e as NullRejectionException).isUndefined);
```
osa1, line 31: Why do we not test this with dart2wasm? We have a similar test in the SDK test suite: https://github.com/dart-lang/sdk/blob/84e279d27968c1555c80411f6fbcdf994ef1afe5/tests/lib/js/js_util/promise_reject_null_test.dart#L59-L66

Is this an issue with dart2wasm?

LibTest/js_interop/NullRejectionException/nullRejectionException_t02.dart
sgrekhov: Replaced with `isJS`

LibTest/js_interop/NullRejectionException/nullRejectionException_t02.dart
sgrekhov: On `dart2wasm`, `isUndefined` always returns `false`, and I skipped it because this behavior is not documented in `isUndefined`. However, the fact that on Wasm `null` and `undefined` are always treated as `null` is documented in `NullableUndefineableJSAnyExtension`. So let's test it. Thank you.

LibTest/js_interop/NullRejectionException/nullRejectionException_t03.dart
sgrekhov: Updated

LibTest/js_interop/NullRejectionException/nullRejectionException_t03.dart
sgrekhov: Updated

sgrekhov (commented):
Thank you! Updated. PTAL.

LibTest/js_interop/NullRejectionException/nullRejectionException_t02.dart
osa1: But there's no `NullableUndefineableJSAnyExtension` in the code being tested here. The exception is defined here: https://github.com/dart-lang/sdk/blob/84e279d27968c1555c80411f6fbcdf994ef1afe5/sdk/lib/js_util/js_util.dart#L137-L148

Note that `isUndefined` is a `bool` field. It doesn't use any extensions.

The class is instantiated here: https://github.com/dart-lang/sdk/blob/84e279d27968c1555c80411f6fbcdf994ef1afe5/sdk/lib/_internal/wasm/lib/js_util_patch.dart#L181-L184

Note that we have an `externref` here, and we can distinguish `undefined`s and `null`s in `externref`s: https://github.com/dart-lang/sdk/blob/84e279d27968c1555c80411f6fbcdf994ef1afe5/sdk/lib/_internal/wasm/lib/js_helper.dart#L139

I can see that the comment in the allocation site says we can't distinguish them, but it's wrong. Maybe it was correct when it was written, I don't know.

We should be able to easily fix this test, with something like: (not tested)

```diff
diff --git a/sdk/lib/_internal/wasm/lib/js_util_patch.dart b/sdk/lib/_internal/wasm/lib/js_util_patch.dart
index bb5cec8069d..d465065f207 100644
--- a/sdk/lib/_internal/wasm/lib/js_util_patch.dart
+++ b/sdk/lib/_internal/wasm/lib/js_util_patch.dart
@@ -180,7 +178,8 @@ Future<T> promiseToFuture<T>(Object jsPromise) {
     // in Dart.
     final e = dartifyRaw(jsError.toExternRef);
     if (e == null) {
-      return completer.completeError(NullRejectionException(false));
+      return completer.completeError(
+          NullRejectionException(isJSUndefined(jsError.toExternRef)));
     }
     return completer.completeError(e);
   }).toJS;
```

In the meantime the special case here for dart2wasm can be removed. I don't see any reason the `isUndefined` field here should be different than dart2js here.

sgrekhov: > But there's no NullableUndefineableJSAnyExtension in the code being tested here

Sorry, I wasn’t clear enough. By "let’s test it" I meant that we should test that `NullRejectionException.isUndefined` always returns `false` on Wasm. And that’s the reason we need a special code branch for Wasm in this test.

eernstg:
Oh, and @srujzs, you may have further comments when you're back from vacation?

LibTest/js_interop/NullRejectionException/nullRejectionException_t02.dart
osa1: Why should `NullRejectionException.isUndefined` always return `false`?

sgrekhov: That’s how it works now, and I believe this behavior should be expected. According to https://github.com/dart-lang/sdk/blob/d3c323e26d1663ab422ff89b65d5a78b0e89c997/sdk/lib/js_interop/js_interop.dart#L465-L468
`undefined` is converted to Dart `null` on Wasm, which is why `isUndefined` returns `false`.

osa1: But I explained in [my comment above][1] that this field (`isUndefined`) doesn't use the extension you link. I also showed the function in dart2wasm to test whether a JS value is `undefined`.

Also, when writing tests, we test against the spec not the implementation, because we want to test spec compliance.

What does the relevant spec say? Is it not possible to implement the spec in dart2wasm?

[1]: https://github.com/dart-lang/co19/pull/3249#discussion_r2185614066

sgrekhov: > But I explained in https://github.com/dart-lang/co19/pull/3249#discussion_r2185614066 that this field (isUndefined) doesn't use the extension you link. I also showed the function in dart2wasm to test whether a JS value is undefined.

I thought that that comment in `NullableUndefineableJSAnyExtension` is not about this extension but about every Wasm to Dart conversion. 

If you believe that the test is wrong let's change it and fill an SDK issue. @srujzs do you agree?

osa1: I did some more digging, I think we'll have to accept the current behavior and test accordingly.

Normally when we get an `externref` from a JS call we can distinguish JS `null`s and `undefined`s. That's what the `isJSUndefined` function is for.

But when we box an `externref` to convert it into `JSAny`, we convert it to `JSAny?`, i.e. make it a nullable value, and use Dart `null` for both JS `undefined` and `null` values.

This is probably for performance reasons: we don't want to allocate values for `undefined`s and `null`s.

But it also means we can't distinguish `undefined`s and `null`s.

So marking this as resolved. Sorry for the noise.

sgrekhov: @osa1 thank you for the review and analisys!

eernstg:
@osa1, it looks like we could land this now. There would then be a failure in 'nullRejectionException_t02.dart' when the update is rolled into the SDK and until https://dart-review.googlesource.com/c/sdk/+/439141 lands. This failure would be approved, so it shouldn't cause any disruptions. Do you agree?

osa1:
@eernstg correct, we can merge this now. The CL will fix the issue.,

I think the current behavior of `isUndefined` is definitely not right. We either want to fix it and return the right value (as in https://dart-review.googlesource.com/c/sdk/+/439141), or throw the same error as in `NullableUndefineableJSAnyExtension.{isUndefined,isNull}`. I added a comment about this in the CL. If we decide throw we can update this test.

PR #3251, opened 2025-07-08
#3180. Add JSExport tests. Part 2.

sgrekhov:
There are some tests in this PR that check the `@JSExport` annotation when applied to extension types and extensions. The current behavior is that this annotation is a no-op when applied to an extension type or extension, and it produces a compile-time error when a member of an extension type or extension is annotated with this annotation. The tests are written accordingly. Please review.

### DIFF
--- a/LibTest/js_interop/JSExport/jsExport_A01_t07.dart
+++ b/LibTest/js_interop/JSExport/jsExport_A01_t07.dart
@@ -1,0 +1,68 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Annotation to allow Dart classes to be wrapped with a JS object
+/// using `dart:js_interop`'s `createJSInteropWrapper`.
+///
+/// When an instance of a class annotated with this annotation is passed to
+/// `createJSInteropWrapper`, the method returns a JS object that contains a
+/// property for each of the class' instance members. When called, these
+/// properties forward to the instance's corresponding members.
+///
+/// @description Checks that when an instance of an extension type is annotated
+/// with `@JSExport()` and passed to `createJSInteropWrapper`, the method
+/// returns a JS object that contains a property for each of the representation
+/// type's instance members.
+/// @author sgrekhov22@gmail.com
+
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+import '../../../Utils/expect.dart';
+import '../js_utils.dart';
+
+String log = "";
+
+@JSExport()
+class C {
+  int variable = 42;
+  String method(String v) => "method($v);";
+  String get getter => "Some getter";
+  void set setter(bool value) {
+    log = "setter($value);";
+  }
+}
+
+@JSExport()
+extension type ET(C c) implements C {
+  String etMethod(String v) => "etMethod($v);";
+  String get etGetter => "Some etGetter";
+  void set etSetter(bool value) {
+    log = "etSetter($value);";
+  }
+}
+
+void main() {
+  var jsET = createJSInteropWrapper<C>(ET(C()));
+  globalContext["jsET"] = jsET;
+  eval(r'''
+    globalThis.v1 = globalThis.jsET.variable;
+    globalThis.v2 = globalThis.jsET.method('x');
+    globalThis.v3 = globalThis.jsET.getter;
+    globalThis.jsET.setter = false;
+
+    globalThis.v4 = globalThis.jsET.c;
+    globalThis.v5 = globalThis.jsET.etMethod;
+    globalThis.v6 = globalThis.jsET.etGetter;
+    globalThis.v7 = globalThis.jsET.etSetter;
+  ''');
+  Expect.equals(42, (globalContext["v1"] as JSNumber).toDartInt);
+  Expect.equals("method(x);", (globalContext["v2"] as JSString).toDart);
+  Expect.equals("Some getter", (globalContext["v3"] as JSString).toDart);
+  Expect.equals("setter(false);", log);
+
+  Expect.isNull(globalContext["v4"]);
+  Expect.isNull(globalContext["v5"]);
+  Expect.isNull(globalContext["v6"]);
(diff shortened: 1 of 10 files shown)


LibTest/js_interop/JSExport/jsExport_A02_t04.dart
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion You can either annotate specific instance members to only wrap
+/// those members or you can annotate the entire class, which will include all
+/// of its instance members.
+///
+/// @description Checks that it is a compile-time error to annotate with
+/// `JSExport()` an extension type instance members.
```
eernstg, line 10: Probably more readable:
```suggestion
/// @description Checks that it is a compile-time error to annotate extension
/// type instance members with `JSExport()`.
```

LibTest/js_interop/JSExport/jsExport_A03_t01.dart
```diff
@@ -0,0 +1,58 @@
...
+  eval(r'''
+    globalThis.v1 = globalThis.jsC.variable;
+    globalThis.v2 = globalThis.jsC.method('x');
+    globalThis.v3 = globalThis.jsC.getter;
+    globalThis.jsC.setter = false;
+
+    globalThis.v4 = globalThis.jsC.annotatedVariable;
+    globalThis.v5 = globalThis.jsC.annotatedMethod;
+    globalThis.v6 = globalThis.jsC.annotatedGetter;
+    globalThis.v7 = globalThis.jsC.annotatedSetter;
```
eernstg, line 47: We wouldn't expect to get a method or getter named `annotatetSetter`. Doesn't that imply that this potential setter should be tested like other setters?
```suggestion
    globalThis.jsC.annotatedSetter = false;
```
.. and then checking the `log` later.

LibTest/js_interop/JSExport/jsExport_A01_t07.dart
```diff
@@ -0,0 +1,68 @@
...
+extension type ET(C c) implements C {
+  String etMethod(String v) => "etMethod($v);";
+  String get etGetter => "Some etGetter";
+  void set etSetter(bool value) {
+    log = "etSetter($value);";
+  }
+}
+
+void main() {
+  var jsET = createJSInteropWrapper<C>(ET(C()));
```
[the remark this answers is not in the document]
eernstg: OK, but given that there is absolutely no difference between an instance of `C` and an instance of `C` which has been given the static type `ET` and upcasted to `C`, I'm not convinced that the use of `ET` shows anything. ;-)

LibTest/js_interop/JSExport/jsExport_A01_t07.dart
```diff
@@ -0,0 +1,68 @@
...
+class C {
+  int variable = 42;
+  String method(String v) => "method($v);";
+  String get getter => "Some getter";
+  void set setter(bool value) {
+    log = "setter($value);";
+  }
+}
+
+@JSExport()
```
srujzs, line 36: We could expect a warning here because I don't see a good reason why we shouldn't.

It's a no-op already, so that leads me to believe that this would only add confusion if allowed.

LibTest/js_interop/JSExport/jsExport_A01_t08.dart
```diff
@@ -0,0 +1,66 @@
...
+class C {
+  int variable = 42;
+  String method(String v) => "method($v);";
+  String get getter => "Some getter";
+  void set setter(bool value) {
+    log = "setter($value);";
+  }
+}
+
+@JSExport()
```
srujzs, line 36: Same comment as for an extension type.

LibTest/js_interop/JSExport/jsExport_A03_t02.dart
```diff
@@ -0,0 +1,30 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion By default, the property will have the same name as the
+/// corresponding instance member. You can change the property name of a member
+/// in the JS object by providing a `name` in the `@JSExport` annotation on the
+/// member.
+///
+/// @description Checks that the `createJSInteropWrapper` throws if an argument
+/// is a class annotated with  a `@JSExport` annotation with a specified `name`.
```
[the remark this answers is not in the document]
srujzs: We print a warning when the class has a non-empty `@JSExport` annotation. We should probably expect that in this test.

srujzs (approved):
LGTM % my comments.

LibTest/js_interop/JSExport/jsExport_A01_t07.dart
eernstg: Very good! @sgrekhov, do we have a known and working way to detect these warnings? If we can do it then I'd recommend that the warning is expected and the test failure is approved when it's rolled, with a reference to an issue about implementing this warning. Similarly for the `extension` case below.

sgrekhov: I filled https://github.com/dart-lang/sdk/issues/61082 and updated the test to expect an `unspecified` error (this also covers warnings).

LibTest/js_interop/JSExport/jsExport_A01_t08.dart
sgrekhov: Updated.

LibTest/js_interop/JSExport/jsExport_A03_t02.dart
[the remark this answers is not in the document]
sgrekhov: Updated to expect a warning. Also filled https://github.com/dart-lang/sdk/issues/61084

LibTest/js_interop/JSExport/jsExport_A01_t07.dart
eernstg: OK, thanks!

eernstg (approved):
LGTM.

PR #3253, opened 2025-07-10
#3180. Add JSExport tests. Part 4.

sgrekhov:


sgrekhov:
I guess that `typedef` tests in this PR are wrong and we simply should expect an error/warnings. But, please, confirm first. And I'll update the tests accordingly if needed.

LibTest/js_interop/JSExport/jsExport_A06_t05.dart
```diff
@@ -0,0 +1,46 @@
...
+class C {
+  int variable = 42;
+  String method(String v) => "method($v);";
+  String get getter => "Some getter";
+  void set setter(bool value) {
+    log = "setter($value);";
+  }
+}
+
+@JSExport()
```
srujzs, line 29: This could be a warning so users don't think it makes a difference, but I don't even know if this declaration appears in the CFE, so we might not be able to bubble up a warning here.

IIRC, typedefs are resolved early, so we would never even see `CAlias` as a type in the kernel.

LibTest/js_interop/JSExport/jsExport_A06_t03.dart
```diff
@@ -0,0 +1,42 @@
...
+
+@JSExport()
+int variable = 42;
+//  ^^^^^^^^
+// [analyzer] unspecified
+// [web] unspecified
+
+@JSExport()
+void func() {}
+//   ^^^^
```
srujzs, line 22: More of a meta-level question, once these get rolled into the SDK, are we skipping them and linking to the issue to implement them next to the skip?

It would be nice to add these warnings/errors for unrelated declarations, but I don't think I'll get to it soon.

srujzs (approved):
LGTM!

eernstg (commented):
Looks good! I think it would be helpful to add a comment about the tests that seem to assume that we can do `super` invocations of members of another object (like `otherObject.super.m()`, or even `otherObject.super<SomeType>.m()` because every superinvocation must specify a class from which the search should start, except that this is implicit when the superinvocation occurs inside that class).

One more thing came up, and I asked for input from @johnniwinther in order to clarify the situation around type aliases and Kernel code.

LibTest/js_interop/JSExport/jsExport_A06_t06.dart
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Only concrete instance members can and will be wrapped, and it's
+/// an error to annotate other members with this annotation.
+///
+/// @description Checks that it is a compile-time error if a class is not
+/// annotated with `@JSExport()`, but its type alias is, and that alias is
+/// passed as a type argument to `createJSInteropWrapper`.
```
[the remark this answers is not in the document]
sgrekhov: Currently compiler produces `Error: Class 'C' does not have a `@JSExport` on it or any of its members.` here.

LibTest/js_interop/JSExport/jsExport_A06_t04.dart
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Only concrete instance members can and will be wrapped, and it's
+/// an error to annotate other members with this annotation.
+///
+/// @description Checks that it is a compile-time error if a local variables or
```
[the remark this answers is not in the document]
sgrekhov: Fixed

LibTest/js_interop/JSExport/jsExport_A05_t01.dart
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion If members are overridden, only the overriding member will be
+/// wrapped as long as it or its class has this annotation.
+///
+/// @description Checks that if members are overridden, only the overriding
+/// members are wrapped.
```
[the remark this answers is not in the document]
sgrekhov: Added more info to the description.

sgrekhov (commented):
Typos fixed and the description updated. PTAL.

LibTest/js_interop/JSExport/jsExport_A06_t07.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Only concrete instance members can and will be wrapped, and it's
+/// an error to annotate other members with this annotation.
+///
+/// @description Checks that it is a compile-time error an old-fashioned
+/// function type is annotated with `JSExport()`.
```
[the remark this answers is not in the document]
eernstg: @johnniwinther is on vacation, will be back on Monday. I've sketched a way to proceed in a top-level comment of this PR.

eernstg:
We could proceed as follows:

All the errors/warnings that we'd ideally want to have are expected in these tests.

Some of those expectations may fail because the implementation of the diagnostics hasn't been completed at this time. Others may fail because it's difficult to do at all (e.g., type aliases may just be gone completely in Kernel code, and it might be a major refactoring effort to change that).

So we'd just approve all the failures with references to suitable issues.

Some tests would then start working as soon as the diagnostics have been implemented. For other tests, we'd clarify the situation; if the desired diagnostic turns out to be impractical to emit (e.g., errors/warnings about type aliases) then the test expectations would be adjusted accordingly.

With a plan along those lines, we could land this PR now.

@sgrekhov, is it true that every desired diagnostic message is expected at this time, or do we have some locations where no message is expected even though we'd like to have it?

LibTest/js_interop/JSExport/jsExport_A06_t06.dart
```diff
@@ -0,0 +1,27 @@
...
+/// @author sgrekhov22@gmail.com
+
+import 'dart:js_interop';
+
+class C {
+  int variable = 42;
+}
+
+@JSExport()
+typedef CAlias = C;
```
eernstg, line 20: If we follow the plan I suggested in the top-level comment then we'd expect to get a warning on `CAlias` because that would be a meaningful and useful behavior. This will fail for now, but the plan would be to clarify the situation and then adjust the test if it turns out to be impossible to get that warning.

LibTest/js_interop/JSExport/jsExport_A06_t07.dart
```diff
@@ -0,0 +1,23 @@
...
+/// @author sgrekhov22@gmail.com
+/// @issue 61093
+
+import 'dart:js_interop';
+
+@JSExport()
+typedef void Foo();
+//           ^^^
+// [analyzer] unspecified
+// [web] unspecified
```
eernstg, line 19: This expectation is good: It fails for now, and we may not be able to fix it, but right now we just want to expect all the diagnostic messages that we _want_ to have.

sgrekhov:
In this PR `jsExport_A06_t04` and `jsExport_A06_t07` have no expected diagnostic messages. See https://github.com/dart-lang/sdk/issues/61093. All other tests are passing, but I’m also keeping in mind that we’re waiting for clarification regarding annotated typedefs. For now, the typedef-related tests are passing, but we’ll likely need to update them to expect a warning and create a corresponding SDK issue.

eernstg (commented):
Added a couple of comments. I found one location where I believe we should expect a diagnostic message (following the plan I suggested).

eernstg:
> In this PR `jsExport_A06_t04` and `jsExport_A06_t07` have no expected diagnostic messages

OK, so they would be in the same category: We want those diagnostic messages because `@JSExport(...)` can't possibly be useful on a local declaration (I think! ;-).

sgrekhov:
Tests updated to expect a warning in case of typedef annotated with `@anonymous`. https://github.com/dart-lang/sdk/issues/61116

LibTest/js_interop/JSExport/jsExport_A06_t05.dart
```diff
@@ -5,14 +5,12 @@
 /// @assertion Only concrete instance members can and will be wrapped, and it's
 /// an error to annotate other members with this annotation.
 ///
-/// @description Checks that alias of a class annotated with `@JSExport()` also
-/// can be annotated with `@JSExport()`.
+/// @description Checks that it is a warning error if a type alias is  annotated
```
eernstg: Would that be?:
```suggestion
/// @description Checks that it is a warning or an error if a type alias is  annotated
```

LibTest/js_interop/JSExport/jsExport_A06_t06.dart
```diff
@@ -5,10 +5,11 @@
 /// @assertion Only concrete instance members can and will be wrapped, and it's
 /// an error to annotate other members with this annotation.
 ///
-/// @description Checks that it is a compile-time error if a class is not
-/// annotated with `@JSExport()`, but its type alias is, and that alias is
-/// passed as a type argument to `createJSInteropWrapper`.
+/// @description Checks that it is a warning error if a class is not annotated
```
eernstg: Ditto.

LibTest/js_interop/JSExport/jsExport_A06_t05.dart
eernstg: No further comments, I'll just fix this.

eernstg (approved):
LGTM, landing!

LibTest/js_interop/JSExport/jsExport_A06_t07.dart
johnniwinther: It should be possible for backends to report errors on typedef *declarations* but not uses through a typedef. So it should be possible to report an error here.

PR #3260, opened 2025-07-22
#3180. Add `instanceof` and `instanceOfString` tests

sgrekhov:
I also tried to test the case where `constructorName` contains a `.` but wasn’t successful. @srujzs, do you know how to import a JS class from a library into the tests? Any advice is appreciated. I tried the approach below, but it doesn’t work (`module.js` contains definitions for classes `A` and `B`).
```dart
// OtherResources=module.js
import 'dart:async';
import 'dart:js_interop';
import 'dart:js_interop_unsafe';
import '../../../Utils/expect.dart';
import '../js_utils.dart';

final completer = Completer<String>();

void complete(String value) {
  completer.complete(value);
}

main() async {
  globalContext["complete"] = complete.toJS;
  eval(r'''
    (async () => { 
      const lib1 = await import('./module.js');
      globalThis.objA = new lib1.A(42, "A form JS");
      globalThis.objB = new lib1.B();
    })().then(function(v) {
      globalThis.complete("");
    });
  ''');
  asyncStart();
  completer.future.then((_) {
    Expect.isTrue(globalContext["objA"].instanceOfString("lib1.A"));
    asyncEnd();
  });
}
```


sgrekhov:
Updated. PTAL.

srujzs (approved):

PR #3263, opened 2025-07-24
#3180. Add `isA` tests. Part 1.

sgrekhov:


LibTest/js_interop/JSAnyUtilityExtension/isA_A02_t01.dart
```diff
@@ -0,0 +1,58 @@
...
+main() {
+  eval(r'''
+    globalThis.s = "JS string";
+    globalThis.b = false;
+    globalThis.n1 = 42;
+    globalThis.n2 = 3.14;
+  ''');
+  Expect.isTrue(globalContext["s"].isA<JSString>());
+  Expect.isTrue(globalContext["s"].isA<JSAny>());
+  Expect.isFalse(globalContext["s"].isA<JSObject>());
```
eernstg, line 29: Oh, so a JavaScript string is not a `JSObject`? Surprising!

The [`JSObject` documentation](https://api.dart.dev/dart-js_interop/JSObject-extension-type.html) does say that it "is the supertype of all JavaScript objects, but not other JS types, like primitives", but the [documentation of `JSString`](https://api.dart.dev/dart-js_interop/JSString-extension-type.html) does not indicate that `JSString` is primitive.

Should it? It could be considered to be implied by the fact that 'Implemented types' does not include `JSObject`, but that's still less than 100% explicit. @srujzs, WDYT?

LibTest/js_interop/JSAnyUtilityExtension/isA_A02_t01.dart
```diff
@@ -0,0 +1,58 @@
...
+  Expect.isFalse(globalContext["s"].isA<JSNumber>());
+  Expect.isFalse(globalContext["s"].isA<JSBoolean>());
+  Expect.isFalse(globalContext["s"].isA<JSSymbol>());
+  Expect.isFalse(globalContext["s"].isA<JSBigInt>());
+
+  Expect.isTrue(globalContext["b"].isA<JSBoolean>());
+  Expect.isTrue(globalContext["b"].isA<JSAny>());
+  Expect.isFalse(globalContext["b"].isA<JSObject>());
```
eernstg, line 37: Same consideration here.

LibTest/js_interop/JSAnyUtilityExtension/isA_A02_t01.dart
```diff
@@ -0,0 +1,58 @@
...
+  Expect.isFalse(globalContext["b"].isA<JSNumber>());
+  Expect.isFalse(globalContext["b"].isA<JSString>());
+  Expect.isFalse(globalContext["b"].isA<JSSymbol>());
+  Expect.isFalse(globalContext["b"].isA<JSBigInt>());
+
+  Expect.isTrue(globalContext["n1"].isA<JSNumber>());
+  Expect.isTrue(globalContext["n1"].isA<JSAny>());
+  Expect.isFalse(globalContext["n1"].isA<JSObject>());
```
eernstg, line 45: And here.

eernstg (commented):
Looks good! I added a couple of comments about the overall organization of the test cases: I think there is a kind of square matrix (that is "for each primitive type, test an instance against each primitive type"), but a couple of cases are missing because it is handled as several separate sets rather than one complete set of primitive types.

LibTest/js_interop/JSAnyUtilityExtension/isA_A03_t03.dart
```diff
@@ -0,0 +1,47 @@
...
+@JS()
+extension type MyArray2<T extends JSAny?>(JSArray<T> ar) implements JSArray<T> {
+}
+
+main() {
+  MyArray1 ma1 = MyArray1(JSArray());
+  globalContext["ma1"] = ma1;
+  Expect.isTrue(globalContext["ma1"].isA<JSArray>());
+  Expect.isFalse(globalContext["ma1"].isA<JSNumber>());
+  Expect.isFalse(globalContext["ma1"].isA<MyArray1>());
```
eernstg, line 36: Perhaps use `JS('Array')` (or something similar, whatever it takes) to add an extra case or two where it is true?

LibTest/js_interop/JSAnyUtilityExtension/isA_A02_t01.dart
sgrekhov: `JSBigInt` and `JSSymbol` are added to the matrix in this test.

LibTest/js_interop/JSAnyUtilityExtension/isA_A02_t01.dart
```diff
@@ -0,0 +1,58 @@
...
+    globalThis.n2 = 3.14;
+  ''');
+  Expect.isTrue(globalContext["s"].isA<JSString>());
+  Expect.isTrue(globalContext["s"].isA<JSAny>());
+  Expect.isFalse(globalContext["s"].isA<JSObject>());
+  Expect.isFalse(globalContext["s"].isA<JSNumber>());
+  Expect.isFalse(globalContext["s"].isA<JSBoolean>());
+  Expect.isFalse(globalContext["s"].isA<JSSymbol>());
+  Expect.isFalse(globalContext["s"].isA<JSBigInt>());
+
```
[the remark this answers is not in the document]
sgrekhov: > - A user-defined interop extension type as the type in isA.

Added `isA_A04_t01.dart`. But please note.
```dart
extension type ET(JSObject _) implements JSObject {
  external String foo();
}

main() {
  eval(r'''
    class ET {
      foo() {
        return "ET.foo()";
      }
    }
    globalThis.et = new ET();
  ''');
  var value = globalContext["et"] as ET;
  Expect.isFalse(value.isA<ET>()); // isA returns false for ET. Is this expected?
...
}
```

> - A nullable JS type as the type in isA.

Added `isA_A02_t03.dart`. It's a copy of `isA_A02_t01.dart`, but with nullable JS types. Please let me know if that's not what you meant.

sgrekhov (commented):
Thank you! Updated. PTAL.

LibTest/js_interop/JSAnyUtilityExtension/isA_A02_t01.dart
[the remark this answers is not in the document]
srujzs: That's expected because the class is never set in `globalThis.ET`. If it were, I'd expect it to be true.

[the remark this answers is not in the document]
sgrekhov: Thank you! Added `isA_A04_t02.dart` checking this.

eernstg (approved):
Looks like this could be ready, added a couple of questions, awaiting responses.

eernstg (approved):
LGTM

PR #3269, opened 2025-07-31
#3180. Add `@JS()` annotation tests. Part 3.

sgrekhov:


LibTest/js_interop/JS/js_A03_t05.dart
```diff
@@ -0,0 +1,47 @@
...
+
+@JS()
+external int answer();
+
+final completer = Completer<String>();
+
+void complete(String value) {
+  completer.complete(value);
+}
+
```
srujzs, line 31: nit: I'd add a test where we're using an extension type as well and make sure its static members/constructors are prefixed.

LibTest/js_interop/JS/js_A03_t05.dart
```diff
@@ -0,0 +1,47 @@
...
+@JS("lib1")
+library;
+
+import 'dart:async';
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+import '../../../Utils/expect.dart';
+import '../js_utils.dart';
+
+@JS()
```
srujzs, line 23: nit: Perhaps an example with a renamed top-level as well?

LibTest/js_interop/JS/js_A03_t02.dart
```diff
@@ -0,0 +1,51 @@
...
+/// @author sgrekhov22@gmail.com
+
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+import '../../../Utils/expect.dart';
+import '../js_utils.dart';
+
+@JS("JSEt")
+extension type ET(JSObject _) implements JSObject {
+  external String foo();
```
srujzs, line 29: nit: Perhaps an `external` constructor and factory so we know that's correctly lowered?

srujzs (approved):
Feel free to ignore my comments if they're going to be addressed in a future PR.

LibTest/js_interop/JS/js_A03_t02.dart
eernstg: Interesting! I'm not sure where we'd find anything in JavaScript which will faithfully emulate a factory constructor! So how is it even possible to lower that? ;-)

LibTest/js_interop/JS/js_A03_t05.dart
eernstg: @sgrekhov, do you plan to do this later on, or would it fit right in with this PR, or do we perhaps already have it elsewhere?

eernstg (commented):
Looks good, but a couple of questions remain.

LibTest/js_interop/JS/js_A03_t02.dart
sgrekhov: @eernstg I believe it is described here https://dart.dev/interop/js-interop/usage#interop-type-members. 

I'm going to add tests for constructors in a separate PR. As far as understand these constructirs perfectly work without `@JS()` annotation, therefore there will be tests for "general interop", not for `@JS()` annotation.

LibTest/js_interop/JS/js_A03_t05.dart
sgrekhov: I tried to do it quickly in this PR, but the test failed. It's probably an issue, or maybe I did something wrong. Let's add it later in a separate PR.

LibTest/js_interop/JS/js_A03_t05.dart
sgrekhov: Again, I'm going to do it in a separate PR for a "general interop".

sgrekhov (commented):
Thank you! A good advices! I'll implement them in a separate PR.

LibTest/js_interop/JS/js_A03_t05.dart
eernstg: OK!

LibTest/js_interop/JS/js_A03_t02.dart
eernstg: Aha, "same semantics .. regardless whether they're a factory". So they are essentially not lowered to "a factory constructor" in JavaScript, the keyword `factory` in the Dart interop code is simply ignored. Makes sense (or it could be a warning because ignoring what a developer writes may create misinterpretations of the situation).

But that will be handled later on, as you say.

PR #3270, opened 2025-08-04
#3180. Add `@JS()` annotation tests. Part 4.

sgrekhov:


LibTest/js_interop/JS/js_A04_t01.dart
```diff
@@ -0,0 +1,81 @@
...
+/// - Adding a JavaScript prefix to all the external top-level declarations,
+///   static members, and constructors of a library by parameterizing the
+///   annotation on the library with `name`.
+/// - Specifying the JavaScript class to use for external static members and
+///   constructors of an interop extension type by parameterizing the annotation
+///   on the interop extension type with `name`.
+/// - Renaming external declarations by parameterizing the annotation on the
+///   member with `name`.
+///
+/// @description Check that `@JS("")` is equivalent `@JS()`.
```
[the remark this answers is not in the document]
eernstg: Perhaps `@JS("")` should be an error? Seems like it should rename the given declaration to the empty name, and that would then give rise to various problems.

@srujzs, WDYT?

LibTest/js_interop/JS/js_A04_t02.dart
```diff
@@ -0,0 +1,81 @@
...
+/// - Adding a JavaScript prefix to all the external top-level declarations,
+///   static members, and constructors of a library by parameterizing the
+///   annotation on the library with `name`.
+/// - Specifying the JavaScript class to use for external static members and
+///   constructors of an interop extension type by parameterizing the annotation
+///   on the interop extension type with `name`.
+/// - Renaming external declarations by parameterizing the annotation on the
+///   member with `name`.
+///
+/// @description Check that `@JS(null)` is equivalent `@JS()`.
```
eernstg, line 18: Typo:
```suggestion
/// @description Check that `@JS(null)` is equivalent to `@JS()`.
```

eernstg: This would probably be a necessity, because `JS()` would accept a positional parameter optionally and with default value null. So that makes sense and should not be an error nor a warning.

eernstg (approved):
LGTM, noting a couple of typos and awaiting further input.

I posed one question. It would presumably be handled as a separate SDK issue if there is anything to handle, but the question arose naturally from the source code in this PR so I thought it was worth considering.

LibTest/js_interop/JS/js_A03_t06.dart
```diff
@@ -0,0 +1,47 @@
...
+library;
+
+import 'dart:async';
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+import '../../../Utils/expect.dart';
+import '../js_utils.dart';
+
+@JS("answer")
+external int myAswer();
```
srujzs, line 24: nit: `myAnswer`

LibTest/js_interop/JS/js_A04_t01.dart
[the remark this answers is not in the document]
srujzs: This is somewhat backwards-compatible with the `package:js` `@JS` annotation, but we could go down that route if we think it's confusing to allow this.

For now, it's probably better to document: https://dart-review.googlesource.com/c/sdk/+/443741

srujzs (approved):

eernstg (approved):
LGTM, landing.

PR #3280, opened 2025-08-15
#3180. Add interop types tests. Part 2.

sgrekhov:


LibTest/js_interop/interop_type_A02_t02.dart
```diff
@@ -0,0 +1,61 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion When interacting with a JS value, you need to provide a Dart type
+/// for it. You can do this by either using or declaring an interop type.
+/// Interop types are either a "JS type" provided by Dart or an extension type
+/// wrapping an interop type.
+///
+/// @description Check that it is a compile-time error to declare an external
+/// constructor on JS interop extension type whose representation type of is not
```
eernstg, line 11: Typos: 'on a JS', 'type is'. 'not a'.

eernstg (commented):
Added a couple of comments, awaiting further input.

I'm suggesting that there might be an extra variant of some tests because they are supposed to cover the primitive JavaScript types, and https://developer.mozilla.org/en-US/docs/Glossary/Primitive mentions null and undefined as primitive, and at least `Null` seems to be expressible in Dart interop code. But perhaps we already have lots of null testing elsewhere?

LibTest/js_interop/interop_type_A01_t10.dart
```diff
@@ -0,0 +1,53 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion When interacting with a JS value, you need to provide a Dart type
+/// for it. You can do this by either using or declaring an interop type.
+/// Interop types are either a "JS type" provided by Dart or an extension type
+/// wrapping an interop type.
+///
+/// @description Check that a JS interop type member may contain `dart:html`
+/// types.
```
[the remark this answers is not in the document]
srujzs: I'll admit it's a bit awkward to have it in an error but not the webpage. I was a little worried about advertising that you can do this as `dart:html` is deprecated, but perhaps it might be useful to have it in the migration docs?

LibTest/js_interop/interop_type_A01_t12.dart
```diff
@@ -0,0 +1,104 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion When interacting with a JS value, you need to provide a Dart type
+/// for it. You can do this by either using or declaring an interop type.
+/// Interop types are either a "JS type" provided by Dart or an extension type
+/// wrapping an interop type.
+///
+/// @description Check that it is a compile-time error if a JS interop type
```
srujzs, line 10: This should be okay (no error) as they're interop types still.

LibTest/js_interop/interop_type_A02_t01.dart
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion When interacting with a JS value, you need to provide a Dart type
+/// for it. You can do this by either using or declaring an interop type.
+/// Interop types are either a "JS type" provided by Dart or an extension type
+/// wrapping an interop type.
+///
+/// @description Check that it is a compile-time error to declare an external
```
srujzs, line 10: Sorry, I led you astray. :)

`JSAny` (and any interop types that wrap it) should be okay. The reason is that it's a valid supertype of `JSObject`.

LibTest/js_interop/interop_type_A02_t02.dart
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion When interacting with a JS value, you need to provide a Dart type
+/// for it. You can do this by either using or declaring an interop type.
+/// Interop types are either a "JS type" provided by Dart or an extension type
+/// wrapping an interop type.
+///
+/// @description Check that it is a compile-time error to declare an external
```
srujzs, line 10: This, however, should indeed be a compile-time error as it will surely crash later.

PR #3282, opened 2025-08-19
#3180. Add importModule() tests

sgrekhov:


LibTest/js_interop/importModule_A01_t02.dart
```diff
@@ -0,0 +1,39 @@
...
+
+main() {
+  var promise = importModule(
+      '/root_dart/tests/co19/src/LibTest/js_interop/module.js'.toJS);
+  var future = promise.toDart;
+  asyncStart();
+  future.then((module) {
+    var m = Module(module);
+    Expect.throws(() {
+      m.notExported();
```
eernstg, line 35: Would the test be more tight if it were to check that `m.notExported` does not exist? (.. something with `isUndefined`, I guess). The point is that a function can throw for lots of different reasons, and in this case we just want to make sure there is no function in the first place.

sgrekhov: It makes sense but tear-off of external JS interop type members is a compile-time error. So `m.notExported` wont't be compiled.

sgrekhov: Hm, a good idea! But doesn't work. Probably @srujzs  can say what I'm doing wrong.
```dart
extension type Module(JSObject _) implements JSObject {
  external String notExported();
  external bool hasOwnProperty(JSString name);
}

// and later
Expect.isFalse(m.hasOwnProperty("notExported".toJS)); // TypeError: module.hasOwnProperty is not a function
// If to try
Expect.isFalse(m.callMethod("hasOwnProperty".toJS, "notExported".toJS)); // TypeError: _this[method] is not a function
```

LibTest/js_interop/importModule_A01_t01.dart
```diff
@@ -0,0 +1,51 @@
...
+
+import 'dart:js_interop';
+import '../../Utils/expect.dart';
+
+extension type Module(JSObject _) implements JSObject {
+  external int answer();
+}
+
+extension type A._(JSObject _) implements JSObject {
+  external A(id, name);
```
srujzs, line 29: I believe the lack of types here will be a compile-time error as `dynamic` is not allowed.

LibTest/js_interop/importModule_A01_t02.dart
srujzs: I believe this is because the resulting object from `import()` has the following restriction:

> The module namespace object is a [sealed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/isSealed) object with [null prototype](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#null-prototype_objects).

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#module_namespace_object

So, `Object`'s `hasOwnProperty` won't be in the resulting object's prototype. Alternatively, I'd just use `dart:js_interop_unsafe`'s `has` which will use the keyword `in`.

srujzs (approved):

PR #3283, opened 2025-08-19
#3180. Add tests for `createJSInteropWrapper()` function

sgrekhov:
Existing tests for `JSExport` also tests `createJSInteropWrapper()` function. So there are only few new tests here.

### DIFF
--- a/LibTest/js_interop/JSExport/jsExport_A07_t02.dart
+++ b/LibTest/js_interop/JSExport/jsExport_A07_t02.dart
@@ -11,9 +11,9 @@
 /// properties forward to the instance's corresponding members.
 ///
 /// @description Check that it is a compile-time error if a class member
-/// annotated with `@JSExport()` has a non-primitive type.
+/// annotated with `@JSExport()` has a non-JS interop compatible type.
 /// @author sgrekhov22@gmail.com
-/// @issue 61379
+/// @issue 56792
 
 import 'dart:js_interop';
 
--- a/LibTest/js_interop/JSExport/jsExport_A07_t03.dart
+++ b/LibTest/js_interop/JSExport/jsExport_A07_t03.dart
@@ -10,27 +10,33 @@
 /// property for each of the class' instance members. When called, these
 /// properties forward to the instance's corresponding members.
 ///
-/// @description Check that it is a compile-time error if a class annotated with
-/// `@JSExport()` contains a non-JS interop compatible type.
+/// @description Check that it is not an error if a class member has a non-JS
+/// interop compatible type but it is not exported.
 /// @author sgrekhov22@gmail.com
-/// @issue 56792
 
 import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+import '../../../Utils/expect.dart';
+import '../js_utils.dart';
 
-@JSExport()
 class C {
   final D d;
-//      ^
-// [analyzer] unspecified
-// [web] unspecified
   C(this.d);
+  @JSExport()
+  String foo() => "foo();";
 }
 
-@JSExport()
 class D {
   int bar() => 42;
 }
 
 void main() {
-  print(C);
+  var jsC = createJSInteropWrapper<C>(C(D()));
+  globalContext["jsC"] = jsC;
+  eval(r'''
+    globalThis.v1 = globalThis.jsC.foo();
+    globalThis.v2 = globalThis.jsC.d;
+  ''');
+  Expect.equals("foo();", (globalContext["v1"] as JSString).toDart);
+  Expect.isNull(globalContext["v2"]);
 }


LibTest/js_interop/createJSInteropWrapper_A02_t01.dart
```diff
@@ -0,0 +1,46 @@
...
+  int bar() => 42;
+}
+
+void main() {
+  D d = D();
+  var jsC = createJSInteropWrapper<C>(C(d));
+  globalContext["jsC"] = jsC;
+  Expect.throws(() {
+    eval(r'''
+      globalThis.jsC.d.bar();
```
srujzs, line 43: This is testing the representation of `D` that the compilers emit, which could totally emit Dart `bar` as just `bar` in JS. DDC may already do this but I'm not sure.

This also isn't really testing `createJSInteropWrapper`. Perhaps you might be thinking that we create wrappers recursively (so `d` here would be wrapped as well)? Currently we do not, and in fact, that's one of the reasons we want to add errors for non-interop types: https://github.com/dart-lang/sdk/issues/56792

LibTest/js_interop/createJSInteropWrapper_A02_t02.dart
```diff
@@ -0,0 +1,48 @@
...
+
+void main() {
+  var d = D();
+  var c = C(d);
+  var jsC = createJSInteropWrapper<C>(c);
+  createJSInteropWrapper<D>(d);
+  globalContext["jsC"] = jsC;
+  Expect.throws(() {
+    eval(r'''
+      globalThis.jsC.d.bar();
```
srujzs, line 45: Similar comment as above.

LibTest/js_interop/createJSInteropWrapper_A01_t01.dart
```diff
@@ -0,0 +1,32 @@
...
+///
+/// The object literal will be a map of properties, which are either the written
+/// instance member names or their renames, to callbacks that call the
+/// corresponding Dart instance members.
+///
+/// See https://dart.dev/interop/js-interop/mock for more details on how to
+/// declare classes that can be used in this method.
+///
+/// @description Check that it is a compile-time error argument of
+/// [createJSInteropWrapper] function does not contain `@JSExport` annotation.
```
eernstg, line 18: Typo:

```suggestion
/// [createJSInteropWrapper] function does not contain a `@JSExport` annotation.
```

eernstg (commented):
Let's take another iteration on this one.

LibTest/js_interop/createJSInteropWrapper_A02_t01.dart
srujzs: Sorry, I'll be more specific.

`jsC.d.bar()` is really testing two things:

1. The result of `createJSInteropWrapper` (`jsC`) has a property `d`.
2. That property `d` does not have a function `bar` in JS or if it does, it throws.

The first one is fine, although it's likely already tested in other tests. The second one is the troublesome one.

`d` here is a Dart object that is either compiled to JS with the JS compilers or is an external reference (which seems to materialize as an opaque JS object) when compiled to Wasm. This distinction as well as the fact that we don't want users using compiled Dart objects directly in JS is why the above issue was filed.

Testing that this compiled Dart object does not contain a `bar` is undefined behavior. The compilers can choose to have a property in there called `bar` if they so choose. 

----

So, I'm not sure what you're testing here:

> Check that it is a JS error to invoke method of the Dart class which was not passed to [createJSInteropWrapper].

There's nothing about the implementation of `createJSInteropWrapper` that guarantees this. What we should probably check instead is that we get a static error as `D` isn't a valid type for a value that you can fetch from `jsC`. This *could* be a warning and we end up just not including that property in the resulting wrapper, but regardless, users should expect an error/warning when writing a class like that and trying to export it.

sgrekhov: Thank you for the detailed explanation!

>  This could be a warning and we end up just not including that property in the resulting wrapper, but regardless, users should expect an error/warning when writing a class like that and trying to export it.

Вoes it mean that if a class/member is annotated with `@JSExport()` then we should expect an error/warning in the case like: the below?
```dart
@JSExport()
class C {
  final D d;
//        ^
// [analyzer] unspecified
// [web] unspecified
  C(this.d);
}

@JSExport()
class D {
  int bar() => 42;
}
```

srujzs: Yes, basically anything that isn't allowed on an `external` interop member should apply here. If the entire class was exported or the specific instance member was exported and the member contains an invalid type in its signature, we should expect an error/warning.

srujzs (approved):
nit: I'd maybe add a test where it's okay to have an exported class that has a member with an invalid type but it never exports it e.g.

```dart
class C {
  final D d; // not exported
  @JSExport()
  int get boo => d.bar();
  C(this.d);
}
```

LibTest/js_interop/JSExport/jsExport_A07_t01.dart
```diff
@@ -0,0 +1,36 @@
...
+/// @assertion Annotation to allow Dart classes to be wrapped with a JS object
+/// using `dart:js_interop`'s `createJSInteropWrapper`.
+///
+/// When an instance of a class annotated with this annotation is passed to
+/// `createJSInteropWrapper`, the method returns a JS object that contains a
+/// property for each of the class' instance members. When called, these
+/// properties forward to the instance's corresponding members.
+///
+/// @description Check that it is a compile-time error if a class annotated with
+/// `@JSExport()` contains non-primitive type.
```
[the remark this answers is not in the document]
sgrekhov: Thank you. Updated.

sgrekhov:
Added `jsExport_A07_t03.dart`. Please take another look.

LibTest/js_interop/createJSInteropWrapper_A02_t01.dart
eernstg: I think this is converging, and it looks good to me. So I'll land this PR based on the assumption that it's in good shape. @srujzs, if you do have further comments then let's handle them in a separate PR.

eernstg (approved):
LGTM.

PR #3290, opened 2025-08-27
#3180. Fix delete_A01_t06.dart to work in both strict/non-strict modes

sgrekhov:


LibTest/js_interop_unsafe/JSObjectUnsafeUtilExtension/delete_A01_t06.dart
```diff
@@ -27,7 +27,21 @@ main() {
...
+    // In case of dart2wasm compiled code is loaded using modules. It
+    // automatically enables the strict mode. In case of the strict mode an
+    // error is thrown.
+    Expect.throws(() {
+      array.delete("length".toJS);
+    });
+    Expect.throws(() {
+      obj.delete("p1".toJS);
+    });
+  }
```
osa1, line 46: Can we run dart2wasm code in non-strict code?

The reason why I ask is because there can be many things that throw an exception, and because all JS exceptions are just one value in dart2wasm (a Wasm limitation, but it will be lifted soon with the new exception instructions) we can't check if the exception is the right one.

srujzs: Being able to opt out of strict mode seems to be a harder task than opting in. I tried calling a `JSFunction` that calls `deleteProperty` within a function that explicitly doesn't have `use strict` in it, but no luck, presumably because the code where `deleteProperty` is compiled is itself in strict mode.

I also think being able to toggle between the two modes for each compiler is a larger task than this is worth. My only nit here is that the current expectations in the test lock in the current mode each compiler is in. If we want to avoid that, we can check for *either* an error or false.

srujzs (approved):

LibTest/js_interop_unsafe/JSObjectUnsafeUtilExtension/delete_A01_t06.dart
sgrekhov: >  My only nit here is that the current expectations in the test lock in the current mode each compiler is in

I think it's fine. If the test runner changes, this test will fail, and the comments should explain why.

osa1 (approved):

LibTest/js_interop_unsafe/JSObjectUnsafeUtilExtension/delete_A01_t06.dart
eernstg: > we can check for _either_ an error or false.

@sgrekhov, I think that's a strategy which is both useful and maintainable. Perhaps we should still land this one now, and then there will be a PR doing the either thing (which might require changes in more than one library). WDYT?

sgrekhov: Updated to expect  either an error or false. PTAL.

PR #3291, opened 2025-08-27
#3180. Add tests for `ObjectToJSBoxedDartObject`

sgrekhov:


LibTest/js_interop/ObjectToJSBoxedDartObject/toJSBox_A01_t02.dart
```diff
@@ -0,0 +1,29 @@
...
+/// [Object].
+///
+/// Throws an [Exception] if this [Object] is a JavaScript value.
+///
+/// Unlike [ObjectToExternalDartReference.toExternalReference], this returns a
+/// JavaScript value. Therefore, the representation is guaranteed to be
+/// consistent across all platforms and interop members can be declared on
+/// [JSBoxedDartObjects].
+///
+/// @description Checks that an [Exception] is thrown if this [Object] is a
```
srujzs, line 18: nit: you're checking that this *doesn't* throw an exception. It currently does on the JS compilers.

LibTest/js_interop/ObjectToJSBoxedDartObject/toJSBox_A01_t03.dart
```diff
@@ -0,0 +1,30 @@
...
+/// [Object].
+///
+/// Throws an [Exception] if this [Object] is a JavaScript value.
+///
+/// Unlike [ObjectToExternalDartReference.toExternalReference], this returns a
+/// JavaScript value. Therefore, the representation is guaranteed to be
+/// consistent across all platforms and interop members can be declared on
+/// [JSBoxedDartObjects].
+///
+/// @description Checks that an [Exception] is thrown if this [Object] is a
```
srujzs, line 18: Same comment as above.

srujzs (approved):
One thing worth checking for: what happens if we call `toJSBox` on an `ExternalDartReference`? I suspect we're inconsistent there in some cases. :) dart2wasm will always throw while ddc/dart2js will throw only if the original value in the `ExternalDartReference` were to throw.

I'm not sure we actually have a good fix for that as we would need to distinguish the result of `toExternalReference` from the original object, and I don't see a good way of doing that with dart2js/ddc unless we wrap the value (defeating the original purpose of performance).

sgrekhov:
> One thing worth checking for: what happens if we call toJSBox on an ExternalDartReference?

https://github.com/dart-lang/sdk/blob/a29e08c72e2ce21813c1edf50cbcdfcac7a7acdd/sdk/lib/js_interop/js_interop.dart#L446-L448

Extension type `ExternalDartReference` doesn't implement `Object` and therefore doesn't have `toJsBox` getter, defined as member of extension on `Object`. 

Descriptions updated. PTAL.

srujzs (approved):
> Extension type ExternalDartReference doesn't implement Object and therefore doesn't have toJsBox getter, defined as member of extension on Object.

Ha! That ended up being convenient. :)

eernstg (commented):
Looks good! I think it would be helpful to add a few words to a description, then land it.

Issue #3292, opened 2025-08-28
Check failures of tests for flow analysis


sgrekhov:
Flow analysis [specification](https://github.com/dart-lang/language/blob/main/resources/type-system/flow-analysis.md) and implementation were updated. Need to check the current co19 test [failures](https://dart-current-results.web.app/#/filter=co19/TypeSystem/flow-analysis) to check if they still indicating issues or simply should be fixed/updated.

PR #3294, opened 2025-08-28
#3292. Update assertions for `promotion_via_assignment_*` tests

sgrekhov:
Please note that this PR updates assertions and descriptions but also fixes `promotion_via_assignment_A03_t07.dart`. Now it passes and I believe this is according to the changed spec. Please review.

### DIFF
--- a/TypeSystem/flow-analysis/promotion_via_assignment_A07_t01.dart
+++ b/TypeSystem/flow-analysis/promotion_via_assignment_A07_t01.dart
@@ -79,8 +79,8 @@
 ///     is a subtype of `T`, the precondition of assign is satisfied, namely
 ///     that the static type of `E1'` must be a subtype of `x`'s declared type.
 ///
-/// @description Checks that a variable is not promotable from the type
-/// `Object?` to the type `T` if `T` is an extension type.
+/// @description Checks that a variable can be promoted from the type `S` to
+/// the type `T` even if `T` is not a subtype of `S`. Test extension types.
 /// @author sgrekhov22@gmail.com
 
 import '../../Utils/static_type_helper.dart';
@@ -90,23 +90,26 @@ extension type ET2(int v) implements Object {}
 extension type ET3(int v) implements int {}
 
 test1(Object? o) {
+  if (o is ET2) {}
   if (o is ET1) {
     o = ET2(42);
-    o.expectStaticType<Exactly<Object>>();
+    o.expectStaticType<Exactly<ET2>>();
   }
 }
 
 test2(Object? o) {
+  if (o is ET3) {}
   if (o is ET2) {
     o = ET3(42);
-    o.expectStaticType<Exactly<Object>>();
+    o.expectStaticType<Exactly<ET3>>();
   }
 }
 
 test3(Object? o) {
+  if (o is ET1) {}
   if (o is ET3) {
     o = ET1(42);
-    o.expectStaticType<Exactly<Object?>>();
+    o.expectStaticType<Exactly<ET1>>();
   }
 }
 


TypeSystem/flow-analysis/promotion_via_assignment_A07_t01.dart
```diff
@@ -2,17 +2,85 @@
...
+/// - Let `before(E1) = before(N)`.
+/// - Let `E1'` be the result of applying type coercion to `E1`, to coerce it to
+///   the declared type of `x`.
+/// - Let `after(N) = assign(x, E1', after(E1))`.
+///   - Since type coercion to type `T` produces an expression whose static type
+///     is a subtype of `T`, the precondition of assign is satisfied, namely
+///     that the static type of `E1'` must be a subtype of `x`'s declared type.
+///
+/// @description Checks that a variable is not promotable from the type `S` to
+/// the type `T` if `T` is not a subtype of `S`. Test extension types.
```
eernstg, line 83: This is somewhat twisted, and I can't really see what `S` and `T` should stand for. ;-)

At first, `o` has the declared type `Object?` and `Object` is a type of interest (based on the declaration alone). At the beginning of line 94, 100, 106, `o` still has type `Object?` because no evidence has been gathered for anything else, but `ET1/2/3` is now a type of interest. Then we assign an `int` to `o` and promote `o` to `Object` because that is a type of interest and a supertype of `int`.

The extension type is a type of interest, but it is not the assigned type and not a supertype of the assigned type, so it isn't a member of `p3`. As far as I can see, `p3` contains exactly one element, namely `Object`. So there's no way the extension type could be a candidate for being the new promoted type of `o`. 

So `S` could be `Object?` and `T` could be `Object`, but this contradicts the fact that we _do_ promote `o` from `Object?` to `Object`. Otherwise `S` could be `Object?` and `T` could be the extension type, but it doesn't make much sense to say that we don't promote `o` to the extension type, because we aren't even trying: the assignment is the only action that could justify a promotion which is in place when the `expectStaticType` is reached. (And if we _did_ do `o = ET1(42)` then the promotion _would_ occur.)

However, I think the test could match the description if the assignments were adjusted to use a different extension type:

```dart
void test1(Object? o) {
  if (o is ET1) {
    o = ET2(42);
    o.expectStaticType<Exactly<Object>>();
  }
}

void test2(Object? o) {
  if (o is ET2) {
    o = ET3(42);
    o.expectStaticType<Exactly<Object>>();
  }
}

void test3(Object? o) {
  if (o is ET3) {
    o = ET1(42);
    o.expectStaticType<Exactly<Object?>>();
  }
}
```

With this, both `S` and `T` is an extension type, and we are actually trying to promote from one of them to the other, and they are not subtype-related. The outcome will then be `Object` or `Object?`, depending on whether or not we can know for sure that it isn't null.

TypeSystem/flow-analysis/promotion_via_assignment_A07_t02.dart
```diff
@@ -2,17 +2,85 @@
...
+/// - Let `before(E1) = before(N)`.
+/// - Let `E1'` be the result of applying type coercion to `E1`, to coerce it to
+///   the declared type of `x`.
+/// - Let `after(N) = assign(x, E1', after(E1))`.
+///   - Since type coercion to type `T` produces an expression whose static type
+///     is a subtype of `T`, the precondition of assign is satisfied, namely
+///     that the static type of `E1'` must be a subtype of `x`'s declared type.
+///
+/// @description Checks that a variable is not promotable from the type `S` to
+/// the type `T` if `T` is not a subtype of `S`. Test extension types.
```
eernstg, line 83: We can't do exactly the same thing here because the extension types aren't subtypes of the declared type. We could adjust the extension types such that it isn't an error, or we could drop the test if it turns out to be to similar to the previous one.

eernstg (commented):
Looks good! I suggested one adjustment of a description. The two tests at the end are more tricky (they must have slipped under the radar back when they were introduced). I'm explaining why I don't think the first one is testing what it says, and suggesting that the last one could be removed if there's no obvious way to make it work well.

sgrekhov (commented):
Thank you. Updated. PTAL

TypeSystem/flow-analysis/promotion_via_assignment_A07_t01.dart
eernstg: If we insist that we're trying to promote from `S` to `T` where `T` is not a subtype of `S` then the new version actually won't do it: It tries to promote from `Object?` to `ET2`/`ET3`/`ET1`, and they _are_ subtypes of `Object?`. That's the reason why I moved the two statements into the body of the `if` statement.

eernstg (commented):
I had one more comment on A07_t01

sgrekhov:
Description updated, statements moved into the body of `if`.

TypeSystem/flow-analysis/promotion_via_assignment_A07_t01.dart
```diff
@@ -79,8 +79,8 @@
 ///     is a subtype of `T`, the precondition of assign is satisfied, namely
 ///     that the static type of `E1'` must be a subtype of `x`'s declared type.
 ///
-/// @description Checks that a variable is not promotable from the type `S` to
-/// the type `T` if `T` is not a subtype of `S`. Test extension types.
+/// @description Checks that a variable is not promotable from the type
+/// `Object?` to the type `T` if `T` is an extension type.
```
eernstg, line 83: Oh, but that's not true, we can certainly promote a variable from a type `S` (here: `Object?`) to an extension type `T` where `T` is a subtype of `S`. There's nothing special about extension types in this respect. In fact, that's exactly what happens in the code in the body of the `if` statement (both before and after the latest commit).

So, in `test1`, commit 1d71776, the `if` body is empty and we proceed after the `if` statement with `ET1` as a type of interest, but an unpromoted `o`. The assignment doesn't promote `o` to `ET2` because `ET2` isn't a type of interest. However, `Object` is a type of interest, and it is a supertype of `ET2`, so we follow the rules (gathering a set of types and choosing the "best" `one) and promote `o` to `Object`.

So it's confusing to say that the variable is not promotable when it is actually promoted, and it is confusing to say that it's about an impossible promotion from a type `S` to a type `T` which isn't a subtype of `S` when `ET2` is actually a subtype of `Object?`.

So my suggestion (about `test1`) was to use `ET1` as `S` and `ET2` as `T` (this means that `T` is not a subtype of `S`). But this implies that `o` must have the type `ET1` in the first place, which means that we need to move the statements into the `if` body.

It's still slightly confusing because `ET2` is not a type of interest, which means that we aren't going to promote `o` to `ET2` _even_ if we change `ET2` such that it is a subtype of `ET1`.

In order to put all elements in place, we should actually make the assigned type a type of interest (and we should preserve the property that the assigned type isn't a subtype of the current type of `o` when the assignment occurs). This means that everything _other than_ the subtype relationship is satisfied, and hence it's the missing subtype relationship that causes the promotion to fail. We will still promote from `Object?` to `Object` in some cases, but it is true that we don't promote from `ET1` to `ET2`.

So let's go back to the previous description and then adjust the code a bit more.

TypeSystem/flow-analysis/promotion_via_assignment_A07_t01.dart
```diff
@@ -90,21 +90,24 @@ extension type ET2(int v) implements Object {}
 extension type ET3(int v) implements int {}
 
 test1(Object? o) {
-  if (o is ET1) {}
-  o = ET2(42);
-  o.expectStaticType<Exactly<Object>>();
+  if (o is ET1) {
+    o = ET2(42);
+    o.expectStaticType<Exactly<Object>>();
+  }
```
eernstg, line 96:
```suggestion
  if (o is ET2) {} // Make `ET2` a type of interest.
  if (o is ET1) { // Promote `o` to `ET1`.
    o = ET2(42); // Everything except `ET2 <: ET1` is ok.
    o.expectStaticType<Exactly<Object>>(); // So we don't promote to `ET2`, only `Object`.
  }
```

eernstg (commented):
Sorry about pushing on this, but let's take one more iteration.

TypeSystem/flow-analysis/promotion_via_assignment_A07_t01.dart
sgrekhov:
```dart
  o.expectStaticType<Exactly<Object>>(); // So we don't promote to `ET2`, only `Object`.
```
In fact, static type of `o` here is `ET2`. Is this a bug? Or may be we have missed anything?

eernstg (approved):
LGTM, landing.

This is a good combination of elements, and it is testing the new approach to `assign` with combined demotions and promotions.

TypeSystem/flow-analysis/promotion_via_assignment_A07_t01.dart
eernstg: Thanks for going through with all this! ;-)

PR #3300, opened 2025-09-03
#3180. Add test for `toDart`.

sgrekhov:


eernstg (approved):
LGTM, awaiting further input.

In another test (and another PR, if that's more convenient), can we get the run-time failure about "not the same runtime" by means of two isolates?

Similarly, we should be able to get the "not a wrapped Dart object" error by casting an expression of type `JSObject` (e.g., the empty object `JSObject()`) to `JSBoxedDartObject` and calling `toDart`.

srujzs (approved):

srujzs:
I believe Erik is referring to an error when calling `toDart` on something like `JSObject() as JSBoxedDartObject`. It's a bit of a cheat because it relies on the representation types being the same but it'll work to test that an exception is thrown.

I think the test runner won't be able to support two Dart runtimes. The error there will test the same behavior as what Erik proposes in that `toDart` will look for a `Symbol` that is unique to the Dart runtime (top-level, declared once), and if it's not there, it will throw.

sgrekhov:
Thank you. The new test added. Please review.

eernstg (approved):
OK, I think this means that we can't test the "two different runtimes" error.
LGTM, and I'll land it.

PR #3301, opened 2025-09-03
Fixes #3297. Add tests about redirecting factory constructor inference

### LINKED ISSUE #3297: Write a couple of tests about redirecting factory constructor inference
The issue https://github.com/flutter/flutter/issues/174503 gave rise to some bug fixes in the analyzer (https://dart-review.googlesource.com/c/sdk/+/447603), concerning the inference of actual type arguments being passed to the redirectee constructor of a redirecting factory. In particular, the following example failed:

```dart
// Library 'a.dart'.

sealed class AsyncResult<T extends Object> {
  const AsyncResult._();
  const factory AsyncResult.ok(T value) = Ok._;
}

class Ok<T extends Object> extends AsyncResult<T> {
  final T value;
  const Ok._(this.value) : super._();
}
```

```dart
// Library 'main.dart'.
import 'a.dart';

void f() {
  const x = AsyncResult<String>.ok('data');
  x;
}
```

With the fix, this succeeds. With the bug, it fails in the analyzer, reporting that

>   error •main.dart:4:13 • A value of type 'Ok<Object>' can't be
>           assigned to a const variable of type 'AsyncResult<String>'.
>           Try using a subtype, or removing the 'const' keyword •
>           variable_type_mismatch

It is not easy to test because it requires multiple libraries (put everything into a single file and it works), but that's presumably because the wrong approach was used when creating summaries, which doesn't occur at all for a one-library example.

sgrekhov:
A strange issue. Sometimes the error was disappiering after renaming of the class. No error with the new SDK.

### DIFF
--- a/Language/Classes/Constructors/Factories/inference_A01_t01.dart
+++ b/Language/Classes/Constructors/Factories/inference_A01_t01.dart
@@ -7,8 +7,8 @@
 /// otherwise the return type `M<T1,..., Tn>` where `T1,...,Tn` are the type
 /// parameters of the enclosing class.
 ///
-/// @description Checks that return type of the factory constructor can be
-/// inferred.
+/// @description Checks that the redirectee type of the factory constructor can
+/// be inferred.
 /// @author sgrekhov22@gmail.com
 
 import 'inference_lib.dart';
--- a/Language/Classes/Constructors/Factories/inference_A01_t01.dart
+++ b/Language/Classes/Constructors/Factories/inference_A01_t01.dart
@@ -1,0 +1,22 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion The return type of a factory whose signature is of the form
+/// `factory M` or the form `factory M.id` is `M` if `M` is not a generic type;
+/// otherwise the return type `M<T1,..., Tn>` where `T1,...,Tn` are the type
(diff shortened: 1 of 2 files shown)


Language/Classes/Constructors/Factories/inference_A01_t01.dart
```diff
@@ -0,0 +1,22 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion The return type of a factory whose signature is of the form
+/// `factory M` or the form `factory M.id` is `M` if `M` is not a generic type;
+/// otherwise the return type `M<T1,..., Tn>` where `T1,...,Tn` are the type
+/// parameters of the enclosing class.
+///
+/// @description Checks that return type of the factory constructor can be
+/// inferred.
```
eernstg, line 11: I don't think the issue was that the return type was now and then not inferred correctly, the issue was that the redirectee `D._` is subject to type inference, but in some situations (such as when the analyzer created a summary of `inference_lib.dart`) no type inference was performed. Instead, the analyzer used instantiation to bound. This caused the redirection to be treated as `D<Object>._` rather than `D<T>._`, and that's an error because the return type of `D<Object>._` is `D<Object>`, which is not a subtype of the return type of the redirecting factory which is `C0<T>`.

So we're actually checking that the redirectee type inference is performed (as opposed to being replaced by instantiation to bound).

eernstg:
@scheglov, do you have further comments?

scheglov:
No, your analysis about instantiation to bounds is correct, this is what was happening before the fix.

PR #3306, opened 2025-09-08
#3180. Add tests for `callAsFunction`

sgrekhov:


LibTest/js_interop/JSFunctionUtilExtension/callAsFunction_A01_t01.dart
```diff
@@ -0,0 +1,45 @@
...
+
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+import '../../../Utils/expect.dart';
+import '../../js_interop/js_utils.dart';
+
+main() {
+  eval(r'''
+    globalThis.f1 = function(p1, p2, p3, p4) {
+      return `${p1}, ${p2}, ${p3}, ${p4}`;
```
eernstg, line 33: In order to verify that `f1.callAsFunction(...)` doesn't just create a function that ignores its first argument and passes the remaining arguments on to the receiver `f1`, the body should contain an expression that uses `this`. For instance, it could include `${this.bar}` in the template literal, and the argument passed in line 38 would then need to be an object that has a `bar` property.

LibTest/js_interop/JSFunctionUtilExtension/callAsFunction_A01_t02.dart
```diff
@@ -0,0 +1,45 @@
...
+  globalContext["jsFoo"] = foo.toJSCaptureThis;
+  JSFunction jsFoo = globalContext["jsFoo"] as JSFunction;
+  var res = jsFoo.callAsFunction(
+    "Value of this".toJS,
+    1.toJS,
+    "two".toJS,
+    3.14.toJS,
+    true.toJS,
+  );
+  Expect.equals("Value of this, 1, two, 3.14, true", (res as JSString).toDart);
```
eernstg, line 44: This is somewhat twisted...

First we have a Dart function `foo`, which is then subjected to `toJSCaptureThis`, which yields a JavaScript function which will pass the value of `this` on to the Dart function as a regular parameter named `pThis`. This means that we're going one step in one direction and then one step backwards again, and checking that we're in the same position where we started (that is, a regular function taking 5 arguments).  ;-)

I don't think it's wrong, but I do think it's a fragile kind of testing when we do two things that are basically the opposite of each other and then check that nothing happened. It is safer to do one thing that has an effect, and then check that the right effect occurred.

Perhaps it's sufficient to do the `bar` thing with the previous test, and then this test might simply be omitted?

eernstg (commented):
I think the first test could be strengthened to check the handling of `this`, and the second test might then be redundant. Awaiting further input as well.

sgrekhov:
Makes sense. Thank you. Updated.

eernstg (approved):
LGTM

PR #3308, opened 2025-09-09
#3180. Add more tests for `toJSCaptureThis`

sgrekhov:
Probably it also makes sense to add tests for passing `null` as `this`. However, in non-strict mode it is replaced by the global object. The test runner executes `dart2js` tests in non-strict mode and `dart2wasm` in strict mode. Previously, we preferred to avoid writing tests that depend on the mode. WDYT?

### DIFF
--- a/LibTest/js_interop/FunctionToJSExportedDartFunction/toJSCaptureThis_A01_t09.dart
+++ b/LibTest/js_interop/FunctionToJSExportedDartFunction/toJSCaptureThis_A01_t09.dart
@@ -20,7 +20,7 @@ import 'dart:js_interop';
 import 'dart:js_interop_unsafe';
 import '../../../Utils/expect.dart';
 
-void foo(Object? pThis) {
+void foo(JSAny? pThis) {
   if (isJS) {
     // Test runner runs `dart2js` tests in non-strict mode. In this case
     // (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call#parameters)


eernstg:
I tend to think that we should test mode-dependent properties as well as other properties: Assume that they are observable properties of the execution, and they are predictable and documented. In this case, they might be correct or buggy behaviors as seen from a developer's point of view, and they would reasonably expect that implementations behave as specified (so we should try to detect situations where they don't).

The main arguments against could be that "this will change anyway" or "that's a tiny detail", or even "this will give rise to an exponentially growing number of tests", but I don't think any of those are particularly convincing.

@srujzs, WDYT?

LibTest/js_interop/FunctionToJSExportedDartFunction/toJSCaptureThis_A01_t07.dart
```diff
@@ -0,0 +1,36 @@
...
+/// `this` value when called.
+///
+/// Identical to `toJS`, except the resulting [JSExportedDartFunction] will pass
+/// `this` from JavaScript as the first argument to the converted [Function].
+/// Any [Function] that is converted with this member should take in an extra
+/// parameter at the beginning of the parameter list to handle this.
+///
+/// @description Check that this property returns a JavaScript function that
+/// wraps this [Function] and captures the `this` value when called. Check that
+/// a [JSString] can be passes as a value of `this`.
```
eernstg, line 16: Typo:

```suggestion
/// a [JSString] can be passed as the value of `this`.
```

LibTest/js_interop/FunctionToJSExportedDartFunction/toJSCaptureThis_A01_t08.dart
```diff
@@ -0,0 +1,36 @@
...
+/// `this` value when called.
+///
+/// Identical to `toJS`, except the resulting [JSExportedDartFunction] will pass
+/// `this` from JavaScript as the first argument to the converted [Function].
+/// Any [Function] that is converted with this member should take in an extra
+/// parameter at the beginning of the parameter list to handle this.
+///
+/// @description Check that this property returns a JavaScript function that
+/// wraps this [Function] and captures the `this` value when called. Check that
+/// an arbitrary object can be passes as a value of `this`.
```
eernstg, line 16: Ditto.

eernstg (approved):
LGTM, modulo one typo

srujzs:
> @srujzs, WDYT?

I don't have an issue with testing strict mode vs non-strict mode semantics if they're easily testable (e.g. have a JS function in `eval` with `'use strict'`). If it requires the compilers to make strict mode configurable, I'm not sure making the various compilers provide that functionality is something the backend teams would want to do.

Relying on the current behavior of each compiler is also suspect, because that may not be something that is guaranteed.

srujzs (changes requested):
Sorry, I approved it instead of requesting changes. :P

sgrekhov:
Updated. PTAL.

eernstg:
@srujzs wrote:
> Similarly, relying on the current behavior of each compiler is also suspect, because that may not be something that is guaranteed

That's what I meant we should do: The compilers _do_ behave in this particular way, and developers may (by accident or by design) rely on the specific behavior, and then it makes sense (to me) to have tests to verify that the compilers don't suddenly change their behavior without anybody in our team noticing anything.

However, we could also adopt a different rationale and avoid testing anything in this area. Different compilers behave differently in specific ways, and we're intentionally not testing anything that reveals this difference.

That's the status that the C and C++ standards have always designated as "unspecified behavior", and they usually say that a program that depends on unspecified behavior is buggy, even in the case where it works. ;-)

The language specification uses this concept (and the word 'unspecified') two times (once in commentary and once in normative text). In both cases it's about the function objects obtained by tearing off an instance method twice (same method and same receiver), and the rule is that they can be `identical` or not. The reason for this is optimization. For example, it should be OK for an implementation to reuse a function object which was obtained by tearing off a certain method from a certain receiver.

In other words, we have almost entirely avoided unspecified behavior in Dart. For example, actual arguments in a function call are evaluated left-to-right, and it can't be changed just because this would allow some kind of optimization. And so on.

That's probably the reason why my gut reaction is that we should avoid unspecified behavior as far as possible.

PR #3309, opened 2025-09-11
#3180. Add tests for `ByteBufferToJSArrayBuffer`, `JSArrayBufferToByteBuffer` and `JSArrayToList`

sgrekhov:
@srujzs a couple of small notes about documentation.

1) https://github.com/dart-lang/sdk/blob/36a78b317fd498f8e1bc85c4b1c4b7de477d7d6c/sdk/lib/js_interop/js_interop.dart#L882

> If it's a wrapper, this method unwraps it and either returns the...

`toJS` and `toDart` are not methods. Probably it is better to use the word "operation" here. And the word "method" also occurs in many places in doc comments in the file `js_interop.dart` when these comments in fact not about methods.

2) https://github.com/dart-lang/sdk/blob/36a78b317fd498f8e1bc85c4b1c4b7de477d7d6c/sdk/lib/js_interop/js_interop.dart#L1218-L1219

> When compiling to JavaScript, [List]s are [JSArray]s and this will be a cast.

But, in fact, the following fails on `dart2js`.
```dart
main() {
  JSArray<JSNumber> a = JSArray();
  Expect.identical(a, a.toDart);
}
```
I believe that this is the same https://github.com/dart-lang/sdk/issues/61327 issue, probably `a.toDart` returns some `CastList` instead of the original object. If so, and this is not an issue but an intended bahavior maybe it makes sense to reflect it in the documentation as well?

### DIFF
--- a/LibTest/js_interop/ByteBufferToJSArrayBuffer/toJS_A03_t03.dart
+++ b/LibTest/js_interop/ByteBufferToJSArrayBuffer/toJS_A03_t03.dart
@@ -23,8 +23,8 @@
 /// > [JSArrayBuffer] and vice versa unless it was instantiated in JavaScript.
 ///
 /// @description Check that on `dart2wasm` if [ByteBuffer] was instantiated in
-/// Dart then this operations clones it and therefore changes in the original
-/// object not affect the cloned one.
+/// Dart then this operations clones it. Hence, changes in the original object
+/// do not affect the cloned one.
 /// @author sgrekhov22@gmail.com
 
 import 'dart:js_interop';


LibTest/js_interop/JSArrayBufferToByteBuffer/toDart_t02.dart
```diff
@@ -0,0 +1,41 @@
...
+  bb.asInt16List()[0] = 1;
+  Expect.equals(1, (ar["0"] as JSNumber).toDartInt);
+
+  if (isJS) {
+    // Casting case.
+    Expect.identical(ab, ab.toDart.toJS);
+  }
+  if (isWasm) {
+    // Wrapping/unwrapping case
+    Expect.notIdentical(ab, ab.toDart.toJS);
```
eernstg, line 39: This particular case is tricky: I don't think there is any harm caused by reusing a wrapper.

I expect the wrapper to have no mutable state (it's all in the wrappee), and the contract with clients simply does not promise that it's the same wrapper, so clients shouldn't test two wrappers with `identical` and expect true. I think this means that good client code will simply avoid testing these objects with `identical`, and it will work just fine also in the case where a wrapper is reused (if the implementation happens to be able to do this).

Following the approach that we've used in several other situations, I think it would be useful to avoid expecting `notIdentical` here. We could test and allow both `identical` and `notIdentical`, but that would probably never crash anyway (and "it doesn't crash" is the only thing we can test when a boolean value can succeed with the value true as well as false). The point is that this is a possible optimization, and we don't want to turn those into test failures for no good reason.

So this test could actually be `ab.toDart.toJS`, and then perhaps a comment saying something like '// This expression does not crash. It may be `identical` or `nonIdentical` to `ab`.'

eernstg (commented):
Looks good, just one issue left.

LibTest/js_interop/JSArrayBufferToByteBuffer/toDart_t02.dart
srujzs: That's fair that we perhaps shouldn't be testing identicality of JS value wrappers. I'm not sure if that's an optimization we'll ever do to reuse the same wrapper, but it makes sense to tell users to not test it (although that means it's undefined behavior! :D).

A similar discussion here is the wrapper lists that we introduce when calling `toDart` on typed arrays and arrays. We introduce a new wrapper each time in dart2wasm (and when needed for casting purposes in dart2js/ddc) but we don't document that. Should we document that it's new?

eernstg: If it is true that the wrappers can be reused (that is, "reasonable" programs won't break if we do this) then this would be one of the cases where we could leave some wiggle room for implementations.

In that case the wrapper object should be documented to exist, but it would be left unspecified (preferably explicitly unspecified ;-) whether it's new or reused.

srujzs: Sounds good. :) I'll leave it unspecified whether we return a new wrapper either in the wrapper list case or the JS box case. Once `==` is fixed, `identical` will be the only way to tell the difference anyways.

PR #3313, opened 2025-09-16
#3180. Add `toJS` tests for typed arrays

sgrekhov:
Just a Nit: for consistency, it would be nice to rename the extension `Uint16ListToJSInt16Array` to `Uint16ListToJSUint16Array`

### DIFF
--- a/LibTest/js_interop/Float32ListToJSFloat32Array/toJS_t02.dart
+++ b/LibTest/js_interop/Float32ListToJSFloat32Array/toJS_t02.dart
@@ -48,11 +48,10 @@ main() {
   }
   if (isWasm) {
     // The `Float32List` was instantiated in JS. This is a wrapping case.
-    // Unwrapping should return the same array but not the same object.
+    // Unwrapping should return the same array.
     var list = a.toDart.toList();
     Expect.equals(1, list[0]);
     Expect.approxEquals(-3.14, list[1]);
     Expect.approxEquals(3.14, list[2]);
-    Expect.notIdentical(l, a.toDart);
   }
 }
--- a/LibTest/js_interop/Float64ListToJSFloat64Array/toJS_t02.dart
+++ b/LibTest/js_interop/Float64ListToJSFloat64Array/toJS_t02.dart
@@ -48,11 +48,10 @@ main() {
   }
   if (isWasm) {
     // The `Float64List` was instantiated in JS. This is a wrapping case.
-    // Unwrapping should return the same array but not the same object.
+    // Unwrapping should return the same array.
     var list = a.toDart.toList();
     Expect.equals(1, list[0]);
     Expect.approxEquals(-3.14, list[1]);
     Expect.approxEquals(3.14, list[2]);
-    Expect.notIdentical(l, a.toDart);
   }
 }
--- a/LibTest/js_interop/Int16ListToJSInt16Array/toJS_t02.dart
+++ b/LibTest/js_interop/Int16ListToJSInt16Array/toJS_t02.dart
@@ -47,8 +47,7 @@ main() {
   }
   if (isWasm) {
     // The `Int16List` was instantiated in JS. This is a wrapping case.
-    // Unwrapping should return the same array but not the same object.
+    // Unwrapping should return the same array.
     Expect.listEquals([1, -2, 3], a.toDart.toList());
-    Expect.notIdentical(l, a.toDart);
   }
 }
--- a/LibTest/js_interop/Int32ListToJSInt32Array/toJS_t02.dart
+++ b/LibTest/js_interop/Int32ListToJSInt32Array/toJS_t02.dart
@@ -47,8 +47,7 @@ main() {
   }
   if (isWasm) {
     // The `Int32List` was instantiated in JS. This is a wrapping case.
-    // Unwrapping should return the same array but not the same object.
+    // Unwrapping should return the same array.
     Expect.listEquals([1, -2, 3], a.toDart.toList());
-    Expect.notIdentical(l, a.toDart);
   }
 }
--- a/LibTest/js_interop/Int8ListToJSInt8Array/toJS_t02.dart
+++ b/LibTest/js_interop/Int8ListToJSInt8Array/toJS_t02.dart
@@ -47,8 +47,7 @@ main() {
   }
   if (isWasm) {
     // The `Int8List` was instantiated in JS. This is a wrapping case.
-    // Unwrapping should return the same array but not the same object.
+    // Unwrapping should return the same array.
     Expect.listEquals([1, -2, 3], a.toDart.toList());
-    Expect.notIdentical(l, a.toDart);
   }
 }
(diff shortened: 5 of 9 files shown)


LibTest/js_interop/Float32ListToJSFloat32Array/toJS_t01.dart
```diff
@@ -0,0 +1,63 @@
...
+    // Casting case. `toDart` returns the same object.
+    Expect.listEquals(l2, a2.toDart.toList());
+  }
+  if (isWasm) {
+    // `a2` was cloned. Changes in `l2` don't affect it.
+    var list = a2.toDart.toList();
+    Expect.equals(1, list[0]);
+    Expect.approxEquals(-3.14, list[1]);
+    Expect.approxEquals(3333.33, list[2]);
+  }
```
srujzs, line 62: I'd check a few additional cases here and everywhere else:

- A typed array that's converted to a typed list via `toDart` is wrapped/casted. Modifications to one should affect the other.
- A typed list which was converted from a typed array that's then converted to a typed array should be unwrapped/casted (`toDart.toJS`). The resulting typed array should be the same object as the original.

srujzs (commented):
> Just a Nit: for consistency, it would be nice to rename the extension Uint16ListToJSInt16Array to Uint16ListToJSUint16Array

Oh that's a good catch, thanks! Looks like the same applies to `JSUint16ArrayToInt16List`.

LibTest/js_interop/Float32ListToJSFloat32Array/toJS_t01.dart
sgrekhov: https://github.com/dart-lang/co19/pull/3314

srujzs: > But it'll be the same object on dart2js only, right? On dart2wasm the resulting typed array shouldn't be identical to the original object.

If the typed list (or list) is the result of converting a typed array (or array) to Dart, in dart2wasm, we wrap the original typed array (or array). Then, when we convert it back into JS, if it's already a wrapper, we unwrap it. So the result of `a.toDart.toJS` should be the same as `a`.

I've added better documentation to hopefully make this more clear: https://dart-review.googlesource.com/c/sdk/+/450140

srujzs (approved):

eernstg (commented):
Looks good! I think a few locations are trying to enforce that two wrappers must be different objects. That's a very special case because it is likely to be benign if they are the same object, and it could be an optimization (it might be cheaper to reuse a wrapper than creating a new one, if the implementation is able to do that). In that case we should allow the optimization rather than making it a test failure.

@srujzs, do you see any reason why it might introduce bugs into a "reasonable" program if it were allowed for an implementation to reuse these wrappers? (Of course, it can be detected, so we can break _some_ programs, I just don't expect programs to need the wrappers to be distinct for correctness reasons in realistic programs).

srujzs:
> But I'm asking because there might be something that I haven't thought about! ;-)

The one oddity that I'd like to either fix or get more clarity about is equality via `==`. https://github.com/dart-lang/sdk/issues/61545 Even if the resolution to this bug means we always check the underlying JS values and therefore `==` will ignore the different wrappers, cast lists on dart2js/ddc throw a wrench into achieving full consistency.

But yes, beyond `==`, I think the only way of detecting this is via `identical`. Depending on the resolution on #61545, I'd want to document the property of equality on the various wrappers we return from `toDart` conversions. As for testing, 🤷 . I see Sergey's point that it might be worth freezing the current state so if we do make that optimization, we update the tests accordingly. With the JS type wrappers, it makes sense not to test because we tell users explicitly that using `identical` is undefined behavior, but no such guarantee exists with these wrappers. That being said, if we do document that it's a "new" wrapper, then identicality is implicitly documented.

eernstg:
> cast lists on dart2js/ddc throw a wrench into achieving full consistency.

Right, that's a tricky case. I do think, though, that the case can be made for the `==` of the wrapper to (1) check that `other` is also a wrapper (I guess there's a type query which will do that), and (2) return `wrappee == other.wrappee`, and also delegate `hashCode` to the wrappee (`hashCode` could also be xor'ed with a constant number to make it different from the `hashCode` of the wrappee and still equal to the `hashCode` of another wrapper with the same wrappee, but it isn't a bug to just return the same `hashCode` because hash codes can always be accidentally equal).

This leaves the cast lists unequal even in the case where they have the same underlying list. However, this is no worse than other cast lists, and it does correspond to the fact that two different cast lists can have different behavior (because they could enforce different element types), so I think it is a meaningful design choice to just let them be unequal.

> if we do make that optimization, we update the tests accordingly

That's definitely a reasonable way to proceed. Still, the reason why I'm pushing (a bit) for the approach where wrapper reuse is allowed is that I want to maintain a certain balance between predictability (so we shouldn't have unspecified behavior if we can avoid it) and performance (.. so we will give implementations a certain amount of wiggle room in order to allow optimizations). The crucial distinction there is that the (bad) unspecified behavior is 'observable by a reasonable program' whereas the (good) wiggle room is not. And it just looks like good wiggle room to allow wrapper reuse. For example, if a developer writes a loop where it is required that a certain object is wrapped, it would be allowed to lift the wrapping out of that loop.

```dart
const iterations = 100_000_000;

class Wrap<X> {
  final List<X> list;
  Wrap(this.list);
}

Wrap? v;

void main() {
  final list = <int>[];
  final wrapList = Wrap(list);
  final sw = Stopwatch()..start();
  print('Start');
  for (var i = 0; i < iterations; ++i) v = Wrap(list);
  print('Done wrapping (${sw.elapsedMilliseconds / 1000})'); // '0.57'.
  sw.reset();
  for (var i = 0; i < iterations; ++i) v = wrapList;
  print('Done wrapped (${sw.elapsedMilliseconds / 1000})'); // '0.111'.
}
```

What I'm really pushing for is to have an approach where "everything, everywhere" is prepared for tolerating reasonable optimizations like this, rather than having to revisit a lot of code at the time where someone actually wants to introduce such an optimization.

srujzs:
Yeah, that's my plan currently wrt `==` and `hashCode` (the referenced bug also references a bug for `hashCode`).

> However, this is no worse than other cast lists, and it does correspond to the fact that two different cast lists can have different behavior (because they could enforce different element types), so I think it is a meaningful design choice to just let them be unequal.

That's a good point. One small contrary behavior to the point here is that equality for a wrapper list and the original JS value will be the same. This is consistent with the JS compilers as there is no wrapper list there, but it does mean on dart2wasm, two separate Dart values with different behavior (the wrapper list and the original JS value box) will compare as equal. Unfortunate, but I don't think there's a better option.

> What I'm really pushing for is to have an approach where "everything, everywhere" is prepared for tolerating reasonable optimizations like this, rather than having to revisit a lot of code at the time where someone actually wants to introduce such an optimization.

Okay, that's a fair argument. I think it really boils down to differences in `identical` anyways, so I'm fine removing them from these tests.

sgrekhov:
`notIdentical()` check removed. PTAL.

srujzs (approved):
LGTM % my comment about `notIdentical` in the `_02` tests.

PR #3316, opened 2025-09-22
#3180. Expect warning for class with @JSExport and non-empty name

sgrekhov:


srujzs (approved):
Woo, we have support for warnings now! LGTM

Issue #3315, opened 2025-09-17
[Primary Constructors] Add co19 tests for the primary constructors feature


kallentu:
Spec: https://github.com/dart-lang/language/blob/main/accepted/future-releases/primary-constructors/feature-specification.md

This issue tracks the tests for primary constructors.
If there's already an issue, LMK!

cc. @sgrekhov @eernstg

sgrekhov:
> [@sgrekhov](https://github.com/sgrekhov) A little update on experiment flags -- I submitted the CL for adding `declaring-constructors` flag a few minutes ago. [dart-lang/sdk@4cdf974](https://github.com/dart-lang/sdk/commit/4cdf9740e9df7028538a0a789f933b349c8b9a73)

Great! Thank you!

sgrekhov:
@kallentu experimental flag renamed to `primary-constructors` (commits 8ba7f2d64786f3575db8bb5470eb291e83d086aa and 3593327be449cb039d06db605c796db0359400a7).

PR #3317, opened 2025-09-22
#3315. Add grammar tests. Part 1.

sgrekhov:


LanguageFeatures/primary-constructors/grammar_A02_t01.dart
```diff
@@ -0,0 +1,60 @@
...
+
+class C4(final int v) {
+  this(this.v);
+//    ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+class C5.someName(final int v) {
+  this.someName(final int v);
```
eernstg, line 48: I think it would be a good idea to test the case where the two declaring constructors have different names: That's still an error, you just can't have more than one declaring constructor.

LanguageFeatures/primary-constructors/grammar_A02_t02.dart
```diff
@@ -0,0 +1,59 @@
...
+///
+/// @description Check that it is a compile-time error if `D` includes a
+/// `<classNamePart>` that contains a `<primaryConstructorNoConst>`, and the
+/// body of `D` contains a `<declaringConstructorSignature>` that contains a
+/// `<declaringParameterList>`. Test an extension type.
+/// @author sgrekhov22@gmail.com
+
+// TODO (sgrekhov) Add `declaring-constructors` experimental flag
+
+extension type ET1(var int v) {
```
eernstg, line 19: `var` should be an error: A plain `int v` means `final int v` in the special case where the declaration is an extension type, and an extension type must declare exactly one final instance variable, so if we try to declare a non-final one then it's an error.

LanguageFeatures/primary-constructors/grammar_A02_t02.dart
```diff
@@ -0,0 +1,59 @@
...
+
+extension type ET4(final int v) {
+  this(this.v);
+//    ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+extension type ET5.someName(final int v) {
+  this.someName(final int v);
```
eernstg, line 48: Again, even with distinct constructor names it is an error.

LanguageFeatures/primary-constructors/grammar_A02_t03.dart
```diff
@@ -0,0 +1,70 @@
...
+  this(this.v);
+//    ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+enum const E5.someName(final int v) {
+  e5.someName(5);
+
+  this.someName(final int v);
```
eernstg, line 58: Again, different names would also be an error.

eernstg: How about the same tests without the `const` modifier in the header? With an `enum` declaration, the constness of the constructor is inferred if it is not specified explicitly (yes, that's a weird exception, but that's what we got ;-).

LanguageFeatures/primary-constructors/grammar_A04_t01.dart
```diff
@@ -0,0 +1,47 @@
...
+
+class const C2(final int v) {
+  const this : assert(v > 0);
+//      ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+class const C3.someName(final int v) {
+  const this.someName : assert(v > 0);
```
eernstg, line 37: Right, in this case it should be the same name.

If we use different names then they are different declaring constructors, which is already an error, no matter whether the body part has a declaring parameter list or not.

But we might want to have (add, if not present already?) a test dealing with multiple declaring constructors where the body part doesn't have a parameter list.

LanguageFeatures/primary-constructors/grammar_A02_t01.dart
sgrekhov: Added.

LanguageFeatures/primary-constructors/grammar_A02_t02.dart
sgrekhov: Added.

LanguageFeatures/primary-constructors/grammar_A02_t03.dart
sgrekhov: Added the case with different names. And the `const` modifier removed for enums with even numbers (I didn't know about this exception, thanks!).

LanguageFeatures/primary-constructors/grammar_A04_t01.dart
sgrekhov: > But we might want to have (add, if not present already?) a test dealing with multiple declaring constructors where the body part doesn't have a parameter list.

Please clarify. Is `C1` in this test is not the case?

sgrekhov (commented):
Thank you. Updated. PTAL.

LanguageFeatures/primary-constructors/grammar_A01_t01.dart
```diff
@@ -0,0 +1,44 @@
...
+
+final class C3(var int v);
+
+mixin class C4(var int v);
+
+abstract class C5(var int v);
+
+sealed class C6(var int v);
+
+extension type ET(var int v);
```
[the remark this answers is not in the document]
eernstg: Got that, thanks!

LanguageFeatures/primary-constructors/grammar_A02_t01.dart
eernstg: Thanks!

LanguageFeatures/primary-constructors/grammar_A02_t03.dart
eernstg: Sounds good!

LanguageFeatures/primary-constructors/grammar_A04_t01.dart
eernstg: In `C1` we have a header part and a body part, and they together specify a declaring constructor named `C1` which is primary (because it has a header part), and it has an initializer list (in the body part). This one would use different names:

```dart
class const C() {
  final int v;
  const this.name : v = 0;
}
```

Here we have an error because there are two declaring constructors, and another error because the in-body constructor does not have a formal parameter list (and there's no `C.name` constructor declaration in the header, so it doesn't have a clue about what this formal parameter list should be like).

Perhaps we shouldn't have this test after all because we don't want to test secondary errors. On the other hand, it's nice to know that it doesn't crash, and it is also reasonably fair to say that the two errors are unrelated. So it should be OK to test it. It might well turn out to have two errors on the same line, but `unspecified` should be happy about that. There may or may not be a question about where to put the carets..

sgrekhov: Added this case for class and enum.

eernstg: Thanks!

eernstg (approved):

PR #3320, opened 2025-09-25
#3315. Add more primary constructors tests

sgrekhov:


LanguageFeatures/primary-constructors/grammar_A02_t02.dart
```diff
@@ -24,7 +24,7 @@ extension type ET1(final int v) {
 }
 
 extension type ET2() {
```
eernstg, line 26: This one is tricky, I think we'd just have to rely on `unspecified` to handle the situation where the same line has multiple errors. Also, we usually just change the tests if they expect a particular error in a given location, and it turns out to be reported somewhere else.

So in this test we'd expect that it is reported as an error (somewhere) that there is a declaring parameter list both in the header and in the body.

However, you could also argue that we have multiple errors: The primary constructor must have exactly one declaring formal parameter (and every parameter of a primary constructor in an extension type is declaring, to avoid breakage in existing code). So it's an error that it has no parameters.

Next, it's an error that we have two constructors with the same name `ET2`. Finally, it's an error that we have two declaring constructors.

So it would probably be a good idea to have `(final int v)` or `(int v)` in the header, and keep `(final int v)` in the in-body declaring constructor.

I think that's just about as correct as we can get it when it needs to show that it's an error to have two declaring formal parameter lists, no matter what.

This could still be reported in line 26 or in line 27, but we'll see what the tools are doing and adjust for that.

LanguageFeatures/primary-constructors/grammar_A04_t04.dart
```diff
@@ -0,0 +1,74 @@
...
+}
+
+extension type const ET1 {
+//             ^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  this(final int v);
+}
+
+extension type const  ET2 {
```
eernstg, line 42:
```suggestion
extension type const ET2 {
```

LanguageFeatures/primary-constructors/grammar_A04_t04.dart
```diff
@@ -0,0 +1,74 @@
...
+  this.someName(final int v);
+}
+
+enum const E1 {
+//   ^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  e0(1);
+
+  this(final int v);
```
eernstg, line 55: We might get an error here, too, because every constructor in an `enum` must be constant, and `const` must always be specified explicitly for every in-body constructor, declaring or not.

We could expect that error here, even though that's not the point in this test, and we could add the required `const` here and avoid the confusion (but break the consistent style of these tests). I'm not sure what's better..

LanguageFeatures/primary-constructors/grammar_A04_t04.dart
```diff
@@ -0,0 +1,74 @@
...
+}
+
+enum const E2 {
+//   ^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  e0;
+
+  this.someName();
```
eernstg, line 64: Same issue.

LanguageFeatures/primary-constructors/static_processing_A04_t04.dart
```diff
@@ -0,0 +1,45 @@
...
+/// a declaring in-body constructor that has one optional declaring parameter
+/// and also non-declaring ones.
+/// @author sgrekhov22@gmail.com
+
+// TODO (sgrekhov) Add `declaring-constructors` experimental flag
+
+import '../../Utils/expect.dart';
+
+extension type const ET1 {
+  this(final int x, iny y, [String z = ""]);
```
eernstg, line 22: Typo?:
```suggestion
  this(final int x, int y, [String z = ""]);
```

LanguageFeatures/primary-constructors/static_processing_A04_t04.dart
```diff
@@ -0,0 +1,45 @@
...
+/// @description Check that it is not an error if an extension type contains
+/// a declaring in-body constructor that has one optional declaring parameter
+/// and also non-declaring ones.
+/// @author sgrekhov22@gmail.com
+
+// TODO (sgrekhov) Add `declaring-constructors` experimental flag
+
+import '../../Utils/expect.dart';
+
+extension type const ET1 {
```
eernstg, line 21: The `const` modifier should be on the in-body constructor declaration.

eernstg: Same below, if there is a `const` modifier in the header.

LanguageFeatures/primary-constructors/static_processing_A05_t01.dart
```diff
@@ -0,0 +1,70 @@
...
+///
+/// @description Check that it is a compile-time error if the name of the
+/// primary constructor is the same as the name of some constructor declared in
+/// the body. Test classes.
+/// @author sgrekhov22@gmail.com
+
+// TODO (sgrekhov) Add `declaring-constructors` experimental flag
+
+class C1(var int v) {
+  C1(this.v);
```
eernstg, line 17: The constructor in the body must be a redirecting generative constructor, or it must be a factory (redirecting or not), because the primary constructor must be the only non-redirecting generative constructor, when it exists.

LanguageFeatures/primary-constructors/static_processing_A05_t01.dart
```diff
@@ -0,0 +1,70 @@
...
+//^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+class C2 {
+  this(var int v);
```
eernstg, line 24: This doesn't quite match the description because there is no primary constructor, but we _do_ have the name clash that the error is all about.

LanguageFeatures/primary-constructors/static_processing_A05_t01.dart
```diff
@@ -0,0 +1,70 @@
...
+  C2(int v);
+//^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+class const C3.someName(final int v) {
+  const C3.someName(this.v);
```
eernstg, line 32: Factory or redirection.

LanguageFeatures/primary-constructors/static_processing_A05_t01.dart
```diff
@@ -0,0 +1,70 @@
...
+//      ^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+class C4 {
+  this.someName(var int v);
+  C4.someName(int v);
```
eernstg, line 40: Again, no primary constructors here, but the name clash is fine!

LanguageFeatures/primary-constructors/static_processing_A05_t01.dart
```diff
@@ -0,0 +1,70 @@
...
+//^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+class C5(var int v) {
+  C5.someName(int v) : this(v);
+  factory C5(int v) = C5.someName;
```
eernstg, line 48: Yep!

LanguageFeatures/primary-constructors/static_processing_A05_t02.dart
```diff
@@ -0,0 +1,70 @@
...
+
+extension type ET1(int v) {
+  ET1(this.v);
+//^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+extension type ET2 {
+  this(int v);
```
eernstg, line 24:
```suggestion
  this(final int v);
```

LanguageFeatures/primary-constructors/static_processing_A05_t02.dart
```diff
@@ -0,0 +1,70 @@
...
+  ET2(int v);
```
eernstg, line 25: This one would have a spurious error because it never initializes the representation variable, so we might have a bug about name clashes and still succeed.

LanguageFeatures/primary-constructors/static_processing_A05_t03.dart
```diff
@@ -0,0 +1,56 @@
...
+/// primary constructor is the same as the name of some constructor declared in
+/// the body. Test enums.
+/// @author sgrekhov22@gmail.com
+
+// TODO (sgrekhov) Add `declaring-constructors` experimental flag
+
+enum const E1(final int v) {
+  e0(1);
+
+  const E1(this.v);
```
eernstg, line 19: Can't have a non-redirecting generative constructor when there is a primary constructor, so the name clash error can be shadowed by this error.

LanguageFeatures/primary-constructors/static_processing_A05_t03.dart
```diff
@@ -0,0 +1,56 @@
...
+//      ^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+enum E2 {
+  e0(2);
+  this(final int v);
```
eernstg, line 27: No in-body constructors (declaring or not) get an implicit `const`, which could shadow the name clash error.

LanguageFeatures/primary-constructors/static_processing_A05_t03.dart
```diff
@@ -0,0 +1,56 @@
...
+  const E2(this.v);
+//      ^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+enum const E3.someName(final int v) {
+  e0(3);
+  const E3.someName(this.v);
```
eernstg, line 36: Similar to line 19.

eernstg (commented):
Bunch of comments. ;-)

LanguageFeatures/primary-constructors/static_processing_A05_t01.dart
sgrekhov: Ah, yes, of course. Thanks! Fixed.

LanguageFeatures/primary-constructors/static_processing_A05_t01.dart
sgrekhov: Again changed to have the primary constructor.

LanguageFeatures/primary-constructors/static_processing_A05_t02.dart
sgrekhov: Fixed.

LanguageFeatures/primary-constructors/static_processing_A05_t02.dart
sgrekhov: Indeed. Changed to `ET2(this.v);`.

LanguageFeatures/primary-constructors/static_processing_A05_t03.dart
sgrekhov: Added the primary constructor. Having two declaring constructors also may shadow the name conflict, but, again, there is "declaring or not" in the assertion text.

LanguageFeatures/primary-constructors/static_processing_A05_t03.dart
sgrekhov: Fixed.

LanguageFeatures/primary-constructors/static_processing_A05_t03.dart
sgrekhov: Fixed.

sgrekhov (commented):
Thank you for noticing all of this. Fixed. PTAL.

LanguageFeatures/primary-constructors/static_processing_A05_t03.dart
```diff
@@ -16,17 +16,17 @@
 enum const E1(final int v) {
   e0(1);
 
-  const E1(this.v);
+  const E1.someName(this.v)
```
eernstg: Typo:
```suggestion
  const E1.someName(this.v);
```

eernstg (approved):
LGTM

LanguageFeatures/primary-constructors/static_processing_A05_t01.dart
```diff
@@ -0,0 +1,70 @@
...
+///
+/// @description Check that it is a compile-time error if the name of the
+/// primary constructor is the same as the name of some constructor declared in
+/// the body. Test classes.
+/// @author sgrekhov22@gmail.com
+
+// TODO (sgrekhov) Add `declaring-constructors` experimental flag
+
+class C1(var int v) {
+  C1.someName(this.v);
```
eernstg, line 17: This one would have to be redirecting as well.

LanguageFeatures/primary-constructors/static_processing_A05_t01.dart
```diff
@@ -0,0 +1,70 @@
...
+
+class C2(int v1) {
+  this(var int v2);
+//^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+class const C3.someName(final int v) {
+  const C3(this.v);
```
eernstg, line 32: And this one would be redirecting as well.

LanguageFeatures/primary-constructors/static_processing_A05_t01.dart
```diff
@@ -0,0 +1,70 @@
...
+  const C3.someName(int v) : this(v);
+//      ^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+class C4.someName(int v1) {
+  this.someName(var int v2);
```
eernstg, line 40: It's the same confusing situation as some time earlier: This could be reported as an error because there are two declaring formal parameter lists in the class (including the header). In this case the in-header constructor and the in-body constructor are the same constructor, and it's an error to have `(var int v2)` in the body.

 It could also be reported as an error because the two declarations are two constructors, and then they can't have the same name.

eernstg (commented):
I landed this one, but I'm afraid there are a couple of things that still need to be fixed in some other PR.

sgrekhov:
I'll fix these issues in the upcoming PR, thank you!

PR #3321, opened 2025-09-26
#3180. Update assertions text

sgrekhov:
No code changes in this PR. Only assertions text updated according to https://dart-review.googlesource.com/c/sdk/+/450140


eernstg (approved):
LGTM. Should be safe, I'll just land it.

PR #3328, opened 2025-10-01
#3315. Add tests for `covariant` keyword

sgrekhov:
Please note:
- primary constructors in `static_processing_A09_t02.dart` have implicit `final`. This may shadow an error in the description (because `covariant` + `final` is also an error). I don't mind to remove these checks.
- missing `final` keywords in `static_processing_A09_t03.dart` cause another error all memvers of an enum must be `final`.  This also may shadow an error that this tests is going to test. Again I don't mind to delete the test if needed.


LanguageFeatures/primary-constructors/static_processing_A09_t02.dart
```diff
@@ -0,0 +1,77 @@
...
+/// `var` nor `final`.
+///
+/// @description Check that it is a compile-time error if a formal parameters of
+/// a declaring constructor has the modifier `covariant`, but neither `var` nor
+/// `final`. Test extension types.
+/// @author sgrekhov22@gmail.com
+
+// TODO (sgrekhov) Add `declaring-constructors` experimental flag
+
+extension type ET1(covariant String x) {}
```
eernstg, line 20: `final` is implicit here, which means that this is not really a test of "neither `final` nor `var`". Perhaps drop the library as a whole? I think it's already covered by other libraries where the focus is on "both `final` and `covariant`".

LanguageFeatures/primary-constructors/static_processing_A09_t03.dart
```diff
@@ -0,0 +1,89 @@
...
+/// constructor, be it in the header or in the body. Let `p` be a formal
+/// parameter of a declaring constructor in a class, enum, or extension type
+/// declaration `D` named `C`:
+/// ...
+/// A compile-time error occurs if `p` has the modifier `covariant`, but neither
+/// `var` nor `final`.
+///
+/// @description Check that it is a compile-time error if a formal parameters of
+/// a declaring constructor has the modifier `covariant`, but neither `var` nor
+/// `final`. Test enums.
```
eernstg, line 15: Having `covariant` but neither `final` nor `var` is an error, and not having `final` is another error in the first place, so it could be claimed that we don't know what we're testing (and a bug could be undetected because we think we're testing one of them, but it's the other that causes the error).

However, I think it's useful to see that the lone `covariant` is handled (and not crashing, for example). So let's keep it.

sgrekhov (commented):
Updated. PTAL.

eernstg (approved):
LGTM

PR #3331, opened 2025-10-02
#3315. Add more tests for declaring constructors.

sgrekhov:


LanguageFeatures/Declaring-constructors/static_processing_A12_t03.dart
```diff
@@ -0,0 +1,49 @@
...
+
+import '../../Utils/static_type_helper.dart';
+
+class C1([var x]);
+
+class C2([final x]);
+
+class C3({final x});
+
+class C4({var x});
```
eernstg, line 42: I just made a small change to the spec: It was never intended that override inference should only apply to optional parameters, so I moved the word 'optional' such that the rule clearly applies to non-optional parameters as well.

Perhaps all the above could just have a mandatory positional declaring parameter as well, such that we can see that they also get their declared type based on these rules?

LanguageFeatures/Declaring-constructors/static_processing_A12_t04.dart
```diff
@@ -0,0 +1,84 @@
...
+/// value then `p` is considered to have the declared type `Object?`.
+///
+/// @description Check that if `p` does not have a declared type and its default
+/// value is `Null` or absent then the declared type of `p` is `Object?` not
+/// `dynamic`.
+/// @author sgrekhov22@gmail.com
+
+// TODO (sgrekhov) Add `declaring-constructors` experimental flag
+
+class C1([var x]);
```
eernstg, line 35: These declarations could also have a mandatory positional declaring parameter as well.

By the way: It is important that this is only concerned with declaring parameters because we can't change the type of non-declaring parameters (without going through a heavy breaking change process). Declaring parameters are new, so we won't break any existing code by using the `Object?` default with them.

LanguageFeatures/Declaring-constructors/static_processing_A14_t01.dart
```diff
@@ -0,0 +1,74 @@
...
+/// @author sgrekhov22@gmail.com
+
+// TODO (sgrekhov) Add `declaring-constructors` experimental flag
+
+import '../../Utils/expect.dart';
+
+class const C1(final int v);
+
+class C2 {
+  const (final int v);
```
eernstg, line 34: Probably:

```suggestion
  const this(final int v);
```

LanguageFeatures/Declaring-constructors/static_processing_A14_t01.dart
```diff
@@ -0,0 +1,74 @@
...
+  const this(final int v);
+}
+
+enum E3(final int v) {
+  e0(3);
+}
+
+enum E4 {
+  e0(4);
+  this(final int v);
```
eernstg, line 58: This one can't omit `const`, just like all other `enum` constructors. It is the intention that the implicit `const` only occurs in the header, did you find something in the spec to the contrary?

LanguageFeatures/Declaring-constructors/static_processing_A11_t01.dart
```diff
@@ -0,0 +1,55 @@
...
+  num get x => 0;
+}
+
+class C1([var x = 1]) extends A1;
+
+class C2([final x = 2]) extends A2;
+
+class C3({final x = 1}) extends A1;
+
+class C4({var x = 2}) extends A2;
```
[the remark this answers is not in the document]
sgrekhov: Makes sense, thank you. Added.

LanguageFeatures/Declaring-constructors/static_processing_A14_t01.dart
sgrekhov: Yes, of course. Fixed.

LanguageFeatures/Declaring-constructors/static_processing_A12_t04.dart
sgrekhov: Added.

LanguageFeatures/Declaring-constructors/static_processing_A12_t03.dart
sgrekhov: Updated.

sgrekhov (commented):
Thank you. Updated. PTAL.

sgrekhov:
Thank you. Missing arguments and `const` added. PTAL.

eernstg (approved):
LGTM

PR #3335, opened 2025-10-03
#3315. Fix spec_parser failures

sgrekhov:


LanguageFeatures/Declaring-constructors/static_processing_A03_t07.dart
```diff
@@ -21,7 +21,7 @@ extension type ET1(int v) {
 }
 
 extension type const ET2.someName(final int v) {
-  factory const ET2(int v) = ET2.someName;
+  factory ET2(int v) = ET2.someName;
```
eernstg, line 24: Same thing here, possibly other places.

eernstg (commented):
Looks good! Couple of comments, just noting that we might preserve `const` in a few cases (I'm assuming that this makes the overall approach more consistent).

sgrekhov:
Thank you. Updated. PTAL.

PR #3350, opened 2025-10-14
#3315. Add tests for the `new` keyword

sgrekhov:


LanguageFeatures/Declaring-constructors/static_processing_A25_t01.dart
```diff
@@ -0,0 +1,101 @@
...
+class C7 {
+  final int v;
+  new({required int v}) : v = v;
+  new.someName({required int v}) : v = v;
+}
+
+class C8 {
+  final int v;
+  new(int v) : v = v;
+  const factory new.someName(int v) = C8;
```
eernstg: This implies that line 70 would need `const`, too.

LanguageFeatures/Declaring-constructors/static_processing_A25_t02.dart
```diff
@@ -0,0 +1,74 @@
...
+}
+
+extension type ET5._(int v) {
+  const new({required int v}) : v = v;
+  new.someName({required int v}) : v = v;
+}
+
+extension type ET6._(int v) {
+  new(int v) : v = v;
+  const factory new.someName(int v) = ET6;
```
eernstg: Needs `const` in line 44.

Issue #3203, opened 2025-05-23
Proposal: write tests that exercise the interaction of expression evaluation with new language features


derekxu16:
This issue tracks a proposal that is planned to be discussed in next week's meeting.

The VM team would like the co19 team to write tests that exercise the interaction of expression evaluation with new language features, when new language features are added. Our plan is for the VM team to write a package that makes it easy to write tests similar to the ones in `package:vm_service`; that is, ones that start a testee VM and make VM Service RPC calls on that testee. For example, [evaluate_uninitialized_late_variable_test](https://github.com/dart-lang/sdk/blob/dfb5f034a90151441314fd30385b622c2c9201ee/pkg/vm_service/test/evaluate_uninitialized_late_variable_test.dart). Then co19 would use that package to write expression evaluation tests.

sgrekhov:
@derekxu16 Could you please provide some more details? As far as I understand, there will be a new package that should be used for future testing. What exactly do you want to test using this new package?

sgrekhov:
Thanks for the example! Do I understand correctly that the requested co19 tests should do roughly the same as the function "// Test interaction of expression evaluation with dot shorthands."? Also, what role is the new package supposed to play? Why not use the `developer.dart` package?

derekxu16:
> Do I understand correctly that the requested co19 tests should do roughly the same as the function "// Test interaction of expression evaluation with dot shorthands."?

Yes, that's correct.

> Also, what role is the new package supposed to play? Why not use the `developer.dart` package?

The package will provide utilities that make it easier to start a testee VM and make VM Service RPC calls on that testee. One of the most important utilities that it will provide is an equivalent of [`runIsolateTests`](https://github.com/dart-lang/sdk/blob/6b53073eae118baec4fd5b9c9b6e8cf743df1587/pkg/vm_service/test/common/test_helper.dart#L457-L506). Most developers will not need to use these utilities, so they are not a good fit for `dart:developer`. Perhaps a new package does not need to be made for them, and they can just be exposed from `package:vm_service`. I'll discuss this with the team.

derekxu16:
Actually, since dependencies are resolved according to the Dart SDK's `package_config.json` when `co19` tests are run, do you think it would be ok if your tests just import the utilities from `'package:vm_service/test/common/test_helper.dart'`, or would you prefer for the utilities to be made public APIs?

sgrekhov:
co19 tests are not part of the Dart SDK.
[...]

PR #3359, opened 2025-10-21
#3203. Add VM - CFE interaction tests for primary initializer scope

sgrekhov:


VM/declaring_constructors_t02.dart
```diff
@@ -0,0 +1,104 @@
...
+/// @assertion A declaring constructor declaration is a declaration that
+/// contains a `<declaringConstructorSignature>` with a
+/// `<declaringParameterList>`, or a declaration that contains a
+/// `<declaringConstantConstructorSignature>`, or it is a
+/// `<primaryConstructorNoConst>` in the header of a class, enum, or extension
+/// type declaration, together with a declaration in the body that contains a
+/// `<declaringConstructorSignature>`.
+///
+/// @description Check that members initialized in primary initializer scope can
+/// be debugged.
```
nshahan: I may be overlooking something but is this testing what is in scope when paused at the primary constructor (`LINE_A`)?

sgrekhov: Hm, I'm afraid not. Seems that you are expecting the test like the below (example is written without primary constructors):
```dart
class C {
  String x;
  C(this.x);          // LINE_A
}

void testeeMain() {
  debugger();         // LINE_B
  var c = C("xxx");   // LINE_C
}

final tests = <IsolateTest>[
  hasStoppedAtBreakpoint,
  stoppedAtLine(LINE_B),
  stepInto,
  stoppedAtLine(LINE_C),
  stepInto,
  stoppedAtLine(LINE_C),
  stepInto,
  stoppedAtLine(LINE_A),

  (VmService service, IsolateRef isolateRef) async {
    final isolateId = isolateRef.id!;
    final isolate = await service.getIsolate(isolateId);
    final lib =
        (await service.getObject(isolateId, isolate.rootLib!.id!)) as Library;
    // Check the value of 'x' here
  },
];
```
Please confirm that it's what you'd expect.

eernstg: Interesting! In `LINE_A` we have the constructor declaration, with no initializer list and no constructor body (but we could have those, too), and `x` in the initializer list would denote the formal parameter whereas `x` in the constructor body would denote the instance variable. I think this means that `LINE_A` is simply too ambiguous to answer questions like "do we have access to `x`?". The parameter and the instance variable may have different values (for instance, with `C(String x): assert(x != ""), x = x { x = ""; }`, the instance variable does not have a value until the second element in the initializer list has been executed, and it gets a new value in the body.

It would be possible to make the location in code more fine grained by inserting line breaks, e.g.:

```dart
class C {
  String x;
  C(
    this.x // LINE_A
  );
}
```

At the comment, the formal parameter `x` is not in scope, and the instance variable `x` may or may not have been initialized (depends on whether we're before or after `this.x` in that line).

However, it certainly makes sense to show the value of a formal parameter at the time where it is used (in any way, including the case where it's used to initialize an instance variable as it is here).

I think this implies that "what's in scope?" is a (slightly) wrong question to ask when it is decided what to show in a debugger, which is probably also reflected by the implemented behavior.

@nshahan, WDYT? Is there a concept which is similar to the language-level notion of a scope which is used for this purpose?

nshahan: > Please confirm that it's what you'd expect.

@sgrekhov 
That looks like what I thought the description comment at the top of the file was saying.

eernstg: @nshahan, I think this thread is clarifying a broader set of topics that aren't blocking for this test to land. Do you agree?

eernstg (approved):
LGTM, awaiting further input.

eernstg (approved):
LGTM

PR #3367, opened 2025-10-27
#2825. Fix resolving extensions tests

sgrekhov:
Now both tests pass


sgrekhov:
> Ah, now I understand! ;-) The explicit resolution is indeed required because there is more than one extension with a member whose name is `id`. OK!

Yes, exactly. Thank you!

PR #3370, opened 2025-10-28
#2825. Add sanity rules tests

sgrekhov:


sgrekhov:
My concern is that older versions don't support the Enhanced Parts feature. For example, there is an issue: https://github.com/dart-lang/sdk/issues/61845. To make the affected test pass, we should use a version of the analyzer that contains the fix.

eernstg:
OK! I think this is not testing things in the 'Explicit Sanity Rules' section, so it isn't blocked by https://github.com/dart-lang/sdk/issues/61859#issuecomment-3461025960, right?

sgrekhov:
Right! It isn't blocked by that issue.

Issue #3329, opened 2025-10-01
[private-named-parameters] Add co19 tests for the "Private Named Parameters" feature


munificent:
The language team has accepted the proposal for ["Private Named Parameters"](https://github.com/dart-lang/language/blob/main/accepted/future-releases/2509-private-named-parameters/feature-specification.md) and implementation is starting soon.

This issue is for adding co19 test coverage. Let me know if you have any questions. Thanks!

cc @sgrekhov @eernstg

PR #3375, opened 2025-10-30
#3329. Add more private named parameters tests

sgrekhov:


LanguageFeatures/Private-named-parameters/static_semantics_A01_t05.dart
```diff
@@ -0,0 +1,84 @@
...
+//                   ^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+enum E1({required final String __}) {
+//                             ^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  e0(_: "");
```
eernstg: Good catch, too!

It makes sense to check that we don't accept funny invocations based on a wrong "corresponding public name" which isn't public.

LanguageFeatures/Private-named-parameters/static_semantics_A02_t04.dart
```diff
@@ -0,0 +1,69 @@
...
+/// @assertion Given a named initializing formal or field parameter (for a
+/// primary constructor) with private name `p` in constructor `C`:
+/// - If `p` has no corresponding public name `n`, then compile-time error. You
+///   can't use a private name for a named parameter unless there is a valid
+///   public name that could be used at the call site.
+/// - If any other parameter in `C` has declared name `p` or `n`, then
+///   compile-time error.
+///
+/// @description Check that it is a compile-time error if any other parameter in
+/// `C` has declared name `p` or `n`.
```
eernstg:
```suggestion
/// `C` has declared name `p` or `n`. Test declaring parameters.
```

LanguageFeatures/Private-named-parameters/super_parameters_A01_t01.dart
```diff
@@ -0,0 +1,44 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A super parameter generates an implicit argument that forwards to
+/// a superclass constructor. The super constructor's argument name is always
+/// public, even if the corresponding constructor parameter uses this feature
+/// and has a private name. Thus, super parameters continue to always use public
+/// names.
+///
+/// @description Check that private super parameter can be invoked using its
+/// public name.
```
eernstg: I think it would make sense (and cover a slightly different amount of code in the tools) to have a similar test where the superinitializer is written explicitly:

```dart
class B1 extends A1 {
  B1(String p) : super(p: p);
  B1.named(String p) : super(p: p);
}
```

LanguageFeatures/Private-named-parameters/static_semantics_A02_t04.dart
sgrekhov: Updated.

PR #3377, opened 2025-10-31
#3203. Add tests for private named parameters

sgrekhov:


VM/private_named_parameters_t01.dart
```diff
@@ -0,0 +1,114 @@
...
+        await service.evaluateInFrame(isolateId, 0, '_x') as InstanceRef;
+    Expect.equals('xxx', xRef1.valueAsString);
+    final xRef2 =
+        await service.evaluateInFrame(isolateId, 0, 'this._x') as InstanceRef;
+    Expect.equals('xxx', xRef2.valueAsString);
+    final xRef3 = await service.evaluateInFrame(isolateId, 0, 'x');
+    Expect.isTrue(xRef3 is ErrorRef);
+    final xRef4 = await service.evaluateInFrame(isolateId, 0, 'this.x');
+    Expect.isTrue(xRef4 is ErrorRef);
+},
```
eernstg: Typo:
```suggestion
  },
```

sgrekhov: Fixed

VM/private_named_parameters_t01.dart
```diff
@@ -0,0 +1,114 @@
...
+    final xRef1 =
+        await service.evaluateInFrame(isolateId, 0, '_x') as InstanceRef;
+    Expect.equals('xxx', xRef1.valueAsString);
+    final xRef2 =
+        await service.evaluateInFrame(isolateId, 0, 'this._x') as InstanceRef;
+    Expect.equals('xxx', xRef2.valueAsString);
+    final xRef3 = await service.evaluateInFrame(isolateId, 0, 'x');
+    Expect.isTrue(xRef3 is ErrorRef);
+    final xRef4 = await service.evaluateInFrame(isolateId, 0, 'this.x');
+    Expect.isTrue(xRef4 is ErrorRef);
```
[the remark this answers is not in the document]
eernstg: Thanks!

VM/private_named_parameters_t01.dart
```diff
@@ -0,0 +1,188 @@
...
+  stepInto,
+  stoppedAtLine(LINE_C),
+  (VmService service, IsolateRef isolateRef) async {
+    final isolateId = isolateRef.id!;
+    final xRef1 =
+        await service.evaluateInFrame(isolateId, 0, '_x') as InstanceRef;
+    Expect.equals("null", xRef1.valueAsString);
+    final xRef2 =
+        await service.evaluateInFrame(isolateId, 0, 'this._x') as InstanceRef;
+    Expect.equals("null", xRef2.valueAsString);
```
nshahan: It is confusing to me that we expect `_x` and `this._x` to both be `null` at this time. Why?

sgrekhov: Based on the current behaviour of
```dart 
class C {
  String x;
  C(this.x);
}
```
`x` is `null` at this step. Therefore, I believe, that in this test `_x` also should be `null`.

eernstg (approved):
(Send an unsent comment.)

VM/private_named_parameters_t01.dart
eernstg: If the current location in code is the initializer list (we could add an initializer list, or we might have stopped at a breakpoint "in the initializer list" even though it's empty, which could be the way the VM sees it) then `x` would denote the parameter. In contrast, at any location with access to `this`, including the body of the constructor (if any), `x` would denote the instance variable, and so would `this.x`.

The same would be true with `_x`.

If the current position in code is the initializer list it _might_ still be possible to access the instance variables of the enclosing object using terms like `this.x` or `this._x`. The language doesn't allow us to do that, but the debugger might.

So I've argued that we should at least know how it's supposed to work, and preferably look at the treatment of both the parameter and the instance variable in situations where it is the intention that both can be viewed and possibly modified.

Issue #3330, opened 2025-10-02
[Static Extensions] Add co19 tests for the Static Extensions feature


chloestefantsova:
This is a meta-issue for the tests for the Static Extensions language feature.

_The language feature is currently being specified. This is a provisional issue and may not be actionable until the specification is written. The current draft of the specification can be found here: https://github.com/dart-lang/language/blob/main/working/0723-static-extensions/feature-specification.md._

PR #3380, opened 2025-11-03
#3330. Add tests for static extensions syntax

sgrekhov:


sgrekhov:
Part of the spec related to constructor isn't stable yet. Waiting for the spec.

PR #3387, opened 2025-11-07
#3330. Add test for the lexical lookup

sgrekhov:
Surprisingly, we have no such test among the tests for extension methods.


eernstg (commented):
Looks good! I've suggested an adjustment of some descriptions.

PR #3388, opened 2025-11-11
#2559. Update `augmenting_constructors_A01_*` tests

sgrekhov:
- Augmentations from parts moved to the main test to improve readability
- Assertion text updated
- Experimental flag changed
- Declarations updated to avoid two completed declarations in the chain of the augmentations


LanguageFeatures/Augmentation-libraries/augmenting_constructors_A01_t02.dart
```diff
@@ -3,37 +3,102 @@
...
-///   constructor. It must have the same number of positional parameters, the
-///   same named parameters, and matching parameters must have the same type,
-///   optionality, and any required modifiers must match. Any initializing
-///   formals and super parameters must also be the same in both constructors.
+/// - The signature of the augmenting function does not match the signature of
+///   the augmented function.
 ///
 /// @description Checks that it is a compile-time error if the signature of the
 /// constructor augmentation does not match the original constructor. Test wrong
 /// number of optional parameters.
```
eernstg: Just to avoid the ambiguity:
```suggestion
/// number of optional positional parameters.
```

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A01_t10.dart
```diff
@@ -37,26 +53,54 @@ enum E {
...
+  ET.bar({required int x, final int y = 0});
+  ET.baz(int x, [final int y = 0]);
+  ET.qux({required int x, final int y = 0});
+}
+
+augment extension type ET {
+  augment ET.foo(this.x, [final y]) {
+    x.expectStaticType<Exactly<int>>();
+    y.expectStaticType<Exactly<int>>();
+    _log = "Augmented: $x, $y";
```
eernstg: The `C` tests above don't write to the `_log` and there's no test case (old lines 50, 53, 56, 59). Is there a specific reason why it is helpful here but unnecessary in the `C` tests?

I'm guessing that it would be fine to add this element to the `C` tests as well.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A01_t02.dart
sgrekhov: Thanks. Updated.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A01_t11.dart
```diff
@@ -3,29 +3,45 @@
...
-/// - The signature of the constructor augmentation does not match the original
-///   constructor. It must have the same number of positional parameters, the
-///   same named parameters, and matching parameters must have the same type,
-///   optionality, and any required modifiers must match. Any initializing
-///   formals and super parameters must also be the same in both constructors.
+/// - The signature of the augmenting function does not match the signature of
+///   the augmented function.
 ///
 /// @description Checks that it is not an error if a constructor augmentation
 /// replaces type of formal parameter by `var`.
```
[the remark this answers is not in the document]
sgrekhov: Without `var` this test becomes a copy of the previous one (with removed `final`). Removed.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A01_t12.dart
```diff
@@ -3,61 +3,104 @@
...
-///   constructor. It must have the same number of positional parameters, the
-///   same named parameters, and matching parameters must have the same type,
-///   optionality, and any required modifiers must match. Any initializing
-///   formals and super parameters must also be the same in both constructors.
+/// - The signature of the augmenting function does not match the signature of
+///   the augmented function.
 ///
 /// @description Checks that it is not an error if a constructor augmentation
 /// specifies a type of formal parameter which were not explicitly specified in
 /// the introductory constructor.
```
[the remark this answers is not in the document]
sgrekhov: Right. This PR is the update of the existing tests. And I didn't checked that this test is not valid any longer. Thank you for catching it. Updated to expect an error in this case.

sgrekhov (commented):
Thank you! Updated. PTAL.

LanguageFeatures/Augmentation-libraries/augmenting_constructors_A01_t02.dart
eernstg: Thanks!

PR #3395, opened 2025-11-14
#3315. Add test checking the special rule for factory constructors

sgrekhov:


LanguageFeatures/Primary-constructors/grammar_A05_t01.dart
```diff
@@ -0,0 +1,74 @@
...
+class C4 {
+  int v;
+  C4.foo(this.v);
+  factory C4(int v);
+
+  augment factory C4(int v) => C4.foo(v + 1);
+}
+
+class C5 {
+  external factory C5();
```
eernstg: So there will be a status file entry preventing executions of this test on configurations where the generated code is JavaScript? How about WASM? (I can't remember how we usually handle this situation where a web compiler requires `@JS(...)` on all externals).

eernstg (commented):
Looks good, just one question: Is there something extra to do for web compilers?

LanguageFeatures/Primary-constructors/grammar_A05_t01.dart
eernstg: Thanks!

PR #3413, opened 2025-11-24
#3182. Update augmenting_constructors_A02-04* tests

sgrekhov:


sgrekhov:
Spec parser doesn't support augmenting enums with no values and therefore fails.

PR #3415, opened 2025-11-25
#3182. Update augmenting_constructors_A09* tests

sgrekhov:
These tests also will be renamed according to their actual assertions

### DIFF
--- a/LanguageFeatures/Augmentations/augmenting_constructors_A09_t02.dart
+++ b/LanguageFeatures/Augmentations/augmenting_constructors_A09_t02.dart
@@ -9,7 +9,7 @@
 /// extra rules to handle features unique to constructors like redirections and
 /// initializer lists.
 ///
-/// @description Checks that if the augmenting constructor may provide a body.
+/// @description Checks that the augmenting constructor may provide a body.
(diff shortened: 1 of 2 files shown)


LanguageFeatures/Augmentations/augmenting_constructors_A09_t03.dart
```diff
@@ -2,29 +2,54 @@
...
 // BSD-style license that can be found in the LICENSE file.
 
-/// @assertion At a high level, a non-redirecting generative constructor marked
-/// `augment` may:
-/// - Augment the constructor with an additional constructor body (bodies are
-///   invoked in augmentation order, starting at the introductory declaration).
+/// @assertion It's a compile-time error if there is no matching declaration.
 ///
 /// @description Checks that it is a compile-time error to augment a default
 /// unnamed constructor (that doesn't exist during augmentation).
```
eernstg: This comment is _very_ useful, especially when the reader may not remember exactly why this is an error!

sgrekhov:
Thanks! Updated. Primary constructors of extension types are replaced by regular constructors. PTAL.

PR #3417, opened 2025-11-26
#3182. Update augmenting_constructors_A11-18* tests

sgrekhov:
Also it's just a first step. The next step is renaming.

### DIFF
--- a/LanguageFeatures/Augmentations/augmenting_constructors_A17_t01.dart
+++ b/LanguageFeatures/Augmentations/augmenting_constructors_A17_t01.dart
@@ -6,8 +6,7 @@
 /// initializer list and/or a body, or by adding a redirection.
 ///
 /// @description Checks that an augmenting redirecting generative constructor
-/// may add redirecting initializer to the augmented constructors initializer
-/// list.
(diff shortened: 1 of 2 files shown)


LanguageFeatures/Augmentations/augmenting_constructors_A17_t01.dart
```diff
@@ -2,21 +2,17 @@
...
-/// @assertion A redirecting generative constructor marked `augment` adds its
-/// redirection to the augmented constructor.
-///
-/// This converts it into a redirecting generative constructor, removing the
-/// potentially non-redirecting property of the constructor.
+/// @assertion An incomplete constructor can be completed by adding an
+/// initializer list and/or a body, or by adding a redirection.
 ///
 /// @description Checks that an augmenting redirecting generative constructor
 /// may add redirecting initializer to the augmented constructors initializer
```
eernstg: I think we should just call it a redirection:
```suggestion
/// may add a redirection to the augmented constructors initializer
```
Otherwise the terminology seems to imply that the redirection is just one element and there can be other initializer list elements. The fact that they both use `:` as a separator doesn't turn the redirection into an initializer list.

PR #3418, opened 2025-11-26
#3182. Update augmenting_constructors_A19-20* tests

sgrekhov:
Again, the next step is the renaming.

### DIFF
--- a/LanguageFeatures/Augmentations/augmenting_constructors_A20_t03.dart
+++ b/LanguageFeatures/Augmentations/augmenting_constructors_A20_t03.dart
@@ -49,6 +49,7 @@ augment extension type ET {
 //                ^
 // [analyzer] unspecified
 // [cfe] unspecified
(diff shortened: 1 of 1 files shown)


LanguageFeatures/Augmentations/augmenting_constructors_A20_t03.dart
```diff
@@ -2,30 +2,58 @@
...
+augment extension type ET {
+  augment factory ET(int x) = ET.bar;
+//                ^
+// [analyzer] unspecified
+// [cfe] unspecified
+  augment factory ET.foo(int x) = ET;
+//                ^
+// [analyzer] unspecified
+// [cfe] unspecified
+//                ^
```
eernstg: Probably:
```suggestion
  augment factory ET.bar(int x) = ET.foo;
//                ^
```

PR #3421, opened 2025-11-28
#3182. Add tests for constructors with initializing formals

sgrekhov:


LanguageFeatures/Augmentations/augmenting_constructors_A09_t03.dart
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion An incomplete constructor can be completed by adding an
+/// initializer list and/or a body, or by adding a redirection.
+///
+/// @description Checks that constructor with the initializing formals can be
```
eernstg: Typo:
```suggestion
/// @description Checks that a constructor with initializing formals can be
```
'initializing formals' would not have the article 'the' because we don't know (and we don't care) exactly which ones we're talking about, we're just making a distinction between the case where the constructor doesn't have any initializing formals and the case where it has _some_ initializing formals (no matter what they are).

Perhaps add 'by an incomplete declaration' to the description in line 9, because it would be an error to have two complete declarations in the same augmentation chain.

LanguageFeatures/Augmentations/augmenting_constructors_A09_t03.dart
```diff
@@ -0,0 +1,82 @@
...
+enum E {
+  e0(0), e1.foo(), e2.foo(2), e3.bar(), e4.bar(x: 3), e5.baz(x: 4);
+  final int x;
+  const E(this.x);
+  const E.foo([this.x]);
+  const E.bar({this.x});
+  const E.baz({required this.x});
+}
+
+augment enum E {
```
eernstg: I can't see grammar support for omitting the part that declares enumerated values, even in an augmenting declaration.

eernstg (commented):
Looks good, with a few comments: One `enum` declaration must be changed, and a couple of descriptions can be made a bit more precise.

LanguageFeatures/Augmentations/augmenting_constructors_A09_t03.dart
eernstg: Yes, opening my eyes helped, I can see it now. ;-D

eernstg (commented):
The value-less enum augmentation _is_ allowed today after all, so one comment is gone.

LanguageFeatures/Augmentations/augmenting_constructors_A09_t04.dart
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion An incomplete constructor can be completed by adding an
+/// initializer list and/or a body, or by adding a redirection.
+///
+/// @description Checks that an incomplete constructor can be completed by the
+/// augmenting constructor with the initializing formals.
```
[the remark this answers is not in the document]
sgrekhov: Done.

PR #3448, opened 2025-12-18
#3182. Add more augmenting getters/setters tests

sgrekhov:


sgrekhov:
The only thing that cames to my ming is:
```dart
enum E {
  e0;
}

augment enum E {
  augment e0; // Error.  Constant is augmenting and augmented.
}
```
But, firsts, we already have such test. Second, here we again have two errors. Constant is augmented and augmenting. 
Does it makes sense to add test chain like 
`final...`
`augment final...`
`augment const ...`
`augment final...`?

PR #3449, opened 2025-12-19
#3182. Add test for late variables

sgrekhov:


LanguageFeatures/Augmentations/augmenting_variables_getters_setters_A01_t13.dart
```diff
@@ -0,0 +1,91 @@
...
+/// otherwise they are complete. For non-abstract variables, the compiler
+/// synthesizes a getter that accesses the backing storage and a setter that
+/// updates it, so these members have bodies.
+///
+/// A getter can be augmented by another getter, and likewise a setter can be
+/// augmented by a setter. This is true whether the getter or setter is
+/// explicitly declared or implicitly declared using a variable declaration.
+///
+/// @description Checks that a `final late` variable may be augmented by a
+/// getter.
```
eernstg: It would probably be helpful to mention the initialization (because late variables with and without an initializing expression are so different):
```suggestion
/// @description Checks that a `final late` variable with an initializing
/// expression may be augmented by a getter.
```

LanguageFeatures/Augmentations/augmenting_variables_getters_setters_A01_t14.dart
```diff
@@ -0,0 +1,99 @@
...
+// SharedOptions=--enable-experiment=augmentations
+
+import '../../Utils/expect.dart';
+
+int get topLevelGetter;
+augment late final int topLevelGetter = 1;
+
+class C {
+  static int get staticGetter;
+  int get instanceSetter;
```
eernstg: Probably:
```suggestion
  int get instanceGetter;
```

LanguageFeatures/Augmentations/augmenting_variables_getters_setters_A01_t14.dart
```diff
@@ -0,0 +1,99 @@
...
+}
+
+augment class C {
+  augment static late final int staticGetter = 2;
+  augment late final int instanceGetter = 3;
+}
+
+mixin M {
+  static int get staticGetter;
+  int get instanceSetter;
```
eernstg: Ditto.

eernstg: A few more like this below.

LanguageFeatures/Augmentations/augmenting_variables_getters_setters_A03_t12.dart
```diff
@@ -0,0 +1,104 @@
...
+/// If the variable is `abstract`, then the getter and setter are incomplete,
+/// otherwise they are complete. For non-abstract variables, the compiler
+/// synthesizes a getter that accesses the backing storage and a setter that
+/// updates it, so these members have bodies.
+///
+/// A getter can be augmented by another getter, and likewise a setter can be
+/// augmented by a setter. This is true whether the getter or setter is
+/// explicitly declared or implicitly declared using a variable declaration.
+///
+/// @description Checks that a `late` variable may be augmented by a setter.
```
eernstg: Given that the other tests have been explicit about the `final` modifier, and given that it makes a difference for the setter/no-setter distinction that non-final variables don't care about the existence of an initializing expression (it's always present), perhaps:
```suggestion
/// @description Checks that a `late` but not `final` variable may be augmented
/// by a setter.
```
This also explains to the reader why the test doesn't contain cases with `final` as well. They might thing "just give us all variants" if they don't see that the cases are different.

LanguageFeatures/Augmentations/augmenting_variables_getters_setters_A06_t02.dart
```diff
@@ -0,0 +1,79 @@
...
+  c.instanceVariable = 1;
+  Expect.equals(1, c.instanceVariable);
+
+  var m = MA();
+  Expect.equals(0, m.instanceVariable);
+  m.instanceVariable = 1;
+  Expect.equals(1, m.instanceVariable);
+
+  Expect.equals(0, E.e0.instanceVariable);
+}
```
eernstg: Hmm, could this be essentially a duplicate of an existing test? It looks familiar.

sgrekhov: I've just double checked and didn't find the duplicate. The earlier tests are really similar but they check variable - augmenting getter/setter (sometimes implicit) and vice versa but there were no this particular case when incomplete variable is augmented by a complete  variable.

sgrekhov:
Updated. PTAL.

PR #3456, opened 2025-12-25
#3182. Update augmenting_types_A06_*.dart tests

sgrekhov:


LanguageFeatures/Augmentations/augmenting_types_A06_t03.dart
eernstg: We wouldn't get an error here any more (because of https://github.com/dart-lang/sdk/issues/58578).

Perhaps change line 23 to declare a method, that's still an error together with the getter.

eernstg (commented):
Looks good, but one test seems to rely on a getter/setter error which is no longer part of the language.

sgrekhov (commented):
Updated. PTAL.

PR #3457, opened 2025-12-25
#3182. Update augmenting_types_A07_*.dart tests

sgrekhov:


eernstg (commented):
Looks good! We need to take one more look at LanguageFeatures/Augmentations/augmenting_types_A07_t07.dart because it relies on an error which is potentially being removed (I think basically everyone in the language team wants to drop this error). I'm not sure, but perhaps it would make sense to add a TODO comment or something like that.

sgrekhov:
I added a TODO item for myself. I also subscribed to the issue and will update the test if the spec change is landed.

eernstg (approved):
LGTM

Issue #3595, opened 2026-02-04
Add tests for dot shorthand access to private members


sgrekhov:
Add dot shorthand access tests according https://github.com/dart-lang/language/issues/4606

PR #3599, opened 2026-02-04
#3595. Add tests for private class-like declarations

sgrekhov:


LanguageFeatures/Static-access-shorthand/type_inference_A09_t01.dart
```diff
@@ -0,0 +1,62 @@
...
+/// access an unavailable class.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=dot-shorthands
+
+import 'type_inference_A09_lib.dart';
+
+main() {
+  var c = cInstance;
+  c = .new(1);
```
eernstg: I'm tempted to suggest that the context type is provided by passing an actual argument to a function declared in the `_lib.dart` library (similarly to an earlier test), such that we don't depend on the context type being the same all the way. If, for example, the type of `c` is "promoted" to `InvalidType` then we might suddenly test the wrong thing in all but the first test case.

LanguageFeatures/Static-access-shorthand/type_inference_A09_t02.dart
```diff
@@ -0,0 +1,58 @@
...
+/// `.new(args)` try access an unavailable class.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=dot-shorthands
+
+import 'type_inference_A09_lib.dart';
+
+main() {
+  PublicC c = cInstance;
+  c = .new(1);
```
eernstg: The same issue could arise here, something like `contextPublic` would be safer.

PR #3607, opened 2026-02-09
#3182. Add tests for augmentation of primary constructors

sgrekhov:
Tests with even numbers in this PR are just copies of the corresponding tests with odd numbers, but with declared parameters. In the case of extension types, they just have an optional `final` in the primary constructor. If you believe this is excessive, I don't mind removing these tests. Please review.

### DIFF
--- a/LanguageFeatures/Augmentations/augmenting_constructors_A10_t02.dart
+++ b/LanguageFeatures/Augmentations/augmenting_constructors_A10_t02.dart
@@ -28,8 +28,8 @@
 ///   the augmented function.
 ///
 /// @description Checks that it is a compile-time error if the signature of the
-/// constructor augmentation does not match the original constructor. Test wrong
-/// number of positional parameters of a declaring primary constructor.
+/// constructor augmentation does not match the original constructor. Test an
+/// incorrect number of declaring positional parameters of a primary constructor
 /// @author sgrekhov22@gmail.com
 
 // SharedOptions=--enable-experiment=augmentations,primary-constructors
--- a/LanguageFeatures/Augmentations/augmenting_constructors_A01_t01.dart
+++ b/LanguageFeatures/Augmentations/augmenting_constructors_A01_t01.dart
@@ -28,8 +28,8 @@
 ///   the augmented function.
 ///
 /// @description Checks that it is a compile-time error if the signature of the
-/// constructor augmentation does not match the original constructor. Test wrong
-/// number of positional parameters.
+/// constructor augmentation does not match the original constructor. Test an
+/// incorrect number of positional parameters.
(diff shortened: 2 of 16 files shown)


LanguageFeatures/Augmentations/augmenting_constructors_A10_t02.dart
```diff
@@ -0,0 +1,104 @@
...
+///     corresponding positional parameter in every preceding declaration that
+///     doesn't have `_` as its name.
+/// ...
+/// It is a compile-time error if:
+/// - The signature of the augmenting function does not match the signature of
+///   the augmented function.
+///
+/// @description Checks that it is a compile-time error if the signature of the
+/// constructor augmentation does not match the original constructor. Test wrong
+/// number of positional parameters of a declaring primary constructor.
```
eernstg: I don't think we have such a notion as a declaring primary constructor. _Every_ primary constructor has a declaring parameter list, which could make it 'a declaring constructor', but the concept isn't useful if it applies to every occurrence. We could use this terminology to denote those constructors which are primary and have at least one declaring parameter, but that seems kind of artificial.

So, perhaps:

```suggestion
/// number of declaring positional parameters of a primary constructor.
```

You could say that "There is only one declaration in the augmentation chain where that parameter has the keyword `final` or `var`", but the parameter as such has this implementation aspect whether or not each of its declarations mentions it (and none of them _can_ mention it except the unique declaration in the chain which is complete).

So I think it's OK to talk about "a declaring positional parameter" even in the case where we are talking about the parameter as such (which means that we're considering the entire augmentation chain that defines it).

eernstg: About 3 more occurrences below.

eernstg (approved):
LGTM

sgrekhov:
> If you believe this is excessive
> 
> I think there is a good chance that the declaring parameters and the non-declaring parameters will cause a different subset of the code to be executed as these signature mismatches are detected, so let's just keep all of them. Thanks for thinking about it!

Thank you. But, I think I wasn't clear enough. So, just in case, I want to make a double check. I meant extension types. Say we have a test for primary constructor without declaring parametes. But we can't have an extension type without declaring parameters, therefore we the test looks like
```dart
extension type ET(int v) {
// The test
}
```
Then we have the same test for declaring parameters. In case of extension types it looks like
```dart
extension type ET(final int v) {
// The same test
}
```
If the second test checks that an optional `final` doesn't introduce any new isssues? Or it is just a duplicate of the first one? I'd keep it but don't mind removing as well.

sgrekhov:
Thank you! https://github.com/dart-lang/co19/pull/3609

PR #3620, opened 2026-02-13
#3182. Add tests for external js interop functions

sgrekhov:
These tests checks that external js interop functions can be augmented. In particular they should prove that metadata can be added by an anugmentation and this metadata works. 

Similar tests for variables, getters, setters and class-like declarations will be added later in separate PR's. 

Please review.


eernstg (approved):
Cool! LGTM.

PR #3624, opened 2026-02-18
#3182. Update constructors test according to the new rules for `_`

sgrekhov:


eernstg (commented):

PR #3643, opened 2026-03-04
#3182. Update existing tests for const and final initialization

sgrekhov:


Language/Variables/final_or_static_initialization_t04.dart
```diff
@@ -0,0 +1,47 @@
...
+/// modifier `late` or the modifier `external`.
+///
+/// @note There is no this statement in the specification, it is defined by the
+/// grammar.
+///
+/// @description Checks that it is not an error if an external library or static
+/// final variable is not initialized in its declaration.
+/// @author sgrekhov22@gmail.com
+
+external final int v;
```
eernstg: This test would need the usual protection against being executed by a JS based configuration. Is that already in place?

eernstg: Sounds good, thanks!

Issue #3641, opened 2026-03-02
Add/update collection literal elements tests


sgrekhov:
Add/update co19 tests according to the https://github.com/dart-lang/language/pull/4617

PR #3647, opened 2026-03-06
#3641. Add more tests for permitted use of type `void`

sgrekhov:
This is the first part of the changes.
- Updated assertions
- Add tests for `[]` and `[]=` operators
- Add tests for elements on top-level in collection literals

### DIFF
--- a/Language/Types/Type_Void/permitted_use_A14_t03.dart
+++ b/Language/Types/Type_Void/permitted_use_A14_t03.dart
@@ -16,7 +16,7 @@
 ///   may have value type `void`.
 ///
 /// @description Checks that in a map literal whose first actual type argument
-/// is `void` an element at the top level may have element type `void`.
+/// is `void`, an element at the top level may have key type `void`.
 /// @author sgrekhov22@gmail.com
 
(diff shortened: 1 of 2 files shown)


Language/Types/Type_Void/permitted_use_A14_t03.dart
```diff
@@ -0,0 +1,28 @@
...
+/// ...
+/// - In a list or set literal `C` whose actual type argument is `void`, an
+///   element at the top level in `C` may have element type `void`. In a map
+///   literal `M` whose first actual type argument is `void`, an element at the
+///   top level in `M` may have key type `void`. In a map literal `M` whose
+///   second actual type argument is `void`, an element at the top level in `M`
+///   may have value type `void`.
+///
+/// @description Checks that in a map literal whose first actual type argument
+/// is `void` an element at the top level may have element type `void`.
```
eernstg:
```suggestion
/// is `void`, an element at the top level may have key type `void`.
```
(I added a comma, too, because the phrase that follows the comma is a main clause -- several other descriptions could do the same.)

PR #3671, opened 2026-03-18
#3315. Fix tests that use obsolete `var` syntax. Part 10.

sgrekhov:


LibTest/collection/LinkedList/every_A01_t02.dart
```diff
@@ -14,7 +14,7 @@ import "../../../Utils/expect.dart";
 import "LinkedList.lib.dart";
 
 //Checks that [every] calls [predicate] a [count] number of times
-check(LinkedList<MyLinkedListEntry> a, bool predicate(e), int count) {
+check(LinkedList<MyLinkedListEntry> a, bool predicate(dynamic e), int count) {
```
eernstg: I think `predicate(e)` was fine, I just noted that it's one of those cases that may benefit from double checking. But this is fine, too. ;-)

PR #3697, opened 2026-03-30
#3182. Add more tests for augmenting parameter with name `_`. Part 2.

sgrekhov:


LanguageFeatures/Augmentations/augmenting_functions_A04_t33.dart
```diff
@@ -0,0 +1,160 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=augmentations
+
+import '../../Utils/expect.dart';
+
+final x = 42;
+
+int topLevelFunction1(int x);
+augment int topLevelFunction1(int _) => x;
```
eernstg: I think the test would be slightly stronger if the `x` declarations had different (unrelated) types. For example, the parameter could have the type `String`. The point is that we will then know that the static analysis considers the `x` in this line to resolve to the top-level declaration. We already know that the dynamic semantics binds the `x` to the top-level variable because of its value, but we might as well make sure that we get a heads-up if the static analysis thinks it is the parameter.

Issue #3698, opened 2026-03-30
Update `Language/Statements/Return` tests according to the current specification


sgrekhov:
Updated and complete tests in `Language/Statements/Return`.

PR #3706, opened 2026-04-03
#3698. Update tests for `return` statement. Part 3.

sgrekhov:


Language/Statements/Return/async_non_generator_function_A02_t01.dart
```diff
@@ -0,0 +1,54 @@
...
+/// Case ⟨Asynchronous non-generator functions⟩. Consider the case where `f` is
+/// an asynchronous non-generator function.
+/// ...
+/// It is a compile-time error if `s` is `return e;`, `flatten(T)` is `void`,
+/// and `flatten(S)` is neither `void`, `dynamic`, `Null`.
+///
+/// @description Checks that it is a compile-time error if a statement of the
+/// form `return e;` occurs in an asynchronous function whose return type is
+/// `void` or `Future<void>` and static type of `e` is not `Future<void>`,
+/// `Future<dynamic>` or `Future<Null>`.
```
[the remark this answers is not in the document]
sgrekhov: Sorry, it's still unclear to me.

> So the point is that we need to combine "`flatten(S)` is neither `void`, `dynamic`, nor `Null`" with "`futureValueType(T)` is either `void`, `dynamic`, or `Null`" to get the error, and negate the part about flatten(S) for the non-error.

Why "`futureValueType(T)` is either `void`, `dynamic`, or `Null`"? We now talking about the statement 

> It is a compile-time error if `s` is `return e;`, `Tv` is `void`,  and `flatten(S)` is neither `void`, `dynamic`, `Null`.

And here `Tv` is a `futureValueType(T)`. If `Tv` is `void` then accordibg to the [definition](https://github.com/dart-lang/language/blob/main/accepted/2.12/nnbd/feature-specification.md#the-future-value-type-of-an-asynchronous-non-generator-function) `T` cannot be `dynamic` or `Null`. But only `void`, `Future<void>`, `FutureOr<void>`, + the same types with `?`. What am I missing?

And it's also not clear to me what is `futureValueType(Null)`? It is `Object?` (according to the "Otherwise, for all `S`, `futureValueType(S) = Object?`")?

Please clarify.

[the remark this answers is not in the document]
eernstg: > talking about ... `return e;`, `Tv` is `void` ...

I think I misread the description and thought that I saw "void, dynamic, or Null" twice. Sorry about that. ;-)

With that, `Tv` is `void`, which means that the possible values of `T` are `void`, `Future<void>`, `FutureOr<void>`, `Future<void>?`, or `FutureOr<void>?`, as you said.

For the error case we'd then need `flatten(S)` to be neither `void`, `dynamic`, nor `Null`. There's no way we can cover any relevant subset for "all types except this handful of types", so we'll  just have to test a couple of examples. We could choose some types such that `flatten(S) == String` or `flatten(S) == Object?`.

This means that `S` could be, e.g., `String`, `Future<String>`, `MySubtypeOfFutureOfString`, `FutureOr<String>`, `X extends U` where `flatten(U) == String`, and similarly for `Object?`, plus `Future<Object>?` and `FutureOr<Object>?`.

(Note that we just need a single function with many return statements in order to test all those variants of `S`, a la `if (2 > 1) return ...;`.)

For the non-error case we'd have similar variants such that `flatten(S)` is `void`, `dynamic`, or `Null`, as a bunch of return statements in a function for each of the chosen possible values of `T`.

eernstg (commented):
I'm afraid this topic calls for a lot of cases. I tried to clarify a few things that I previously messed up, and outlined some possible choices of cases.

sgrekhov:
Updated. More types `T` and `S` added. PTAL.

Language/Statements/Return/async_non_generator_function_A02_t01.dart
```diff
@@ -0,0 +1,428 @@
...
+/// an asynchronous non-generator function with future value type `Tv`.
+/// ...
+/// It is a compile-time error if `s` is `return e;`, `Tv` is `void`,
+/// and `flatten(S)` is neither `void`, `dynamic`, `Null`.
+///
+/// @description Checks that it is a compile-time error if a statement of the
+/// form `return e;` occurs in an asynchronous function whose return type is
+/// `void`, `Future<void>`, `FutureOr<void>`, `Future<void>?`, or
+/// `FutureOr<void>?` and `flatten(S)` where `S` is the static type of `e` is
+/// neither `void`, `dynamic`, `Null`.
```
eernstg: I think this would be more readable:
```suggestion
/// `FutureOr<void>?`, and `flatten(S)` is neither `void`, `dynamic`, 
/// nor`Null`, where `S` is the static type of `e`.
```

Language/Statements/Return/async_non_generator_function_A02_t03.dart
```diff
@@ -0,0 +1,151 @@
...
+/// an asynchronous non-generator function with future value type `Tv`.
+/// ...
+/// It is a compile-time error if `s` is `return e;`, `Tv` is `void`,
+/// and `flatten(S)` is neither `void`, `dynamic`, `Null`.
+///
+/// @description Checks that it is not an error if a statement of the form
+/// `return e;` occurs in an asynchronous function whose return type is `void`,
+/// `Future<void>`, `FutureOr<void>`, `Future<void>?`, or `FutureOr<void>?` and
+/// the static type of `e` is `dynamic`, `Future<dynamic>`, `FutureOr<dynamic>`,
+/// `Future<dynamic>?`, or `FutureOr<dynamic>?`.
```
eernstg: I'm tempted to be lazy here. ;-D

The cases for the return type are fine. However, the treatment of `?` in the static type of the returned expression is more tricky.

The spec language doesn't mention `dynamic?`, but `flatten(Future<dynamic>?)` is actually `dynamic?` (so it's not "`void`, `dynamic`, or `Null`").

The implementations probably normalize the type such that we can't see the difference between `dynamic` and `dynamic?`, but I just don't think it's worth the trouble to insist that we should be allowed to return an expression of type `S` such that `flatten(S)` is `dynamic?`, and also not to insist that it is an error.

The underlying problem is that the normalization of `dynamic?` to `dynamic` is specified rather vaguely. A similar treatment is used for `X?` in a context where `X` is a type variable whose value is known to be `int?`; that's considered to be `int?` even though it is more strictly spoken the type `int??` which is then (vaguely ;-) normalized to `int?`. If the tools report an error then we'll get a slightly stronger type checking on some things involving `dynamic`, which is not bad, and if the don't report an error then it's just behaving as (vaguely) specified.

So I'd recommend that we simply don't include those `dynamic?` cases, namely `Future<dynamic>` and `FutureOr<dynamic>?`.

eernstg: Perhaps add a comment to say that we omitted those cases intentionally? ;-)

Language/Statements/Return/async_non_generator_function_A02_t04.dart
```diff
@@ -0,0 +1,151 @@
...
+/// an asynchronous non-generator function with future value type `Tv`.
+/// ...
+/// It is a compile-time error if `s` is `return e;`, `Tv` is `void`,
+/// and `flatten(S)` is neither `void`, `dynamic`, `Null`.
+///
+/// @description Checks that it is not an error if a statement of the form
+/// `return e;` occurs in an asynchronous function whose return type is `void`,
+/// `Future<void>`, `FutureOr<void>`, `Future<void>?`, or `FutureOr<void>?` and
+/// the static type of `e` is `Null`, `Future<Null>`, `FutureOr<Null>`,
+/// `Future<Null>?`, or `FutureOr<Null>?`.
```
eernstg: Same issue here: We don't want to spend the time resolving exactly how to treat `Null?`, so `Future<Null>?` and `FutureOr<Null>?` can be omitted.

Language/Statements/Return/async_non_generator_function_A03_t01.dart
```diff
@@ -0,0 +1,202 @@
...
+/// an asynchronous non-generator function with future value type `Tv`.
+/// ...
+/// It is a compile-time error if `s` is `return e;`, `Tv` is neither `void` nor
+/// `dynamic`, and `flatten(S)` is `void`.
+///
+/// @description Checks that it is a compile-time error if a statement of the
+/// form `return e;` occurs in an asynchronous function whose return type is
+/// `Object?`, `Future<Object?>`, `FutureOr<Object?>`, `Future<Object?>?`, or
+/// `FutureOr<Object?>?` and the static type of `e` is `void`, `Future<void>`,
+/// `FutureOr<void>`, `Future<void>?`, or `FutureOr<void>?`.
```
eernstg, line 19: (In this case I think it makes sense to keep the `void?` related cases: We do want to ensure that the heads-up about the value of a `void`-ish expression being used does not get lost. Hence, no changes here.)

Language/Statements/Return/async_non_generator_function_A02_t02.dart
```diff
@@ -0,0 +1,154 @@
...
+/// an asynchronous non-generator function with future value type `Tv`.
+/// ...
+/// It is a compile-time error if `s` is `return e;`, `Tv` is `void`,
+/// and `flatten(S)` is neither `void`, `dynamic`, `Null`.
+///
+/// @description Checks that it is not an error if a statement of the form
+/// `return e;` occurs in an asynchronous function whose return type is `void`,
+/// `Future<void>`, `FutureOr<void>`, `Future<void>?`, or `FutureOr<void>?` and
+/// the static type of `e` is `void`, `Future<void>`, `FutureOr<void>`,
+/// `Future<void>?`, or `FutureOr<void>?`.
```
eernstg: The comment below about `dynamic?` applies to `void?` as well.

Language/Statements/Return/async_non_generator_function_A03_t02.dart
```diff
@@ -0,0 +1,167 @@
...
+/// an asynchronous non-generator function with future value type `Tv`.
+/// ...
+/// It is a compile-time error if `s` is `return e;`, `Tv` is neither `void` nor
+/// `dynamic`, and `flatten(S)` is `void`.
+///
+/// @description Checks that it is a compile-time error if a statement of the
+/// form `return e;` occurs in an asynchronous function whose return type is
+/// `Future<Null>`, `FutureOr<Null>`, `Future<Null>?`, or `FutureOr<Null>?` and
+/// the static type of `e` is `void`, `Future<void>`, `FutureOr<void>`,
+/// `Future<void>?`, or `FutureOr<void>?`.
```
eernstg, line 19: (Same here, no changes).

Language/Statements/Return/async_non_generator_function_A03_t03.dart
```diff
@@ -0,0 +1,156 @@
...
+/// an asynchronous non-generator function with future value type `Tv`.
+/// ...
+/// It is a compile-time error if `s` is `return e;`, `Tv` is neither `void` nor
+/// `dynamic`, and `flatten(S)` is `void`.
+///
+/// @description Checks that it is not an error if a statement of the form
+/// `return e;` occurs in an asynchronous function whose return type is `void`,
+/// `Future<void>`, `FutureOr<void>`, `Future<void>?`, or `FutureOr<void>?` and
+/// the static type of `e` is `void`, `Future<void>`, `FutureOr<void>`,
+/// `Future<void>?`, or `FutureOr<void>?`.
```
eernstg: This is again one of those situations where the cases about `Future<void>?` and `FutureOr<void>?` for the type of `e` can be omitted, because we don't want to insist that "`flatten(S)` is `void?`" should be accepted. If the tools do it so be it, but if they report an error we shouldn't punish them.

Language/Statements/Return/async_non_generator_function_A03_t04.dart
```diff
@@ -0,0 +1,156 @@
...
+/// an asynchronous non-generator function with future value type `Tv`.
+/// ...
+/// It is a compile-time error if `s` is `return e;`, `Tv` is neither `void` nor
+/// `dynamic`, and `flatten(S)` is `void`.
+///
+/// @description Checks that it is not an error if a statement of the form
+/// `return e;` occurs in an asynchronous function whose return type is
+/// `dynamic`, `Future<dynamic>`, `FutureOr<dynamic>`, `Future<dynamic>?`, or
+/// `FutureOr<dynamic>?` and the static type of `e` is `void`, `Future<void>`,
+/// `FutureOr<void>`, `Future<void>?`, or `FutureOr<void>?`.
```
eernstg: Same situation, we don't want to punish tools if they do report an error for the `Future<void>?` or `FutureOr<void>?` cases for the type of `e`.

Language/Statements/Return/async_non_generator_function_A03_t05.dart
```diff
@@ -0,0 +1,171 @@
...
+/// an asynchronous non-generator function with future value type `Tv`.
+/// ...
+/// It is a compile-time error if `s` is `return e;`, `Tv` is neither `void` nor
+/// `dynamic`, and `flatten(S)` is `void`.
+///
+/// @description Checks that it is a compile-time error if a statement of the
+/// form `return e;` occurs in an asynchronous function whose return type is
+/// `Future<Never>`, `FutureOr<Never>`, `Future<Never>?`, or `FutureOr<Never>?`
+/// and the static type of `e` is `void`, `Future<void>`, `FutureOr<void>`,
+/// `Future<void>?`, or `FutureOr<void>?`.
```
eernstg: No changes concerning `Future<void>?` and `FutureOr<void>?`.

Language/Statements/Return/async_non_generator_function_A03_t05.dart
```diff
@@ -0,0 +1,171 @@
...
+/// ...
+/// Case ⟨Asynchronous non-generator functions⟩. Consider the case where `f` is
+/// an asynchronous non-generator function with future value type `Tv`.
+/// ...
+/// It is a compile-time error if `s` is `return e;`, `Tv` is neither `void` nor
+/// `dynamic`, and `flatten(S)` is `void`.
+///
+/// @description Checks that it is a compile-time error if a statement of the
+/// form `return e;` occurs in an asynchronous function whose return type is
+/// `Future<Never>`, `FutureOr<Never>`, `Future<Never>?`, or `FutureOr<Never>?`
```
eernstg: That won't work because we will then have to return something whose `flatten` has type `Never`, and that's just a normal assignability error. We're looking for an error which is specifically emitted for improper usage of a value whose static type is `void`.

The future value type _must_ then be a top type, because otherwise we just get the "not assignable" situation when we're "returning `void`".

Language/Statements/Return/async_non_generator_function_A04_t01.dart
```diff
@@ -0,0 +1,135 @@
...
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+Future<num> f14(Future<num?> v) async => v;
+//                                       ^
+// [analyzer] unspecified
+// [cfe] unspecified
+
+Future<num> f15(Future<dynamic> v) async {
```
eernstg, line 106: Yes! This is an interesting case.

Language/Statements/Return/async_non_generator_function_A02_t01.dart
sgrekhov: Updated, thanks.

Language/Statements/Return/async_non_generator_function_A02_t03.dart
sgrekhov: Updated.

Language/Statements/Return/async_non_generator_function_A02_t02.dart
sgrekhov: Updated below and here.

Language/Statements/Return/async_non_generator_function_A02_t04.dart
sgrekhov: Updated, the same comment added.

Language/Statements/Return/async_non_generator_function_A03_t01.dart
sgrekhov: Ok, no changes here :)

Language/Statements/Return/async_non_generator_function_A03_t03.dart
sgrekhov: Updated.

Language/Statements/Return/async_non_generator_function_A03_t04.dart
sgrekhov: Updated.

Language/Statements/Return/async_non_generator_function_A03_t05.dart
sgrekhov: And this is what's happen in the analyzer and CFE now. Thank you. Removed this test.

Language/Statements/Return/async_non_generator_function_A02_t02.dart
```diff
@@ -15,8 +15,10 @@
 /// @description Checks that it is not an error if a statement of the form
 /// `return e;` occurs in an asynchronous function whose return type is `void`,
 /// `Future<void>`, `FutureOr<void>`, `Future<void>?`, or `FutureOr<void>?` and
-/// the static type of `e` is `void`, `Future<void>`, `FutureOr<void>`,
-/// `Future<void>?`, or `FutureOr<void>?`.
+/// the static type of `e` is `void`, `Future<void>`, `FutureOr<void>` (cases
+/// `Future<void>?` and `FutureOr<void>?` are omitted intentionally because
+/// normalization of `void?` to `void` is specified vaguely and it is Ok if
+/// tools don't report an errors in these cases).
```
eernstg, line 21: Thanks!

Language/Statements/Return/async_non_generator_function_A02_t01.dart
eernstg: Thanks!

PR #3724, opened 2026-04-10
Fixes #3723. Add more tests for constants and ternary operator

### LINKED ISSUE #3723: Add more tests for constants
Add `Language/Expressions/Constants/` tests according to the https://github.com/dart-lang/sdk/issues/63144#issuecomment-4214774978

sgrekhov:


Language/Expressions/Constants/ternary_operator_t04.dart
```diff
@@ -16,9 +16,9 @@
 ///   1. `e1` evaluates to true and `e2` is a constant expression, or
 ///   2. `e1` evaluates to false and `e3` is a constant expression.
 ///
-/// @description Checks that an expression of the form `e1?e2:e3` is a constant
-/// if `e1` and `e2` are constants, `e1` evaluates to `true` and `e3` is a
-/// potentially constant.
+/// @description Checks that an expression of the form `e1 ? e2 : e3` is a
+/// constant if `e1` and `e2` are constants, `e1` evaluates to `true` and `e3`
+/// is a potentially constant.
```
eernstg:
```suggestion
/// is potentially constant.
```
We could introduce the phrase "a potential constant" meaning an expression which is potentially constant, but we don't have that and I don't think it's going to be worth the trouble to define it.

However, we can't really make it "a potentially constant" because that's an incomplete phrase: "potentially" is an adverb and it needs an adjective to qualify, so "constant" in this context is an adjective, and then there's no noun for the article 'a' to belong to.

We could do the standard thing and make it "`e3` is a potentially constant expression", but we could also use an abbreviated form where we rely on an implicit backreference to associate the adjective phrase with the noun (that is "`e3` is potentially constant", using the same structure as "the car is extremely red", not "the car is an extremely red").

There are some additional locations with a similar phrase below.

Language/Expressions/Constants/ternary_operator_t08.dart
```diff
@@ -0,0 +1,102 @@
...
+/// following:
+/// ...
+/// - An expression of the form `e1?e2:e3` is potentially constant if `e1`,
+///   `e2`, and `e3` are all potentially constant expressions. It is further
+///   constant if `e1` is a constant expression and either
+///   1. `e1` evaluates to true and `e2` is a constant expression, or
+///   2. `e1` evaluates to false and `e3` is a constant expression.
+///
+/// @description Checks that an expression of the form `e1 ? e2 : e3` is not a
+/// constant if `e1` and `e3` are constants but `e1` is not potentially constant
```
eernstg: Perhaps?:

```suggestion
/// constant if `e1` and `e3` are constants but `e2` is not potentially constant
```

eernstg (commented):
Looks good, just a couple of typos.

Language/Expressions/Constants/ternary_operator_t04.dart
sgrekhov: Thank you. Updated here and below.

Language/Expressions/Constants/ternary_operator_t08.dart
sgrekhov: Yes, of course. Thank you! fixed.

Language/Expressions/Constants/ternary_operator_t04.dart
eernstg: Thanks!

eernstg (approved):
LGTM

PR #3727, opened 2026-04-10
#3315. Add tests for `var` and `final` parameters of operators

sgrekhov:


eernstg (approved):
LGTM

This is one of those situations where we have _many_ test cases because there is a test case for each of the operators and each of the type introducing declarations (class, mixin, etc). It's tempting to say "if it's working with `>` then it will also work with `<`, so we just test one of them". However, that's based on a graybox assumption that it will be exactly the same code in the analyzer/CFE which is being tested in both cases, and we really can't be sure about that. So it's OK to have many test cases here.

Issue #3709, opened 2026-04-06
Check failing co19 tests


sgrekhov:
Check co19 tests that are failing now,  fix if needed.
https://dart-current-results.web.app/?filter=co19/Language

PR #3735, opened 2026-04-14
#3709. Add more tests for named optional parameters

sgrekhov:


Language/Functions/Formal_Parameters/Optional_Formals/name_t01.dart
```diff
@@ -23,7 +23,6 @@ main() {
 
   f1(_: 1);
 //   ^
-// [analyzer] unspecified
 // [cfe] unspecified
```
eernstg: This one is tricky. It is not an error for the named actual argument to have a name that starts with `_`, but it _is_ an error for a named actual argument of a statically checked function invocation to have a name that isn't declared by the statically known type of the callee. But this amounts to "no error" at the `f1` call site.

@johnniwinther, do you have a motivation for reporting this error? Perhaps it could be argued that it's an error to refer to a formal parameter declaration which is itself an error, but we don't usually report this kind of situation. For example:

```dart
class A extends dynamic {} // Error at `dynamic`.
class B extends A {} // No errors reported, even though `A` has an error.
```

eernstg (commented):
Looks good, just one thing needs to be clarified. I'm asking for a comment from @johnniwinther about a corner case for error reporting.

Language/Functions/Formal_Parameters/Optional_Formals/name_t01.dart
johnniwinther: The parameter is wrongfully considered a wildcard, which means we mangle the name and therefore cannot find it upon lookup.

It seems fixable, so let's not expect a CFE error here and file an issue with the CFE instead.

eernstg: Sounds good, thanks!

sgrekhov:
Updated. PTAL.

eernstg (approved):
LGTM

PR #3739, opened 2026-04-17
#3698. Update test for async non-generator functions

sgrekhov:
These tests will be renamed in the next PR to follow the naming of other tests.


Language/Statements/Return/no_expression_function_t07.dart
```diff
@@ -2,27 +2,36 @@
...
-/// @assertion Consider a return statement s of the form return e?;
+/// @assertion Consider a return statement `s` of the form `return e?;`. Let `S`
+/// be the static type of `e`, if `e` is present, let `f` be the immediately
+/// enclosing function, and let `T` be the declared return type of `f`.
 /// ...
-/// It is a compile-time error if s is
-///  return;, unless T is void, dynamic, or Null
+/// Case ⟨Asynchronous non-generator functions⟩. Consider the case where `f` is
+/// an asynchronous non-generator function. It is a compile-time error if `s` is
+/// `return;`, unless `flatten(T)` is `void`, `dynamic`, or `Null`.
```
eernstg: This looks like the old rules (the new ones should talk about the future value type, never about `flatten(T)` where `T` is the return type).

eernstg (commented):
I'm afraid the assertion shows obsolete text (the NNBD feature spec has a newer version of the given rule). Let's adjust that first, and then I'll do a full review.

sgrekhov:
Updated. PTAL.

Language/Statements/Return/no_expression_function_t07.dart
```diff
@@ -2,27 +2,37 @@
...
 
-Future<int> bar() {
+Future<Object?> foo() async {
+  return;
+//^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+Future<Never> bar() async {
```
eernstg: This is tricky. In this case we'd get the error because null isn't assignable to `Never`. I think the test would be stronger if we don't have an assignability based reason to reject the `return;`, we only have the reason which is the negation of the description in line 9-12.

So perhaps we should use something like `Future<int?>` as he declared return type, such that the future value type admits null. Then we will know that the error is based on `return;` having that shape (and, for example `return null;` would be fine).

This is an example of the general approach that we've used in various other locations: In order to test a specific error E1 we should have a program which is wrong according to the definition of E1, but it shouldn't be wrong for any other reason.

Language/Statements/Return/no_expression_function_t08.dart
```diff
@@ -2,23 +2,31 @@
...
-  C() { }
-  Future<bool> get foo async {
+  Future<Object?> get foo async {
+    return;
+//  ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  }
+
+  Future<Never> get bar async {
```
eernstg:
```suggestion
  Future<int?> get bar async {
```

Language/Statements/Return/no_expression_function_t07.dart
```diff
@@ -2,27 +2,37 @@
...
+/// `Null`.
 ///
-/// @description Checks that a compile error occurs if a statement of the form
-/// "return;" is used in a top-level asynchronous function whose declared return
-/// type is Future<int>.
-///
-/// @Issue 42459
+/// @description Checks that it is a compile-time error if a statement of the
+/// form `return;` is used in a top-level asynchronous function whose declared
+/// return type is `Future<Object?>` or `Future<Never>`.
```
eernstg: This description would need to be adjusted to take the issue below into account, as well as the description in a few other tests below.

Language/Statements/Return/no_expression_function_t09.dart
```diff
@@ -2,23 +2,31 @@
...
-  C() { }
-  static Future<C> foo() async {
+  static Future<Object?> foo() async {
+    return;
+//  ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  }
+
+  static Future<Never> bar() async {
```
eernstg: Ditto.

Language/Statements/Return/no_expression_function_t10.dart
```diff
@@ -2,23 +2,31 @@
...
-  C() { }
-  Future<int> foo() async {
+  Future<Object?> foo() async {
+    return;
+//  ^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+  }
+
+  Future<Never> bar() async {
```
eernstg: Ditto.

Language/Statements/Return/no_expression_function_t08.dart
```diff
@@ -27,5 +35,5 @@ class C {
 }
```
eernstg, line 35: Looks like we could have a static getter case here, or an additional test library for that.

Language/Statements/Return/no_expression_function_t11.dart
```diff
@@ -2,34 +2,47 @@
...
       return 1;
     } else {
       return;
 //    ^^^^^^
 // [analyzer] unspecified
+// [cfe] unspecified
+    }
+  }
+
+  Future<Never> bar() async {
```
eernstg: Something like `Future<int?>` again.

eernstg (commented):
Looks good, with a few comments about a possible spurious success (where we get a compile-time error as expected, but about a different issue).

Also, I think it would be good to have a getter version of each test. They do exist in one or two cases, but the handling of return statements in a getter and in a method/function with parameters may go to different parts of the implementation (I don't know, but it's not unthinkable).

Language/Statements/Return/no_expression_function_t07.dart
sgrekhov: Updated.

Language/Statements/Return/no_expression_function_t07.dart
sgrekhov: Thanks. My approach here is to check all top and bottom types and use Object? to cover all other possible types. If we already have tests for Object?, then int? is unlikely to add much. Let’s also check Future<ET?>, where ET is an extension type that doesn’t implicitly implement any other type.

Language/Statements/Return/no_expression_function_t08.dart
sgrekhov: Changed to  `ET?`

Language/Statements/Return/no_expression_function_t08.dart
sgrekhov: Thanks. One of the getters made static.

Language/Statements/Return/no_expression_function_t09.dart
sgrekhov: Fixed as the above.

Language/Statements/Return/no_expression_function_t10.dart
sgrekhov: Fixed.

Language/Statements/Return/no_expression_function_t11.dart
sgrekhov: Fixed.

sgrekhov (commented):
Thanks. Updated. Let's test an extension type instead of  type `Never`. PTAL.

Language/Statements/Return/no_expression_function_t08.dart
eernstg: Thanks!

Language/Statements/Return/no_expression_function_t07.dart
eernstg: Makes sense, thanks!

eernstg (approved):
LGTM

PR #3744, opened 2026-04-20
#3698. Update `no_expression_*` tests

sgrekhov:


Language/Statements/Return/no_expression_t02.dart
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2011, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Executing a return statement with no expression,` return;`
+/// returns without an object.
+///
+/// @description Checks that a return statement of the form `return;` in a
+/// generative constructor returns without an object.
```
eernstg: We probably can't observe that. In this case we're executing a generative constructor, in which case we can't even observe what the execution evaluates to because the instance creation expression will evaluate to the newly created object no matter what the constructor returns (it will return null, but that's discarded).

So what are we actually testing here? I think we're testing that the execution of `return;` will complete the execution of the nearest enclosing function body, and we can see this by  confirming that `x = 1` wasn't executed.

I think it would be helpful to put a clearer emphasis on this perspective, perhaps something like this could work:

  Checks that a return statement of the form `return;` in a generative constructor
  will complete the execution of the body such that subsequent code (like `x = 1`)
  will not be executed.

If we put the emphasis on "without an object" then we're just promising that we are testing a thing that we actually can't know.

eernstg (commented):
Looks good! I think it would be helpful to adjust one description, though.

eernstg (approved):
Thanks, LGTM.

sgrekhov:
Thank you. Updated. Also one additional test case added for the factory constructors. PTAL.

PR #3747, opened 2026-04-21
#3709. Delete two tests that check unspecified behavior

sgrekhov:
Accodding to https://github.com/dart-lang/sdk/issues/29353#issuecomment-4283739790, these tests check unspecified behaviour. Any outcome (pass or fail) is acceptable. Therefore, these tests have been removed.


eernstg (approved):
LGTM

PR #3750, opened 2026-04-23
#2979. Update static errors tests according to the spec

sgrekhov:


LanguageFeatures/nnbd/static_errors_A01_t02.dart
```diff
@@ -2,24 +2,51 @@
...
+  A? a = 2 > 1 ? A() : null;
+  M? m = 2 > 1 ? M() : null;
+  E? e = 2 > 1 ? E.e0 : null;
+  ET? et = 2 > 1 ? ET(0) : null;
+
   Expect.isNotNull(a.hashCode);
   Expect.isNotNull(a.toString());
   Expect.isNotNull(a.runtimeType);
-  Expect.isFalse(a == new A());
+  Expect.isFalse(a == A());
```
eernstg, line 37: Let's just include something like `Expect.throws(() => a.noSuchMethod(i));` where `i` is some suitable invocation.

LanguageFeatures/nnbd/static_errors_A01_t03.dart
```diff
@@ -2,27 +2,75 @@
...
+  m.baz = 2;
+  m + m;
+
+  e.m;
+  e.foo();
+  e.bar;
+  e.baz = 2;
+  e + e;
+
+  if (1 > 2) { // The runtime type of et is `int`, so it doesn't have members of ET.
```
eernstg, line 70: OK!

LanguageFeatures/nnbd/static_errors_A01_t05.dart
```diff
@@ -2,27 +2,56 @@
...
-  AAlias x = A();
-  Expect.isNotNull(x.hashCode);
-  Expect.isNotNull(x.toString());
-  Expect.isNotNull(x.runtimeType);
-  Expect.isFalse(x == A());
+  AAlias a = A();
+  Expect.isNotNull(a.hashCode);
+  Expect.isNotNull(a.toString());
+  Expect.isNotNull(a.runtimeType);
+  Expect.isFalse(a == A());
```
eernstg, line 39: Again, we can have a case for `noSuchMethod`.

LanguageFeatures/nnbd/static_errors_A01_t06.dart
eernstg, line 23: `noSuchMethod`.

LanguageFeatures/nnbd/static_errors_A01_t07.dart
```diff
@@ -2,14 +2,15 @@
...
+/// expression whose type is potentially nullable and not `dynamic`, except for
+/// the methods, setters, getters, and operators on `Object`, and except when
+/// said member is an extension member or the receiver type is an extension type
 ///
 /// @description Check that it is no compile-time error to call a method, setter,
 /// getter or operator on an expression whose type is potentially nullable if
-/// they are methods, setters, getters, and operators on Object. Test
-/// that methods of Object are allowed for the class FutureOr
+/// they are methods, setters, getters, and operators on `Object`. Test that
+/// methods of `Object` are allowed for the class `FutureOr`.
```
eernstg, line 13: Good catch! I hadn't thought about `FutureOr`.

LanguageFeatures/nnbd/static_errors_A01_t07.dart
eernstg: Hmm, 2019, that would be in pre-nnbd times. So we may need to double check that it is still expressing the situation that we want to test. I think it would be helpful to readers to have a type annotation which isn't a raw type (because they tend to be tricky to read precisely).

So it could be `FutureOr<dynamic>` if we wish to preserve the meaning of the raw type, or it could be `FutureOr<Object?>` if the point is that we're testing top types, or it could be `FutureOr<Object>` based on the initializing expression, or it could be each of them.

LanguageFeatures/nnbd/static_errors_A01_t13.dart
eernstg, line 26: `noSuchMethod`.

LanguageFeatures/nnbd/static_errors_A01_t17.dart
```diff
@@ -0,0 +1,47 @@
...
+import "../../Utils/expect.dart";
+
+extension type ET(int v) {}
+
+extension on ET? {
+  int foo() => 1;
+  int get bar => 2;
+  void set baz(int i) {}
+  ET? operator +(ET? other) => other;
+}
```
eernstg, line 25: It would actually be consistent with the other tests if the members were declared on the extension type itself rather than on the extension. The point is that an extension type that does not declare any superinterfaces will implicitly have `Object?` as their only superinterface, which makes the extension type itself potentially nullable.

So we should have a test like that.

In addition to this, it does make sense to test the case where the topic is `extension` members, but in that case it's actually already enough to have `on ET` rather than `on ET?`, because `ET` is already a potentially nullable type.

I guess this could be done by duplicating this test such that t18 is this "extension method called on extension type" test and t17 is adjusted to have an `extension type` (and no `extension`) and test that we don't get an error just because `ET` itself is potentially nullable. Of course that's a thing which is tested in _every_ test that uses an extension type for essentially anything, but it still makes sense to have this test which is checking that very basic property of extension types.

LanguageFeatures/nnbd/static_errors_A01_t18.dart
```diff
@@ -0,0 +1,47 @@
...
+/// setter, getter or operator on an expression whose type is potentially
+/// nullable if they are methods, setters, getters, and operators on an
+/// extension. Test an extension type which implements its representation type.
+/// @author sgrekhov22@gmial.com
+
+import "../../Utils/expect.dart";
+
+extension type ET(int v) implements int {}
+
+extension on ET? {
```
eernstg, line 20: (Ah, we have a t18 already. OK, some other number should work. ;-)

Yes! This makes sense because `ET` is non-nullable.

LanguageFeatures/nnbd/static_errors_A01_t19.dart
```diff
@@ -0,0 +1,42 @@
...
+  void set baz(int i) {}
+  ET1 operator +(ET1 other) => other;
+}
+
+extension type ET2(Object v) implements Object {
+  int foo() => 1;
+  int get bar => 2;
+  void set baz(int i) {}
+  ET2 operator +(ET2 other) => other;
+}
```
eernstg, line 21: This one doesn't really match the description because `ET2` is a non-nullable type.

I think the clean approach would be to omit ET2 and the associated test cases.

eernstg (commented):
Looks good, with a few comments, especially about extension types.

LanguageFeatures/nnbd/static_errors_A01_t02.dart
sgrekhov: Forgot about it. Thank you. Added.

LanguageFeatures/nnbd/static_errors_A01_t05.dart
sgrekhov: Added.

LanguageFeatures/nnbd/static_errors_A01_t06.dart
sgrekhov: Added.

LanguageFeatures/nnbd/static_errors_A01_t07.dart
sgrekhov: Let's choose "each of them" option :) Added type annotation and `FutureOr<Object?>` and `FutureOr<Object>`.

LanguageFeatures/nnbd/static_errors_A01_t13.dart
sgrekhov: Added.

LanguageFeatures/nnbd/static_errors_A01_t19.dart
sgrekhov: Thank you. Removed.

LanguageFeatures/nnbd/static_errors_A01_t17.dart
sgrekhov: > So we should have a test like that.

`static_errors_A01_t19.dart` does that.  In this test `on E?` replaced by `on E`.

sgrekhov (commented):
Updated. `noSuchMethod` cases added. PTAL.

LanguageFeatures/nnbd/static_errors_A01_t02.dart
eernstg: Thanks!

LanguageFeatures/nnbd/static_errors_A01_t17.dart
eernstg: Ah, of course - thanks!

eernstg (commented):
LGTM

eernstg (approved):

LanguageFeatures/nnbd/static_errors_A01_t07.dart
eernstg: SG!

PR #3753, opened 2026-04-27
Update Language/Functions/Formal_Parameters/Optional_Formals/name_t04

biggs0125:
Skip symbol-based expectations for minified builds.

Similar to other tests that use symbols, we need to skip these expectations on minified builds where symbol names are mangled.


sgrekhov:
Use constant `Symbol` names should fix this issue. Please see https://github.com/dart-lang/co19/pull/3755

biggs0125:
Ah that works, thanks @sgrekhov!

Issue #870, opened 2020-07-28
Migrate LibTest\io\HttpClient to null safety


sgrekhov:
Migrate LibTest\io\HttpClient to null safety

PR #3765, opened 2026-04-28
#3698. Add/update tests for execution

sgrekhov:


sgrekhov:
@eernstg please review

Language/Statements/Return/execution_A01_t04.dart
```diff
@@ -0,0 +1,54 @@
...
+/// @assertion The execution of a return statement proceeds as follows.
+///
+/// Case ⟨Synchronous non-generator functions and factory constructors⟩. Let `s`
+/// be a statement of the form `return e;`. Let `f` be the immediately enclosing
+/// function, and consider the case where `f` is a synchronous non-generator or
+/// a factory constructor. Execution of `s` proceeds as follows.
+///
+/// The expression `e` is evaluated to an object `o`. A dynamic error occurs
+/// unless the dynamic type of `o` is a subtype of the actual return type of `f`.
+/// Then the return statement s completes returning `o`.
```
eernstg: Might as well backquote "s" here, too:
```suggestion
/// Then the return statement `s` completes returning `o`.
```
There are a few other occurrences in tests below.

Language/Statements/Return/execution_A01_t04.dart
```diff
@@ -0,0 +1,54 @@
...
+num? f3() {
+  return null as dynamic;
+}
+
+FutureOr<num> f4() {
+  return 4;
+}
+
+FutureOr<num> f5() {
+  return Future.value(5);
```
eernstg: We're not testing inference of the omitted type arguments here, so the test that most precisely fails if the tested behavior is wrong (and for the smallest practically achievable set of other reasons) would specify the type argument explicitly:
```suggestion
  return Future<num>.value(5);
```
It could also be `Future<int>.value(5)`, but we're not testing the subtype relationship as such, so we'd just choose one of them.

Language/Statements/Return/execution_A01_t05.dart
```diff
@@ -0,0 +1,34 @@
...
+import 'dart:async';
+import '../../../Utils/expect.dart';
+
+int foo(v) {
+  return v;
+}
+
+main() async {
+  Expect.throws(() {
+    foo(3.14 as num);
```
eernstg: I'm trying to understand why there's a cast here. It does ensure that the actual argument has a static type which is a supertype of the type required in the return statement in line 24. However, passing a double as an actual argument to a parameter of type `dynamic` is fine, and shouldn't incur any compile-time diagnostics. So it's not needed in order to avoid (correct) compile-time errors or warnings.

If there's a special consideration here then it would be great to have a brief explanation in a comment.

Language/Statements/Return/execution_A02_t01.dart
```diff
@@ -0,0 +1,54 @@
...
+  return r<int>(2);
+}
+
+Future<num> baz() async {
+  return 3;
+}
+
+FutureOr<num> qux() async {
+  return 4;
+}
```
eernstg, line 43: We definitely have to include the case where the run-time type of `o` is a subtype of `Future<Tv>` _and also_ a subtype of `Tv` itself. For example, `Tv` could be `Object` and `o` could have type `Future<Object>` (which is a subtype of both `Tv` and `Future<Tv>`), and the correct behavior would be to await this future, in spite of the fact that `o` itself could also be used to complete the already-returned future.

Language/Statements/Return/execution_A02_t01.dart
```diff
@@ -0,0 +1,54 @@
...
+
+main() async {
+  Expect.notEquals(await r(1), foo());
```
eernstg: It does make sense! ;-)

But it would probably be helpful for future readers if there's a comment saying something like "The expected values are directly based on the specified behavior".

Language/Statements/Return/execution_A02_t01.dart
```diff
@@ -0,0 +1,54 @@
...
+  Expect.equals(await r(1), await foo());
+  Expect.notEquals(await r(2), bar());
+  Expect.equals(await r(2), await bar());
+  Expect.notEquals(await r(3), baz());
```
eernstg: But for this case and the 3 following cases, `3` and `4` will do just fine.

Language/Statements/Return/execution_A02_t02.dart
```diff
@@ -0,0 +1,49 @@
...
+/// Execution of `s` proceeds as follows.
+///
+/// The expression `e` is evaluated to an object `o`. If the run-time type of
+/// `o` is a subtype of `Future<Tv>`, let `v` be a fresh variable bound to `o`
+/// and evaluate `await v` to an object `r`; otherwise let `r` be `o`. A dynamic
+/// error occurs unless the dynamic type of `r` is a subtype of the actual value
+/// of `Tv`. Then the return statement `s` completes returning `r`.
+///
+/// @description Checks that if the run-time type of `o` is a not a subtype of
+/// `Future<Tv>` then dynamic error occurs.
```
eernstg:
```suggestion
/// `Future<Tv>` and not a subtype of `Tv` then dynamic error occurs.
```

Language/Statements/Return/execution_A02_t02.dart
```diff
@@ -0,0 +1,49 @@
...
+import '../../../Utils/expect.dart';
+
+Future<T> r<T>(T r) async => r;
+
+Future<int> foo() async {
+  return r<num>(1) as dynamic;
+}
+
+Future<int> bar() async {
+  return (2 as num) as dynamic;
```
eernstg: This one will execute without any run-time failures: `int` (type of `o`) is a subtype of `int` (`Tv`).

Language/Statements/Return/execution_A02_t04.dart
```diff
@@ -0,0 +1,49 @@
...
+/// `Future<Tv>` then dynamic error occurs.
+/// @author sgrekhov22@gmail.com
+
+import '../../../Utils/expect.dart';
+
+Future<T> r<T>(T r) async => r;
+
+Future<int> foo() async {
+  try {
+    return r<num>(1) as dynamic;
```
eernstg, line 31: This is testing a situation which is very similar to the bug which is reported in https://github.com/dart-lang/sdk/issues/44395.

I'm not sure how to handle this: I haven't tried this particular example, but I expect that it's going to fail. All attempts to fix #44395 have been dropped because of the very subtle (and bug-prone) breaking change.

The best approach would probably be to keep this test as written, and approve that it fails, and mention that this is testing (essentially) the situation of interest in #44395.

Strictly speaking, the situation in #44395 involves a future which has been completed with an error, which means that an exception should be thrown when it is (implicitly) awaited by the return statement inside the try block. In this example the future has been completed with a value, and it's even a value whose run-time type allows it to complete the already-returned future from the enclosing function. However, at the return statement we can see that the value of the returned expression does not have a type which is a subtype of `Future<Tv>`, nor a subtype of `Tv`, so the return statement execution itself incurs a run-time failure (it throws a `TypeError`), and this still means that execution should proceed to the catch block and 2 should be used to complete the already-returned future.

The actual behavior will probably be that execution leaves the try block (as in `break L` where `L` is a label on the outermost statement), and a `TypeError` is thrown because that `Future<num>` is tested against `Future<Tv>` and `Tv` and fails.

Language/Statements/Return/execution_A02_t04.dart
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+Future<int> bar() async {
+  try {
+    return r<num>(1) as dynamic;
+  } on Error {
+    return 2;
+  } finally {
+    return 3;
```
eernstg, line 43: Interesting! So the bug reported in #44395 can be eliminated by adding a `finally` block. Anyway, this will be sorted out when #44395 is clarified. The most likely outcome, I think, is that we adopt #870. If we do that then all of #44395 will go away.

eernstg (commented):
Couple of issues!

Language/Statements/Return/execution_A01_t04.dart
sgrekhov: Thank you. Fixed.

Language/Statements/Return/execution_A01_t04.dart
sgrekhov: Fixed.

Language/Statements/Return/execution_A01_t05.dart
sgrekhov: Yes, cast is excessive here. It doesn't make the test stronger. Removed.

Language/Statements/Return/execution_A02_t01.dart
sgrekhov: Added
```dart
FutureOr<Object> quux() async {
  return r<Object>(5);
}
```

Language/Statements/Return/execution_A02_t01.dart
sgrekhov: Added.

Language/Statements/Return/execution_A02_t01.dart
sgrekhov: Really. Thank you. Updated.

Language/Statements/Return/execution_A02_t02.dart
sgrekhov: Updated.

Language/Statements/Return/execution_A02_t02.dart
sgrekhov: Yes. But `Expect.fail('Run-time error expected');` on success forced `onError()` part and therefore the test successed. Thank you! Fixed.

Language/Statements/Return/execution_A02_t04.dart
sgrekhov: The test pass but I added link to 44395.

sgrekhov (commented):
Thank you. Updated. PTAL.

Language/Statements/Return/execution_A01_t04.dart
eernstg: Thanks!

Language/Statements/Return/execution_A01_t05.dart
eernstg: Thanks! (That is, thanks for being so careful about what it takes to write a strong and correct test, and also thanks for the fixes when something turns out to need an extra iteration.)

Language/Statements/Return/execution_A02_t01.dart
eernstg: Thx!

Language/Statements/Return/execution_A02_t04.dart
eernstg: That's a surprise! Anyway, we're not debugging the bug, and the behavior is actually correct in this case, so who are we to complain. ;-)

Language/Statements/Return/execution_A02_t02.dart
eernstg: Very good!

eernstg (approved):
LGTM

PR #3766, opened 2026-04-28
#2979. Update assertions in NNBD static errors tests

sgrekhov:
`static_errors_A08_*` tests will be renamed to `local_variable_read_A05_*` in a separate PR

### DIFF
--- a/LanguageFeatures/nnbd/static_errors_A08_t01.dart
+++ b/LanguageFeatures/nnbd/static_errors_A08_t01.dart
@@ -2,13 +2,13 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-/// @assertion It is an error if a potentially non-nullable local variable which
-/// has no initializer expression and is not marked late is used before it is
-/// definitely assigned
+/// @assertion It is a compile time error to read a local variable when the
+/// variable is potentially unassigned unless the variable is non-`final` and
+/// has nullable type, or is `late`.
 ///
 /// @description Check that it is an error if a potentially non-nullable local
-/// variable which has no initializer expression and is not marked late is used
-/// before it is definitely assigned. Test Never
+/// variable which has no initializer expression and is not marked `late` is
+/// read before it is definitely assigned. Test a variable of type `Never`.
 /// @author sgrekhov@unipro.ru
 /// @issue 40946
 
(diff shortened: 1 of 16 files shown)


LanguageFeatures/nnbd/static_errors_A08_t01.dart
```diff
@@ -2,13 +2,13 @@
...
-/// definitely assigned
+/// @assertion It is a compile time error to read a local variable when the
+/// variable is potentially unassigned unless the variable is non-`final` and
+/// has nullable type, or is `late`.
 ///
 /// @description Check that it is an error if a potentially non-nullable local
-/// variable which has no initializer expression and is not marked late is used
-/// before it is definitely assigned. Test Never
+/// variable which has no initializer expression and is not marked `late` is
+/// read before it is definitely assigned. Test a variable of type `Never`.
```
eernstg, line 11: This one is tricky, similarly to the situation in a different PR recently, because there will be a diagnostic message in response to the use of a variable of type `Never` (e.g., in line 18).

I mentioned a meta-rule that a test which is intended to demonstrate that a certain error message is emitted, should not have other faults. If it does have other faults then we may get a diagnostic message and conclude that it works just fine, even though the diagnostic message was about some other problem.

Perhaps that's not possible in this case because error messages and warnings will never be considered to be the same outcome? So we won't see a `dead_code` warning and fail to get a 'potentially not assigned' error, and thinking that it's all good?

In any case, I think it would be helpful to adjust the test such that it doesn't give rise to this kind of doubt and scrutiny. Couldn't it just as well be a variable of type `Null`? Or perhaps we could just omit this particular test, given that the remaining tests in this PR do similar things with a number of other types?

eernstg (commented):
Looks good! Just one test has a comment, because I'm worried about detecting a fault in the code which is not the purpose of the test.

LanguageFeatures/nnbd/static_errors_A08_t01.dart
sgrekhov: No, there are no any other diagnostic messages in this test now. There were ones before this change. But now code has been changed and there is no any code after use of type `Never`.

eernstg: OK, it's surprising that it doesn't consider `toString()` to be dead code (it _is_ code that has a run-time semantics, and it won't be reached). So we might get that diagnostic at some point if and when they fix this.

Anyway, it's unlikely that this test will create a "oh we thought it was working, but it was broken all the time" problem because we have all those other tests with types that aren't causing any diagnostics.

eernstg (approved):
LGTM

PR #3772, opened 2026-04-30
#3315. Add more tests for constants

sgrekhov:


LanguageFeatures/Primary-constructors/static_processing_A34_t01.dart
```diff
@@ -0,0 +1,75 @@
...
+}
+
+class B {
+  final A? a;
+  const B(this.a);
+}
+
+class const C1(final A a) {
+  final B b;
+  this : b = const B(a);
```
eernstg: We can't do that because `const B(...)` is only a constant expression (never an only-potentially-constant expression) when each argument is a constant expression. It would be nice to have this, but today we don't, so `const B(a)` is just an error.

eernstg: What we _can_ do is using `a` as a top-level expression (`b = a`) or as a component in a record literal (`b = (a, a)`). We could also have other forms (e.g., `a + a` or `a ? a : a`), but there's no need to test constant expressions in general here, we just need to demonstrate that the primary parameters are potentially constant, so, for example, a plain `b = a` in the initializer list will do just fine for that.

eernstg: See also https://github.com/dart-lang/language/issues/4571. ;-)

LanguageFeatures/Primary-constructors/static_processing_A34_t02.dart
```diff
@@ -0,0 +1,91 @@
...
+class B {
+  final A? a;
+  const B(this.a);
+}
+
+enum const E1(final A a) {
+  e0(A());
+
+  final B b;
+  this : b = const B(a);
```
eernstg: Ditto.

eernstg (commented):
Looks good, but some expressions of the form `const B(a)` would have to be modified before they can be expected to compile without errors.

LanguageFeatures/Primary-constructors/static_processing_A34_t01.dart
sgrekhov: > ...const B(a) is just an error.

I'll add a test for this (in a separate PR)! :) Uptadet here and below. PTAL.

eernstg (approved):
LGTM

Issue #3779, opened 2026-05-04
Fix failing `co19/LibTest/io` tests


sgrekhov:
See https://dart-current-results.web.app/?filter=co19/LibTest/io

PR #3780, opened 2026-05-04
#3779. Fix isDirectory and isFile tests. Add tests for `typeSync`

sgrekhov:
`isDirectory` and `isFile` tests were failing. But these properties behave exactly as specified. Therefore tests were updated and new tests for `typeSync()` added. `typeSync()` works as intended.

### DIFF
--- a/LibTest/io/FileSystemEntity/isDirectorySync_A01_t03.dart
+++ b/LibTest/io/FileSystemEntity/isDirectorySync_A01_t03.dart
@@ -20,6 +20,8 @@ main() async {
 }
 
 void _main(Directory sandbox) {
+  // dirLink will be a link to a temp directory in the sandbox because no target
+  // is passed
   final dirLink = getTempLinkSync(parent: sandbox);
   Expect.isTrue(FileSystemEntity.isDirectorySync(dirLink.path));
   Expect.equals(FileSystemEntityType.directory,
@@ -31,6 +33,7 @@ void _main(Directory sandbox) {
   Expect.equals(FileSystemEntityType.file,
       FileSystemEntity.typeSync(fileLink.path));
 
+  // notExisting will be a path to a non-existing file in the sandbox
   final notExisting = getTempFilePath(parent: sandbox);
   final brokenLink = getTempLinkSync(parent: sandbox, target: notExisting);
   Expect.isFalse(FileSystemEntity.isDirectorySync(brokenLink.path));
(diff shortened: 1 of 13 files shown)


LibTest/io/FileSystemEntity/isDirectory_A01_t03.dart
```diff
@@ -19,11 +19,20 @@ main() async {
...
 
 void _main(Directory sandbox) async {
-  Link link = getTempLinkSync(parent: sandbox);
-  asyncStart();
-  bool result = await FileSystemEntity.isDirectory(link.path);
-  Expect.isFalse(result);
-  FileSystemEntityType t = await FileSystemEntity.type(link.path);
-  Expect.equals(FileSystemEntityType.link, t);
-  asyncEnd();
+  final dirLink = getTempLinkSync(parent: sandbox);
```
eernstg, line 24: Hmm. The behavior of `getTempLinkSync` isn't particularly obvious.

Just a few words like '// `dirLink` will be a link to the sandbox directory because no `target` is passed.' would probably be helpful (especially if I got that wrong and the explanation needs to be different ;-).

LibTest/io/FileSystemEntity/isDirectorySync_A01_t03.dart
```diff
@@ -20,8 +20,20 @@ main() async {
 }
 
 void _main(Directory sandbox) {
-  Link link = getTempLinkSync(parent: sandbox);
-  Expect.isFalse(FileSystemEntity.isDirectorySync(link.path));
-  Expect.equals(FileSystemEntityType.link,
-      FileSystemEntity.typeSync(link.path));
+  final dirLink = getTempLinkSync(parent: sandbox);
```
eernstg, line 25: Ditto, as well as in several tests below. One comment like this in each library that uses this function would be fine, but we can't know which library somebody will need to understand at some point in the future because it starts to fail, so it should probably not be fewer than that.

eernstg (commented):
Looks good! I'm suggesting that a few comments are added because some of the helper functions are honestly not self-explanatory. ;-)

LibTest/io/FileSystemEntity/isDirectory_A01_t03.dart
eernstg: (I have no idea why Github thinks I'm suggesting that 3 lines of code should be deleted, it's only a suggestion that a comment could be helpful just before the declaration of `dirLink`.)

LibTest/io/FileSystemEntity/isDirectorySync_A01_t03.dart
eernstg: (Same github bug here.)

sgrekhov:
Updated. Comments added. PTAL.

eernstg (approved):
LGTM, thanks!

PR #3781, opened 2026-05-05
#3779. Fix `contentChanged_A01_t01.dart` if no event was fired

sgrekhov:
On Linux if file attribute was changed then `FileSystemModifyEvent` is fired with `contentChanged` set to `false`. On Windows and Mac no event is fired in this case. The test now pass on these platforms.


eernstg (approved):
LGTM

PR #3787, opened 2026-05-06
#3203. Test primary constructor parameter in initializing expression

sgrekhov:
Currently, this test passes. Please note that it expects `x == null` at `LINE_A` and `c1` after `stepInto`. The more correct behavior would probably be to expect `Global` instead of `null`. However, the test is written according to the current behavior of the debugger. Please review


eernstg:
@nshahan, WDYT? I'll just follow the discussion and learn. ;-)

VM/primary_constructors_t07.dart
```diff
@@ -0,0 +1,77 @@
...
+  stepInto,
+  stoppedAtLine(LINE_A),
+  (VmService srv, IsolateRef isolateRef) async {
+    final islId = isolateRef.id!;
+    final xRef1 = await srv.evaluateInFrame(islId, 0, 'x') as InstanceRef;
+    Expect.equals('null', xRef1.valueAsString);
+    final xRef2 = await srv.evaluateInFrame(islId, 0, 'y') as InstanceRef;
+    Expect.equals('null', xRef2.valueAsString);
+  },
+  stepInto,
```
nshahan, line 61: Where are we paused after this stepInto?

VM/primary_constructors_t07.dart
```diff
@@ -0,0 +1,77 @@
...
+  stoppedAtLine(LINE_C),
+  stepInto,
+  stoppedAtLine(LINE_A),
+  (VmService srv, IsolateRef isolateRef) async {
+    final islId = isolateRef.id!;
+    final xRef1 = await srv.evaluateInFrame(islId, 0, 'x') as InstanceRef;
+    Expect.equals('null', xRef1.valueAsString);
+    final xRef2 = await srv.evaluateInFrame(islId, 0, 'y') as InstanceRef;
+    Expect.equals('null', xRef2.valueAsString);
```
nshahan, line 59: I agree `null` here feels odd but I think it is fine to land to match the current implementation here until we figure out if the value passed as the argument can be "seen" during the expression evaluation in the debugger.

nshahan (approved):
Based on the ongoing discussion and work associated with https://github.com/dart-lang/sdk/issues/62645, my understanding is that the existing behavior is not what we want but we also don't have a full picture of what is feasible in a reasonable amount of work. I believe we will be making another pass over these expectations in the near future once some more implementation work is done.

VM/primary_constructors_t07.dart
sgrekhov: Good question! Added `stoppedAtLine(LINE_A),` after this `stepInto`. The test passes.

sgrekhov (commented):
Added additional `stoppedAtLine()` check.

eernstg (approved):
Very good, thanks!

PR #3789, opened 2026-05-07
Fixes #3788. App tests for `FileSystemEntity.parentOf()`

### LINKED ISSUE #3788: Add missing tests for `FileSystemEntity`
There are no tests for `FileSystemEntity.parentOf()`.

sgrekhov:


LibTest/io/FileSystemEntity/parentOf_A01_t01.dart
```diff
@@ -0,0 +1,54 @@
...
+
+main() {
+  test('a/b/c', 'a/b');
+  test('a/b/c/', 'a/b');
+  test('a', '.');
+  test('../a/', '..');
+  test('.', '.');
+  test('', '.');
+  test('..', '.');
+  test('...', '.');
```
eernstg, line 33: I think this illustrates that the function returns `.` for every input which can be a file name (with no path), and it makes no attempt whatsoever to verify that any part of the path exists. I'd think `..` should be mapped to `../..`, but apparently it's considered to be a file name in the current directory (which is actually true, but perhaps surprising to many people who are used to thinking about `..` as if it's the parent of the current working directory -- like me ;-). So we get the current working directory rather than the parent-of-parent directory.

I guess the documentation would be the text in the `@assertion`, so perhaps this documentation should be extended a bit to say what it does to `..` (and why)? ;-)

eernstg (approved):
I think we can land this test.

However, it might be useful to create an issue about the documentation of `parentOf`, to be expanded slightly such that it informs developers using this function that "`..` doesn't work".

LibTest/io/FileSystemEntity/parentOf_A01_t01.dart
sgrekhov: I filed https://github.com/dart-lang/sdk/issues/63342 when I was working on this.

PR #3790, opened 2026-05-08
Add tests for LUB calculation

sgrekhov:
This is an item from my TODO list. Added test according https://github.com/dart-lang/language/issues/1618

### DIFF
--- a/TypeSystem/upper-lower-bounds/up_A02_t01.dart
+++ b/TypeSystem/upper-lower-bounds/up_A02_t01.dart
@@ -1,0 +1,32 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types T1 and T2 to be UP(T1,T2)
+/// as follows.
+/// - UP(X1 & B1, T2) =
+///   - T2 if X1 <: T2
+///   - otherwise X1 if T2 <: X1
+///   - otherwise UP(B1a, T2) where B1a is the greatest closure of B1 with
+///     respect to X1, as defined in inference.md.
+/// - UP(T1, X2 & B2) =
+///   - X2 if T1 <: X2
+///   - otherwise T1 if X2 <: T1
+///   - otherwise UP(T1, B2a) where B2a is the greatest closure of B2 with
+///     respect to X2, as defined in inference.md.
+///
+/// @description Check that upper bound of intersection types is calculated as
+/// expected.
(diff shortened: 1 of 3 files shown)


TypeSystem/upper-lower-bounds/up_A02_t01.dart
```diff
@@ -0,0 +1,36 @@
...
+
+class C1 extends D implements B {}
+class C2 extends D implements B {}
+
+main() {
+  B b = (1 > 2) ? C1() : C2();
+  D d = (1 > 2) ? C1() : C2();
+  var x = (1 > 2) ? C1() : C2();
+  x.expectStaticType<Exactly<A>>();
+}
```
eernstg, line 33: This test actually doesn't have any expressions whose type is an intersection type. They only arise based on promotion of a variable whose type is a type variable. For example:

```dart
void f<X extends num>(X x1, X x2) {
  if (x1 is int && x2 is int) { // Promote variables to `X & int`.
    (x1 + x2).isEven; // OK, so `x1` and `x2` have a subtype of `int`.
    if (2 > 1) {
      x1 = x2; // OK, so `x2` has a subtype of `X`.
    } else {
      x2 = x1; // OK, same for `x1`.
    }
  }
}
```

So we need to get hold of an expression (that is, a promoted variable) whose type at the given location is `X1 & B1`, and another expression whose type is the bound of `X1` or a top type to match the first bullet of the first `UP` rule. Next, we need a case where the second expression has a subtype of `X1` (say, `X1` itself or `Never`) to match the second bullet, and another case (e.g., `B1?`) to match the third case.

eernstg (commented):
I think this one needs to be different in order to cover the two `UP` rules mentioned in the assertion (or perhaps there could be several tests).

TypeSystem/upper-lower-bounds/up_A02_t01.dart
sgrekhov: The asserion text is wrong. The idea of this PR is to add a test checking https://github.com/dart-lang/language/issues/1618. Where can I find the right assertion?

eernstg:
OK! I'll go now, but I'll take a look on Monday. Have a nice weekend!

eernstg:
Here's the best description of the change I could find: https://dart-review.googlesource.com/c/sdk/+/353440. https://github.com/dart-lang/language/issues/4689 keeps track of the work to write a markdown document for this.

sgrekhov:
Assertion text for conditional statement is taken from https://github.com/dart-lang/language/issues/1618#issuecomment-1507241494. PTAL.

eernstg (approved):
LGTM.

Btw, this approach to inference on branching constructs was a very, very nice improvement!

PR #3796, opened 2026-05-12
Fixes #2162. Add more member conflict tests and test cases.

### LINKED ISSUE #2162: Add more member conflicts tests
For the statement

> Let C be a class. It is a compile-time error if C declares a constructor C named C.n and a static member with basename n

we have only one test, testing generative constructors and ststic members with the same names. No tests for constant, redirecting or factory constructors. Chect other statements as well

sgrekhov:


Language/Classes/Class_Member_Conflicts/method_and_setter_t02.dart
```diff
@@ -0,0 +1,45 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Let `C` be a class.
+/// ...
+/// It is a compile-time error if the interface of C has a method named `n` and
+/// a setter with basename `n`.
+///
+/// @description Check that it is a compile-time error if an interface has a
+/// method named `n` and an implicit setter with basename `n`.
```
eernstg: I think it's going to be really difficult to test this: If we have an implicitly induced setter whose name is `n=` then we probably can't avoid having a getter whose name is `n`, and then we can't claim to test that the error reports the "pseudo name clash" between the method name and the setter name, because it's certainly going to be a name clash that both the method and the getter have the name `n`.

And we already have another test to verify that the name clash between the getter and the method is reported as an error.

Perhaps the most reader-friendly approach would be to (1) omit this test, and (2) add a comment in Language/Classes/Class_Member_Conflicts/method_and_setter_t01.dart saying that we can't test the implicit case because we can't avoid the plain name clash.

Language/Classes/Class_Member_Conflicts/static_method_and_setter_t02.dart
```diff
@@ -0,0 +1,95 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Let `C` be a class.
+/// ...
+/// It is a compile-time error if `C` declares a static method named `n` and a
+/// static setter with basename `n`.
+///
+/// @description Check that it is a compile-time error if a class, mixin, enum,
+/// extension or extension type declares a static method named `n` and a static
+/// implicit setter with basename `n`.
```
eernstg: I think this one can be treated like Language/Classes/Class_Member_Conflicts/method_and_setter_t02.dart.

eernstg (commented):
Looks good! Just a couple of tests seem to try to test something which is not really possible (unless you have better ideas than I had about how to do it ;-).

sgrekhov:
Yes, I was thinking about that. In these tests, we can't avoid having a getter that might shadow a setter as a reason of the conflict. Let's remove these tests. PTAL.

eernstg (approved):
LGTM

PR #3800, opened 2026-05-14
Fixes #3798. Add tests for `bitwiseOrExpression` in `guardedPattern`

### LINKED ISSUE #3798: Add tests for operators in constant patterns
See https://github.com/dart-lang/sdk/issues/63356 for the details.

sgrekhov:


LanguageFeatures/Patterns/constant_A05_t01.dart
eernstg, line 53: We already have a "case =2". We should probably make all cases distinct such that duplicate values aren't spuriously succeeding.

I think a similar issue comes up in several tests below.

LanguageFeatures/Patterns/constant_A05_t03.dart
```diff
@@ -0,0 +1,103 @@
...
+  final x = switch (value) {
+    const (1 + 2) => "case =3",
+    const (2 - 1) => "case =1",
+    const (2 * 2) => "case =4",
+    const (2 / 1) => "case =2",
+    const (30 ^ 2) => "case =28",
+    const (1 | 4) => "case =5",
+    const (1 & 4) => "case =0",
+    const (15 % 8) => "case =7",
+    const (19 ~/ 8) => "case =2",
```
eernstg, line 44: Here's one of those duplicates.

LanguageFeatures/Patterns/if_case_A03_t03_lib.dart
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We replace the existing ifStatement rule with:
+/// ```
+/// ifStatement ::= ifCondition statement ('else' statement)?
+/// ifCondition :== 'if' '(' expression ( 'case' guardedPattern )? ')'
+/// ```
+/// @description Checks that pure `bitwiseOrExpression` constant expressions are
```
eernstg, line 10: Maybe this library should not have a description? It could say "Used by ..." (assuming that each of these libraries are actually only used from one test library).

LanguageFeatures/Patterns/if_case_A03_t01.dart
```diff
@@ -0,0 +1,95 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We replace the existing ifStatement rule with:
+/// ```
+/// ifStatement ::= ifCondition statement ('else' statement)?
+/// ifCondition :== 'if' '(' expression ( 'case' guardedPattern )? ')'
+/// ```
+/// @description Checks that pure `bitwiseOrExpression` constant expressions are
+/// not valid in a `guardedPattern` and produce a compile-time error.
```
eernstg: The non-terminal `bitwiseOrExpression` actually derives a lot of different expressions (including `a.b.c()` and `await Future.value(0)` and `{true: false}`) which are not the topic of this test.

Perhaps add a couple of words to indicate that we're talking about expressions of the form `e1 op e2` where `op` is a binary operator. That's somewhat informal, but I think it helps the reader see which expressions are of interest here. It could be enough to just say 'Check that binary operator expressions derived from `bitwiseOrExpression` ...'.

eernstg: Several similar cases occur below.

LanguageFeatures/Patterns/if_case_A03_t03_lib.dart
eernstg: Ah, it looks somewhat surprising that the `@description` doesn't say anything which is a meaningful description of the contents of the file, but I now remember that we've have similar cases before (especially with libraries with a tree of part files below them), and we've had exactly the same `@description` in all the files that constitute a single test.

So if that's the approach then there's no need to change anything.

LanguageFeatures/Patterns/constant_A05_t01.dart
eernstg: With this comment, I'm assuming that the purpose of this test includes verifying that the evaluation of the `const (...)` syntax yields the expected results (that is, if `case const (19 ~/ 8):` is treated as `case -1:` rather than `case 2:` then we want a test failure).

_If_ that's true then we want to avoid the duplicates.

eernstg (commented):
I think it would be useful to be slightly more explicit about what's being tested in all those 'bitwiseOrExpression' tests.

LanguageFeatures/Patterns/constant_A05_t03.dart
sgrekhov: Fixed.

LanguageFeatures/Patterns/if_case_A03_t01.dart
sgrekhov: Thanks! Fixed here and below.

LanguageFeatures/Patterns/if_case_A03_t03_lib.dart
sgrekhov: Leaved as is.

sgrekhov (commented):
Thank you. Updated. PTAL

sgrekhov:
@eernstg please review

LanguageFeatures/Patterns/if_case_A03_t03_lib.dart
eernstg: Sounds good!

eernstg (approved):
LGTM

PR #3804, opened 2026-05-19
Fixes #3803. Updated tests for abstract variable augmenting a getter/setter

### LINKED ISSUE #3803: co19/LanguageFeatures/Augmentations/type_annotation_inheritance_A03_t02
```dart
 5 /// @assertion It's a compile-time error if an abstract variable augments a
 6 /// getter and setter that don't have a combined signature.
 7 ///
 8 /// @description Checks that it's a compile-time error if an abstract variable
 9 /// augments a getter and setter that don't have a combined signature.
10 /// @author sgrekhov22@gmail.com
11
12 // SharedOptions=--enable-experiment=augmentations
13
14 abstract class A {
15   int get instanceVariable;
16 }
17
18 abstract class C implements A {
19   void set instanceVariable(String _);
20 }
21
22 augment abstract class C {
23   augment abstract var instanceVariable;
24 //                     ^^^^^^^^^^^^^^^^
25 // [analyzer] unspecified
26 // [cfe] unspecified
27 }
```

I have a local change, not in any CL yet, that started to verify that `augment int foo` has both getter and setter to augment. This test started to pass. I was curious what it was, but see something surprising. My understanding is that `augment abstract var instanceVariable` inside `augment abstract class C` says: augment *declarations* of getter and setter `instanceVariable` inside *declaration* of `class C`. That we work on declarations level, not on interface level.

So, I probably need a clarification from the language team.

sgrekhov:


LanguageFeatures/Augmentations/type_annotation_inheritance_A03_t02.dart
```diff
@@ -5,27 +5,78 @@
...
+  void set instanceVariable(String _);
+}
+
+abstract class C3 with A {
   void set instanceVariable(String _);
 }
 
-augment abstract class C {
+augment abstract class C1 {
+  augment abstract var instanceVariable;
```
eernstg: I would expect an error here because the abstract variable needs to have a type. Given that there is no type annotation it has to be inferred, somehow. But there's no rule that yields an inferred type that works for the situation where the superinterfaces have an `int` getter and a `String` setter.

'inference.md' has this: The type of a non-final field which overrides/implements both a setter and a getter is inferred to be the parameter type of the combined member signature of said setter in the direct superinterfaces, if this type is the same as the return type of the combined member signature of said getter in the direct superinterfaces. If the types are not the same then inference fails with an error.

The underlying rationale would be that it must be possible to write the result of type inference explicitly, so if we can't do that (because it's a variable declaration and its implicitly induced getter and setter have different and even unrelated types) then it's a compile-time error.

LanguageFeatures/Augmentations/type_annotation_inheritance_A03_t04.dart
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It's a compile-time error if an abstract variable augments a
+/// getter and setter that don't have a combined signature.
+///
+/// @description Checks that it's not an error if an abstract variable augments
+/// a getter declaration and there is a setter with the same name in the
+/// superinterface and they don't have a combined signature.
```
eernstg: Same difficulty here.

LanguageFeatures/Augmentations/type_annotation_inheritance_A03_t05.dart
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion It's a compile-time error if an abstract variable augments a
+/// getter and setter that don't have a combined signature.
+///
+/// @description Checks that it's not an error if an abstract variable augments
+/// a setter declaration and there is a getter with the same name in the
+/// superinterface and they don't have a combined signature. Test a mixin.
```
eernstg: Same again.

eernstg (commented):
I think the fact that there is no error when a getter and a setter have different return/parameter types is insufficient to allow the examples in these tests.

I have to go to a meeting now, but I'll write an issue on this tomorrow.

LanguageFeatures/Augmentations/type_annotation_inheritance_A03_t02.dart
sgrekhov: Based on https://github.com/dart-lang/co19/issues/3803#issuecomment-4475414832, I was thinking that an abstract variable augments only the _declaration_. In class `C1`, we have a declaration of the setter but no _declaration_ of a getter (it is in the superinterface). So, we should probably expect an error here because `C1` doesn't have a declaration of a getter to augment but not bacause of the combined member signature. Please clarify the expected result.

eernstg:
About the error at an abstract non-final variable declaration with no type annotation which is augmenting a getter and a setter such that the return type of the getter is not the same as the parameter type of the setter&mdash;we already have that error here: https://github.com/dart-lang/language/blob/main/working/augmentations/feature-specification.md#inheriting-combined-getter-setter-signatures.

The way this feature specification uses the phrase 'combined member signature' is not quite standard (the combined member signature is the shared subtype among a set of function types, it doesn't require all the parameter and return types to be "the same
type", but in this context we only accept "the same type"). But that shouldn't matter much, there's basically only one way to interpret the rule.

eernstg:
Will you write an issue on the analyzer about this when this PR has been landed?

The error message from the CFE is confusing. It does report an error as it should, but the wording of the error message seems to imply that the underlying expectation is off track:

```
n019.dart:7:24: Error: Augmentation member 'x' doesn't match a member in the augmented class.
Try changing the name to an existing member or removing the 'augment' modifier.
  augment abstract var x;
                       ^
```

I would expect something like "The type of augmentation member 'x' can not be inferred because it augments a getter and setter whose return respectively parameter type are different".

sgrekhov:
Let's go back to the initial issue https://github.com/dart-lang/co19/issues/3803. What should be the expected result below?
```dart
14 abstract class A {
15   int get instanceVariable;
16 }
17
18 abstract class C implements A {
19   void set instanceVariable(String _);
20 }
21
22 augment abstract class C {
23   augment abstract var instanceVariable;
24 //                     ^^^^^^^^^^^^^^^^
25 // [analyzer] unspecified
26 // [cfe] unspecified
27 }
```
As I understand it from https://github.com/dart-lang/co19/issues/3803#issuecomment-4475414832, in the code above the augmentation augments _declaration_ in class `C`, in this case the setter only. There is an error then, that the augmentation has no getter _declaration_ to augment (because it is in the superinterface) but this is a separate issue. So we should here expect an error that we are augmenting not existing getter _declaration_ but not an error about combined member signature.

Am I missing anything?

LanguageFeatures/Augmentations/type_annotation_inheritance_A03_t02.dart
eernstg: Ah, that's right! This declaration is an error because it attempts to augment a getter declaration `C1.instanceVariable`, and there is no introductory (or augmenting) declaration of `C1.instanceVariable`.

If we add a getter declaration `int get instanceVariable` to `C1` then we'll have the error that I mentioned, based on the fact that the return/parameter types are different.

In any case, it's an error.

eernstg: Same in line 38, 42.

eernstg:
> So we should here expect an error that we are augmenting not existing getter declaration but not an error about combined member signature.

Indeed!

eernstg (commented):
I didn't see the actual error, added comments in response to that.

sgrekhov:
https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Augmentations/type_annotation_inheritance_A03_t01.dart tests the case when getter and setter don't have a combined signature. New tests checks thet it is a compile-time error when we are augmenting declaration that doesn't exists. PTAL.

eernstg (approved):
LGTM!

PR #3805, opened 2026-05-20
#2119. Remove invalid test attributeChanged_A01_t01.dart

sgrekhov:
In this tests `IFrameElement` was extended. But since https://github.com/dart-lang/sdk/issues/53264 this is a compile-time error. Removed this test.


eernstg (approved):
LGTM

Issue #2303, opened 2023-10-11
Add tests for `File/Directory/Link.create/createSync()` when the path exists


sgrekhov:
We have co19 tests checking what should happen if we call `create/createSync()` and `FileSystemEntity` of the same type exists. But we have no tests checking what will happen when, say, we are creating a `File` but `Directory/Link` with the same path exists. Need to test all possible combinations

PR #3809, opened 2026-05-21
#2303. Make `Link.create()` tests stronger

sgrekhov:
It's important to tests `create()` with both possible values of `recursive`. Especially when the target of a link is a not existing entity and recursive is true (check that this entity is not created by mistake).


eernstg (approved):
LGTM

PR #3815, opened 2026-05-26
#2145. Add more tests for variables

sgrekhov:


Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A03_t01.dart
```diff
@@ -0,0 +1,72 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A non-late static variable declaration `D` named `id` that has
+/// the modifier `final` or the modifier `const` does not induce a setter.
+/// However, an assignment to `id` at a location where `D` is in scope is not
+/// necessarily a compile-time error. For example, a setter named `id=` could be
+/// found by lexical lookup.
+///
+/// @description Checks that it is not an error to assign to a `const` variable
```
eernstg: The assignment is actually an invocation of the setter, which means that the wording 'assign to a `const` variable' is misleading (the reader might think that we're _actually_ allowed to mutate the variable and the setter is just a kind of catalyst, sitting on the side and watching it happen). I think a couple of double quotes would suffice:
```suggestion
/// @description Checks that it is not an error to "assign to a `const` variable"
```
A similar situation occurs in other tests.

eernstg (commented):
I think it would be helpful for future readers to make a small adjustment in the descriptions.

Language/Variables/Implicitly_Induced_Getters_and_Setters/final_A03_t01.dart
sgrekhov: Updated. PTAL.

eernstg (approved):
LGTM

PR #3816, opened 2026-05-26
#2145. Add more tests for evaluation of implicit variable getters

sgrekhov:
It's possible to add some more tests for certain statements. However, these tests would be trivial — they would only check that the getter returns the object to which it is bound. This behavior is already covered by thousands of lines of code.

### DIFF
--- a/Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A03_t01.dart
+++ b/Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A03_t01.dart
@@ -58,7 +58,7 @@ extension type ET(int _) {
 void main() {
   Expect.throws(() {
     v = (() {
-        print(v);
+      print(v);
       return 42;
     })();
   });
--- a/Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A03_t01.dart
+++ b/Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A03_t01.dart
@@ -1,0 +1,95 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A late-uninitialized getter is a getter `g` which is implicitly
+/// induced by a non-local variable `v` that does not have an initializing
+/// expression. An invocation of `g` proceeds as follows: If the variable `v`
+/// has not been bound to an object then a dynamic error occurs. If `v` has been
+/// bound to an object `o′` then the invocation of `g` completes immediately,
+/// returning `o′`.
+/// ...
+/// Case ⟨Static or library variable⟩. If `d` declares a static or library
+/// variable, the implicitly induced getter of `id` executes as follows:
+/// ...
+/// - Variable without an initializer. If `d` declares a variable `id` without
+///   an initializing expression and does not have the modifier `late`, an
+///   invocation of the implicitly induced getter of `id` evaluates to the
+///   object that `id` is bound to.
+///   The variable is always bound to an object in this case. This may be the
+///   `null` object, which is the initial value of some variable declarations
+///   covered by this case.
+///   If `d` declares a variable `id` without an initializing expression and has
+///   the modifier `late`, the implicitly induced getter is a late-uninitialized
+///   getter. This determines the semantics of an invocation.
+///
+/// @description Checks that if `d` declares a variable `id` without an
+/// initializing expression and has the modifier `late` it is a dynamic error if
+/// `v` has not been bound to an object.
+/// @author sgrekhov22@gmail.com
+
+import "../../../Utils/expect.dart";
+
+late int v;
+
+class C {
+  static late final int v;
+}
+
+mixin M {
+  static late final int v;
+}
+
+enum E {
+  e0;
+  static late final int v;
+}
+
+class A {}
+
+extension Ext on A {
+  static late final int v;
+}
+
+extension type ET(int _) {
+  static late final int v;
+}
+
+void main() {
+  Expect.throws(() {
+    v = (() {
+        print(v);
+      return 42;
+    })();
+  });
+  Expect.throws(() {
+    C.v = (() {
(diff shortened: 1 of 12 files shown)


Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A03_t01.dart
```diff
@@ -0,0 +1,95 @@
...
+}
+
+extension type ET(int _) {
+  static late final int v;
+}
+
+void main() {
+  Expect.throws(() {
+    v = (() {
+        print(v);
```
eernstg: Indentation.

Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A04_t01.dart
```diff
@@ -0,0 +1,50 @@
...
+
+class MA = Object with M;
+
+void main() {
+  C c = C();
+  Expect.throws(() {
+    c.v = (() {
+      print(c.v);
+      return 42;
+    })();
```
eernstg, line 40: It looks like the test could still do its job according to the description if lines 37-40, including both, were replaced by `c.v;` (or `print(c.v);`).

This could be a tighter test because it would throw if and only if the evaluation of the uninitialized lazy variable throws. The current test seems to have the potential ability to look OK for the wrong reasons, including the situation where it throws at the assignment to `c.v`.

Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A05_t01.dart
```diff
@@ -0,0 +1,92 @@
...
+/// `g` completes returning `o`. If the evaluation of `e` throws then the
+/// invocation of `g` completes throwing the same object and stack trace, and
+/// does not change the binding of `v`.
+/// ...
+/// Case ⟨Late, initialized instance variable⟩. If `d` declares an instance
+/// variable `id` which has the modifier `late` and an initializing expression,
+/// the implicitly induced getter of `id` is a late-initialized getter. This
+/// determines the semantics of an invocation.
+///
+/// @description Checks the implicit getter of an initialized instance variable
```
eernstg:
```suggestion
/// @description Checks the implicit getter of an initialized late instance variable
```

Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A05_t02.dart
```diff
@@ -0,0 +1,51 @@
...
+/// invocation of `g` completes throwing the same object and stack trace, and
+/// does not change the binding of `v`.
+/// ...
+/// Case ⟨Late, initialized instance variable⟩. If `d` declares an instance
+/// variable `id` which has the modifier `late` and an initializing expression,
+/// the implicitly induced getter of `id` is a late-initialized getter. This
+/// determines the semantics of an invocation.
+///
+/// @description Checks that it is not an error if during an evaluation of the
+/// initializing expression `id` is invoked.
```
eernstg: Probably easier to read:
```suggestion
/// @description Checks that it is not an error if `id` is invoked
/// during an evaluation of the initializing expression.
```

Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A05_t03.dart
```diff
@@ -0,0 +1,54 @@
...
+/// does not change the binding of `v`.
+/// ...
+/// Case ⟨Late, initialized instance variable⟩. If `d` declares an instance
+/// variable `id` which has the modifier `late` and an initializing expression,
+/// the implicitly induced getter of `id` is a late-initialized getter. This
+/// determines the semantics of an invocation.
+///
+/// @description Checks that it is a compile-time error if during an evaluation
+/// of the initializing expression `id` is invoked and the type of the variable
+/// is not specified.
```
eernstg: Perhaps:
```suggestion
/// @description Checks that it is a compile-time error if `id` is invoked
/// during an evaluation of the initializing expression, and the type of the
/// variable is not specified.
```

eernstg (commented):
Couple of comments!

Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A03_t01.dart
sgrekhov: Fixed.

Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A04_t01.dart
sgrekhov: There are such tests, for example https://github.com/dart-lang/co19/blob/master/LanguageFeatures/nnbd/late_A02_t01.dart. This test checks the case when read occurs in the initializer. I didn't find such test.

Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A05_t01.dart
sgrekhov: Thank you. Fixed.

Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A05_t02.dart
sgrekhov: Updated.

Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A05_t03.dart
sgrekhov: Updated.

sgrekhov (commented):
Thank you. Updated. One place left unchanged. PTAL.

Language/Variables/Evaluation_of_Implicit_Variable_Getters/definition_A04_t01.dart
eernstg: OK!

eernstg (approved):
LGTM

Issue #3817, opened 2026-05-27
Change specific error messages to "unspecified"


sgrekhov:
Some former multi tests include specific error messages, we should change those to “unspecified” (exception: tests for specific warnings).

PR #3819, opened 2026-05-28
#3817. Replace specific error messages by 'unspecified'. Part 1.

sgrekhov:


Language/Expressions/Maps/equal_keys_t01.dart
```diff
@@ -4,22 +4,20 @@
...
 
-
 main() {
   const m1 = <int, String>{1: "val1", 1: "val2"};
 //                        ^
-// [cfe] Constant evaluation error:
-//                                    ^
-// [analyzer] COMPILE_TIME_ERROR.EQUAL_KEYS_IN_CONST_MAP
+// [cfe] unspecified
+// [analyzer] unspecified
```
eernstg, line 18: Just curious, does this work? I thought the diagnostic message expectation comments had to be specified in a canonical order (`[analyzer]` first, then `[cfe]`, and so on alphabetically).

https://github.com/dart-lang/sdk/blob/6c88c6f2ba475fe519bdd7377ac51462b2581dfd/pkg/test_runner/lib/src/static_error.dart#L23

I don't actually know what would go wrong, but I think the canonical order is intended to help avoiding unnecessary diffs (especially if the comments are ever generated or updated by a script).

eernstg: It looks like we have this in a few more locations. If it turns out to be useful to make many changes in order to maintain this ordering consistently then it would probably be a good idea to do it in a separate PR (or multiple PRs, if there are a lot of them).

However, the conclusion might also be that "we don't ever generate these comments automatically, and the non-canonical ordering really doesn't hurt".

eernstg (commented):
Just one question!

Language/Expressions/Maps/equal_keys_t01.dart
sgrekhov: Interesting! I never knew about that. I ran this updated test against a freshly built SDK in both the analyzer and the CFE, and the test passed. I think this comment is obsolete and misleading. I'll debug the SDK and remove the comment if that is indeed the case.

sgrekhov: I suggest to land it now. If the comment is actual we'll see it on the roll (if it is important for any other configuration, for example web).

eernstg: Removing the comment certainly wouldn't break anyone's code or test runs, but it might still be a useful rule for others to know: Some tests (in particular, in `$SDK/tests/language`) have had their diagnostic expectation comments updated by a script, and this would presumably create a lot of unnecessary diff elements if they aren't ordered canonically. So even if we don't care, others might.

I don't think co19 tests would ever have generated diagnostic expectation comments, in particular because they use `unspecified` which is presumably never generated by the relevant scripts; but also just because the automatically updated test expectation comments are used when the text of a diagnostic changes (usually: slightly), and that again doesn't make a difference for `unspecified`. We only change the expectation comments when the location of a diagnostic message changes so much that it moves to a different line of code.

You could say that the comment should be removed, or the behavior of the test runner should include reporting that some expectation comments are misordered. The current approach is half-done, and confusing.

However, if we have decided that these matters are irrelevant to co19 then we probably shouldn't do anything that might be inconvenient for some users.

An issue on `test_runner` indicating that the importance of this requirement ought to be clarified might be the best way ahead.

eernstg (approved):
LGTM

PR #3837, opened 2026-06-02
#3315. Add some more tests for primary constructors

sgrekhov:


LanguageFeatures/Primary-constructors/static_processing_A09_t02.dart
```diff
@@ -0,0 +1,43 @@
...
+
+/// @assertion The following errors apply to formal parameters of a primary
+/// constructor. Let `p` be a formal parameter of a primary constructor in a
+/// class, mixin class, enum, or extension type declaration `D` named `C`:
+/// ...
+/// A compile-time error occurs if `p` has the modifier covariant, but not `var`.
+///
+/// @description Check that it is a compile-time error if a formal parameters of
+/// a primary constructor has the modifier `covariant`, but not `var`. Test
+/// classes.
```
eernstg, line 13: I think it would be useful to check that it is still an error if the parameter has the modifier `final` (which is another kind of "not `var`").

LanguageFeatures/Primary-constructors/static_processing_A12_t05.dart
```diff
@@ -0,0 +1,66 @@
...
+///   `Object?`.
+///
+/// @description Check that if p has neither a declared type nor a `var` or
+/// `final` modifier, then its static type is `dynamic`. Test classes.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=primary-constructors
+
+class C1(x) {
+  this {
```
eernstg, line 34: Perhaps introduce `var didThrow = false;` and then make it true in the catch block and check that it's true at the end. It's unlikely, but the current code could presumably have a spurious success if the tools consider `x` to have a type that has a `checkDynamic` getter or method in its interface.

LanguageFeatures/Primary-constructors/static_processing_A12_t06.dart
```diff
@@ -0,0 +1,72 @@
...
+/// declared type, its type is `Object?` because of the implicit `final`
+/// modifier.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=primary-constructors
+
+import '../../Utils/static_type_helper.dart';
+
+extension type ET1(x) {
+  this {
```
eernstg, line 34: In this case it's actually more likely that we could have a spuriously succeeding test run: If `x` has _any_ type other than `dynamic` then we can execute `expectStaticType<...>`, and it will succeed if that type is a top type (except `void`).

LanguageFeatures/Primary-constructors/static_processing_A12_t07.dart
```diff
@@ -0,0 +1,72 @@
...
+///
+/// @description Checks that, in an extension type, if `p` does not have a
+/// declared type, its type is `Object?` rather than `dynamic` because of the
+/// implicit `final` modifier.
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=primary-constructors
+
+extension type ET1(x) {
+  this {
```
eernstg, line 32: Ditto.

eernstg (commented):
Couple of enhancements suggested.

LanguageFeatures/Primary-constructors/static_processing_A09_t02.dart
sgrekhov: [static_processing_A08_t01.dart](https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Primary-constructors/static_processing_A08_t01.dart)
[static_processing_A08_t02.dart](https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Primary-constructors/static_processing_A08_t02.dart)
[static_processing_A08_t03.dart](https://github.com/dart-lang/co19/blob/master/LanguageFeatures/Primary-constructors/static_processing_A08_t03.dart)
do that.

LanguageFeatures/Primary-constructors/static_processing_A12_t05.dart
sgrekhov: Makes sense, thank you. Updated.

LanguageFeatures/Primary-constructors/static_processing_A12_t06.dart
sgrekhov: Removed `try-catch`. If we have  a spurious `dynamic` here this test won't detect it. But the next tests will do (it expects)
```dart
    x.testDynamic;
//    ^^^^^^^^^^^
// [analyzer] unspecified
// [cfe] unspecified
```

LanguageFeatures/Primary-constructors/static_processing_A12_t07.dart
sgrekhov: Please see the above.

sgrekhov (commented):
Updated. PTAL.

LanguageFeatures/Primary-constructors/static_processing_A12_t05.dart
eernstg: Thx!

LanguageFeatures/Primary-constructors/static_processing_A12_t06.dart
eernstg: OK!

LanguageFeatures/Primary-constructors/static_processing_A09_t02.dart
eernstg: OK, thanks!

eernstg (approved):
LGTM

PR #3838, opened 2026-06-02
#3315. Add tests for abbrevations. Part 1.

sgrekhov:


eernstg (commented):
The language team hasn't agreed on the terminology at this time, but the language specification consistently uses `MyClass` and `MyClass.someIdentifier` as the names of constructors. They may be declared using that name directly (like the traditional `MyClass();` and like primary constructors) or they may be declared using the new abbreviated syntax (like `factory();` or `const new someIdentifier();`).

If we switch to say that the name of `MyClass();` is empty or `new`, and the name of `MyClass.name();` is `name`, we have a new naming scheme, and I believe this can be done consistently as well.

We'll need to talk about "the constructor named `new` in the class `MyClass`" rather than "the constructor `MyClass.new`", but this can be done.

However, none of these considerations will allow a constructor to have the name `new name` (or `factory name` for that matter).

Perhaps the constructor naming war could be avoided altogether by saying that the constructor is "declared using `new name` / `const factory` / etc"?

sgrekhov:
Updated. PTAL.

eernstg (approved):
LGTM, thanks!

PR #3849, opened 2026-06-08
Update VM tests to use new IsolateTestHarness

biggs0125:
In order to improve the maintainability of these tests I have redesigned the IsolateTest infrastructure.

This involves separating the test itself from the code under test by putting them in separate files. The expectations are now all set via the harness class and line numbers are no longer hard-coded but instead are derived from the source code itself via comment labels.

Associated SDK commit: https://dart-review.googlesource.com/c/sdk/+/509700


VM/primary_constructors_t01_lib.dart.stops
```diff
@@ -0,0 +1,14 @@
+primary_constructors_t01_lib.dart:57:10
```
sgrekhov: May be add a comment explaining how this file is used and how it should be modified (I guess it is generated)?

VM/primary_constructors_t01_lib.dart.stops
```diff
@@ -0,0 +1,14 @@
...
+primary_constructors_t01_lib.dart:57:12
```
sgrekhov: Are these numbers correct? As far as I can see there is no line 57 in `primary_constructors_t01_lib.dart` file.

sgrekhov (commented):
Please see a copuple of questions below regarding `VM/primary_constructors_t01_lib.dart.stops`.

VM/primary_constructors_t01_lib.dart.stops
biggs0125: Good catch, I removed a bunch of newlines but didn't regenerate the goldens. They're updated now.

VM/primary_constructors_t01_lib.dart.stops
```diff
@@ -0,0 +1,14 @@
+primary_constructors_t01_lib.dart:27:10
```
sgrekhov: Is it possible to add a comment at the beginning of this file with the command used to update it? Something like `// Update this file by running: ...`

VM/primary_constructors_t01_lib.dart.stops
biggs0125: Done. Added it to the start of every golden.

nshahan (approved):
LGTM 

Overall, I'm going to defer to @sgrekhov since they will be maintaining these kinds of tests going forward.

VM/primary_constructors_t01_lib.dart.stops
biggs0125: I added a similar header to the top of the golden file. It doesn't say the exact test but it should be easily inferable given the file name.

sgrekhov:
Thanks! LGTM.

Issue #3847, opened 2026-06-08
Add tests for type inference


sgrekhov:
[Specification](https://github.com/dart-lang/language/blob/main/resources/type-system/inference.md)

sgrekhov:
Waiting for https://github.com/dart-lang/language/issues/4718 before completion.

PR #3850, opened 2026-06-09
#3847. Add method override inference tests

sgrekhov:


TypeSystem/inference/method_override_inference_A01_t02.dart
```diff
@@ -0,0 +1,46 @@
...
+
+/// @assertion A method `m` of a class `C` is subject to override inference if
+/// it is missing one or more component types of its signature, and one or more
+/// of the direct superinterfaces of `C` has a member named `m` (that is, `C.m`
+/// overrides one or more declarations). Each missing type is filled in with the
+/// corresponding type from the combined member signature `s` of `m` in the
+/// direct superinterfaces of `C`.
+///
+/// @description Checks that in case of extension missing components of a method
+/// signature are not inferred form a superinterface.
```
eernstg: Good catch! We should test this. However, an extension does not introduce a type and hence it can't have any superinterfaces. So we just talk about the 'on-type' as usual:

```suggestion
/// @description Checks that with an extension, missing components of a method
/// signature are not inferred from the on-type.
```

TypeSystem/inference/method_override_inference_A01_t01.dart
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A method `m` of a class `C` is subject to override inference if
```
eernstg, line 5: The language team definitely supports the broadening of the specified rule such that override inference applies to membered declarations other than classes: `mixin`, `mixin class`, and `enum` are all covered.

On the other hand, the language team does _not_ support override inference with instance members of extension types.

I think this test and the following tests are already in line with this.

TypeSystem/inference/method_override_inference_A01_t03.dart
```diff
@@ -0,0 +1,40 @@
...
+
+/// @assertion A method `m` of a class `C` is subject to override inference if
+/// it is missing one or more component types of its signature, and one or more
+/// of the direct superinterfaces of `C` has a member named `m` (that is, `C.m`
+/// overrides one or more declarations). Each missing type is filled in with the
+/// corresponding type from the combined member signature `s` of `m` in the
+/// direct superinterfaces of `C`.
+///
+/// @description Checks that in case of extension types missing components of a
+/// method signature are not inferred form a superinterface.
```
eernstg:
```suggestion
/// @description Checks that with an extension type, missing components of a
/// method signature are not inferred from a superinterface.
```

We should cover both implemented extension types and implemented types of other kinds (`class`, `mixin`, `mixin class`, and `enum`).

TypeSystem/inference/method_override_inference_A01_t04.dart
```diff
@@ -0,0 +1,72 @@
...
+  void m5({required v}) {}
+}
+
+class C2 implements B {
+  m1() => 0;
+  void m2(v) {}
+  void m3([v = 1]) {}
+  void m4({v = 1}) {}
+  void m5({required v}) {}
+}
```
eernstg, line 42: We should be able to have a `mixin class` and a `mixin` with an `implements B` clause or (with the `mixin`) an `on` clause.

TypeSystem/inference/method_override_inference_A01_t05.dart
```diff
@@ -0,0 +1,72 @@
...
+  void m5({required v}) {}
+}
+
+class C2 implements B {
+  m1() => 0;
+  void m2(v) {}
+  void m3([v = 1]) {}
+  void m4({v = 1}) {}
+  void m5({required v}) {}
+}
```
eernstg, line 42: Again, we should be able to test that override inference can take place in a `mixin`, and a `mixin class` based on an `implements/on B` clause.

eernstg (commented):
Looks good, couple of comments about the terminology and potential additional cases.

TypeSystem/inference/method_override_inference_A01_t02.dart
sgrekhov: Thanks! Updated.

TypeSystem/inference/method_override_inference_A01_t03.dart
sgrekhov: Description updated.

> We should cover both implemented extension types and implemented types of other kinds (class, mixin, mixin class, and enum).

Please clarify. I added the forgotten case `mixin M2 implements A ...` to `method_override_inference_A01_t01.dart`. So, we now have all kinds of items with the `implements` keyword here (except `mixin class`, we currently just have `class` and `mixin`. Does it make sense to add `mixin class` as well?). Is that what you meant?

TypeSystem/inference/method_override_inference_A01_t04.dart
sgrekhov: Thank you! Missed these cases. Added `mixin M1 implements B` and `mixin M2 on B`. Again, do we need a separate `mixin class` case?

TypeSystem/inference/method_override_inference_A01_t05.dart
sgrekhov: Yes! Added.

sgrekhov (commented):
Thank you. Updated. There is a question about if we really need a separate `mixin class` case. PTAL.

sgrekhov:
I'm also going to add cases when the type is inferred from abstract and external methods but in a separate PR.

TypeSystem/inference/method_override_inference_A01_t03.dart
```diff
@@ -0,0 +1,40 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion A method `m` of a class `C` is subject to override inference if
+/// it is missing one or more component types of its signature, and one or more
+/// of the direct superinterfaces of `C` has a member named `m` (that is, `C.m`
+/// overrides one or more declarations). Each missing type is filled in with the
+/// corresponding type from the combined member signature `s` of `m` in the
+/// direct superinterfaces of `C`.
+///
+/// @description Checks that with ab extension type, missing components of a
```
eernstg: Typo:

```suggestion
/// @description Checks that with an extension type, missing components of a
```

TypeSystem/inference/method_override_inference_A01_t03.dart
eernstg: We should cover all kinds of type introducing membered declarations because they are the ones that have member declarations and superinterfaces, and this means that they can be subject to override inference.

The specification currently only says that override inference applies to class declarations, but the language team agrees that this is an oversight, and other type introducing membered declarations should be included as well.

However, the language team does not want to have override inference with extension types (the mean reason is that the 'redeclare' relation is so different from an 'override' relation that we shouldn't expect a redeclaring member to have any particular relationship with the redeclared member).

The membered type introducing declarations are class, mixin, mixin class, enum, and extension type declarations.

I prefer to describe the semantics of a mixin class as a double effect: When a mixin class declaration is encountered, a class as well as a mixin is introduced into the enclosing library scope, and each lookup will choose the class when the given name is used as a class (almost always), and it will choose the mixin when it is used as a mixin (which is basically only in `with` clauses).

I'm not sure, but I do think this makes a mixin class so special that it should be considered as a separate kind of declaration. Implementers of the tools could disagree, but we usually don't consider the implementations as the end of all questions &mdash; even if a mixin class is treated essentially like a class (and we just don't get an error when it is applied as a mixin, and we do get some errors if it has a non-trivial constructor etc.), the next version of the analyzer or CFE could handle mixin classes differently.

In summary, I do think mixin classes should be counted as a kind of type introducing membered declaration, just like the other 4.

In addition to all these, we have `extension` declarations, and they certainly don't have anything that corresponds to an override relationship for any member, so they're treated differently.

TypeSystem/inference/method_override_inference_A01_t04.dart
eernstg: (See the comment above.)

TypeSystem/inference/method_override_inference_A01_t05.dart
eernstg: Thx!

eernstg (commented):
Some discussion about mixin classes and a typo.

sgrekhov:
Thank you. Updated. Typos fixed, `mixin class` cases added. PTAL.

eernstg (approved):
LGTM, thanks!

PR #3858, opened 2026-06-13
Update VM tests to match fixed AOT-compatible framework setup.

biggs0125:
Through further testing I realized that the previous setup of this framework was incompatible with the way test_runner + AOT works. This change maintains the clarity of having the code under test be in a separate library from the test entrypoint while also making the test compatible with AOT by having all code be a single program (via the import and tearoff).


sgrekhov:
LGTM. Tnanks.

biggs0125:
@sgrekhov Thanks, mind adding approval as well?

nshahan (approved):

PR #3884, opened 2026-06-18
Fixes #3877. Add missing parameter in augmenting_constructors_A01_t04.dart

sgrekhov:

PR #3889, opened 2026-06-18
Fixes #3884. Fix private parameter names in augmenting_constructors_A01_t24.dart

sgrekhov:


eernstg (approved):
LGTM

eernstg:
Note that the specification documents seem to leave no option open to augment a private named parameter, and the approach used in this test (and a couple of others) may change when https://github.com/dart-lang/language/issues/4710 is discussed and resolved.

PR #3890, opened 2026-06-18
#3847. Add more tests for getters and setters in generic classes

sgrekhov:


eernstg:
That's an interesting title! ;-)

I've proposed the introduction of generic getters (enabling things like `e.as<int>` as an invocation of an `as` getter with a type argument), but I haven't thought about generic setters. Perhaps `e.mySetter =<int> 'Hello';` for `set mySetter<T>(String s) => someVariable = s is T;`?

sgrekhov:
> That's an interesting title! ;-)
> 
> I've proposed the introduction of generic getters (enabling things like `e.as<int>` as an invocation of an `as` getter with a type argument), but I haven't thought about generic setters. Perhaps `e.mySetter =<int> 'Hello';` for `set mySetter<T>(String s) => someVariable = s is T;`?

Yes, the title is wrong. Renamed to "Add more tests for getters and setters in generic classes".

TypeSystem/inference/getter_setter_override_inference_A03_t06.dart
```diff
@@ -0,0 +1,106 @@
...
+
+class B {
+  void set m1(String _) {}
+  void set m2(String _) {}
+}
+
+class C1 implements A, B {
+  get m1 => 0;
+  get m2 => 0;
+  void set m1(_) {} // We need to implement setters as well
```
eernstg, line 29: Thanks for the comment, it's helpful!

eernstg (approved):
LGTM

Issue #3912, opened 2026-06-23
Move `LanguageFeatures/Super-mixins` tests to `Language`


sgrekhov:
The tests under `LanguageFeatures/Super-mixins` were written according to the [feature specification](https://github.com/dart-lang/language/blob/main/accepted/2.1/super-mixins/feature-specification.md). The assertions in these tests were derived from that specification. Since then, the feature has been fully integrated into the Dart language specification, which is now the official source of truth.

We need to:

- update the assertion texts in these tests to match the wording of the current Dart language specification;
- rename the tests, if necessary;
- move them to the appropriate folder under Language.

This issue can serve as a proof of concept (PoC) to estimate the cost of this migration.

PR #3913, opened 2026-06-23
#3912. Updated assertions in tests for super-mixins. Part 1.

sgrekhov:
This is an experimental PR demonstrating how co19 tests from `LanguageFeatures` can be migrated to the `Language` folder. The full process will be as follows:
- For language features that are already integrated into the current specification, update the assertions to use the wording from the specification.
- Rename the tests (if necessary) and move them to the appropriate folder under `Language`.


eernstg:
(Same comment as [this one](https://github.com/dart-lang/co19/pull/3919#issuecomment-4787453806).)

sgrekhov:
Added the description.

PR #3919, opened 2026-06-24
#3912. Updated assertions in tests for super-mixins. Part 2.

sgrekhov:
This is an experimental PR demonstrating how co19 tests from `LanguageFeatures` can be migrated to the `Language` folder. The full process will be as follows:
- For language features that are already integrated into the current specification, update the assertions to use the wording from the specification.
- Rename the tests (if necessary) and move them to the appropriate folder under `Language`.


eernstg:
Hello @sgrekhov, I think we should mention in the description of this PR that it is an experiment intended to show how we could proceed, and it is not intended to be landed as long as we're still discussing how to proceed.

sgrekhov:
Added the description.

PR #3924, opened 2026-06-26
Fixes #3915. Add tests for augmentation with `with` clause

### LINKED ISSUE #3915: Add tests for augmentations with `with` clause
See https://github.com/dart-lang/sdk/issues/63644

sgrekhov:
`augmenting_class_like_declarations_A02_t20-27.dart` check that augmentation may add `with` clause. These tests add missing cases.

### DIFF
--- a/LanguageFeatures/Augmentations/augmenting_class_like_declarations_A02_t29.dart
+++ b/LanguageFeatures/Augmentations/augmenting_class_like_declarations_A02_t29.dart
@@ -19,9 +19,9 @@ import '../../Utils/expect.dart';
 StringBuffer log = StringBuffer();
 
 class A {
-  String v = 'M1';
-  String get g => 'M1';
-  String m() => 'M1';
+  String v = 'A';
+  String get g => 'A';
+  String m() => 'A';
   void set s(String v) {
     log.write('$v;');
   }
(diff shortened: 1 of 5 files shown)


LanguageFeatures/Augmentations/augmenting_class_like_declarations_A02_t29.dart
```diff
@@ -0,0 +1,49 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=augmentations
+
+import '../../Utils/expect.dart';
+
+StringBuffer log = StringBuffer();
+
+class A {
+  String v = 'M1';
```
eernstg: Could be `'A'` rather than `'M1'`.

LanguageFeatures/Augmentations/augmenting_class_like_declarations_A02_t30.dart
```diff
@@ -0,0 +1,49 @@
...
+
+augment class C with M1 implements M2 {}
+
+main() {
+  Expect.equals('M1', C().v);
+  Expect.equals('M1', C().g);
+  Expect.equals('M1', C().m());
+  C().s = 'M1';
+  Expect.equals('M1;', log.toString());
+}
```
eernstg, line 49: OK, this test may look funny at first, but it makes sense to cover the property "an implemented mixin isn't mixed in, it is just implemented". Good catch!

eernstg (commented):
Looks good, just one matter of string contents could be adjusted for consistency.

LanguageFeatures/Augmentations/augmenting_class_like_declarations_A02_t29.dart
sgrekhov: Yes, thank you for noticing. Updated.

sgrekhov (commented):
Updated. PTAL.

LanguageFeatures/Augmentations/augmenting_class_like_declarations_A02_t29.dart
eernstg: Thx!

eernstg (approved):
LGTM

PR #3935, opened 2026-06-30
#3847. Add function literal inference tests for parameters

sgrekhov:
This is a very small PR containing a couple of tests for function literal inference. The goal is to check whether the general approach is OK. Any advice is appreciated.

### DIFF
--- a/TypeSystem/inference/function_literal_inference_A01_t01.dart
+++ b/TypeSystem/inference/function_literal_inference_A01_t01.dart
@@ -1,0 +1,249 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion Function literals which are inferred in an non-empty typing
+/// context where the context type is a function type are inferred as described
+/// below.
+///
+/// Each parameter is assumed to have its declared type if present. If no type
+/// is declared for a parameter and there is a corresponding parameter in the
+/// context type schema with type schema `K`, the parameter is given an inferred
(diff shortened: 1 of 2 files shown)


TypeSystem/inference/function_literal_inference_A01_t01.dart
```diff
@@ -0,0 +1,249 @@
...
+/// context type schema with type schema `K`, the parameter is given an inferred
+/// type `T` where `T` is derived from `K` as follows. If the greatest closure
+/// of `K` is `S` and `S` is a subtype of `Null`, then without null safety `T`
+/// is `dynamic`, and with null safety `T` is `Object?`. Otherwise, `T` is `S`.
+/// If there is no corresponding parameter in the context type schema, the
+/// variable is treated as having type `dynamic`.
+///
+/// @description Checks if no type is declared for a parameter and there is a
+/// corresponding parameter in the context type schema with a type schema, the
+/// parameter type is inferred from the type schema.
```
eernstg, line 20: "Is inferred from that type schema" is fine here, but type inference is in general done "in the context of a type schema" rather than "from a type schema" because the type schema doesn't get to dictate the result. For example, inference of `1` in the context of `List<List<_>>` will provide the context schema to the inference algorithm, but it is ignored completely and the expression is typed as an `int`. This may be an error, but it could also be an assignment to a local variable which is then subject to demotion.

Inference of the declared type of a formal parameter of a function literal is a very special situation because we can use _any_ type as the declared type of the parameter. This is the reason why "inferred from" is OK right here: The context type _does_ actually get to make the final decision.

However, it's a good habit in general to use a wording that allows for inference to use the context type in whatever way it wishes.

eernstg (approved):
LGTM, landing.
I added one comment, just to mention a small issue about terminology.

PR #3937, opened 2026-07-01
#3847. Add function literal inference tests for return type. Part 1.

sgrekhov:


TypeSystem/inference/function_literal_inference_A02_t02.dart
```diff
@@ -0,0 +1,264 @@
...
+/// the typing context for the function body as computed above from the imposed
+/// return type schema.
+/// - For each `return e;` statement in the block, let `S` be the inferred type
+///   of `e`, using the local type inference algorithm described below with
+///   typing context `K`, and update `T` to be `UP(flatten(S), T)` if the
+///   enclosing function is `async`, or `UP(S, T)` otherwise.
+///
+/// @description Check that return type of an async function with the only
+/// `return e;` statement is `UP(flatten(S), Never)`, where `S` is inferred type
+/// of `e`.
```
eernstg, line 20: I think we should cover the case where `flatten(S)` doesn't just return `S`, that is, cases where the `e` in `return e;` has a static type which is `Future<...>` or `FutureOr<...>`. I think one extra case using `Future<int>` would be fine: We just need to see that `flatten` is indeed used (some other tests will have different cases verifying that `flatten` works correctly, so we don't need to duplicate those cases here).

TypeSystem/inference/function_literal_inference_A02_t04.dart
```diff
@@ -0,0 +1,94 @@
...
+/// any control path reaches the end of the block without exiting the block, as
+/// computed by the definite completion analysis specified elsewhere. Let `K` be
+/// the typing context for the function body as computed above from the imposed
+/// return type schema.
+/// - For each `return e;` statement in the block, let `S` be the inferred type
+///   of `e`, using the local type inference algorithm described below with
+///   typing context `K`, and update `T` to be `UP(flatten(S), T)` if the
+///   enclosing function is `async`, or `UP(S, T)` otherwise.
+///
+/// @description Check that return type of an async function with the only
```
eernstg: Ditto.

TypeSystem/inference/function_literal_inference_A02_t04.dart
```diff
@@ -0,0 +1,94 @@
...
+/// `return e;` statement is `UP(flatten(S), Never)`, where `S` is inferred
+/// using the local type inference algorithm.
```
eernstg, line 20: We should again have a case where `flatten(S)` makes a difference.

TypeSystem/inference/function_literal_inference_A03_t01.dart
```diff
@@ -0,0 +1,118 @@
...
+/// any control path reaches the end of the block without exiting the block, as
+/// computed by the definite completion analysis specified elsewhere. Let `K` be
+/// the typing context for the function body as computed above from the imposed
+/// return type schema.
+/// ...
+/// - If the enclosing function is not marked `sync*` or `async*`: For each
+///   `return;` statement in the block, update `T` to be `UP(Null, T)`.
+///
+/// @description Check that return type of a function with the only `return;` or
+/// with no `return` statement is `UP(Null, Never)`.
```
eernstg: Ditto.

TypeSystem/inference/function_literal_inference_A03_t02.dart
```diff
@@ -0,0 +1,118 @@
...
+/// the definite completion analysis specified elsewhere. Let `T` be `Null` if
+/// any control path reaches the end of the block without exiting the block, as
+/// computed by the definite completion analysis specified elsewhere. Let `K` be
+/// the typing context for the function body as computed above from the imposed
+/// return type schema.
+/// ...
+/// - If the enclosing function is not marked `sync*` or `async*`: For each
+///   `return;` statement in the block, update `T` to be `UP(Null, T)`.
+///
+/// @description Check that return type of a function marked `async` with the
```
eernstg: Ditto.

eernstg (commented):
Looks good, could have a couple of extra cases to ensure that we will detect if `flatten(S)` isn't being done when it should.

TypeSystem/inference/function_literal_inference_A02_t02.dart
sgrekhov: Yes, thank you. I forgot about `Future`. Added.

TypeSystem/inference/function_literal_inference_A02_t04.dart
sgrekhov: Added.

TypeSystem/inference/function_literal_inference_A03_t01.dart
sgrekhov: This test checks that the retur type of the function literal `() { return; };` is inferred as `Null`. How can we check `flutten()` here?

TypeSystem/inference/function_literal_inference_A03_t02.dart
sgrekhov: The same as the above. `flutten()` chack cannot be applied to this test.

sgrekhov (commented):
Updated. More tests cases added but the last two tests leaved unchanged. PTAL

TypeSystem/inference/function_literal_inference_A02_t02.dart
eernstg: Thanks!

eernstg (approved):
LGTM

TypeSystem/inference/function_literal_inference_A03_t01.dart
eernstg: Oh, I should have written a longer comment. 'Ditto' was just intended to repeat the comment about the missing article and the terminology about 'only'. I made those changes.

eernstg (approved):
LGTM

TypeSystem/inference/function_literal_inference_A03_t01.dart
sgrekhov: Thank you!

PR #3939, opened 2026-07-02
#3847. Add function literal inference tests for return type. Part 2.

sgrekhov:
For review it's enough to review top-level functions only. The other code repeats these functions as static and instance members in classes, mixins, mixin classes, enums, extensions and extension types.

### DIFF
--- a/TypeSystem/inference/function_literal_inference_A04_t01.dart
+++ b/TypeSystem/inference/function_literal_inference_A04_t01.dart
@@ -1,0 +1,344 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion The actual returned type of a function literal with a block body
+/// is computed as follows. Let `T` be `Never` if every control path through the
+/// block exits the block without reaching the end of the block, as computed by
+/// the definite completion analysis specified elsewhere. Let `T` be `Null` if
+/// any control path reaches the end of the block without exiting the block, as
+/// computed by the definite completion analysis specified elsewhere. Let `K` be
+/// the typing context for the function body as computed above from the imposed
+/// return type schema.
+/// ...
+/// - For each `yield e;` statement in the block, let `S` be the inferred type
+///   of `e`, using the local type inference algorithm described below with
+///   typing context `K`, and update `T` to be `UP(S, T)`.
+/// - If the enclosing function is marked `sync*`, then for each `yield* e;`
+///   statement in the block, let `S` be the inferred type of `e`, using the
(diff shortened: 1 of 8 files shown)


TypeSystem/inference/function_literal_inference_A04_t01.dart
```diff
@@ -0,0 +1,344 @@
...
+///   of `e`, using the local type inference algorithm described below with
+///   typing context `K`, and update `T` to be `UP(S, T)`.
+/// - If the enclosing function is marked `sync*`, then for each `yield* e;`
+///   statement in the block, let `S` be the inferred type of `e`, using the
+///   local type inference algorithm described below with a typing context of
+///   `Iterable<K>;` let `E` be the type such that `Iterable<E>` is a
+///   super-interface of `S;` and update `T` to be `UP(E, T)`.
+///
+/// @description Check that the return type of a `sync*` function with a single
+/// `yield e;` is `Iterable<X>;`, where `X` is the static type of `e`.
```
eernstg, line 24: I'd actually expect the current specification and implementation to make the return type `Iterable<X?>` because of the spurious `Null` mentioned in dart-lang/language#4718.

Perhaps this particular test should be postponed until this has been resolved? Otherwise, we should expect the `?` to be included in all cases because all function bodies can be completed normally.

eernstg: The same consideration holds for the following tests about generators where the body can complete normally (I think that's all of them, they don't have any return statements or other constructs that prevent the body from completing normally).

TypeSystem/inference/function_literal_inference_A05_t02.dart
```diff
@@ -0,0 +1,100 @@
...
+/// ...
+/// - If the enclosing function is marked `async*`, then for each `yield* e;`
+///   statement in the block, let `S` be the inferred type of `e`, using the
+///   local type inference algorithm described below with a typing context of
+///   `Stream<K>`; let `E` be the type such that `Stream<E>` is a
+///   super-interface of `S;` and update `T` to be `UP(E, T)`.
+///
+/// @description Check that the return type of an `async*` function with a
+/// single `yield e;` is `Stream<X>;`, where `X` is the static type of `e`
+/// inferred using the local type inference algorithm.
```
eernstg, line 26: OK, this particular test already has return type `Stream<X?>` everywhere because it's the same as `Stream<X>`, so we don't have to block it on dart-lang/language#4718.

eernstg (commented):
Almost every test in this PR depends on dart-lang/language#4718 (about the spurious `?` which is added to the element type in all return types of generators when the end of the body can be reached).

Perhaps we should just block this until that issue has been clarified.

sgrekhov:
Blocked by dart-lang/language#4718.

PR #3940, opened 2026-07-03
#3847. Add function literal inference tests for return type. Part 3.

sgrekhov:
For review, it's enough to review only the top-level functions. The rest of the code repeats these functions as static and instance members of classes, mixins, mixin classes, enums, extensions, and extension types.


sgrekhov:
@eernstg please review

TypeSystem/inference/function_literal_inference_A06_t01.dart
```diff
@@ -0,0 +1,308 @@
...
+  Ext.sf1.expectStaticType<Exactly<void Function()>>();
+  Ext.sf2.expectStaticType<Exactly<void Function()>>();
+  Ext.sf3.expectStaticType<Exactly<void Function()>>();
+  Ext.sf4.expectStaticType<Exactly<void Function()>>();
+
+  ET.sf1.expectStaticType<Exactly<void Function()>>();
+  ET.sf2.expectStaticType<Exactly<void Function()>>();
+  ET.sf3.expectStaticType<Exactly<void Function()>>();
+  ET.sf4.expectStaticType<Exactly<void Function()>>();
+}
```
eernstg, line 76: Very nice!

However, we'd need to know that the return type is actually `void`, and the `expectStaticType` check doesn't quite achieve this, it only confirms that the return type is a top type.

It isn't convenient because this test should be runnable, and a real `void` confirmation test would probably have to expect errors for a situation where the returned object is used. Perhaps the declarations of functions could be moved to a '...lib.dart' file and then this test could import it, and a new test that does the `Object? _ = myVoidValue;` and expects an error would be another library that imports the same file. This would also be nice because we don't have to double check that the two tests are actually working on the same set of function literals.

TypeSystem/inference/function_literal_inference_A06_t02.dart
```diff
@@ -0,0 +1,518 @@
...
+  Ext.sf7.expectStaticType<Exactly<Future<void> Function()>>();
+
+  ET.sf1.expectStaticType<Exactly<Future<void> Function()>>();
+  ET.sf2.expectStaticType<Exactly<Future<void> Function()>>();
+  ET.sf3.expectStaticType<Exactly<Future<void> Function()>>();
+  ET.sf4.expectStaticType<Exactly<Future<void> Function()>>();
+  ET.sf5.expectStaticType<Exactly<Future<void> Function()>>();
+  ET.sf6.expectStaticType<Exactly<Future<void> Function()>>();
+  ET.sf7.expectStaticType<Exactly<Future<void> Function()>>();
+}
```
eernstg, line 196: Similarly, a test like `void main() async { Object? _ = await myFutureVoidValue; }` expecting an error would verify that the return type is actually `Future<void>` and not just `Future<SomeOtherTopType>`.

TypeSystem/inference/function_literal_inference_A06_t03.dart
```diff
@@ -0,0 +1,518 @@
...
+  Ext.sf7.expectStaticType<Exactly<Future<void> Function()>>();
+
+  ET.sf1.expectStaticType<Exactly<Future<void> Function()>>();
+  ET.sf2.expectStaticType<Exactly<Future<void> Function()>>();
+  ET.sf3.expectStaticType<Exactly<Future<void> Function()>>();
+  ET.sf4.expectStaticType<Exactly<Future<void> Function()>>();
+  ET.sf5.expectStaticType<Exactly<Future<void> Function()>>();
+  ET.sf6.expectStaticType<Exactly<Future<void> Function()>>();
+  ET.sf7.expectStaticType<Exactly<Future<void> Function()>>();
+}
```
eernstg, line 107: Similar again.

eernstg (commented):
Looks good!

One thing may slip through, though: I think a compiler/analyzer could conclude that the return type of a number of function literals can be `Object?` or some other top type which is not `void`, and the existing tests will not detect the issue. So we will probably need to have some `void` specific tests confirming that it is an error to use those returned values..

sgrekhov:
Updates as suggested. PTAL.

eernstg (approved):
LGTM.

Nice, thank you!

PR #3941, opened 2026-07-06
#3847. Add tests for local function return type inference

sgrekhov:


TypeSystem/inference/local_return_type_inference_A02_t02.dart
```diff
@@ -0,0 +1,141 @@
...
+
+/// @assertion With null safety, local function body inference is changed so
+/// that the local function name is not considered available for inference while
+/// performing inference on the body. As a result, any recursive calls to the
+/// function for which the result type is required for inference to complete
+/// will no longer be treated as having return type `dynamic`, but will instead
+/// result in an inference failure.
+///
+/// @description Checks that the omitted return type of a non-local function is
+/// inferred to be `dynamic`.
```
eernstg: I'd expect the assertion to say something about non-local functions, but it doesn't.

Isn't there a location in a specification document where this kind of inference is specified? Perhaps it's all the way back to the rule that a type which isn't written is inferred as `dynamic` (which is superceded by all kinds of inference, but pops up now and then when nothing else is applicable)?

eernstg (commented):
One thing needs to be clarified: How do we know that the return type of a non-local function declaration with no return type annotation is `dynamic`?

I'm afraid it can be tricky to find it, but this just means that it is very useful to resolve that question at this time where we need it.

TypeSystem/inference/local_return_type_inference_A02_t02.dart
sgrekhov: Yes, I meant that rule.  Updated the description accordingly. PTAL.

eernstg (commented):
I'm still confused... ;-)

sgrekhov:
The assertion these tests check is about local functions. The test `local_return_type_inference_A02_t02.dart` checks that this rule does _not_ apply to non-local functions. In particular, if the return type is not specified, then the rule "If a function declaration does not declare a return type explicitly, its return type is dynamic" is applied. I updated the description to reflect this. Maybe it is still not clear enough? Or... Why are you still confused?

eernstg:
The specification documents very rarely specify any properties using an implied negative version of a rule. That is, we may specify that a local function can get its return type from type inference, but the inference will fail if there are any recursive invocations of the function itself in the function body. However, it does _not_ follow that any particular treatment is given to non-local functions (that is, we don't assume that a non-X entity has a not-P property as a consequence of X entities being specified to have the property P).

In contrast, whatever is specified for non-local functions would be specified separately in spec sections about non-local functions (or subsets thereof, e.g., instance methods or top-level functions).

This means that the test TypeSystem/inference/local_return_type_inference_A02_t02.dart might be a very useful and meaningful test to have in some other set of tests (not about local return type inference), but it would be associated with an assertion about non-local functions (or about the relevant subset), not an assertion about local return type inference. And the assertion (and description) would positively say that something holds for non-local functions, it wouldn't be using a double negation to conclude anything (like "not-X entities such as non-local functions have a not-P property such as not being able to obtain a return type by type inference, recursive or not").

sgrekhov:
Now it's clear. Yes, makes sense. Moved this test to the appropriate folder (with the new assertion). PTAL.

eernstg (approved):
LGTM, thanks!

PR #3943, opened 2026-07-07
#3847. Update constant type variable elimination tests

sgrekhov:


LanguageFeatures/nnbd/const_type_var_elimination_A01_t01.dart
```diff
@@ -7,23 +7,24 @@
...
+/// is initialized by `const` literal of type `List<Never>`.
 /// @author iarkh@unipro.ru
 
+import '../../Utils/static_type_helper.dart';
+
 class G<T> {
   void foo() {
-    const List<T> l1 = [];
+    const List<T> l1 = []; // `[]` has type List<Never> but it cannot be checked
+    l1.expectStaticType<Exactly<List<T>>>(); // l1 is still List<T>
```
eernstg, line 29: Why would it not work to query it like `l1 is List<Never>`?

eernstg (commented):
Looks good! It does look like it would be possible to verify that the given list is actually a `List<Never>`.

sgrekhov:
Indeed. Thank you. Updated. PTAL.

eernstg (approved):
LGTM

PR #3944, opened 2026-07-07
#3847. Add least and greatest closure tests. Part 1.

sgrekhov:


TypeSystem/inference/least_and_greatest_closure_A03_t01.dart
```diff
@@ -0,0 +1,27 @@
...
+///   - The greatest closure of `S` is `S`
+///
+/// @description Check that if `S` is a base type then the greatest closure of
+/// `S` is `S`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+test(int x, int y) {
+  var v = 1 > 2 ? x : y;
```
eernstg, line 21: I don't think this situation involves a computation of the greatest or least closure of anything.

However, with `<X extends int>(X x, Never y)` it should work.

TypeSystem/inference/least_and_greatest_closure_A05_t01.dart
```diff
@@ -0,0 +1,32 @@
...
+/// respect to `L` is `U?` where `U` is the greatest closure of `T` with respect
+/// to `L`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+class C<X> {}
+
+test<X extends C<X>, Y extends C<Y>>(X? x, Y? y) {
+  var v = 1 > 2 ? x : y;
```
eernstg, line 25: We're taking the greatest closure of the bound, so we should probably use `<X extends C<X>?>(X x, Never y)`.

TypeSystem/inference/least_and_greatest_closure_A01_t01.dart
```diff
@@ -0,0 +1,29 @@
...
+///
+/// @description Check that if `S` is `X` where `X` is in `L`, then the greatest
+/// closure of `S` with respect to `L` is `Object?`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+class C<X> {}
+
+test<X extends C<X>, Y extends C<Y>>(X x, Y y) {
```
eernstg, line 21: Actually, it should be sufficient to use a single type parameter bound. The other type can be `Never` because this will never make a difference when `UP` is computed.

Hence, `test<X extends C<X>>(X x, Never y)` should allow `UP` to compute the greatest closure of `X` with respect to `{X}`, and then we can directly look at the result because `UP(that-type, Never)` is `that-type`.

This should make all the tests a bit simpler, and it should make them a bit more precisely targeted at their actual goal.

eernstg: Even when the left and right hand types are so different, we don't have to test `x : y` as well as `y : x` because this is a test of 'greatest closure' and 'least closure', not a test of `UP`.

TypeSystem/inference/least_and_greatest_closure_A04_t01.dart
```diff
@@ -0,0 +1,27 @@
...
+///   - The greatest closure of `S` is `S`
+///
+/// @description Check that if `S` is a base type then the least closure of `S`
+/// is `S`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+test(void Function(int) x, void Function(int) y) {
+  var v = 1 > 2 ? x : y;
```
eernstg, line 21: We wish to compute the least closure of `int`, and this would be the result of computing the greatest closure of `void Function(int)` and extracting the parameter type from the result. So `<X extends void Function(int)>(X x, Never y)` should work.

TypeSystem/inference/least_and_greatest_closure_A05_t02.dart
```diff
@@ -0,0 +1,30 @@
...
+///
+/// @description Check that if `S` is `T?` then the greatest closure of `S` with
+/// respect to `L` is `U?` where `U` is the greatest closure of `T` with respect
+/// to `L`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+test<X extends num, Y extends int>(X? x, Y? y) {
+  var v = 1 > 2 ? x : y;
```
eernstg, line 23: `<X extends num?>(X x, Never y)`

TypeSystem/inference/least_and_greatest_closure_A06_t01.dart
```diff
@@ -0,0 +1,32 @@
...
+/// respect to `L` is `U?` where `U` is the least closure of `T` with respect to
+/// `L`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+class C<X> {}
+
+test<X extends C<X>, Y extends C<Y>>(void Function(X?) x, void Function(Y?) y) {
+  var v = 1 > 2 ? x : y;
```
eernstg, line 23: `<X extends void Function(C<X>?)>(X x, Never y)`?

TypeSystem/inference/least_and_greatest_closure_A06_t01.dart
```diff
@@ -0,0 +1,32 @@
...
+  v.expectStaticType<Exactly<void Function(Never?)>>();
```
eernstg, line 24: `Exactly<void Function(C<Never>?)>`?

TypeSystem/inference/least_and_greatest_closure_A06_t02.dart
```diff
@@ -0,0 +1,29 @@
...
+///
+/// @description Check that if `S` is `T?` then the least closure of `S` with
+/// respect to `L` is `U?` where `U` is the least closure of `T` with respect to
+/// `L`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+test(void Function(num?) x, void Function(int?) y) {
+  var v = 1 > 2 ? x : y;
```
eernstg, line 23: `<X extends void Function(num?)>(X x, Never y)`

TypeSystem/inference/least_and_greatest_closure_A06_t02.dart
```diff
@@ -0,0 +1,29 @@
...
+  v.expectStaticType<Exactly<void Function(int?)>>();
```
eernstg: `Exactly<void Function(num?)>`

TypeSystem/inference/least_and_greatest_closure_A02_t01.dart
```diff
@@ -0,0 +1,29 @@
...
+///
+/// @description Check that the least closure of `S` with respect to `L` is
+/// `Never`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+class C<X> {}
+
+test<X extends C<X>, Y extends C<Y>>(void Function(X) x, void Function(Y) y) {
```
eernstg, line 21: `<X extends void Function(C<X>)>(X x, Never y)` should work.

TypeSystem/inference/least_and_greatest_closure_A02_t01.dart
```diff
@@ -0,0 +1,29 @@
...
+  var v = 1 > 2 ? x : y;
+  v.expectStaticType<Exactly<void Function(Never)>>();
```
eernstg, line 23: `Exactly<void Function<C<Never>>>`

eernstg (commented):
A number of things need to be adjusted because we only have a few situations where the greatest closure is used. With the approach that uses some cases in `UP`, we need to make sure that the type which is subject to closure is in the type variable bound, not anywhere else.

TypeSystem/inference/least_and_greatest_closure_A01_t01.dart
sgrekhov: Done.

TypeSystem/inference/least_and_greatest_closure_A02_t01.dart
sgrekhov: Left unchanged, the reason explained in the next comment.

TypeSystem/inference/least_and_greatest_closure_A02_t01.dart
sgrekhov: In fact, in this case we have. 
```dart
test<X extends void Function(C<X>)>(X x, Never y) {
  var v = 1 > 2 ? x : y;
  v.expectStaticType<Exactly<X>>();
}
```
I guess that we should have two function parameters to trigger least closure calculation.

TypeSystem/inference/least_and_greatest_closure_A03_t01.dart
sgrekhov: Yes. thank you. Updated.

TypeSystem/inference/least_and_greatest_closure_A04_t01.dart
sgrekhov: In this case we have:
```dart
test<X extends void Function(int)>(X x, Never y) {
  var v = 1 > 2 ? x : y;
  v.expectStaticType<Exactly<X>>();
}
```
We need two function types. Left unchanged.

TypeSystem/inference/least_and_greatest_closure_A05_t01.dart
sgrekhov: In this case the actual result is
```dart
test<X extends C<X>?>(X x, Never y) {
  var v = 1 > 2 ? x : y;
  v.expectStaticType<Exactly<X>>();
}
```
Does it proof that if `S` is `T?` then the greatest closure is `U?`. Left unchanged for a while.

TypeSystem/inference/least_and_greatest_closure_A05_t02.dart
sgrekhov: The actual result is
```dart
test<X extends num?>(X x, Never y) {
  var v = 1 > 2 ? x : y;
  v.expectStaticType<Exactly<X>>();
}
```
I think that the old version with `v.expectStaticType<Exactly<num?>>();` is more obvoius. Left unchanged.

TypeSystem/inference/least_and_greatest_closure_A06_t01.dart
sgrekhov: Again, I think we need two function types here. The actual result in the comment below.

TypeSystem/inference/least_and_greatest_closure_A06_t01.dart
sgrekhov: The actual result in this case is
```dart
test<X extends void Function(C<X>?)>(X x, Never y) {
  var v = 1 > 2 ? x : y;
  v.expectStaticType<Exactly<X>>();
}
```
Left unchanged.

TypeSystem/inference/least_and_greatest_closure_A06_t02.dart
sgrekhov: Again, see below.

TypeSystem/inference/least_and_greatest_closure_A06_t02.dart
sgrekhov: The actual result
```dart
test<X extends void Function(num?)>(X x, Never y) {
  var v = 1 > 2 ? x : y;
  v.expectStaticType<Exactly<X>>();
}
```
Left unchanged.

sgrekhov (commented):
Partially implemented. I think that we need two function types for the least closure. PTAL.

sgrekhov:
Changes reverted. PTAL.

TypeSystem/inference/least_and_greatest_closure_A01_t01.dart
eernstg: Summary of lessons learned:

Having thought about this some more, it seems necessary to have two type variables with the same structure (one is just an alpha conversion of the other)), such that neither is a subtype of the other. This will force the use of the `UP` rule that includes the computation of the greatest closure, on both sides of `UP(typeOfX, typeOfY)`, which means that we'll compute the `UP(erasedType, erasedType)` which is `erasedType`, which will allow us to inspect the result returned by the greatest closure algorithm.

We can then investigate the result returned by the least closure algorithm on `S` via the greatest closure result on `void Function(S)`.

This result, call it `U`, will then again be produced twice (one time for each type variable), yielding the same result (because the type variables only differ in the choice of name for the type variable which is being eliminated), and `UP(U, U)` yields that same result, which can then be inspected as the type of `v`.

So this should work!

TypeSystem/inference/least_and_greatest_closure_A02_t01.dart
eernstg: Yep, that should work!

TypeSystem/inference/least_and_greatest_closure_A02_t01.dart
eernstg: Exactly.

TypeSystem/inference/least_and_greatest_closure_A03_t01.dart
eernstg: Fine!

TypeSystem/inference/least_and_greatest_closure_A01_t01.dart
eernstg: This thread is left unresolved just because it contains useful analysis of the situation.

TypeSystem/inference/least_and_greatest_closure_A04_t01.dart
eernstg: `UP` is only going to compute the greatest/least closure of anything if we enforce that one of the relevant cases is used. In particular, the case about `X extends B` will compute the greatest closure of `B` if the other argument isn't a subtype or supertype.

So we have to use `test<X extends void Function(int), Y extends void Function(int)>(X x, Y y)` then the static type of the conditional expression will find the greatest closure of `void Function(int)`, twice, take `UP` of those types (that type), yielding `void Function(int)`. So this can be used to confirm that the greatest closure of `void Function(int)` is `void Function(int)`, or it can be used to confirm that the least closure of `int` is `int`.

TypeSystem/inference/least_and_greatest_closure_A05_t01.dart
eernstg: We can test the greatest closure of `X?` by testing `<X extends List<X?>, Y extends List<Y?>>(X x, Y y)`: The conditional type is found by taking the greatest closure of `List<X?>` with respect to `{X}`, which is `List<V>` where `V` is the greatest closure of `X?`.

We need to use `List` (or some other generic class) in order to avoid having a type variable whose bound is trivial (such as `X extends X?`), because those type variables _should be_ an error in the first place, so we don't want them to sit all over the place and explode if and when we do that.

So this one still needs a small adjustment.

TypeSystem/inference/least_and_greatest_closure_A05_t02.dart
eernstg: Needs `<X extends num?, Y extends num?>(X x, Y y)`, then the type of the conditional expression will be the greatest closure of `num?` with respect to `X`.

TypeSystem/inference/least_and_greatest_closure_A06_t01.dart
eernstg: Right, we do need two function types.

eernstg: `<X extends void Function(X?), Y extends void Function(Y?)>(X x, Y y)` should work, and the resulting type should be `void Function(Never?)`.

TypeSystem/inference/least_and_greatest_closure_A06_t02.dart
eernstg: `<X extends void Function(num?), Y extends void Function(num?)>(X x, Y y)` should work, and the resulting type should be `void Function(num?)`.

eernstg (commented):
Yet more comments. ;-)

TypeSystem/inference/least_and_greatest_closure_A05_t01.dart
sgrekhov: Thank you! Updated.

TypeSystem/inference/least_and_greatest_closure_A05_t02.dart
sgrekhov: Updated.

TypeSystem/inference/least_and_greatest_closure_A06_t01.dart
sgrekhov: Updated.

TypeSystem/inference/least_and_greatest_closure_A06_t02.dart
sgrekhov: Changed to `<X extends void Function(num?), Y extends void Function(int?)>(X x, Y y)` and the resulting type `void Function(int?)`.

sgrekhov (commented):
Thank you! This PR was a great learning experience! Updated. PTAL.

TypeSystem/inference/least_and_greatest_closure_A05_t01.dart
eernstg: Thx!

TypeSystem/inference/least_and_greatest_closure_A05_t02.dart
eernstg: I think we should stick to using two type parameters with a bound which are alpha equivalent, such that we are testing the greatest/least closure algorithm, and not a mixture of the greatest/least closure algorithm and the `UP` algorithm. So that's the reason why I suggested `<X extends num?, Y extends num?>`.

It would also be confusing if the README says that the two type parameters should be alpha equivalent, and then some of them aren't, even though we may be able to argue that we can use the result even when they differ more than by renaming.

TypeSystem/inference/least_and_greatest_closure_A06_t01.dart
eernstg: Thanks!

TypeSystem/inference/least_and_greatest_closure_A06_t02.dart
eernstg: Same thing here, anything other than alpha equivalence brings `UP` into the test, which is noise.

eernstg (commented):
I think we should be careful about only using alpha equivalent type parameter declarations, because this is the simplest way to ensure that we're testing greatest/least closure and not `UP` or `DOWN`.

TypeSystem/inference/least_and_greatest_closure_A05_t02.dart
sgrekhov: Updated.

TypeSystem/inference/least_and_greatest_closure_A06_t02.dart
sgrekhov: Done.

sgrekhov (commented):
Thanks! Updated. PTAL.

eernstg (approved):
Thanks! LGTM.

eernstg (commented):
Oh, TypeSystem/inference/least_and_greatest_closure_A04_t01.dart still doesn't have bounded type parameters, just plain parameter types.

sgrekhov:
> Oh, TypeSystem/inference/least_and_greatest_closure_A04_t01.dart still doesn't have bounded type parameters, just plain parameter types.

`least_and_greatest_closure_A04_t01.dart` checks the case when "`S` is a base type (or in general, if it does not contain any variable from `L`)". Does this case also include type variables? What is a "base type" then?

eernstg:
The important requirement is that the given type does not contain any type parameters from `L`. We don't need to clarify what a base type is (it could be exactly that, or it could be any type which is denoted by a single identifier, but in that case it gets weird if we say that `F` is a base type when we have `typedef F = void Function<X>(List<X>);`).

So: The type does not contain any type parameters from `L`.

With that, we can investigate the greatest closure of a type that satisfies this requirement, such as `void Function(int)`, and we can use the outcome to test the least closure (of `int`).

However, we still need the same setup in order to be able to obtain the greatest closure of this "base" type in the first place:

```dart
test<X extends void Function(int), Y extends void Function(int)>(X x, Y y) {
  var v = 1 > 2 ? x : y;
  v.expectStaticType<Exactly<void Function(int)>>;
}
```

So we're using the type variables in order to force `UP` to use the required case (the one that uses the greatest closure algorithm), and this causes the greatest closure algorithm to be executed with input `void Function(int)` (look, no type parameters! ;-), and then we investigate the outcome in order to see that the least closure of `int` was indeed computed as `int`.

sgrekhov:
Thank you for the detailed explanation. Updated. PTAL.

eernstg (approved):
LGTM!

PR #3945, opened 2026-07-09
#3847. Add least and greatest tests for `Future` and `FutureOr`

sgrekhov:


TypeSystem/inference/least_and_greatest_closure_A07_t01.dart
```diff
@@ -0,0 +1,29 @@
...
+/// ...
+/// - if `S` is `Future<T>`
+///   - The least closure of `S` with respect to `L` is `Future<U>` where `U` is
+///     the least closure of `T` with respect to `L`
+///   - The greatest closure of `S` with respect to `L` is `Future<U>` where `U`
+///     is the greatest closure of `T` with respect to `L`
+///
+/// @description Check that if `S` is `Future<T>` then the greatest closure of
+/// `S` with respect to `L` is `Future<U>` where `U` is the greatest closure of
+/// `T` with respect to `L`.
```
eernstg, line 17: This is kind of superfluous because that's the treatment that every generic class gets (there is no reason for the spec to say anything about `Future`). It might have been introduced in order to make the document "closed" in the sense that the rule about `FutureOr` (which _does_ need to be specified explicitly) mentions `Future`, and then we need to cover `Future` as well.

Anyway, we'll just test this because we have this rule, redundant or not.

TypeSystem/inference/least_and_greatest_closure_A07_t02.dart
```diff
@@ -0,0 +1,41 @@
...
+///     is the greatest closure of `T` with respect to `L`
+///
+/// @description Check that if `S` is `Future<T>` then the greatest closure of
+/// `S` with respect to `L` is `Future<U>` where `U` is the greatest closure of
+/// `T` with respect to `L`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+test1<X extends Future<int>, Y extends Future<num>>(X x, Y y) {
```
eernstg: Alpha equivalent:
```suggestion
test1<X extends Future<int>, Y extends Future<int>>(X x, Y y) {
```

eernstg: Same for several other tests in this PR.

TypeSystem/inference/least_and_greatest_closure_A08_t02.dart
```diff
@@ -0,0 +1,62 @@
...
+/// with respect to `L` is `Future<U>` where `U` is the least closure of `T`
+/// with respect to `L`.
+/// @author sgrekhov22@gmail.com
+/// @issue https://github.com/dart-lang/language/issues/4720
+
+import '../../Utils/static_type_helper.dart';
+
+test1<
+  X extends void Function(Future<int>),
+  Y extends void Function(Future<num>)
```
eernstg: Alpha.

TypeSystem/inference/least_and_greatest_closure_A09_t02.dart
```diff
@@ -0,0 +1,43 @@
...
+/// @description Check that if `S` is `FutureOr<T>` then the greatest closure of
+/// `S` with respect to `L` is `FutureOr<U>` where `U` is the greatest closure
+/// of `T` with respect to `L`.
+/// @author sgrekhov22@gmail.com
+
+import 'dart:async';
+
+import '../../Utils/static_type_helper.dart';
+
+test1<X extends FutureOr<int>, Y extends FutureOr<num>>(X x, Y y) {
```
eernstg: Alpha.

TypeSystem/inference/least_and_greatest_closure_A10_t02.dart
```diff
@@ -0,0 +1,64 @@
...
+/// @author sgrekhov22@gmail.com
+/// @issue https://github.com/dart-lang/language/issues/4720
+
+import 'dart:async';
+
+import '../../Utils/static_type_helper.dart';
+
+test1<
+  X extends void Function(FutureOr<int>),
+  Y extends void Function(FutureOr<num>)
```
eernstg: Alpha.

eernstg (commented):
Some of these tests need to be adjusted to ensure that we're working on two alpha equivalent type parameter declarations.

TypeSystem/inference/least_and_greatest_closure_A07_t02.dart
sgrekhov: Thanks. Fixed.

TypeSystem/inference/least_and_greatest_closure_A08_t02.dart
sgrekhov: Fixed.

TypeSystem/inference/least_and_greatest_closure_A09_t02.dart
sgrekhov: Fixed.

TypeSystem/inference/least_and_greatest_closure_A10_t02.dart
sgrekhov: Fixed.

sgrekhov (commented):
Thank you. Updated. PTAL.

eernstg (approved):
LGTM

PR #3946, opened 2026-07-10
#3847. Add least and greatest tests for interface types

sgrekhov:


TypeSystem/inference/least_and_greatest_closure_A11_t02.dart
```diff
@@ -0,0 +1,43 @@
...
+/// @description Check that if `S` is `C<T0, ..., Tk>` then the greatest closure
+/// of `S` with respect to `L` is `C<U0, ..., Uk>` where `Ui` is the greatest
+/// closure of `Ti` with respect to `L`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+class C<X, Y> {}
+
+test1<X extends C<int, List<num>>, Y extends C<num, List<int>>>(X x, Y y) {
```
eernstg: In order to test greatest/least closure and not `UP`/`DOWN`:
```suggestion
test1<X extends C<int, List<num>>, Y extends C<int, List<num>>>(X x, Y y) {
```

eernstg: Similarly for the remaining cases: They should use two alpha equivalent type parameter declarations.

TypeSystem/inference/least_and_greatest_closure_A12_t01.dart
```diff
@@ -0,0 +1,34 @@
...
+import '../../Utils/static_type_helper.dart';
+
+class C<X, Y> {}
+
+test<X extends void Function(C<X, List<X>>), Y extends Function(C<Y, List<Y>>)>(
+  X x,
+  Y y,
+) {
+  var v = 1 > 2 ? x : y;
+  v.expectStaticType<Exactly<void Function(C<Never, List<Never>>)>>();
```
eernstg, line 29: Interesting! The least closure definition _does_ include the variance rule that `DOWN` does not have. So this should indeed work!

TypeSystem/inference/least_and_greatest_closure_A12_t02.dart
```diff
@@ -0,0 +1,73 @@
...
+/// of `Ti` with respect to `L`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+class C<X, Y> {}
+
+test1<
+  X extends void Function(C<num, List<num>>),
+  Y extends Function(C<int, List<int>>)
```
eernstg: These bounds would be OK if we're testing `UP` or `DOWN`, but they're introducing noise into the test if it is intended to test the greatest/least closure. So they should be alpha equivalent (and there's no point in testing that `dynamic` and `void` are mutual subtypes, just use the same return type in both bounds).

eernstg: A similar comment could be made for the remaining tests.

eernstg (commented):
We have again some tests where the type parameter declarations aren't alpha equivalent, even though the description and file name claims that we're testing the greatest/least closure.

TypeSystem/inference/least_and_greatest_closure_A11_t02.dart
sgrekhov: Updated.

TypeSystem/inference/least_and_greatest_closure_A12_t02.dart
sgrekhov: Updated all.

sgrekhov (commented):
Thanks. Done. PTAL.

eernstg (approved):
LGTM

TypeSystem/inference/least_and_greatest_closure_A11_t02.dart
eernstg: Thanks!

eernstg (approved):
LGTM

PR #3947, opened 2026-07-13
#3847. Add README.md

sgrekhov:
I'll add links to this readme file later in a separate PR

### DIFF
--- a/TypeSystem/inference/README.md
+++ b/TypeSystem/inference/README.md
@@ -1,0 +1,28 @@
+Some tests in this directory (in particular, `least_and_greatest_closure_A*.dart`) 
+check the least and greatest closure of a type. To do this, we examine the results 
+of the [UP](https://github.com/dart-lang/language/blob/main/resources/type-system/upper-lower-bounds.md#upper-bounds) 
+and [DOWN](https://github.com/dart-lang/language/blob/main/resources/type-system/upper-lower-bounds.md#lower-bounds) 
+functions. However, we need to be careful because these functions may return a type 
+that was not produced by the least/greatest closure computation itself.
+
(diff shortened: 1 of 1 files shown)


TypeSystem/inference/README.md
```diff
@@ -0,0 +1,28 @@
+Some tests in this directory (in particular, `least_and_greatest_closure_A*.dart`) 
+check the least and greatest closure of a type. To do this, we examine the results 
+of the [UP](https://github.com/dart-lang/language/blob/main/resources/type-system/upper-lower-bounds.md#upper-bounds) 
+and [DOWN](https://github.com/dart-lang/language/blob/main/resources/type-system/upper-lower-bounds.md#lower-bounds) 
+functions. However, we need to be careful because these functions may return a type 
+that was not produced by the least/greatest closure computation itself.
+
+The least and greatest closure of type is [defined](https://github.com/dart-lang/language/blob/main/resources/type-system/inference.md#type-variable-elimination-least-and-greatest-closure-of-a-type)
+with respect to a set of type variables. Therefore, each test that verifies this 
+behavior must involve at least two type variables. However, the definitions of
```
eernstg: I think it's safer to just say this:
```suggestion
with respect to a set of type variables. The definitions of 
```

The point is that the need for two type parameters doesn't arise because greatest/least closure is defined with respect to a set of type parameters. We're actually using a singleton set in every one of these tests.

The actual reason why we need two type parameters is described very well further down in this file: We need to avoid the situation where `UP` uses one of the subtype rules, so we must ensure that the type of `x` and the type of `y` are incomparable (neither is a subtype of the other), which is true when both the type of `x` and the type of `y` is a type parameter.

eernstg (commented):
Looks great! Just one sentence is slightly misleading.

sgrekhov:
Thanks! Updated. PTAL.

eernstg (approved):
LGTM

PR #3948, opened 2026-07-13
#3847. Add tests for generic function literals

sgrekhov:


TypeSystem/inference/function_literal_inference_A01_t05.dart
```diff
@@ -0,0 +1,112 @@
...
+/// type `T` where `T` is derived from `K` as follows. If the greatest closure
+/// of `K` is `S` and `S` is a subtype of `Null`, then without null safety `T`
+/// is `dynamic`, and with null safety `T` is `Object?`. Otherwise, `T` is `S`.
+/// If there is no corresponding parameter in the context type schema, the
+/// variable is treated as having type `dynamic`.
+///
+/// @description Check that if there is no corresponding parameter in the
+/// context type schema, the variable is treated as having type `dynamic`. Test
+/// the case when the type parameter of a function literal is eliminated by a
+/// generic function instantiation algorithm.
```
eernstg: This will do:
```suggestion
/// generic function instantiation.
```
We don't care how the generic function instantiation is performed, we just know that the initializing expression is a generic function and the context type is non-generic, and this implies that nothing in the two function types is "corresponding", so there is no corresponding parameter no matter how the parameters are declared.

(If we wish to talk about the generic function instantiation algorithm in some other context then we'd probably say _the_ generic function instantiation algorithm because we only have one.)

eernstg: I think it would make sense to have another test very similar to this one where there is no corresponding parameter for some other reason, e.g.,

```dart
void Function() f1 = ([v]) {}; // `v` has type `dynamic`.
```

TypeSystem/inference/function_literal_inference_A02_t06.dart
```diff
@@ -0,0 +1,140 @@
...
+///   enclosing function is `async`, or `UP(S, T)` otherwise.
+///
+/// @description Check that the return type of an async function with a single
+/// `return e;` statement is `UP(flatten(S), Never)`, where `S` is inferred type
+/// of `e`. Test generic type literals.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+var f1 = <X>(X x) async {
```
eernstg: I think it would be useful to test a case where `flatten(S)` isn't just `S` every single time.

Perhaps: `var f1 = <X>(Future<X> x) async { return x; }`.

However, this would be the 'actual return type', which is different from the inferred return type of the function literal (we're adding `Future` and other top-level constructs to get the return type `Future<S>` from the actual return type `S`).

That shouldn't be a problem though, because we can still check that the return type is `Future<X>` for the generic function literal `<X>(Future<X> x) async { return x; }`, which is obtained by finding the actual returned type `X` and then wrapping it in a `Future`.

TypeSystem/inference/function_literal_inference_A02_t08.dart
```diff
@@ -0,0 +1,134 @@
...
+///   enclosing function is `async`, or `UP(S, T)` otherwise.
+///
+/// @description Check that the return type of an async function with a single
+/// `return e;` statement is `UP(flatten(S), Never)`, where `S` is inferred
+/// using the local type inference algorithm. Test generic function literals.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+var f1 = <T>(T t) async {
```
eernstg: Again, it would be useful to return an expression whose static type isn't mapped to itself by `flatten`, such that we can actually get a test failure if, for example, the implementation entirely forgets to apply `flatten`.

We could use `var f1 = <T>(Future<T> t) async {...}`. We could also try other types (`<T>(FutureOr<T> t) ...`, for example), but that would amount to a more detailed test of `flatten`, and that should be covered by some existing tests already. So `Future<T>` is fine.

TypeSystem/inference/function_literal_inference_A02_t08.dart
```diff
@@ -0,0 +1,134 @@
...
+  return 2 > 1 ? t : (1 as num?);
```
eernstg: We could use `Future<num?>.value(1)` in order to have `Future<num?>` which is then reduced to `num?` by `flatten`, which would yield the inferred return type `Future<num?>`.

TypeSystem/inference/function_literal_inference_A02_t07.dart
```diff
@@ -0,0 +1,133 @@
...
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+var f1 = <T>(T t) {
+  return 2 > 1 ? t : (1 as num?);
+};
+var f2 = <T extends num>(T t) {
+  return 2 > 1 ? t : (1 as num?);
+};
```
eernstg, line 30: Nice!

We get to see that `UP()` is used in two distinct ways (one where the operands are unrelated, and the other one where one is a subtype of the other), which is enough to support the assumption that we're actually computing the given `S` as described.

eernstg (commented):
Looks good, with a couple of comments!

In particular, some of the tests are verifying that a particular type (the actual return type) is obtained using `flatten`. However, in all the cases tested, `flatten(S)` will just return `S`, so we wouldn't detect a bug where the implementation simply forgot to call `flatten`.

TypeSystem/inference/function_literal_inference_A01_t05.dart
sgrekhov: https://github.com/dart-lang/co19/blob/master/TypeSystem/inference/function_literal_inference_A01_t02.dart checks this case.

TypeSystem/inference/function_literal_inference_A02_t06.dart
sgrekhov: Thanks. Updated.

TypeSystem/inference/function_literal_inference_A02_t08.dart
sgrekhov: Done.

sgrekhov (commented):
Thank you. Updated. PTAL.

TypeSystem/inference/function_literal_inference_A01_t05.dart
eernstg: Thanks!

TypeSystem/inference/function_literal_inference_A02_t06.dart
eernstg: Thanks!

eernstg (approved):
LGTM

PR #4138, opened 2026-07-24
#3847. Add more test for least and greatest closure

sgrekhov:


TypeSystem/inference/least_and_greatest_closure_A18_t01.dart
```diff
@@ -0,0 +1,48 @@
...
+test2<
+  X extends void Function(
+    void Function<T1 extends X, T2 extends num>(int, [num]),
+  ),
+  Y extends void Function(
+    void Function<T1 extends Y, T2 extends num>(int, [num]),
+  )
+>(X x, Y y) {
+  var v = 1 > 2 ? x : y;
+  v.expectStaticType<Exactly<Function(Never)>>();
```
eernstg: We might as well write the return type explicitly:
```suggestion
  v.expectStaticType<Exactly<void Function(Never)>>();
```
It should work because the return type is otherwise implicitly `dynamic`, but (as usual) this test isn't intended to test the behavior of function types with no explicit return type.

eernstg: Same in line 30, and in A18_t02.

eernstg (approved):
LGTM!

I had a single comment, but I just made the change.

sgrekhov:
Thank you!

PR #4154, opened 2026-07-27
#3847. Add test for upper bound

sgrekhov:
@eernstg, this is a test for the upper bound rules described in https://github.com/dart-lang/language/blob/main/resources/type-system/inference.md#upper-bound. If you have any advice on how to add more tests for this part of the specification, it would be greatly appreciated.

### DIFF
--- a/TypeSystem/inference/upper_bound_A01_t01.dart
+++ b/TypeSystem/inference/upper_bound_A01_t01.dart
@@ -14,6 +14,9 @@
 ///
 /// @description Check that the subtyping algorithm takes the least/greatest
 /// closure into account.
+/// @note We can't systematically test `UP` and `DOWN` on type schemas, but here
+/// is an example from the feature specification that illustrates that it can
+/// occur.
 /// @author sgrekhov22@gmail.com
 /// @issue 63883
 
--- a/TypeSystem/inference/upper_bound_A01_t01.dart
+++ b/TypeSystem/inference/upper_bound_A01_t01.dart
@@ -1,0 +1,32 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We write `UP(T0, T1)` for the upper bound of `T0` and `T1` and
+/// `DOWN(T0, T1)` for the lower bound of `T0` and `T1`. This extends to type
+/// schema as follows:
+/// - We add the axiom that `UP(T, _) == T` and the symmetric version.
+/// - We replace all uses of `T1 <: T2` in the `UP` algorithm by `S1 <: S2`
+///   where `Si` is the least closure of `Ti` with respect to `_`.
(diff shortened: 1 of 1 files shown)


eernstg (commented):
I just discussed this with @chloestefantsova, confirming that there is no realistic way to tailor a test such that it is testing the non-grounded versions of the constraint solution algorithms. This implies that we can't expect to observe the handling of type schemas during type inference, we can only inspect that the types (not schemas) which are produced as the output from the type inference algorithm are as expected.

In particular, we don't need to test the generalization of `UP` or `DOWN` to type schemas, because we can't. ;-)

It can be tested by the tool specific whitebox tests where they can do arbitrary things to the analyzer or CFE in order to observe the execution of internal algorithms, but it can't actually be done as a Dart language test.

So I think we should just drop this particular test and this particular testing topic.

sgrekhov:
@eernstg  this test is taken from the [specification](https://github.com/dart-lang/language/blob/main/resources/type-system/inference.md#upper-bound) and according to it "...illustrates the effect of taking the least/greatest closure in the subtyping algorithm". I'd like to preserve it, because it allowed to find https://github.com/dart-lang/sdk/issues/63883. Maybe it could be used as a test for another assertion - perhaps as an additional test for the `UP` function? If so, I'll move it to my set of tests for the `UP` function (which I haven't committed yet).

eernstg:
OK, presumably @leafpetersen found a way to squeeze some information about the behavior of `UP` and `DOWN` on type schemas out of an actual program. Impressive! ;-)

The inference of `check(C((List<int> x) {}))` has context type schema `_`, so there are no downwards inference results, it is all obtained during upwards inference. For that, we need to ensure that `X` is chosen such that `(List<int> x) {}` is assignable to a parameter of type `void Function(X)`, that is, `void Function(List<int>) <: void Function(X)`, that is, `X <: List<int>`. Next, `C((List<int> x) {})` must be assignable to `C<List<T>>`, that is, `C<X> <: C<List<T>>`, that is, `X <: List<T>`.

However, `T` is not being inferred during inference of `C((List<int> x) {})`, so it is represented as the unknown type `_`. We compute the merged constraint by taking the least closure of `List<_>` and `List<int>` which is `List<int>`, so `X` is bound to `List<int>`.

Then the type of `C<List<int>>((List<int> x) {})` can be determined: `C<List<int>>`.

This yields the constraint `C<List<T>> <: C<List<int>>`, that is, `List<T> <: List<int>`, that is, `T <: int`, which is then the result because we have no other constraints (and we don't choose `_ <: T <: int` because `_` isn't known).

So we _do_ use `DOWN` in a situation where one of the operands is a type schema. However, I also think it illustrates that it's being tested in a very indirect manner, and it doesn't generalize to a systematic test of `UP` and `DOWN` on type schemas.

So let's keep this one, preferably with a comment that says something along the lines of "We can't systematically test `UP` and `DOWN` on type schemas, but here is an example from the feature specification that illustrates that it can occur."

sgrekhov:
Thank you! The comment has been added.

eernstg (commented):
LGTM, thanks!

eernstg (approved):

PR #4164, opened 2026-07-28
#3847. Add test for the closure of type constraints

sgrekhov:


TypeSystem/inference/closure_of_type_constraints_A01_t01.dart
```diff
@@ -0,0 +1,43 @@
...
+/// @description Check that the closure of a type constraint `Pb <: X <: Pt`
+/// with respect to a set of type variables `L` is the subtype constraint
+/// `Qb <: X <: Qt` where `Qb` is the greatest closure of `Pb` with respect to
+/// `L`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+class C<T> {
+  C(T Function<X>(X x));
```
eernstg: Should probably be a regular declaration of a formal parameter whose type is a function type:
```suggestion
  C(T Function<X>(X) _);
```
The name of the parameter _can_ be `Function`, but it would not be a reader-friendly choice.

TypeSystem/inference/closure_of_type_constraints_A01_t02.dart
```diff
@@ -0,0 +1,42 @@
...
+///
+/// @description Check that the closure of a type constraint `Pb <: X <: Pt`
+/// with respect to a set of type variables `L` is the subtype constraint
+/// `Qb <: X <: Qt` where `Qt` is the least closure of `Pt` with respect to `L`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+class C<T> {
+  C(List<X> Function<X>(T x));
```
eernstg:
```suggestion
  C(List<X> Function<X>(T) _);
```

eernstg (commented):
Looks good! Just one suggestion: We should probably not have a parameter whose name is `Function`.

sgrekhov:
Thanks. Updated. PTAL

eernstg (approved):
LGTM

PR #4172, opened 2026-07-29
#3847. Add tests for the constraint solving

sgrekhov:


TypeSystem/inference/constraint_solving_A02_t03.dart
```diff
@@ -0,0 +1,28 @@
...
+/// @description Check that if `Mb` is known and `Mt` is unknown then the
+/// solution is `Mb`.
+/// @author sgrekhov22@gmail.com
+
+import '../../Utils/static_type_helper.dart';
+
+X f<X extends num>(X x) => x;
+
+void main() {
+  f(1 as dynamic).expectStaticType<Exactly<num>>(); // num <: X <: _
```
eernstg: I think the constraint would be `dynamic <:  X <: _` because the static type of `1 as dynamic` is `dynamic`.

You could say that a `dynamic` expression is assignable to any type, but that would justify many possible inferred types like (`Never`, `double`, `extension type ET(int i) implements int;` and whatnot). I think inference will use the actual type even when the argument is `dynamic`.

Checking https://github.com/dart-lang/language/blob/master/resources/type-system/inference.md#rules: That is true, there is no special rule for `dynamic <: X <: [Whatever]`, it is just treated like any other type `T` as `T <: X <: [Whatever]` where the requirement is subtyping, not assignability.

eernstg (commented):
Do we have a case like this one coming in some other PR?

```dart
List<X> f<X extends num>() => <Never>[];

void main() {
  f().expectStaticType<Exactly<List<num>>>; // _ <: X <: num
}
```
Otherwise I think we should have it.

sgrekhov:
Thank you. Added the new test case (thank you for the advise) and the comment updated. PTAL.

eernstg (approved):
Thanks, LGTM

Issue #4173, opened 2026-07-30
Add tests for `UP` and `DOWN` functions


sgrekhov:
Specification https://github.com/dart-lang/language/blob/main/resources/type-system/upper-lower-bounds.md#upper-bounds

PR #4178, opened 2026-08-03
#4173. Add tests checking UP(`T`, `T`) = `T`

sgrekhov:


eernstg (approved):
LGTM!

eernstg:
These test cases are actually more demanding than they look at first: The implementations do not apply the reflexivity rule directly on any other types than the atomic ones (class types and other types like `Record` or `void` that don't have internal structure). So we're actually checking that the structural traversal applied to non-atomic types all arrive at the same result as a direct application of `UP(T, T) = T` would yield.

PR #4182, opened 2026-08-05
#4173. Add tests for UP function with one top type, part 2

sgrekhov:


TypeSystem/upper-lower-bounds/up_A04_t01.dart
```diff
@@ -0,0 +1,287 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `T2`) = `T2` if TOP(`T2`)
+///
+/// @description Check that UP(`T1`, `T2`) = `T2` if TOP(`T2`). Test `void`.
```
eernstg: I think we'd need to make one thing explicit here. The rules in the definition of `UP` are ordered, and it is part of the definition that the rule which should be applied in any given situation is the _first_ one in the definition which is applicable.

This means that we have to look at every earlier rule for every test which is claimed to use any particular rule, in order to ensure that it isn't one of those cases where we must use one of those earlier rules.

I think we've had a couple of cases already where it was demonstrated that this particular check is tricky and must be performed carefully. ;-)

Here's a list of criteria to rule out the application of the rules (please double check that they are both necessary and sufficient!):

- **UP**(`T`, `T`): The operands are not the same type (strictly, not even normalized).
- **UP**(`T1`, `T2`) where **TOP**(`T1`) and **TOP**(`T2`): An operand is not **TOP**.
- **UP**(`T1`, `T2`) where **TOP**(`T1`): The first operand is not **TOP**.
- **UP**(`T1`, `T2`) where **TOP**(`T2`): The second operand is not **TOP**.
- **UP**(`T1`, `T2`) where **BOTTOM**(`T1`) and **BOTTOM**(`T2`): An operand is not **BOTTOM**.
- **UP**(`T1`, `T2`) where **BOTTOM**(`T1`): The first operand is not **BOTTOM**.
- **UP**(`T1`, `T2`) where **BOTTOM**(`T2`): The second operand is not **BOTTOM**.
- **UP**(`X1 & B1`, `T2`): The first operand is not an intersection type.
- **UP**(`T1`, `X2 & B2`): The second operand is not an intersection type.
- **UP**(`T1`, `T2`) where **NULL**(`T1`) and **NULL**(`T2`): An operand is not **NULL**.
- **UP**(`T1`, `T2`) where **NULL**(`T1`): The first operand is not **NULL**.
- **UP**(`T1`, `T2`) where **NULL**(`T2`): The second operand is not **NULL**.
- **UP**(`T1`, `T2`) where **OBJECT**(`T1`) and **OBJECT**(`T2`): An operand is not **OBJECT**.
- **UP**(`T1`, `T2`) where **OBJECT**(`T1`): The first operand is not **OBJECT**.
- **UP**(`T1`, `T2`) where **OBJECT**(`T2`): The second operand is not **OBJECT**.
- **AT THIS POINT** all cases above can be ruled out by checking: The operands are not the same type and none of them is **TOP**, **BOTTOM**, **NULL**, **OBJECT**, or an intersection type.
- **UP**(`T1?`, `T2?`): An operand is not of the form `U?`.
- **UP**(`T1?`, `T2`): The first operand is not of the form `U?`.
- **UP**(`T1`, `T2?`): The second operand is not of the form `U?`.
- **UP**(`X1 extends B1`, `T2`): The first operand is not a type parameter.
- **UP**(`T1`, `X2 extends B2`): The second operand is not a type parameter.
- **UP**(`T Function<...>(...)`, `Function`): The first operand is not a function type or the second operand is not `Function`.
- **UP**(`Function`, `T Function<...>(...)`): The first operand is not `Function` or the second operand is not a function type.
- **UP**(`T0 Function<X0 extends B00, ... Xm extends B0m>(P00, ... P0k)`,
         `T1 Function<X0 extends B10, ... Xm extends B1m>(P10, ... P1l)`): An operand is not a positional-parameter-only function type.
- **UP**(`T0 Function<X0 extends B00, ... Xm extends B0m>(P00, ... P0k, Named0)`,
         `T1 Function<X0 extends B10, ... Xm extends B1m>(P10, ... P1k, Named1)`): An operand is not a named-parameter function type.
- **UP**(`T Function<...>(...)`, `S Function<...>(...)`): An operand is not a function type.
- **UP**(`T Function<...>(...)`, `T2`): The first operand is not a function type.
- **UP**(`T1`, `T Function<...>(...)`): The second operand is not a function type.
- **UP**(`(...)`, `Record`): The first operand is not a record type, or the second operand is not `Record`.
- **UP**(`Record`, `(...)`): The first operand is not `Record`, or the second operand is not a record type.
- **UP**(`(S0, ... Sk, {T0 d0, ..., Tn dn})`,
         `(S0', ... Sk', {T0' d0, ..., Tn' dn})`): An operand is not a record type, or the operands are record types with a different shape.
- **UP**(`(...)`, `(...)`): An operand is not a record type..
- **UP**(`(...)`, `T2`): The first operand is not a record type.
- **UP**(`T1`, `(...)`): The second operand is not a record type.
- **UP**(`FutureOr<T1>`, `FutureOr<T2>`): An operand is not `FutureOr<...>`.
- **UP**(`Future<T1>`, `FutureOr<T2>`): The first operand is not `Future<...>`, or the second operand is not `FutureOr<...>`.
- **UP**(`FutureOr<T1>`, `Future<T2>`): The first operand is not `FutureOr<...>`, or the second operand is not `Future<...>`.
- **UP**(`T1`, `FutureOr<T2>`): The second operand is not `FutureOr<...>`.
- **UP**(`FutureOr<T1>`, `T2`): The first operand is not `FutureOr<...>`.
- **UP**(`T1`, `T2`) where `T1` <: `T2`: The first operand is not a subtype of the second operand.
- **UP**(`T1`, `T2`) where `T2` <: `T1`: The second operand is not a subtype of the first operand.
- **UP**(`C<T0, ..., Tn>`, `C<S0, ..., Sn>`): An operand is not an interface type, or the operands are not instances of the same generic type (including the non-generic case).
- **UP**(`C0<T0, ..., Tn>`, `C1<S0, ..., Sk>`): Last case, nothing to exclude.

TypeSystem/upper-lower-bounds/up_A04_t02.dart
```diff
@@ -0,0 +1,318 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `T2`) = `T2` if TOP(`T2`)
+///
+/// @description Check that UP(`T1`, `T2`) = `T2` if TOP(`T2`).
```
eernstg: Perhaps:
```suggestion
/// @description Check that UP(`T1`, `T2`) = `T2` if TOP(`T2`). Test `Object?`.
```

TypeSystem/upper-lower-bounds/up_A04_t03.dart
```diff
@@ -0,0 +1,137 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `T2`) = `T2` if TOP(`T2`)
+///
+/// @description Check that UP(`T1`, `T2`) = `T2` if TOP(`T2`).
```
eernstg:
```suggestion
/// @description Check that UP(`T1`, `T2`) = `T2` if TOP(`T2`). Test `dynamic`.
```

TypeSystem/upper-lower-bounds/up_A04_t04.dart
```diff
@@ -0,0 +1,154 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `T2`) = `T2` if TOP(`T2`)
+///
+/// @description Check that UP(`T1`, `T2`) = `T2` if TOP(`T2`).
```
eernstg:
```suggestion
/// @description Check that UP(`T1`, `T2`) = `T2` if TOP(`T2`). Test `FutureOr<dynamic>`.
```

eernstg (commented):
I created a list of criteria which can be used to systematically rule out the ability to use each of the cases of the **UP** function (it's in the first comment).

Luckily, it is often enough to determine that one of the operands has a specific property (e.g., the first operand is not **TOP**), and the other operand can be any type whatsoever. It is often the case that a later criterion implies an earlier one (e.g., "first operand is not TOP" => "an operand is not TOP"), so if we've confirmed the former then we have the latter for free.

Also luckily, we can rule out the 4 first rules (when we get to rule 5) by confirming that no operand is **TOP** and the operands are distinct types. Similarly for **TOP** and **BOTTOM** when we get to rule 8, and so on. I added an item **AT THIS POINT** to make this point later on when we have covered all the extreme cases (**TOP**, **BOTTOM**, and the "nearly so" cases **OBJECT** and **NULL**).

I think it would be useful to turn this into an *.md file (perhaps 'README.md' can be used) in the directory where these tests are stored. For each of the situations where we claim to be testing a specific case `N` in the definition of **UP** it should be possible to take this list of criteria and confirm for every case before `N` that it is not applicable. Only then can we assume that we're actually testing case `N`, which is confirmed by checking that all requirements for `N` are satisfied by the given code.

I've confirmed that the tests do actually test the case that they claim to test (e.g., looking at each case and seeing that the operands are not the same type, then looking at each case and seeing that the first operand is not **TOP**, which implies that "an operand is not **TOP**").

Let's create that *.md file, possibly with some extra **AT THIS POINT** bullets if they are helpful, and then hope that we don't actually drown in cases as we proceed. ;-)

sgrekhov:
Thank you. Makes sense. I added some more criteria in this PR and will add the `README.md` file in a separate PR. PTAL.

TypeSystem/upper-lower-bounds/up_A02_t01.dart
```diff
@@ -9,8 +9,9 @@
 ///   - `T1` if MORETOP(`T1`, `T2`)
 ///   - `T2` otherwise
 ///
-/// @description Check that UP(`T1`, `T2`) = `T1` if TOP(`T1`) and TOP(`T2`) and
-/// MORETOP(`T1`, `T2`). Test that `void` is more top type than others.
+/// @description Check that UP(`T1`, `T2`) = `T1` if `T1 != T2` and TOP(`T1`)
+/// and TOP(`T2`) and MORETOP(`T1`, `T2`). Test that `void` is more top type
+/// than others.
```
eernstg, line 14: Nice!

I'm afraid this can be verbose when we get to rule 12 or so, but we can use the shortcuts (relying on "AT THIS POINT" rules) to keep it a bit less verbose, and it does actually describe some requirements that we _must_ keep in mind in order to test **UP** correctly.

TypeSystem/upper-lower-bounds/up_A03_t01.dart
```diff
@@ -7,7 +7,8 @@
 /// ...
 /// - UP(`T1`, `T2`) = `T1` if TOP(`T1`)
 ///
-/// @description Check that UP(`T1`, `T2`) = `T1` if TOP(`T1`). Test `void`.
+/// @description Check that UP(`T1`, `T2`) = `T1` if TOP(`T1`) and
+/// TOP(`T2`) == `false`. Test `void`.
```
eernstg: This rule also requires `T1 != T2` in order to be applicable, but **TOP**(`T1`) but not **TOP**(`T2`) _implies_ that `T1 != T2`. So perhaps this can be expressed 

```suggestion
/// TOP(`T2`) == `false` (which implies `T1 != T2`). Test `void`.
```

eernstg: The same comment is applicable to other tests of the same rule.

eernstg (commented):
Adding the "rule out earlier cases" document in a later PR is fine.

I'm trying to understand how it can be ensured that we don't end up with any tests of a rule "**UP**(`...`) ... = ..." which is really a test of an earlier rule (which might yield a different result). I think it's a very useful idea to express all the requirements in the test (as in the updated description in up_A02_t01.dart), and I've suggested some adjustments in order to have a similar level of explicitness for 'up_A03_t01.dart' and the like.

sgrekhov:
Updated. While we're still at the beginning of the list, it makes sense to state all the conditions explicitly for better understanding. Further down the list, that may become more difficult :) PTAL.

eernstg (approved):
LGTM!

PR #4184, opened 2026-08-07
Fixes #4183. Update expected results in reachability_for_in_A01_t0*.dart

### LINKED ISSUE #4183: Update tests co19/TypeSystem/flow-analysis/reachability_for_in_A01_t0{7-9} to reflect https://dart-review.googlesource.com/c/sdk/+/531160
The following tests currently fail on the CFE. After https://dart-review.googlesource.com/c/sdk/+/531160 is landed, they will fail on the analyzer too:
- co19/TypeSystem/flow-analysis/reachability_for_in_A01_t07
- co19/TypeSystem/flow-analysis/reachability_for_in_A01_t08
- co19/TypeSystem/flow-analysis/reachability_for_in_A01_t09

Each test needs adjusting for a slightly different reason:
- reachability_for_in_A01_t07 assumes that if the loop variable has type `Never`, then the iterable expression will not be reached. This is not true. For example, this code:
  ```dart
  main() {
    int i = 0;
    try {
      for (Never n in <dynamic>[i = 1]) {
        print('loop body');
      }
    } catch (_) {
      print('i=$i');
    }
  }
  ```
  prints `i=1`.
- reachability_for_in_A01_t08 assumes that if the loop variable has type `Never`, then the body of the loop will not be reached. Although this is true, the current flow analysis rules do not detect this particular kind of unreachability. (The fact that the analyzer currently passes this test is a consequence of https://github.com/dart-lang/sdk/issues/60424.)
- reachability_for_in_A01_t09 assumes that if the loop variable has type `Never`, then code that follows the loop will not be reached. This is not true. For example, this code:
  ```dart
  main() {
    for (Never n in []) {
      print('loop body');
    }
    print('after loop');
  }
  ```
  prints `after loop`.

See also https://github.com/dart-lang/sdk/issues/60394.

sgrekhov:


TypeSystem/flow-analysis/reachability_for_in_A01_t08.dart
```diff
@@ -30,10 +30,7 @@ test2(Never n) {
       i = 42;
     }
   }
-  i; // Definitely unassigned.
-//^
-// [analyzer] unspecified
-// [cfe] unspecified
+  i; // Possibly assigned. https://github.com/dart-lang/sdk/issues/60424
```
eernstg: I'm not quite sure how the entries in `@issue` are chosen, but wouldn't it be more consistent to add 60424 to `@issue` and omit these two URIs in a code comment?

eernstg (commented):
Looks good, just a tiny nit question.

TypeSystem/flow-analysis/reachability_for_in_A01_t08.dart
```diff
@@ -9,7 +9,7 @@
 /// - Let `after(N) = join(before(S), break(S))`
 ///
 /// @description Checks that if `X` has type `Never` then an assignment in `S`
-/// is unreachable.
+/// is not unreachable.
```
stereotype441: Consider expanding this comment to clarify that while the code _is_ in fact unreachable, the flow analysis specification does detect that it is unreachable. (What we want to test is that the flow analysis algorithm follows its spec, not that it detects unreachable code in all circumstances, which would of course be [impossible](https://en.wikipedia.org/wiki/Halting_problem#Undecidability) 😃).

For instance, you might say:
```dart
/// @description Checks that if `X` has type `Never` then an assignment in `S`
/// is not considered unreachable. (Technically it _is_ unreachable, but flow
/// analysis does not detect unrechability in this scenario.)
```

stereotype441 (approved):
Lgtm. Your call whether to accept my suggestion below.

TypeSystem/flow-analysis/reachability_for_in_A01_t08.dart
eernstg: Oh, yes, I'd support a clarification at this point as well!

sgrekhov:
Thanks! Updated. PTAL.

eernstg (approved):
Thanks, LGTM!

PR #4189, opened 2026-08-11
#4173. Add tests for bottom types

sgrekhov:


TypeSystem/upper-lower-bounds/up_A05_t01.dart
```diff
@@ -0,0 +1,34 @@
...
+// ignore_for_file: dead_code, unused_local_variable
+
+void f1<X extends Never>(Never n1, X n2) {
+  var v = (1 > 2) ? n1 : n2; // MOREBOTTOM(Never, X) = true
+  // It's not possible to check that `v` is `X` rather than `Never`. We just
+  // check that it is subtype of `Never` (i.e., it is assignable to both
+  // `String` and `int` and is not `dynamic`).
+  int i = v;
+  String s = v;
+  var v2 = (1 > 2) ? v : 42 as Object?;
```
eernstg, line 28: Nice! However, you have to look closely to see why `v2` is needed.

It would probably be helpful to have a comment here saying that `v2` has type `dynamic` if `v` has type `dynamic`, and it has type `Object?` if `v` has any non-top type (including the bottom type that we expect it to have).

TypeSystem/upper-lower-bounds/up_A05_t01.dart
```diff
@@ -0,0 +1,34 @@
...
+/// @description Check that UP(`T1`, `T2`) = `T2` if BOTTOM(`T1`) and
+/// BOTTOM(`T2`) and MOREBOTTOM(`T1`, `T2`).
+/// @author sgrekhov22@gmail.com
+
+// ignore_for_file: dead_code, unused_local_variable
+
+void f1<X extends Never>(Never n1, X n2) {
+  var v = (1 > 2) ? n1 : n2; // MOREBOTTOM(Never, X) = true
+  // It's not possible to check that `v` is `X` rather than `Never`. We just
+  // check that it is subtype of `Never` (i.e., it is assignable to both
```
eernstg: Typo:
```suggestion
  // check that it is a subtype of `Never` (i.e., it is assignable to both
```
Same issue below in at least one test.

TypeSystem/upper-lower-bounds/README.md
```diff
@@ -0,0 +1,151 @@
+We define the upper bound of two types T1 and T2 to be **UP**(`T1`,`T2`) as follows.
```
eernstg: Oh, was this the README file mentioned in [this comment](https://github.com/dart-lang/co19/pull/4182#issuecomment-5192236680)?

I don't think it's useful to repeat the definition of `UP`. The part which does carry new information is the set of minimal criteria which are sufficient to determine that any particular case is guaranteed to _not_ be applicable.

We need this because it is necessary in order to recognize, for example, that **UP**(`Never`, `Never`) is _not_ defined by an application of the rule '**UP**(`T1`, `T2`) where **BOTTOM**(`T1`) and **BOTTOM**(`T2`)'. It is instead defined by an application of the first rule, '**UP**(`T`, `T`)'.

I created a proposal for a set of criteria which would be usable in this manner, [here](https://github.com/dart-lang/co19/pull/4182#discussion_r3720025319).

However, I think it should be expanded a bit in order to be as useful as possible: I added an entry marked **AT THIS POINT** which shows a criterion which is sufficient to rule out an application of all the rules up to that point.

So when we're testing the rule just after **AT THIS POINT** it is sufficient to check that each test case satisfies the **AT THIS POINT** criteria, which is considerably simpler than checking every single rule which occurs before **AT THIS POINT**.

There can be more simplifications, though. For example, we could add the following after the 4 first rules:

> - **AT THIS POINT* all cases above can be ruled out by checking: The operands are not the same type, and none of them is TOP.

This should make it much more convenient to ensure that we don't have any test cases which are intended to test some specific rule (say, rule 11, whatever that is) when it is actually defined by an earlier rule (say, rule 6). This is important because an application of rule 11 might succeed locally (that is, we're satisfying all requirements in the rule itself), but the result obtained from rule 6 might be different than the result obtained from rule 11, and it's the result from the earlier rule which is valid.

Oh, and it would be very useful to have 4 eyes rather than two eyes checking that the criteria are all correct (that is, each rule definitely does not apply if the criterion is satisfied, and also: each rule definitely does apply if the criterion is not satisfied).

TypeSystem/upper-lower-bounds/up_A05_t01.dart
```diff
@@ -0,0 +1,35 @@
...
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `T2`) where BOTTOM(`T1`) and BOTTOM(`T2`) =
+///   - `T2` if MOREBOTTOM(`T1`, `T2`)
+///   - `T1` otherwise
+///
+/// @description Check that UP(`T1`, `T2`) = `T2` if BOTTOM(`T1`) and
+/// BOTTOM(`T2`) and MOREBOTTOM(`T1`, `T2`) and the earlier conditions listed in
+/// README.md are not met.
```
eernstg: This is the part that I think is too complex. The reader _could_ go to the definition of **UP** and find this rule, conclude what 'earlier conditions' means, and then check for each test case below that each of them actually fails to apply.

However, I think this is so tricky (so easy to get wrong) that it is worth the effort to compute the intermediate result, which can be checked against README.md, which can in turn be checked against the definition of **UP**. These checks can be performed separately. So let's assume that we have a correct README.md, and it says something about what `T1` and `T2` must satisfy in order to ensure that the earlier rules are not applicable, for instance, "`T1` and `T2` are not the same type, and none of them is **TOP**".

We can then check that `T1` and `T2` are distinct types in each test case, and that none of them is **TOP**.

This makes it much easier to support the claim that we have actually checked these rule-order-dependencies carefully. ;-)

eernstg: This comment is relevant to other tests below, too.

eernstg (commented):
Lots of things are great!

However, I do think we should be more explicit about the handling of the order-dependent ruleset that is the definition of **UP**. The reason for this is that a plain reference to the definition of **UP** yields very little help if we do get something wrong here and there. I explained this in more detail in a comment on README.md, and also in a couple of locations in the tests.

sgrekhov:
Updated. It's indeed very easy to miss a case that is covered by an earlier rule. For example, I removed the tests where `void` was used as the type. This case is covered by the rule where one of the operands is `TOP()`, and the corresponding test already covers the case where the other operand is `Never`.

I removed README.md for now. Until we reach **AT THIS POINT**, it is more reliable to "manually" list all the conditions that should be met. Let's see how long the list becomes by the time we reach **AT THIS POINT** :)

PTAL

eernstg (commented):
The simplest way to ensure that all the requirements are expressed explicitly is probably to state them every time. In particular, the current `@description`s seem to rely on an unstated lemma that no type `T` can satisfy TOP(`T`) and `BOTTOM(`T`) at the same time. This is probably easy to prove, but I think it's just as easy to require "not TOP(`T`)" in a number of locations.

At first, I thought this would just be needed in a single test, but then I discovered that it's needed in several.

I edited a couple of cases, and then decided that I shouldn't edit a whole bunch of files. 

So let's take another round with this one.

sgrekhov:
Thank you! Updated. PTAL.

eernstg (approved):
LGTM!

PR #4191, opened 2026-08-12
#4173. Add tests for intersection types

sgrekhov:


TypeSystem/upper-lower-bounds/up_A09_t01.dart
```diff
@@ -0,0 +1,150 @@
...
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `X2 & B2`) =
+///   - `X2` if `T1 <: X2`
+///   - otherwise `T1` if `X2 <: T1`
+///   - otherwise UP(`T1`, `B2a`) where `B2a` is the greatest closure of `B2`
+///     with respect to `X2`, as defined in inference.md.
+///
+/// @description Check that UP(`T1`, `X2 & B2`) = `X2` if `T1 != X2 & B2`,
+/// TOP(`T1`) `== false`, BOTTOM(`T1`) `== false` and `T1 <: X2`.
```
eernstg: Similar story, the safe approach is to make all the requirements explicit:
```suggestion
/// not TOP(`T1`), not BOTTOM(`T1`), not BOTTOM(`T2`), and `T1 <: X2`.
//// Note that TOP(`X2 & B2`) can never hold.
```
The last part serves as a local lemma indicating that we don't even need to check this for each test case below. It's needed because otherwise we might use UP(`T1`, `T2`) = `T2` if TOP(`T2`).

Another matter is the use of mathematical terminology and notation vs. source code notation: We could say "BOTTOM(`T1`) == false", but this is basically relying on the expectation that we have a programming language available at the meta level (the specification level). That's in principle a dangerous assumption because we might well start using a feature F to specify the feature F (perhaps in some complex, indirect, but circular way), which would then provide F with an almost unlimited power to use the recursion as we please, that is, F could mean almost anything.

A mathematical notation could be something like "BOTTOM(`T1`)" (stating that the meta-level predicate BOTTOM is satisfied by `T1`), or the negation expressed using "not BOTTOM(`T1`)". In contrast, a notation like "BOTTOM(`T1`) == false" is less safe because it mixes the levels.

For the 'same type' concept, I haven't commented. It's tempting to use `S != T` because it is so concise, and the non-source-code version would be something like "`S` is not the same type as `T`", which is again in need of a definition (at least once).  So let's just use `!=` for that purpose.

TypeSystem/upper-lower-bounds/up_A09_t01.dart
```diff
@@ -0,0 +1,150 @@
...
+
+void f2<X2 extends A, T1 extends X2>(T1 t1, X2 x2) {
+  if (x2 is B2) { // `x2` promoted to `X2 & B2`.
+    var v = (1 > 2) ? t1 : x2; // UP(T1, X2 & B2) = X2 because T1 <: X2
+    v.expectStaticType<Exactly<X2>>();
+  }
+}
+
+void f3(int t1, num x2) {
+  if (x2 is B2) { // `x2` promoted to `num & B2`.
```
eernstg: That won't work: An intersection type `S & T` cannot exist if `S` is not a type variable. So `x2` will probably always have to have a type which is a type variable, and then it needs to be promoted, because `x2` is always going to be used to find UP(`...`, `X2 & B2`).

eernstg: OK, a lot of cases below have the same issue, so they all need to be adjusted.

TypeSystem/upper-lower-bounds/up_A09_t02.dart
```diff
@@ -0,0 +1,151 @@
...
+/// ...
+/// - UP(`T1`, `X2 & B2`) =
+///   - `X2` if `T1 <: X2`
+///   - otherwise `T1` if `X2 <: T1`
+///   - otherwise UP(`T1`, `B2a`) where `B2a` is the greatest closure of `B2`
+///     with respect to `X2`, as defined in inference.md.
+///
+/// @description Check that UP(`T1`, `X2 & B2`) = `T1` if `T1 != X2 & B2`,
+/// not TOP(`T1`), not BOTTOM(`T1`) and `X2 <: T1`.
+/// @author sgrekhov22@gmail.com
```
eernstg, line 17: Same issues as the previous test.

eernstg (commented):
This one contains a bunch of intersection types that cannot exist, so they need to be adjusted.

sgrekhov:
Thanks! Updated. PTAL.

TypeSystem/upper-lower-bounds/up_A09_t01.dart
```diff
@@ -38,113 +38,103 @@ void f2<X2 extends A, T1 extends X2>(T1 t1, X2 x2) {
   }
 }
 
-void f3(int t1, num x2) {
-  if (x2 is B2) { // `x2` promoted to `num & B2`.
-    var v = (1 > 2) ? t1 : x2; // UP(int, num & B2) = num because int <: num
-    v.expectStaticType<Exactly<num>>();
+void f3<X2 extends num, T1 extends X2>(T1 t1, X2 x2) {
```
eernstg: I think we need this:
```suggestion
void f3<X2 extends A, T1 extends X2>(T1 t1, X2 x2) {
```
Otherwise `X2 & B2` cannot exist (an intersection type must be of the form `X & B` such that `B` is a subtype of the declared bound of `X`).

But this just means that `f3` is the same as `f2` (at least, I can't see how they differ).

TypeSystem/upper-lower-bounds/up_A09_t01.dart
eernstg: We need to show why `UP(T1, T2) = T2 if TOP(T2)` and  `UP(T1, T2) = T1 if BOTTOM(T2)` cannot be relevant:
```suggestion
/// not TOP(`T1`), not BOTTOM(`X2 & B2`), `T1` is not
/// an intersection type, not BOTTOM(`T1`), and `T1 <: X2`.
/// We also use the fact that TOP(`X2 & B2`) never holds.
```
Please fix the ordering of the requirements if I'm not following your style.

TypeSystem/upper-lower-bounds/up_A09_t01.dart
```diff
@@ -38,113 +38,103 @@ void f2<X2 extends A, T1 extends X2>(T1 t1, X2 x2) {
...
   }
 }
 
-void f4(String? t1, Null x2) {
-  if (x2 is B2) { // `x2` promoted to `Null & B2`.
-    var v = (1 > 2) ? t1 : x2;
-    // UP(String?, Null & B2) = String? because String? <: Null
-    v.expectStaticType<Exactly<String?>>();
+void f4<X2 extends Null, T1 extends X2>(T1 t1, X2 x2) {
+  if (x2 is B2) { // `x2` promoted to `X2 & B2`.
```
eernstg: Again, we can't promote `x2` to `X2 & B2` because `B2` is not a subtype of `Null`. Looking down, I can see that `X2` gets various different bounds, but I'm afraid that won't work.

The task at this point is to find a systematic way to come up with test cases to cover in this context.

I think it makes sense to say that we are not testing the subtyping relation here, which means that it is a non-goal to cover different ways to satisfy `X1 <: T2`.

`X1` is a type variable, so there's only one kind of type to cover there, and `X1` without a bound plus `X1 extends A` is fine for that.

I think this implies that the coverage is already fine with `f1` and `f2`, and there's no sufficiently relevant coverage in having `f3` .. `f14`.

TypeSystem/upper-lower-bounds/up_A09_t02.dart
```diff
@@ -39,113 +39,103 @@ void f2<T1 extends A, X2 extends T1>(T1 t1, X2 x2) {
   }
 }
 
-void f3(num t1, int x2) {
-  if (x2 is B2) { // `x2` promoted to `int & B2`.
-    var v = (1 > 2) ? t1 : x2; // UP(num, int & B2) = num because int <: num
-    v.expectStaticType<Exactly<num>>();
+void f3<T1 extends num, X2 extends T1>(T1 t1, X2 x2) {
```
eernstg: I'm afraid we have the same situation here: `f1` and `f2` are covering the cases which are inherently relevant to the given rule, but we are not testing subtype relationships, and this means that `f3` .. `f14` are basically not needed.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
```diff
@@ -19,31 +19,32 @@
...
 class A {}
-class T1 extends A {}
-class B2 extends A {}
-class X2 extends A {}
+class A2 extends A {}
+class B extends A2 {}
+class B2 {}
 
-void f1(T1 t1, X2 x2) {
+void f1<T1 extends A2, X2 extends A, B2>(T1 t1, X2 x2) {
```
eernstg: Perhaps the type parameter declaration `B2` should be removed? It's confusing that `B2` is both the name of a type parameter and also the name of a class, and the fact that it is a type parameter with no bound causes the promotion to `X2 & B2` to fail because `B2` is not a subtype of `A`.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
```diff
@@ -19,31 +19,32 @@
...
   if (x2 is B2) { // `x2` promoted to `X2 & B2`.
     // UP(T1, X2 & B2) = UP(T1, B2a), where B2a is
-    // the greatest closure of `B2` with respect to `X2`.
+    // the greatest closure of `B2` with respect to `X2` (`A` in this case).
```
eernstg: We already have tests covering the greatest and least closure algorithm elsewhere, so there's no need to push anything fancy into that algorithm here.

This means that `B2` could just be the class, which means that the greatest closure of `B2`  is again `B2`, and the standard upper bound of `B2a` and `T1` would be `A2`, which should then be expected below.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
```diff
@@ -19,31 +19,32 @@
...
     v.expectStaticType<Exactly<A>>();
   }
 }
 
-void f2<TT1 extends A, XX2 extends A>(TT1 t1, XX2 x2) {
-  if (x2 is B2) { // `x2` promoted to `XX2 & B2`.
-    // UP(TT1, XX2 & B2) = UP(TT1, B2a), where `B2a` is
-    // the greatest closure of `B2` with respect to `XX2`.
+void f2<T1 extends A2, X2 extends A2, B2 extends A>(T1 t1, X2 x2) {
+  if (x2 is B2) { // `x2` promoted to `X2 & B2`.
```
eernstg, line 38: `B2` is still not a subtype of the bound of `X2`, so there's no promotion.

Given that there's no need to explore the greatest/least closure algorithm, and given that this test will be identical to `f1` if we just refer to the class `B2` rather than the type variable `B2`, I think `f2` can be removed.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
```diff
@@ -19,31 +19,32 @@
...
+    // the greatest closure of `B2` with respect to `X2` (`A2` in this case).
     var v = (1 > 2) ? t1 : x2;
-    v.expectStaticType<Exactly<A>>();
+    v.expectStaticType<Exactly<A2>>();
   }
 }
 
-void f3(T1? t1, X2? x2) {
+void f3<T1 extends A, X2 extends A>(T1? t1, X2? x2) {
   if (x2 is B2) { // `x2` promoted to `X2? & B2`.
```
eernstg: We actually can't have that, the left operand of an intersection type can only be a type variable, not even the corresponding nullable type.

However, I created https://github.com/dart-lang/language/issues/4747 to clarify whether the language should be generalized to allow promotion from a type like `X2?` to a type like `X2 & int` in one step (we can do it in two steps today). Note that we still don't have a type of the form `X2? & int`, we're just exploring a kind of promotion which is taking two steps simultaneously.

In any case, the rule that we're testing here only applies to the situation where the promotable variable has a type of the form `X2 & B2`, and we aren't covering anything which is directly a part of this rule by obtaining this type in a bunch of different ways.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
```diff
@@ -19,31 +19,32 @@
...
 import '../../Utils/static_type_helper.dart';
+import 'up_lib.dart';
 
 class A {}
-class T1 extends A {}
-class B2 extends A {}
-class X2 extends A {}
+class A2 extends A {}
+class B extends A2 {}
+class B2 {}
```
eernstg: Would need to be something like `class B2 extends/implements B;` in order to allow promotion below.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
```diff
@@ -80,10 +81,10 @@ void f6(T1 Function() t1, X2 Function() x2) {
 }
```
eernstg: I'm afraid `f2` .. `f6` aren't providing relevant coverage for the given `UP` rule, in addition to the fact that they are talking about intersection types that don't exist.

eernstg (commented):
I'm afraid a bunch of test cases should be removed because they use intersection types that don't exist, and/or because they are trying to contribute extra coverage, but they are doing so with respect to relations (like subtyping) that aren't being tested here.

TypeSystem/upper-lower-bounds/up_A09_t01.dart
sgrekhov: Thank you. Updated for all tests.

TypeSystem/upper-lower-bounds/up_A09_t01.dart
sgrekhov: f3-f14 are removed.

TypeSystem/upper-lower-bounds/up_A09_t01.dart
sgrekhov: Thanks. Removed.

TypeSystem/upper-lower-bounds/up_A09_t02.dart
sgrekhov: Updated.

TypeSystem/upper-lower-bounds/up_A09_t02.dart
sgrekhov: Removed.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
sgrekhov: Updated.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
sgrekhov: `f3`–`f6` were removed, but `f2` was updated and kept. The idea is to check that the implementation of the UP function uses `B2a` at this step, rather than simply `B2`. If the implementation erroneously uses `B2` instead of B2a, the test should detect it. Is this a valid test now?

sgrekhov (commented):
Thank you. Updated. Wrong test cases removed. But one in `up_A09_t03` was kept. PTAL.

TypeSystem/upper-lower-bounds/up_A09_t01.dart
eernstg: Sounds good!

TypeSystem/upper-lower-bounds/up_A09_t03.dart
```diff
@@ -24,18 +25,17 @@ import 'up_lib.dart';
 class A {}
 class A2 extends A {}
 class B extends A2 {}
```
eernstg: Might as well keep the naming such that the code gets to be similar to the UP rules, and it matches comments here and there.
```suggestion
class B2 extends A2 {}
```

TypeSystem/upper-lower-bounds/up_A09_t03.dart
```diff
@@ -0,0 +1,50 @@
...
+import 'up_lib.dart';
+
+class A {}
+class A2 extends A {}
+class B extends A2 {}
+
+void f1<T1 extends A2, X2 extends A>(T1 t1, X2 x2) {
+  if (x2 is B) { // `x2` promoted to `X2 & B2`.
+    // UP(T1, X2 & B2) = UP(T1, B2a), where B2a is
+    // the greatest closure of `B2` with respect to `X2` (`A2` in this case).
```
eernstg: If `class B` is renamed to `B2` then the greatest closure of `B2` with respect to anything is `B2`.

This is a simple case, and just one case, but that's not a problem because we have already tested greatest/least closure elsewhere. In any case, `UP(T1, B2a)` is `A2`.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
```diff
@@ -0,0 +1,50 @@
...
+    var v = (1 > 2) ? t1 : x2;
+    v.expectStaticType<Exactly<A2>>();
+  }
+}
+
+void f2<T1 extends A2, X2 extends A2, B2 extends A2>(T1 t1, X2 x2) {
+  if (x2 is B2) { // `x2` promoted to `X2 & B2`.
+    // UP(T1, X2 & B2) = UP(T1, B2a), where B2a is
+    // the greatest closure of `B2` with respect to `X2` (`A2` in this case).
```
eernstg: I think the greatest closure is just `B2` again. `B2` is a type variable which is not in `L`, which is `{X2}`, and this is the "does not contain any type variable from `L` case.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
eernstg: I think so, I added a comment about the greatest closure of `B2`, but the expectation is fine as far as I can see.

eernstg (commented):
Couple of tiny issues.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
sgrekhov: Thank you. Updated.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
sgrekhov: Is there a way to check that `B2a` is used rather that simply `B2`?

sgrekhov (commented):
Updated for `f1`. For `f2`, is there a way to check that `B2a` is used rather than simply `B2`? If not, I'll remove `f2`. PTAL.

TypeSystem/upper-lower-bounds/up_A09_t03.dart
eernstg: That's a good question!

However, I do think that is being tested in another series of tests about greatest/least closure that we had recently, because those tests were using precisely this case and the next one in order to get to look at the greatest/least closure (that is, _all_ cases). So I think it's fine to just have a very simple case here.

eernstg: I'll adjust the comment such that it mentions the actual greatest closure -- with that, we can land this PR. :tada:

eernstg (approved):
LGTM

PR #4192, opened 2026-08-13
#4173. Add tests for NULL types

sgrekhov:


TypeSystem/upper-lower-bounds/up_A10_t01.dart
```diff
@@ -0,0 +1,59 @@
...
+/// @description Check that UP(`T1`, `T2`) = `T2` if `T1 != T2`, `T1` and `T2`
+/// are both not a TOP, BOTTOM or intersection types and NULL(`T1`), NULL(`T2`)
+/// and MOREBOTTOM(`T1`, `T2`).
+/// @author sgrekhov22@gmail.com
+/// @issue 63908
+
+// ignore_for_file: unused_local_variable
+
+void f1(Never n) {
+  var v = (1 > 2) ? n : null;
```
eernstg: `NULL(Never)` doesn't hold. The types that satisfy `NULL` are `Null`, `Null?`, and `T?` where `BOTTOM(T)` (that is, `Never?` and `X?` where `X extends Never`; `(X & B)?` where `BOTTOM(B)` doesn't count because no types of the form `(X & B)?` exist).

So we'd basically have a matrix with those entries on both axes, and then we'd filter out the combinations which have already been handled in earlier rules. In particular, the main diagonal is gone because it is handled by `UP(T, T)`.

TypeSystem/upper-lower-bounds/up_A10_t01.dart
```diff
@@ -0,0 +1,59 @@
...
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `T2`) where NULL(`T1`) and NULL(`T2`) =
+///   - `T2` if MOREBOTTOM(`T1`, `T2`)
+///   - `T1` otherwise
+///
+/// @description Check that UP(`T1`, `T2`) = `T2` if `T1 != T2`, `T1` and `T2`
+/// are both not a TOP, BOTTOM or intersection types and NULL(`T1`), NULL(`T2`)
```
eernstg: I think it's easier to see that it is a list of cases if only the last element has 'and'.
```suggestion
/// are both not a TOP, BOTTOM, or intersection type, if NULL(`T1`), NULL(`T2`),
```
We could add 'if' before every element in the list, but I think it's OK to have it only at the spot where it gets hard to remember that we're "iffing" every element in the list. ;-)

TypeSystem/upper-lower-bounds/up_A10_t01.dart
```diff
@@ -0,0 +1,59 @@
...
+/// @author sgrekhov22@gmail.com
+/// @issue 63908
+
+// ignore_for_file: unused_local_variable
+
+void f1(Never n) {
+  var v = (1 > 2) ? n : null;
+  // UP(Never, Null) = Null, because MOREBOTTOM(Never, T) = true
+  // Let's check that `v` is not assignable to a non-nullable variable
+  int i = v;
```
eernstg: Wouldn't it work to test directly that `v.expectStaticType<Exactly<Null>>`?

TypeSystem/upper-lower-bounds/up_A10_t02.dart
```diff
@@ -0,0 +1,59 @@
...
+/// ...
+/// - UP(`T1`, `T2`) where NULL(`T1`) and NULL(`T2`) =
+///   - `T2` if MOREBOTTOM(`T1`, `T2`)
+///   - `T1` otherwise
+///
+/// @description Check that UP(`T1`, `T2`) = `T1` if `T1 != T2`, `T1` and `T2`
+/// are both not a TOP, BOTTOM or intersection types and NULL(`T1`), NULL(`T2`)
+/// and not MOREBOTTOM(`T1`, `T2`).
+/// @author sgrekhov22@gmail.com
+/// @issue 63908
```
eernstg, line 16: Similar to A10_ t01.

TypeSystem/upper-lower-bounds/up_A11_t01.dart
```diff
@@ -0,0 +1,168 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `T2`) where NULL(`T1`) =
+///   - `T2` if `T2` is nullable
+///   - `T2*` if `Null <: T2` or `T1 <: Object` (that is, `T1` or `T2` is
+///     legacy)
```
eernstg: I'd recommend that we simply remove every rule or subrule which is dealing with legacy types. They do not exist in relevant versions of Dart.
```suggestion
```

TypeSystem/upper-lower-bounds/up_A11_t01.dart
```diff
@@ -0,0 +1,168 @@
...
+///   - `T2?` otherwise
+///
+/// @description Check that UP(`T1`, `T2`) = `T2` if `T1 != T2`, `T1` and `T2`
+/// are both not a TOP, BOTTOM or intersection types, NULL(`T1`) is `true`, not
+/// NULL(`T2`) and `T2` is nullable.
```
eernstg:
```suggestion
/// @description Check that UP(`T1`, `T2`) = `T2` if `T1 != T2`, `T1` and `T2`
/// are both not a TOP, BOTTOM or intersection type, if NULL(`T1`), not
/// NULL(`T2`), and `T2` is nullable.
```

TypeSystem/upper-lower-bounds/up_A11_t01.dart
```diff
@@ -0,0 +1,168 @@
...
+import 'dart:async';
+import '../../Utils/static_type_helper.dart';
+import 'up_lib.dart';
+
+void f1(num? n) {
+  var v = (1 > 2) ? null : n; // UP(Null, num?) = num?
+  v.expectStaticType<Exactly<num?>>();
+}
+
+void f2(Never? t1, num? n) {
```
eernstg, line 26: OK, so we're just checking the case NULL(`T1`) using the types `Null` and `Never?`. That's fine, given that the two A10 tests have all cases.

TypeSystem/upper-lower-bounds/up_A11_t01.dart
eernstg: A similar comment would apply to the `@description` of more tests below.

eernstg (commented):
Looks good! There is one test where a type should be NULL but isn't, so that one needs an adjustment. There's another comment on the descriptions.

TypeSystem/upper-lower-bounds/up_A10_t01.dart
sgrekhov: Updated.

TypeSystem/upper-lower-bounds/up_A10_t01.dart
sgrekhov: Ah yes, indeed! Replaced by `X?` where `X extends Never` and moved to the test where we are checking not MOREBUTTOM(because MOREBOTTOM(X?, Null) == false).

TypeSystem/upper-lower-bounds/up_A10_t01.dart
sgrekhov: Yes, `v.expectStaticType<Exactly<X?>>();`.

TypeSystem/upper-lower-bounds/up_A10_t02.dart
sgrekhov: Updated.

TypeSystem/upper-lower-bounds/up_A11_t01.dart
sgrekhov: Removed here and below.

TypeSystem/upper-lower-bounds/up_A11_t01.dart
sgrekhov: Updated here and below.

sgrekhov (commented):
Thank you. Updated. PTAL.

TypeSystem/upper-lower-bounds/up_A10_t01.dart
```diff
@@ -0,0 +1,50 @@
...
+}
+
+void f2(Null? n) { // ignore: unnecessary_question_mark
+  // UP(Null, Null?) = dynamic for historical reasons though MOREBOTTOM(Null, Null?) = true
+  var v = (1 > 2) ? null : n;
+  if (1 > 2) {
+    v.checkDynamic;
+  }
+}
+
```
eernstg, line 48: OK, so we have `(Null, Never?)`, `(Null, Null?)`, but not `(Null, X?)` where `X extends Never`, and not the opposite direction. But as I understand it, they're covered in A10_t02.

Still, I can't find `(Null, X?)`. Also I can't find pairs that don't include `Null` itself (like `(Null?, Never?)` or `(Null?, X?)` or `(Null?, Never?)`). Are they somewhere else?

I know, it's extremely nitpickish, but I don't think we have any other test that covers `NULL` and `UP` in those cases.

TypeSystem/upper-lower-bounds/up_A10_t02.dart
eernstg: Thanks!

TypeSystem/upper-lower-bounds/up_A10_t02.dart
```diff
@@ -0,0 +1,58 @@
...
+
+import '../../Utils/static_type_helper.dart';
+
+// ignore_for_file: unused_local_variable
+
+void f1<X extends Never>(X? n) {
+  var v = (1 > 2) ? n : null;
+  // UP(X?, Null) = X?, because MOREBOTTOM(X?, Null) == false
```
eernstg: If `MOREBOTTOM(X?, Null)` is false, wouldn't that yield `UP(X?, Null) = Null`?

TypeSystem/upper-lower-bounds/up_A10_t02.dart
```diff
@@ -0,0 +1,58 @@
...
+  // Let's check that `v` is not assignable to a non-nullable variable
+  v.expectStaticType<Exactly<X?>>();
```
eernstg: We really can't see the difference, but I think it should be `Null`.

TypeSystem/upper-lower-bounds/up_A11_t01.dart
eernstg: Thanks!

eernstg (commented):
I'm afraid we still have a couple of things here.

TypeSystem/upper-lower-bounds/up_A10_t01.dart
sgrekhov: Added, thank you.

TypeSystem/upper-lower-bounds/up_A10_t02.dart
sgrekhov: UP(`T1`, `T2`) = `T1` if MOREBOTTOM(X?, Null) == false. `X?` in this case.

TypeSystem/upper-lower-bounds/up_A10_t02.dart
sgrekhov: Please see my comment above.

sgrekhov (commented):
Updated, the missing cases added. PTAL

TypeSystem/upper-lower-bounds/up_A10_t01.dart
```diff
@@ -0,0 +1,77 @@
...
+  var v2 = (1 > 2) ? v : 42 as Object?;
+  v2.proofNotDynamic;
+//   ^^^^^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+void f2(Null? n) {
+  // UP(Null, Null?) = dynamic for historical reasons though MOREBOTTOM(Null, Null?) = true
+  var v = (1 > 2) ? null : n;
```
eernstg: It's true that the inferred type of `v` is `dynamic`, but the type of the initializing expression is still `Null`, so we should be able to test that directly:

```dart
  ((1 > 2) ? null : n).expectStaticType<Exactly<Null>>;
```
With that, we shouldn't need lines 42-44.

TypeSystem/upper-lower-bounds/up_A10_t01.dart
eernstg: Thanks!

TypeSystem/upper-lower-bounds/up_A10_t01.dart
```diff
@@ -0,0 +1,77 @@
...
+
+import '../../Utils/static_type_helper.dart';
+
+// ignore_for_file: unused_local_variable, unnecessary_question_mark
+
+void f1(Never? n) {
+  var v = (1 > 2) ? null : n;
+  // UP(Null, Never?) = Never?, because MOREBOTTOM(Null, T) = true
+  // It's not possible to check that `v` is `Never?`. Let's check that it is
+  // assignable to both `String` and `int` and is not `dynamic`
```
eernstg: But `Never?` is not assignable to `int` and `String`, we'd need to strip off the `?` part first.

For example, before line 27: `if (v == null) return-or-fail;`. When `v` has been promoted to a non-null type we can assign to `int` and `String`.

If `v` has the type `dynamic` then it still works because the `if ...` lets `v` keep the type `dynamic`, so we will still detect that case and fail as we should.

TypeSystem/upper-lower-bounds/up_A10_t02.dart
```diff
@@ -0,0 +1,89 @@
...
+
+import '../../Utils/static_type_helper.dart';
+
+// ignore_for_file: unused_local_variable, unnecessary_question_mark
+
+void f1(Never? n) {
+  var v = (1 > 2) ? n : null;
+  // UP(Never?, Null) = Never?, because MOREBOTTOM(T, Null) = false
+  // It's not possible to check that `v` is `Never?`. Let's check that it is
+  // assignable to both `String` and `int` and is not `dynamic`
```
eernstg, line 26: Needs to return/fail if `v == null`, as in the previous test.

TypeSystem/upper-lower-bounds/up_A10_t02.dart
```diff
@@ -0,0 +1,89 @@
...
+  var v2 = (1 > 2) ? v : 42 as Object?;
+  v2.proofNotDynamic;
+//   ^^^^^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+void f2(Null? n) {
+  // UP(Null?, Null) = dynamic for historical reasons though MOREBOTTOM(Null?, Null) = false
+  var v = (1 > 2) ? n : null;
```
eernstg: We can test the type of the expression `(1 > 2) ? n : null` directly.

TypeSystem/upper-lower-bounds/up_A10_t02.dart
eernstg: Ah, of course!

TypeSystem/upper-lower-bounds/up_A10_t02.dart
eernstg: My bad again.

eernstg (commented):
Very nice! Just a couple of comments. In particular, the case where the variable type is `dynamic` because the initializing expression has static type `Null` can be tested directly. Another test seems to expect that `Never?` can be assigned to `int` and `String`, but that should be rejected as a compile-time error.

TypeSystem/upper-lower-bounds/up_A10_t01.dart
sgrekhov: How interesting. Thanks! Updated.

TypeSystem/upper-lower-bounds/up_A10_t01.dart
sgrekhov: Added. The actual type of `v` here is `dynamic`. It's sdk#63908

TypeSystem/upper-lower-bounds/up_A10_t02.dart
sgrekhov: Added.

TypeSystem/upper-lower-bounds/up_A10_t02.dart
sgrekhov: Updated.

sgrekhov (commented):
Thanks. Updated. PTAL.

TypeSystem/upper-lower-bounds/up_A10_t01.dart
eernstg: Thanks!

TypeSystem/upper-lower-bounds/up_A10_t01.dart
eernstg: Very good, thanks!

TypeSystem/upper-lower-bounds/up_A10_t02.dart
```diff
@@ -37,11 +40,10 @@ void f1(Never? n) {
...
 void f2(Null? n) {
-  // UP(Null?, Null) = dynamic for historical reasons though MOREBOTTOM(Null?, Null) = false
-  var v = (1 > 2) ? n : null;
-  if (1 > 2) {
-    v.checkDynamic;
-  }
+  // In the case var v = (1 > 2) ? n : null; the inferred type of `v` is
+  // `dynamic` for historical reasons. Though MOREBOTTOM(Null?, Null) = false
+  // and the expression should has type `Null`.
+  ((1 > 2) ? n : null).expectStaticType<Exactly<Null>>();
```
eernstg, line 46: I think we should ensure that the type is not `dynamic` as well:

```dart
  ((1 > 2) ? n : null).proofNotDynamic;
//                     ^^^^^^^^^^^^^^^ and so on.
```

eernstg (commented):
Just one location needs to check that a type isn't dynamic!

TypeSystem/upper-lower-bounds/up_A10_t02.dart
sgrekhov: Updated.

eernstg (approved):
LGTM!

PR #4193, opened 2026-08-18
#4173. Add tests for OBJECT

sgrekhov:


TypeSystem/upper-lower-bounds/up_A13_t01.dart
```diff
@@ -0,0 +1,41 @@
...
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `T2`) where OBJECT(`T1`) and OBJECT(`T2`) =
+///   - `T1` if MORETOP(`T1`, `T2`)
+///   - `T2` otherwise
+///
+/// @description Check that UP(`T1`, `T2`) = `T1` if `T1 != T2` and OBJECT(`T1`)
+/// and OBJECT(`T2`) and MORETOP(`T1`, `T2`). Test that `Object` is more top
+/// than `FutureOr<Object>`.
```
eernstg: We're relying on a lemma here as well:

```suggestion
/// than `FutureOr<Object>`. Note that none of TOP(`T`), BOTTOM(`T`), or
/// NULL(`T`) holds when OBJECT(`T`), and `T` is not an intersection type.
```
This is relevant for both `T1` and `T2`, but the fresh meta-variable `T` allows us to say it just once.

TypeSystem/upper-lower-bounds/up_A13_t01.dart
```diff
@@ -0,0 +1,41 @@
...
+/// @author sgrekhov22@gmail.com
+
+import 'dart:async';
+import '../../Utils/static_type_helper.dart';
+
+void f1(Object x, FutureOr<Object> y) {
+  var v = (1 > 2) ? x : y; // MORETOP(Object, FutureOr<Object>) = true
+  v.expectStaticType<Exactly<Object>>();
```
eernstg: `Object` and `FutureOr<Object>` are subtypes of each other, which means that we can't see the difference using `expectStaticType`.

We also can't see the difference using an extension method because it's based on finding actual type arguments of the extension such that a subtype relation exists.

However, we could use the following as a starting point for code that makes the distinction:

```dart
import 'dart:async';

Future<X> f<X>() {
  print(X);
  return Future<Never>.error(0);
}

void main() {
  Object o1 = 42;
  FutureOr<Object> o2 = 42;
  
  o1 = f(); // Prints 'dynamic'.
  o2 = f(); // Prints 'Object'.
}
```
The "taster" function `f` would then check the value of `X` and expect that it's `== dynamic` when we expect the type to be `Object`.

TypeSystem/upper-lower-bounds/up_A13_t02.dart
```diff
@@ -0,0 +1,41 @@
...
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `T2`) where OBJECT(`T1`) and OBJECT(`T2`) =
+///   - `T1` if MORETOP(`T1`, `T2`)
+///   - `T2` otherwise
+///
+/// @description Check that UP(`T1`, `T2`) = `T2` if `T1 != T2` and OBJECT(`T1`)
+/// and OBJECT(`T2`) and not MORETOP(`T1`, `T2`). Test that `Object` is more top
+/// than `FutureOr<Object>`.
```
eernstg: Similar adjustments as in t01.

eernstg (commented):
We have to use an extra trick in order to be able to see a difference between types of the form `FutureOr<...FutureOr<Object>...>`. Also, I commented on the description because we're relying on a lemma (that is, a rule that doesn't need to be checked for each test case, it's just known to be satisfied once and for all) in order to know that none of the earlier rules could be applied.

The later tests would use a similar trick, but I only commented on the first few tests because the overall idea is the same.

TypeSystem/upper-lower-bounds/up_A13_t01.dart
sgrekhov: Added to all descriptions in this PR.

TypeSystem/upper-lower-bounds/up_A13_t01.dart
sgrekhov: Thank you. Updated everywhere.

sgrekhov (commented):
I had to go through several iterations in this PR. First, I replaced `expectStaticType()` with new code that can distinguish between `Object` and `FutureOr<Object>`, but then I realized that `expectStaticType()` is also needed to check that the result is not, say, `Object?`.

In some cases, the new checks may look excessive, but let's keep them in these tests. If, say, `UP(num, Object)` surprisingly returns `FutureOr<Object>`, these tests will be able to detect it.

PTAL.

TypeSystem/upper-lower-bounds/up_lib.dart
```diff
@@ -21,3 +22,39 @@ extension type ET(int _) {}
...
+/// Future<X> f<X>() {
+///   print(X);
+///   return Future<Never>.error(0);
+/// }
+///
+/// void main() {
+///   Object o1 = 42;
+///   FutureOr<Object> o2 = 42;
+///
+///   o1 = f(); // Prints 'dynamic'.
```
eernstg: I'm afraid we can't really trust the result to be `dynamic`: It is currently, but the specified behavior calls for `Object?` rather than `dynamic`, and the implementation will probably approach this behavior. We still have the distinction, though, because `Object` is different from the top types no matter which kind of equality we're using.

TypeSystem/upper-lower-bounds/up_lib.dart
```diff
@@ -21,3 +22,39 @@ extension type ET(int _) {}
...
+///   o2 = f(); // Prints 'Object'.
+/// }
+///```
+Future<X> checkObject<X>() {
+  checkObjectLog = '$X';
```
eernstg: Using the textual representation of a type is never safe, but we don't have to do that: We just need to confirm that `X` is a top type when we expect the context type to be `Object`, and `X` is `Object` when we expect the context type to be `FutureOr<Object>`.

One way to do it would be something like the following:

```dart
Future<X> confirmObjectContext<X>() {
  // Confirm that `X` is a top type.
  if (<Object?>[] is List<X>) return Future<X>.value(0 as dynamic);
  throw "Something which is commonly used to indicate test failure";
}

Future<X> confirmFutureOrObjectContext<X>() {
  // Confirm that `X` is `Object`.
  if (X == Object) return Future<X>.value(0 as dynamic);
  throw "...";
}
```

Example usage:

```dart
void main() async {
  Object o1 = 42;
  FutureOr<Object> o2 = 42;
  FutureOr<FutureOr<Object>> o3 = 42;
  FutureOr<FutureOr<FutureOr<Object>>> o4 = 42;
  
  o1 = confirmObjectContext();
  o2 = confirmFutureOrObjectContext();
  var flattenedO3 = await o3;
  flattenedO3 = confirmFutureOrObjectContext();
  var flattenedFlattenedO4 = await (await o4);
  flattenedFlattenedO4 = confirmFutureOrObjectContext();
}
```

The evaluations of `await ...` should be safe: If the type is _not_ of the form `FutureOr<...>` or `Future<...>` then the result after flattening would not be of the form `FutureOr<...>`; next, if the type is `Future<...>` then it is not an OBJECT type.

There may still be a loophole somewhere, but I haven't noticed any, and it's at least a lot better than not testing for `FutureOr`ness at all.

eernstg (commented):
We still have some tricky stuff here!

I'm sorry about the continued nitpicking, but as long as we _do_ have the ability to distinguish `Object` from `FutureOr<Object>` we should have at least one location in our test suite where it is tested.

This review just points out that it is possible to avoid testing types by looking at their `toString()` and shows one possible approach.

So please take a look at it!

sgrekhov:
Thank you! Updated. PTAL.

TypeSystem/upper-lower-bounds/up_A13_t01.dart
```diff
@@ -24,24 +24,21 @@ void f1(Object x, FutureOr<Object> y) {
   // Object and FutureOr<Object> are subtypes of each other, which means that we
   // can't see the difference using `expectStaticType()` function.
   v.expectStaticType<Exactly<Object>>(); // Check that `v` is not `Object?`
```
eernstg: We're actually checking that the type of `v` is `OBJECT`. It is true that we're also checking that its type isn't a top type, but saying that it is `OBJECT` would be more informative for future readers.

TypeSystem/upper-lower-bounds/up_A13_t01.dart
```diff
@@ -24,24 +24,21 @@ void f1(Object x, FutureOr<Object> y) {
...
-  v = checkObject();
-  expectObject();
+  v = confirmObjectContext();
 }
 
 void f3(FutureOr<Object> x, FutureOr<FutureOr<Object>> y) {
   // MORETOP(FutureOr<Object>, FutureOr<FutureOr<Object>>) =
   // MORETOP(Object, FutureOr<Object>) = true
   var v = (1 > 2) ? x : y;
   v.expectStaticType<Exactly<FutureOr<Object>>>(); // Check that `v` is not `FutureOr<Object?>`
```
eernstg: We might as well just check `expectStaticType<Exactly<Object>>` because the purpose is to ensure that the type of `v` is `OBJECT`, and `FutureOr<FutureOr<.......` makes no difference for this kind of test).

eernstg: Some tests below are similar.

TypeSystem/upper-lower-bounds/up_A13_t02.dart
```diff
@@ -24,24 +24,21 @@ void f1(FutureOr<Object> x, Object y) {
...
-  v = checkObject();
-  expectObject();
+  v = confirmObjectContext();
 }
 
 void f3(FutureOr<FutureOr<Object>> x, FutureOr<Object> y) {
   // MORETOP(FutureOr<FutureOr<Object>>, FutureOr<Object>) =
   // MORETOP(FutureOr<Object>, Object) = false
   var v = (1 > 2) ? x : y;
   v.expectStaticType<Exactly<FutureOr<Object>>>(); // Check that `v` is not `FutureOr<Object?>`
```
eernstg: This one is particularly confusing. ;-)

It's easier to follow if we just test for `Object` which implies `OBJECT`, and then say in the comment that it's all about `OBJECT`.

TypeSystem/upper-lower-bounds/up_lib.dart
```diff
@@ -38,23 +38,43 @@ typedef Rec = (int, String, {bool b});
...
-Future<X> checkObject<X>() {
-  checkObjectLog = '$X';
-  return Future<X>.value(0 as dynamic);
+Future<X> confirmObjectContext<X>() {
+  // Confirm that `X` is a top type.
+  if (<Object?>[] is List<X>) {
+    return Future<X>.value(0 as dynamic);
+  }
+  Expect.fail('The context is not a top type');
+  return Future<X>.value(null); // We need to return something
```
eernstg, line 61: Why would this be possible? I'd expect something like `null as X`, not just `null`. By the way, the value could also be null (`null as X`, that is) in line 48.

eernstg: Similar issues below.

TypeSystem/upper-lower-bounds/up_lib.dart
```diff
@@ -38,23 +38,43 @@ typedef Rec = (int, String, {bool b});
 ///   Object o1 = 42;
 ///   FutureOr<Object> o2 = 42;
 ///
-///   o1 = f(); // Prints 'dynamic'.
-///   o2 = f(); // Prints 'Object'.
+///   o1 = confirmObjectContext(); // Throws if `o1` is `FutureOr<Object>`
```
eernstg: `o1 is FutureOr<Object>` actually evaluates to true. Perhaps it would be safer to say that "Throws if the static type of `o1` is `FutureOr<Object>`.".

eernstg: Similar issues in a couple of places below.

eernstg (commented):
More nitpicking! I think it will be helpful for the `expectStaticType` tear-offs to test for a plain `Object` because it can't determine anything about `Object` vs. `FutureOr<Object>`, but it _can_ confirm that the static type is `OBJECT` (and that's a step that we need to have because `confirmObjectContext` and its brethren need to be invoked on relevant types, they do not work as intended for arbitrary types).

TypeSystem/upper-lower-bounds/up_A13_t01.dart
sgrekhov: Updated here and below

TypeSystem/upper-lower-bounds/up_A13_t01.dart
sgrekhov: Added here and below.

TypeSystem/upper-lower-bounds/up_lib.dart
sgrekhov: Updated.

TypeSystem/upper-lower-bounds/up_lib.dart
sgrekhov: This is a dead code that will never be excuted, but the function cannot be compiled without it. `Future.value()` constructor is declared as `Future<X>.value([FutureOr<X>? value])` and accepts `null`.
`(null as X)` in line 48 leads to "type 'Null' is not a subtype of type 'Object' in type cast" error. No change in this area.

sgrekhov (commented):
Thank you. Updated. PTAL.

TypeSystem/upper-lower-bounds/up_A14_t02.dart
```diff
@@ -0,0 +1,212 @@
...
+import 'up_lib.dart';
+
+void f1(Object o, num? n) {
+  var v = (1 > 2) ? o : n; // UP(Object, num?) = Object?
+  v.expectStaticType<Exactly<Object?>>(); // Check that v is not `Object`
+  // `Object` and `FutureOr<Object> `are subtypes of each other, which means
+  // that we can't see the difference using `expectStaticType()` function.
+  // `v.expectStaticType<Exactly<FutureOr<Object?>>>();` also succeeds. Let's
+  // check that `v` is not `FutureOr<Object?>`
+  v = confirmObjectContext();
```
eernstg: This one is different: We need to confirm that `v` has type `Object?`, and we'd like to know that it isn't `FutureOr<Object>?`.

So we need something like this:

```dart
void main() {
  Object? o1 = 42 as dynamic; // Avoid immediate promotion to `Object`.
  FutureOr<Object>? o2 = 42 as dynamic; // Ditto.

  // Confirm that their static types are all `TOP`.
  o1.expectStaticType<Exactly<Object?>>; 
  o2.expectStaticType<Exactly<Object?>>;

  // Remove null from the variables' static types.
  if (o1 == null || o2 == null) return;

  // Confirm that those static types do/do-not include `FutureOr`.
  o1 = confirmObjectContext(); // OK. Throws if we do `o2 = ...`.
  o2 = confirmFutureOrObjectContext(); // OK. Throws if we do `o1 = ...`.
}
```

I don't think there's an easy way to abstract over this:  We must test that the static types of the variables are `TOP` because they must be `T?` where `OBJECT(T)`, and then we need to strip off the `?` part of the type in order to be able to test the next step, but this step is just like the OBJECT/OBJECT case so we just do the same things.

For the run-time check to be performed, we need to know that the `return;` statement isn't actually executed, so perhaps the body of the `if` statement should be `Expect.fail()`.

TypeSystem/upper-lower-bounds/up_A14_t02.dart
```diff
@@ -0,0 +1,212 @@
...
+/// intersection type.
+/// @author sgrekhov22@gmail.com
+
+import 'dart:async';
+import '../../Utils/static_type_helper.dart';
+import 'up_lib.dart';
+
+void f1(Object o, num? n) {
+  var v = (1 > 2) ? o : n; // UP(Object, num?) = Object?
+  v.expectStaticType<Exactly<Object?>>(); // Check that v is not `Object`
```
eernstg: Similarly to the previous tests, the comment should describe the property which is being confirmed. We know that the type should be `T?` where `OBJECT(T)`, but this implies that the type must be `TOP`. A comment like "// Check that `v`'s type is `TOP`." says this clearly, and it's a lot less ambiguous than "v is not `Object`" (which could mean `(v is Object) == false`).

TypeSystem/upper-lower-bounds/up_A14_t02.dart
```diff
@@ -0,0 +1,212 @@
...
+  // `Object` and `FutureOr<Object> `are subtypes of each other, which means
+  // that we can't see the difference using `expectStaticType()` function.
+  // `v.expectStaticType<Exactly<FutureOr<Object?>>>();` also succeeds. Let's
+  // check that `v` is not `FutureOr<Object?>`
+  v = confirmObjectContext();
+}
+
+void f2(FutureOr<Object> o, num? n) {
+  var v = (1 > 2) ? o : n; // UP(FutureOr<Object>, num?) = FutureOr<Object>?
+  v.expectStaticType<Exactly<FutureOr<Object>?>>();
```
eernstg: All the `expectStaticType` tests can't see `FutureOr`, so they're just testing that the type is `TOP`, which means that they're more readable if they are actually written as `expectStaticType<Exactly<Object?>>`.

TypeSystem/upper-lower-bounds/up_A15_t01.dart
```diff
@@ -0,0 +1,193 @@
...
+/// ...
+/// - UP(`T1`, `T2`) where OBJECT(`T2`) =
+///   - `T2` if `T1` is non-nullable
+///   - `T2?` otherwise
+///
+/// @description Check that UP(`T1`, `T2`) = `T2` if `T1 != T2`, `T1` and `T2`
+/// are neither TOP, BOTTOM, NULL, nor intersection type, if OBJECT(`T2`), not
+/// OBJECT(`T1`) and `T1` is non-nullable. Note that none of TOP(`T`),
+/// BOTTOM(`T`), or NULL(`T`) holds when OBJECT(`T`), and `T` is not an
+/// intersection type.
```
eernstg, line 16: Calls for changes similar to the earlier tests.

TypeSystem/upper-lower-bounds/up_A15_t02.dart
```diff
@@ -0,0 +1,212 @@
...
+/// ...
+/// - UP(`T1`, `T2`) where OBJECT(`T2`) =
+///   - `T2` if `T1` is non-nullable
+///   - `T2?` otherwise
+///
+/// @description Check that UP(`T1`, `T2`) = `T2?` if `T1 != T2`, `T1` and `T2`
+/// are neither TOP, BOTTOM, NULL, nor intersection type, if OBJECT(`T2`), not
+/// OBJECT(`T1`) and `T1` is not non-nullable. Note that none of TOP(`T`),
+/// BOTTOM(`T`), or NULL(`T`) holds when OBJECT(`T`), and `T` is not an
+/// intersection type.
```
eernstg, line 16: Calls for changes similar to earlier tests, too.

eernstg (commented):
More adjustments. I made some changes to the files to illustrate what I meant, and the remaining tests would need similar changes.

In particular, I think it's helpful to make usages of `expectStaticType` very explicit about what it's testing (and, conversely, what it cannot test), namely "the type of this variable is OBJECT or TOP". Next, we can use other techniques (`confirm...`) to make the distinction between types with  and without `FutureOr...`.

TypeSystem/upper-lower-bounds/up_A13_t01.dart
eernstg: Thanks!

TypeSystem/upper-lower-bounds/up_lib.dart
eernstg: Ah, that's true, the `Future` API has some elements from the stone age, which makes `Future<T>.value(null)` OK during static analysis even though it's not intended to work. It simply wasn't considered realistic to change the parameter declaration to handle the case where the type argument of the future is non-nullable.

I don't know if it's better to throw at run time rather than continuing to run the test when `Expect.fail` has actually already been executed.

Anyway, now I understand why it's even possible to use `null` in that context.

TypeSystem/upper-lower-bounds/up_A14_t02.dart
sgrekhov: Thanks. Updated here and below.

TypeSystem/upper-lower-bounds/up_A14_t02.dart
sgrekhov: Updated here and below.

TypeSystem/upper-lower-bounds/up_A14_t02.dart
sgrekhov: Fixed.

TypeSystem/upper-lower-bounds/up_A15_t01.dart
sgrekhov: Done.

TypeSystem/upper-lower-bounds/up_A15_t02.dart
sgrekhov: Done.

sgrekhov (commented):
Thank you! Updated. PTAL.

TypeSystem/upper-lower-bounds/up_A13_t01.dart
eernstg: Thanks!

TypeSystem/upper-lower-bounds/up_A15_t01.dart
```diff
@@ -0,0 +1,182 @@
...
+  v.expectStaticType<Exactly<Object>>();
+  v = confirmObjectContext();
+}
+
+void f22(E n, FutureOr<Object> o) {
+  var v = (1 > 2) ? n : o; // UP(E, FutureOr<Object>) = FutureOr<Object>
+  v.expectStaticType<Exactly<Object>>();
+  v = confirmFutureOrObjectContext();
+}
+
```
eernstg, line 158: ET?

TypeSystem/upper-lower-bounds/up_A14_t02.dart
```diff
@@ -0,0 +1,317 @@
...
+
+void f4<X extends num>(FutureOr<Object> o, X? n) {
+  var v = (1 > 2) ? o : n; // UP(FutureOr<Object>, X?) = FutureOr<Object>?
+  v.expectStaticType<Exactly<Object?>>();
+  if (n == null) {
+    Expect.fail('The actual value must be non-null for the test to complete.');
+    return;
+  }
+  v = confirmFutureOrObjectContext();
+}
```
eernstg, line 66: See the comment on 'TypeSystem/upper-lower-bounds/up_A15_t02.dart': We'd need an extra copy of `f3` and `f4` covering the case where the 'not non-nullable' type is _potentially_ nullable, not just nullable.

TypeSystem/upper-lower-bounds/up_A14_t02.dart
```diff
@@ -0,0 +1,317 @@
...
+import 'up_lib.dart';
+
+void f1(Object o, num? n) {
+  var v = (1 > 2) ? o : n; // UP(Object, num?) = Object?
+  v.expectStaticType<Exactly<Object?>>(); // Check that v's type is TOP.
+  // `Object` and `FutureOr<Object> `are subtypes of each other, which means
+  // that we can't see the difference using `expectStaticType()` function.
+  // `v.expectStaticType<Exactly<FutureOr<Object?>>>();` also succeeds. Let's
+  // check that `v` is not `FutureOr<Object?>`
+  if (n == null) { // Strip the `?`
```
eernstg: We should strip the `?` from the static type of `v`, not `n`. Same applies in many cases below, and the other 'not non-nullable' teste.

TypeSystem/upper-lower-bounds/up_A15_t02.dart
```diff
@@ -0,0 +1,317 @@
...
+  // `ET` is neither non-nullable (`ET <: Object` is false) nor nullable
+  var v = (1 > 2) ? n : o; // UP(ET, FutureOr<Object>) = FutureOr<Object>?
+  v.expectStaticType<Exactly<Object?>>();
+  if (n == null) { // Strip the `?`
+    Expect.fail('The actual value must be non-null for the test to complete.');
+    return;
+  }
+  v = confirmFutureOrObjectContext();
+}
+
```
eernstg, line 309: Just one more case: We need to cover the situation where 'not non-nullable' is satisfied because the given type is _potentially_ nullable &mdash; at this time every case uses a plain nullable type. This could be done by having an extra copy of `f3` and `f4` where the type variable is declared as `X extends num?` and the first parameter is declared as `X n`.

This case is interesting because the addition of `?` to the resulting type occurs in a situation where null _might_ be included in the type `T1`, but it might not. It's just not possible for us to know that null isn't included, so we have to consider it as included.

eernstg (commented):
Really close, couple of comments.

TypeSystem/upper-lower-bounds/up_A14_t02.dart
sgrekhov: Ups. Sorry. Updated everywhere.

TypeSystem/upper-lower-bounds/up_A15_t01.dart
sgrekhov: In this test we are checking non-nullable types. `ET` was tested in `up_A15_t02.dart` with the comment that it is nor nullable  nor non-nullable. Moved these tests to this file and in `up_A15_t02.dart` test `ET?`.

sgrekhov: Same for `up_A14_t*.dart`.

TypeSystem/upper-lower-bounds/up_A15_t02.dart
sgrekhov: Added (and in `up_A14.t02.dart` as well).

sgrekhov (commented):
Thanks. Done.

TypeSystem/upper-lower-bounds/up_A15_t01.dart
eernstg: Oh, I'm sorry, `ET` is a potentially nullable type, that is, it is 'not non-nullable', which means that it goes together with `ET?` in the test about 'not non-nullable' types.

eernstg (commented):
Very last adjustment: I got the placement of the `ET` tests wrong: This _is_ a 'not non-nullable' type, and it was already correctly located together with the other 'not non-nullable' type tests.

TypeSystem/upper-lower-bounds/up_A15_t02.dart
eernstg: Thanks!

sgrekhov:
No problem. Moved back.

eernstg (approved):
Long process, thanks for enduring it! ;-)

LGTM

sgrekhov:
Thank you too! :)

PR #4195, opened 2026-08-19
#4173. Add tests for bounded type variable

sgrekhov:


TypeSystem/upper-lower-bounds/up_A19_t01.dart
```diff
@@ -0,0 +1,108 @@
...
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`X1 extends B1`, `T2`) =
+///   - `T2` if `X1 <: T2`
+///   - otherwise `X1` if `T2 <: X1`
+///   - otherwise UP(`B1a`, `T2`) where `B1a` is the greatest closure of `B1`
+///     with respect to `X1`, as defined in inference.md.
+///
+/// @description Check that UP(`X1 extends B1`, `T2`) = `T2` if `X1 != T2`, none
+/// of `X1` and `T2` is TOP, BOTTOM, NULL, OBJECT, nullable or an intersection
```
eernstg: When it comes to 'nullable' we have to say that `T2` isn't a type of the form `T?`. For example, `FutureOr<int?>` is nullable, but it isn't of the form `T?`.

eernstg: A similar adjustment is needed in subsequent descriptions.

TypeSystem/upper-lower-bounds/up_A19_t01.dart
```diff
@@ -0,0 +1,108 @@
...
+  var v = (1 > 2) ? x1 : t2; // UP(X1 extends E, Enum) = Enum
+  v.expectStaticType<Exactly<Enum>>();
+}
+
+void f13<X1 extends ET2>(X1 x1, ET t2) {
+  var v = (1 > 2) ? x1 : t2; // UP(X1 extends ET2, ET) = ET
+  v.expectStaticType<Exactly<ET>>();
+}
+
+void f14<X1 extends FutureOr<int>>(X1 x1, FutureOr<num> t2) {
```
eernstg, line 88: It would be interesting to see how `FutureOr<int?>` and `FutureOr<num?>` are treated, just so we have the case where we use a nullable type which is not of the form `T?` for any `T`.

This could be a new `f15` (I'm not sure, but perhaps it yields a slightly greater coverage when we have both).

TypeSystem/upper-lower-bounds/up_A19_t02.dart
```diff
@@ -0,0 +1,96 @@
...
+/// ...
+/// - UP(`X1 extends B1`, `T2`) =
+///   - `T2` if `X1 <: T2`
+///   - otherwise `X1` if `T2 <: X1`
+///   - otherwise UP(`B1a`, `T2`) where `B1a` is the greatest closure of `B1`
+///     with respect to `X1`, as defined in inference.md.
+///
+/// @description Check that UP(`X1 extends B1`, `T2`) = `X1` if `X1 != T2`, none
+/// of `X1` and `T2` is TOP, BOTTOM, NULL, OBJECT, nullable or an intersection
+/// type, `X1 <: T2` is `false` and `T2 <: X1`.
```
eernstg:
```suggestion
/// type, `X1 <: T2` does not hold, and `T2 <: X1`.
```

TypeSystem/upper-lower-bounds/up_A19_t02.dart
```diff
@@ -0,0 +1,96 @@
...
+/// @author sgrekhov22@gmail.com
+
+import 'dart:async';
+import '../../Utils/static_type_helper.dart';
+import 'up_lib.dart';
+
+void f1<X1 extends num, T2 extends X1>(X1 x1, T2 t2) {
```
eernstg, line 23: That's nice! We know there's no reason to consider a BOTTOM type as `T2` because that would be handled by an earlier rule, and that keeps the current test smaller, which is nice when we know that it is big enough.

TypeSystem/upper-lower-bounds/up_A20_t02.dart
```diff
@@ -0,0 +1,108 @@
...
+/// ...
+/// - UP(`T1`, `X2 extends B2`) =
+///   - `X2` if `T1 <: X2`
+///   - otherwise `T1` if `X2 <: T1`
+///   - otherwise UP(`T1`, `B2a`) where `B2a` is the greatest closure of `B2`
+///     with respect to `X2`, as defined in inference.md.
+///
+/// @description Check that UP(`T1`, `X2 extends B2`) = `T1` if `T1 != X2`, none
+/// of `T1` and `X2` is TOP, BOTTOM, NULL, OBJECT, nullable or an intersection
+/// type, `T1 <: X2` is `false` and `X2 <: T1`.
```
eernstg:
```suggestion
/// type, `T1 <: X2` does not hold, and `X2 <: T1`.
```

TypeSystem/upper-lower-bounds/up_A20_t02.dart
```diff
@@ -0,0 +1,108 @@
...
+
+void f13<X2 extends ET2>(ET t1, X2 x2) {
+  var v = (1 > 2) ? t1 : x2; // UP(ET, X2 extends ET2) = ET
+  v.expectStaticType<Exactly<ET>>();
+}
+
+void f14<X2 extends FutureOr<int>>(FutureOr<num> t1, X2 x2) {
+  var v = (1 > 2) ? t1 : x2; // UP(FutureOr<num>, X2 extends FutureOr<int>) = FutureOr<num>
+  v.expectStaticType<Exactly<FutureOr<num>>>();
+}
```
eernstg, line 91: Again, we should have a variant `f15` that uses `FutureOr<int?>` and `FutureOr<num?>`.

eernstg (commented):
Looks good! Just one case needs to be handled: The prior rules will handle types of the form `T?`, but that isn't 100% the same thing as a nullable type.

TypeSystem/upper-lower-bounds/up_A19_t01.dart
sgrekhov: Updated everywhere.

TypeSystem/upper-lower-bounds/up_A19_t01.dart
sgrekhov: Added `void f15<X1 extends FutureOr<int?>>(X1 x1, FutureOr<num?> t2) {...}`. And also the same for the mirroring rule in up_A20_t02.dart.

TypeSystem/upper-lower-bounds/up_A19_t02.dart
sgrekhov: Fixed.

TypeSystem/upper-lower-bounds/up_A20_t02.dart
sgrekhov: Fixed.

TypeSystem/upper-lower-bounds/up_A20_t02.dart
sgrekhov: Added.

sgrekhov (commented):
Thank you. Updated. PTAL.

TypeSystem/upper-lower-bounds/up_A19_t01.dart
eernstg: Thanks!

TypeSystem/upper-lower-bounds/up_A19_t01.dart
eernstg: Thanks!

eernstg (approved):
LGTM

PR #4196, opened 2026-08-20
#4173. Add tests for function types

sgrekhov:


TypeSystem/upper-lower-bounds/up_A21_t01.dart
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T Function<...>(...)`, `Function`) = `Function`
+///
+/// @description Check that UP(`T Function<...>(...)`, `Function`) = `Function`.
```
eernstg, line 10: We should keep checking that no earlier case could be applied:
```suggestion
/// @description Check that UP(`T Function<...>(...)`, `Function`) = `Function`. Note that
/// `Function` and a function type are not TOP, OBJECT, NULL, or BOTTOM, and they are
/// not intersection types, or of the form `T?`
```
A similar comment should fit in each of these tests.

TypeSystem/upper-lower-bounds/up_A26_t01.dart
```diff
@@ -0,0 +1,97 @@
...
+/// @author sgrekhov22@gmail.com
+
+import 'dart:async';
+import '../../Utils/static_type_helper.dart';
+import 'up_lib.dart';
+
+void f1(int Function() t1, num t2) {
+  var v = (1 > 2) ? t1 : t2; // UP(int Function(), num) = UP(Object, num) = Object
+  v.expectStaticType<Exactly<Object>>();
+  v = confirmObjectContext(); // Check that `v` is not `FutureOr<Object>`
```
eernstg, line 25: Nice!

eernstg (commented):
Looks good! Just one thing: The application of these rules relies on all earlier rules to be inapplicable, but some of these tests do not have a rule (we called it a "lemma") which makes it easy to see that this is indeed true.

sgrekhov:
Updated. Also removed tests for intersection types like
```dart
void test<T1 extends num, T2 extends T1>(
  T1 Function(T1 x) f1,
  T2 Function(T2 x) f2,
) {
  if (f1 is T2 Function(T1 x)) {
    var v1 = (1 > 2) ? f1 : f2;
    v1.expectStaticType<Exactly<T2 Function(T2 x)>>();
  }
}
```
I believe that these tests should be added to the set of tests for intersection types.

Question. Is the test below a valid test for the rule **UP**(`T Function<...>(...)`, `Function`) = `Function` ?

Or it first tests the rule **UP**(`X1 extends B1`, `T2`) = ... ?

```dart
void f9<X extends num>(X Function(X) t1, Function t2) {
  var v = (1 > 2) ? t1 : t2; // UP(X Function(X), Function) = Function
  v.expectStaticType<Exactly<Function>>();
}
```

eernstg:
> these tests should be added to the set of tests for intersection types

Exactly, thanks!

> Is the test below a valid test for the rule **UP**(`T Function<...>(...)`, `Function`) = `Function`?

Yes, the "UP(functionType, Function)" rule you mention is the one which is used first, so that is the best classification.

eernstg (commented):
Looks good! A "lemma" could be added to some descriptions, just so we keep treating the inability to apply an earlier rule explicitly.

sgrekhov:
> Looks good! A "lemma" could be added to some descriptions, just so we keep treating the inability to apply an earlier rule explicitly.

Please clarify. All descriptions in this PR already contain the "lemma".

eernstg (approved):
LGTM

PR #4197, opened 2026-08-24
#4173. Add tests for record types

sgrekhov:


TypeSystem/upper-lower-bounds/up_A30_t01.dart
```diff
@@ -0,0 +1,182 @@
...
+void f7((dynamic, {String s}) t1, (Object?, {Object s}) t2) {
+  var v = (1 > 2) ? t1 : t2;
+  v.expectStaticType<Exactly<(dynamic, {Object s})>>();
+  if (1 > 2) {
+    v.$1.proofDynamic;
+  }
+}
+
+void f8(Rec t1, (num, Object, {Object b}) t2) {
+  var v = (1 > 2) ? t1 : t2; // Rec = (int, String, {bool b})
```
eernstg, line 68: Nice comment, thanks!

TypeSystem/upper-lower-bounds/up_A28_t01.dart
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`(...)`, `Record`) = `Record`
+///
+/// @description Check that UP(`(...)`, `Record`) = `Record`.
```
eernstg, line 10: It would be good to have something like this lemma:
```suggestion
/// @description Check that UP(`(...)`, `Record`) = `Record`. Note that no earlier
/// rule applies because record types and `Record` are never TOP, OBJECT, NULL,
/// BOTTOM, intersection types, types of the form `T?`, type variables, function
/// types, or `Function`.
```

TypeSystem/upper-lower-bounds/up_A30_t01.dart
```diff
@@ -0,0 +1,182 @@
...
+/// ...
+/// - UP(`(S0, ... Sk, {T0 d0, ..., Tn dn})`,
+///      `(S0', ... Sk', {T0' d0, ..., Tn' dn})`) =
+///   `(Q0, ...,Qk, {R0, ..., Rn})` if:
+///     - `Qi` is UP(`Si`, `Si'`)
+///     - `Ri` is UP(`Ti`, `Ti'`)
+///
+/// @description Check that UP of two record types with the same shape is a
+/// record type of that shape where each field type is UP of the corresponding
+/// field types.
```
eernstg, line 16: The lemma would be slightly different here and in tests below because no operand to `UP` is `Record`.

TypeSystem/upper-lower-bounds/up_A32_t01.dart
```diff
@@ -0,0 +1,113 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`(...)`, `T2`) = UP(`Object`, `T2`)
+///
+/// @description Check that UP(`(...)`, `T2`) = UP(`Object`, `T2`) = `Object`
+/// if `T2` is non-nullable and is neither a record type, TOP, BOTTOM, NULL,
+/// OBJECT, function type nor an intersection type.
```
eernstg: The lemma would be needed here, too, even though it's very tempting to think that it is redundant.

TypeSystem/upper-lower-bounds/up_A33_t01.dart
```diff
@@ -0,0 +1,113 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `(...)`) = UP(`T1`, `Object`)
+///
+/// @description Check that UP(`T1`, `(...)`) = UP(`T1`, `Object`) = `Object`
+/// if `T1` is non-nullable and is neither a record type, TOP, BOTTOM, NULL,
+/// OBJECT, function type nor an intersection type.
```
eernstg: Ditto.

eernstg (commented):
Looks good! Could add a bunch of "lemmas".

TypeSystem/upper-lower-bounds/up_A28_t01.dart
sgrekhov: Thank you. Updated everywhere.

TypeSystem/upper-lower-bounds/up_A30_t01.dart
sgrekhov: Removed `Record` from the lemma.

TypeSystem/upper-lower-bounds/up_A32_t01.dart
sgrekhov: Added.

TypeSystem/upper-lower-bounds/up_A33_t01.dart
sgrekhov: Added.

sgrekhov (commented):
Thank you. Updated. PTAL.

eernstg (approved):
LGTM

PR #4200, opened 2026-08-28
#4173. Add tests for `FutureOr` types

sgrekhov:


TypeSystem/upper-lower-bounds/up_A02_t02.dart
```diff
@@ -29,16 +29,30 @@ void f1(dynamic y) {
 void f2(Object? y) {
   var v = (1 > 2) ? getFutureOrVoid() : y; // MORETOP(FutureOr<void>, Object?) = true
   v.expectStaticType<Exactly<FutureOr<void>>>();
+  // v.expectStaticType<Exactly<Object?>>(); also succeeds.
```
eernstg, line 32: Good catch, thanks!

TypeSystem/upper-lower-bounds/up_A03_t02.dart
```diff
@@ -8,14 +8,17 @@
 /// - UP(`T1`, `T2`) = `T1` if TOP(`T1`)
 ///
 /// @description Check that UP(`T1`, `T2`) = `T1` if TOP(`T1`) and
-/// TOP(`T2`) == `false` (which implies `T1 != T2`). Test `void`.
+/// TOP(`T2`) == `false` (which implies `T1 != T2`). Test types `Object?` and
```
eernstg: Let's keep the levels separate:
```suggestion
/// not TOP(`T2`) (which implies `T1 != T2`). Test types `Object?` and
```

TypeSystem/upper-lower-bounds/up_A03_t02.dart
```diff
@@ -8,14 +8,17 @@
...
+/// `FutureOr<Object?>`.
 /// @author sgrekhov22@gmail.com
 
 import 'dart:async';
 import '../../Utils/static_type_helper.dart';
 import 'up_lib.dart';
 
-void f1(Object? o, num n) {
+// TODO(sgrekhov): we should distinguish between `dynamic, `Object?` and `FutureOr<Object?>`
```
eernstg: Typo:
```suggestion
// TODO(sgrekhov): we should distinguish between `dynamic`, `Object?` and `FutureOr<Object?>`
```
It probably doesn't matter because no rendering of the code quote occurs in code, but it still looks better.

### So what's missing?

The implication would presumably be that each of the tests below would need to contain something like `v = confirmObjectContext()` or `v = confirmFutureOrObjectContext()`), except that it should work with `Object?` rather than `Object`, and this kind of test needs to run, and the current library will be rejected at compile time.

I think it's going to be tricky to come up with a way to rule out the type `FutureOr<Object?>` when the type `Object?` is expected.

First, `FutureOr<Object?>` and `Object?` compare as equal according to `==`, and in particular, when reified to `Type` objects in constant expressions.

Next, when `FutureOr<Object?>` is the context type of an expression which is `Future<X>` for an `X` which is inferred, `X` will be inferred as `Object?` (because we're using the `Future` branch of the union `Future<Object?> | Object?`), but type inference _should_ also bind `X` to `Object?` when the context type is a plain `Object?`. It will currently yield `dynamic`, but that should be fixed at some point in the future, which means that we can only rely on this type to be a top type at this time.

Next, `await v` will have static type `Object?` both when the static type of `v` is `Object?` and when it is `FutureOr<Object?>`. This implies that it is not a useful technique to extract the actual type argument of the `FutureOr` type.

Next again, it is also useless to extract the other branch of the `FutureOr` type, because we would then just obtain `Object?`, and we wouldn't know whether this is the other branch of the `FutureOr` type or the original type which was unaffected by the attempt to extract a type argument.

### Desperate moves

We could rely on `await_only_futures` to reveal that an expression `e` does not have static type `FutureOr<S>` for any `S` (because it'll flag `await e` as not awaiting a future). This is not so reliable because it relies on a warning which is by definition not a part of the language (so warnings might change). However, it works for now, if we have a way to enable certain lints for a specific test run.

Do we? ;-)

TypeSystem/upper-lower-bounds/up_A03_t02.dart
```diff
@@ -8,14 +8,17 @@
...
+
+void f1a(Object? o, num n) {
   var v = (1 > 2) ? o : n;
   v.expectStaticType<Exactly<Object?>>();
   v?.checkNotDynamic;
```
eernstg: I think it's safer to use `v.checkNotDynamic`. When the static type of `v` is `dynamic` this will compile (no null-check required). When the static type of `v` is `Object?` it will complain about the need to check for null, and when it's non-nullable it will complain because there is no statically known member whose name is `checkNotDynamic`, but it will always be an error.

On top of that, I'm worried that type inference could in some cases strip off the `Null` part of the type `dynamic` and yield `Object`, in which case the test will not succeed even though the static type of `v` is `dynamic`.

Here is a rule in inference.md that does such things:

> If Q is Q0? the match holds under constraint set C:
> 
> - If P is P0? and P0 is a subtype match for Q0 under constraint set C.
> - Or if P is dynamic or void and Object is a subtype match for Q0 under constraint set C.
> - Or if P is a subtype match for Q0 under non-empty constraint set C.
> - Or if P is a subtype match for Null under constraint set C.
> - Or if P is a subtype match for Q0 under empty constraint set C.

So `v.expectStaticType<Exactly<Object?>>();` will do the right thing and reads more clearly than `v?.expectStaticType<Exactly<Object?>>();`.

TypeSystem/upper-lower-bounds/up_A03_t02.dart
```diff
@@ -24,7 +27,7 @@ void f1(Object? o, num n) {
 // [cfe] unspecified
 }
 
-void f2(FutureOr<Object?> o, num n) async {
+void f1b(FutureOr<Object?> o, num n) async {
   var v = (1 > 2) ? o : n;
   v.expectStaticType<Exactly<Object?>>();
   (await v)?.checkNotDynamic;
```
eernstg: Similar, we want to make sure `await v` does not have type `dynamic`, which is safer without the `?`.

TypeSystem/upper-lower-bounds/up_A03_t02.dart
eernstg: Haha!

There is a way to do it after all:

```dart
import 'dart:async';

// Uncomment one of the following two:
// typedef T = Object?;
// typedef T = FutureOr<Object?>;

X nonNull<X>(X? x) => x as X;

void main() {
  T v = 1 as dynamic; // Prevent immediate promotion of `v`.
  var v1 = nonNull(v);
  Object _ = v1; // OK?
}
```

This will succeed when `T` is `Object?` because `nonNull` will be inferred to return `Object` when the actual argument type is `Object?`, and `Object` is assignable to `Object`. However, it will return `FutureOr<Object?>` when the parameter type is `FutureOr<Object?>` because `FutureOr` doesn't support eliminating the null in general, and the rules do not take into account this special case where it could actually express the resulting non-nullable type. Hence, the initialization of `v1` is an error.

TypeSystem/upper-lower-bounds/up_A03_t02.dart
eernstg: This occurs in a few more tests below.

TypeSystem/upper-lower-bounds/up_A04_t02.dart
```diff
@@ -8,14 +8,18 @@
...
+/// `FutureOr<Object?>`.
 /// @author sgrekhov22@gmail.com
 
 import 'dart:async';
+
 import '../../Utils/static_type_helper.dart';
 import 'up_lib.dart';
 
-void f1(Object? o, num n) {
+// TODO(sgrekhov): we should distinguish between `dynamic, `Object?` and `FutureOr<Object?>`
```
eernstg: Again, the hard case `Object?` vs. `FutureOr<Object?>` should be doable using the non-null trick.

TypeSystem/upper-lower-bounds/up_A04_t02.dart
```diff
@@ -8,14 +8,18 @@
...
+
+void f1a(Object? o, num n) {
   var v = (1 > 2) ? n : o;
   v.expectStaticType<Exactly<Object?>>();
   v?.checkNotDynamic;
```
eernstg: Again, the safe approach is to avoid `?`.

TypeSystem/upper-lower-bounds/up_A04_t02.dart
```diff
@@ -24,16 +28,16 @@ void f1(Object? o, num n) {
 // [cfe] unspecified
 }
 
-void f2(FutureOr<Object?> o, num n) async {
+void f1b(FutureOr<Object?> o, num n) async {
   var v = (1 > 2) ? n : o;
-  v.expectStaticType<Exactly<Object?>>();
-  (await v)?.checkNotDynamic;
-//           ^^^^^^^^^^^^^^^
+  v.expectStaticType<Exactly<FutureOr<Object?>>>();
```
eernstg: Considering that `expectStaticType` cannot distinguish different top types, we can only really test whether it's TOP or not. I think we might as well stick to the style which is used in almost all cases below: Test for `Exactly<Object?>` as the canonical way to test for TOP.

TypeSystem/upper-lower-bounds/up_A04_t02.dart
```diff
@@ -285,35 +289,55 @@ void f30(FutureOr<Object?> o, ET n) async {
...
+  v.expectStaticType<Exactly<Object?>>();
+  v?.checkNotDynamic;
+//   ^^^^^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+}
+
+void f16b(FutureOr<Object?> o, Future<dynamic> n) async {
+  var v = (1 > 2) ? n : o;
+  v.expectStaticType<Exactly<FutureOr<Object?>>>();
```
eernstg:
```suggestion
  v.expectStaticType<Exactly<Object?>>();
```

TypeSystem/upper-lower-bounds/up_A34_t01.dart
```diff
@@ -0,0 +1,140 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`FutureOr<T1>`, `FutureOr<T2>`) = `FutureOr<T3>` where `T3` = UP(`T1`,
+///   `T2`)
+///
+/// @description Check that UP(`FutureOr<T1>`, `FutureOr<T2>`) = `FutureOr<T3>`
+/// where `T3` = UP(`T1`, `T2`).
```
eernstg:
```suggestion
/// where `T3` = UP(`T1`, `T2`), when further: 
/// `FutureOr<T1> != FutureOr<T2>` and neither `FutureOr<T1>` nor
/// nor `FutureOr<T2>` is TOP or OBJECT. Note that `FutureOr<...>`
/// is never BOTTOM, NULL, an intersection type, a type of the form
/// `U?`, a type variable, `Function`, a function type, `Record`, or a
/// record type.
```
Please double-check that we have them all! ;-)  Similar additions may be needed in other tests below.

TypeSystem/upper-lower-bounds/up_A34_t01.dart
```diff
@@ -0,0 +1,140 @@
...
+void f1(FutureOr<int> t1, FutureOr<num> t2) {
+  // UP(FutureOr<int>, FutureOr<num>) = FutureOr<UP(int, num)> = FutureOr<num>
+  var v = (1 > 2) ? t1 : t2;
+  v.expectStaticType<Exactly<FutureOr<num>>>();
+}
+
+void f2(FutureOr<String> t1, FutureOr<int> t2) {
+  // UP(FutureOr<String>, FutureOr<int>) = FutureOr<Object>
+  var v = (1 > 2) ? t1 : t2;
+  v.expectStaticType<Exactly<FutureOr<Object>>>();
```
eernstg: Let's just stick to `Exactly<Object>` as the canonical way to test whether a type is OBJECT. Similar cases occur below.

TypeSystem/upper-lower-bounds/up_A34_t01.dart
```diff
@@ -0,0 +1,140 @@
...
+  // Object and FutureOr<Object> are subtypes of each other, which means that we
+  // can't see the difference using `expectStaticType()` function.
```
eernstg, line 37: It's still good to have this comment!

TypeSystem/upper-lower-bounds/up_A35_t01.dart
```diff
@@ -0,0 +1,279 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`Future<T1>`, `FutureOr<T2>`) = `FutureOr<T3>` where `T3` = UP(`T1`,
+///   `T2`)
+///
+/// @description Check that UP(`Future<T1>`, `FutureOr<T2>`) = `FutureOr<T3>`
+/// where `T3` = UP(`T1`, `T2`).
```
eernstg: Slightly edited conditions and "lemma" needed.

TypeSystem/upper-lower-bounds/up_A35_t01.dart
```diff
@@ -0,0 +1,279 @@
...
+void f1b(Future<num> t1, FutureOr<int> t2) {
+  // UP(Future<num>, FutureOr<int>) = FutureOr<UP(num, int)> = FutureOr<num>
+  var v = (1 > 2) ? t1 : t2;
+  v.expectStaticType<Exactly<FutureOr<num>>>();
+}
+
+void f2a(Future<String> t1, FutureOr<int> t2) {
+  // UP(Future<String>, FutureOr<int>) = FutureOr<Object>
+  var v = (1 > 2) ? t1 : t2;
+  v.expectStaticType<Exactly<FutureOr<Object>>>();
```
eernstg:
```suggestion
  v.expectStaticType<Exactly<Object>>();
```

TypeSystem/upper-lower-bounds/up_A36_t01.dart
```diff
@@ -0,0 +1,278 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`FutureOr<T1>`, `Future<T2>`) = `FutureOr<T3>` where `T3` = UP(`T1`,
+///   `T2`)
+///
+/// @description Check that UP(`FutureOr<T1>`, `Future<T2>`) = `FutureOr<T3>`
+/// where `T3` = UP(`T1`, `T2`).
```
eernstg: Another slightly different requirement and "lemma" needed here.

TypeSystem/upper-lower-bounds/up_A37_t01.dart
```diff
@@ -0,0 +1,141 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `FutureOr<T2>`) = `FutureOr<T3>` where `T3` = UP(`T1`, `T2`)
+///
+/// @description Check that UP(`T1`, `FutureOr<T2>`) = `FutureOr<T3>` where
+/// `T3` = UP(`T1`, `T2`). `T1` is neither TOP, BOTTOM, NULL, OBJECT, a function
+/// type, a record type, `Future`/`FutureOr`, `T?` nor an intersection type.
```
eernstg: We still need to add a couple of extra requirements: `FutureOr<T2>` must not be TOP or OBJECT, etc, and `T1` must not be `Function` or `Record`. Perhaps the safe bet would be to write the requirements from scratch just because that's a systematic process, that might be easier than adding the missing cases to the current text. Or perhaps not, anything that works is good. ;-)

TypeSystem/upper-lower-bounds/up_A37_t01.dart
```diff
@@ -0,0 +1,141 @@
...
+void f1b(num t1, FutureOr<int> t2) {
+  // UP(num, FutureOr<int>) = FutureOr<UP(num, int)> = FutureOr<num>
+  var v = (1 > 2) ? t1 : t2;
+  v.expectStaticType<Exactly<FutureOr<num>>>();
+}
+
+void f2a(String t1, FutureOr<int> t2) {
+  // UP(String, FutureOr<int>) = FutureOr<UP(String, int)> = FutureOr<Object>
+  var v = (1 > 2) ? t1 : t2;
+  v.expectStaticType<Exactly<FutureOr<Object>>>();
```
eernstg:
```suggestion
  v.expectStaticType<Exactly<Object>>();
```

TypeSystem/upper-lower-bounds/up_lib.dart
```diff
@@ -88,3 +79,26 @@ Future<X> confirmFutureOrObjectContext<X>() {
...
+/// void main() {
+///   Object? o1 = 42;
+///   FutureOr<Object?> o2 = 42;
+///
+///   o1 = confirmFutureOrNullableObjectContext(); // Throws
+///   o2 = confirmFutureOrNullableObjectContext(); // Ok
+/// }
+///```
+Future<X> confirmFutureOrNullableObjectContext<X>() {
+  // Confirm that `X` is `Object`.
```
eernstg: That would be `Object?`.

TypeSystem/upper-lower-bounds/up_lib.dart
```diff
@@ -88,3 +79,26 @@ Future<X> confirmFutureOrObjectContext<X>() {
...
+  if (X == NullableObject) {
+    return Future<X>.value(0 as dynamic);
+  }
+  Expect.fail('The context is not `Object?`');
```
eernstg: This would probably be

```suggestion
  Expect.fail('The context is not `FutureOr<Object?>`');
```

TypeSystem/upper-lower-bounds/up_lib.dart
```diff
@@ -88,3 +79,26 @@ Future<X> confirmFutureOrObjectContext<X>() {
...
+  return Future<X>.value(null);
+}
```
eernstg: This is not quite safe. When the context type is `Object?`, `X` is inferred as `dynamic`, but the specification says that it should be inferred as `Object?`.

Hence, we shouldn't have tests here and there which will break if and when that discrepancy is corrected by inferring `Object?`. The language team wants to avoid unneeded occurrences of `dynamic`, in particular invisible ones, so this might very well happen even though it is likely to be quite breaking.

One approach that would work is the `nonNull` trick that I mentioned earlier, which will even work in the analyzer.

TypeSystem/upper-lower-bounds/up_A38_t01.dart
```diff
@@ -0,0 +1,141 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`FutureOr<T1>`, `T2`) = `FutureOr<T3>` where `T3` = UP(`T1`, `T2`)
+///
+/// @description Check that UP(`FutureOr<T1>`, `T2`) = `FutureOr<T3>` where
+/// `T3` = UP(`T1`, `T2`). `T2` is neither TOP, BOTTOM, NULL, OBJECT, a function
+/// type, a record type, `Future`/`FutureOr`, `T?` nor an intersection type.
```
eernstg: Similar to the previous library TypeSystem/upper-lower-bounds/up_A37_t01.dart.

TypeSystem/upper-lower-bounds/up_A38_t01.dart
```diff
@@ -0,0 +1,141 @@
...
+void f1b(FutureOr<num> t1, int t2) {
+  // UP(FutureOr<num>, int) = FutureOr<UP(num, int)> = FutureOr<num>
+  var v = (1 > 2) ? t1 : t2;
+  v.expectStaticType<Exactly<FutureOr<num>>>();
+}
+
+void f2a(FutureOr<String> t1, int t2) {
+  // UP(FutureOr<String>, int) = FutureOr<UP(String, int)> = FutureOr<Object>
+  var v = (1 > 2) ? t1 : t2;
+  v.expectStaticType<Exactly<FutureOr<Object>>>();
```
eernstg:
```suggestion
  v.expectStaticType<Exactly<Object>>();
```

eernstg (commented):
That was a long one!

Lots of things look good. I've mentioned the need to have a more complete set of requirements and "lemmas" in the description, and a few other things.

One thing was quite tricky: How do we distinguish `Object?` from `FutureOr<Object?>` such that we can complain about getting one when the other is expected?

It turns out to be possible via a step that eliminates null from the type: This turns `Object?` into `Object`, but `FutureOr<Object?>` remains the same, so we can do this and then test that the type is or isn't nullable.

I commented on the new function in up_lib.dart. It works for now, but it is dangerous because it relies on a spec-non-conformant behavior that we'll hopefully fix. Soon. Or whatever.  ;-D

TypeSystem/upper-lower-bounds/up_A03_t02.dart
sgrekhov: Updated everywhere.

TypeSystem/upper-lower-bounds/up_A03_t02.dart
sgrekhov: Thank you, Erik!!! After several unsuccessful attempts to distinguish between `dynamic`, `Object?`, and `FutureOr<Object?>`, I gave up and decided to just add a TODO to this PR and come back to the problem after taking a break. But you did it and saved me a lot of time!!!

This is a very nice solution that works statically! Thank you!!!

TypeSystem/upper-lower-bounds/up_A03_t02.dart
sgrekhov: Updated.

TypeSystem/upper-lower-bounds/up_A03_t02.dart
sgrekhov: Updated. Please note that in the `fXb` functions that check the `FutureOr<Object?>` type, I also changed `v.expectStaticType<Exactly<Object?>>();` to `v.expectStaticType<Exactly<FutureOr<Object?>>>();`. Otherwise, it looks like a typo when we first check that the static type is `Object?` and then check that the static type is neither `dynamic` nor `Object?`.

I can revert it if needed, but I don't see any harm in it.

TypeSystem/upper-lower-bounds/up_A04_t02.dart
sgrekhov: Done.

TypeSystem/upper-lower-bounds/up_A04_t02.dart
sgrekhov: Yes!

TypeSystem/upper-lower-bounds/up_A04_t02.dart
sgrekhov: Please see my earlier comment. The tests looks like a typo when we first do `v.expectStaticType<Exactly<Object?>>();` and after that checking that the static type `v` is not `Object?`.

TypeSystem/upper-lower-bounds/up_A34_t01.dart
sgrekhov: Thank you. Updated. Seems nothing was forgotten :)

TypeSystem/upper-lower-bounds/up_A34_t01.dart
sgrekhov: Updated. Added comment that explains it.

TypeSystem/upper-lower-bounds/up_A35_t01.dart
sgrekhov: Added.

TypeSystem/upper-lower-bounds/up_A35_t01.dart
sgrekhov: Fixed.

TypeSystem/upper-lower-bounds/up_A36_t01.dart
sgrekhov: Added.

TypeSystem/upper-lower-bounds/up_A37_t01.dart
sgrekhov: Updated.

TypeSystem/upper-lower-bounds/up_A38_t01.dart
sgrekhov: Updated.

TypeSystem/upper-lower-bounds/up_A37_t01.dart
sgrekhov: Done, please check.

TypeSystem/upper-lower-bounds/up_lib.dart
sgrekhov: Removed. Replaced by `nonNull`.

sgrekhov (commented):
I was expecting this one to be long :) Updated! Thank you for the great nonNull advice! PTAL.

eernstg (approved):
LGTM!

I adjusted several occurrences of `.expectStaticType<Exactly<FutureOr<Object?>>>` to `.expectStaticType<Exactly<Object?>>` in order to maintain a consistent notation for a test which is really about "is this a TOP type?", and similarly for a couple of OBJECT types expressed as `FutureOr<Object>`.

TypeSystem/upper-lower-bounds/up_A03_t02.dart
eernstg: Thanks!

TypeSystem/upper-lower-bounds/up_A03_t02.dart
eernstg: Thanks!

TypeSystem/upper-lower-bounds/up_A03_t02.dart
eernstg: Oh, I overlooked this one, and landed the PR.

I understand that the test for `FutureOr<Object?>` makes sense relative to a consistency check locally in the test itself, which is definitely useful.

However, I still think it is useful to be particularly careful about this kind of test (that is, having many different syntactic shapes with the same meaning, namely "is this type TOP/OBJECT/SOMETHING?"). It might even be helpful to have a standard comment like `// The type is TOP.` or `// The type is OBJECT.` on this particular kind of test. The point is that it is incredibly easy to think that `expectStaticType` can _actually_ see the difference between different TOP/OBJECT/SOMETHING types.

If we make it easy for the reader to understand that we have _only_ tested that the type is TOP or whatever then it makes more sense that the test is immediately followed by another test that allows us to distinguish between the types in the given set of types.

I thought about having a special test like `v.expectStaticTopType()` etc, but I haven't tried to write a declaration that would actually work.

This doesn't work, but it might be possible to morph it into something that works:

```dart
extension ExpectStaticTopType on Object? {
  void expectStaticTopType() {}
}

extension ExpectStaticTopTypeFail<X> on X {
  void get expectStaticTopType => null;
}
```

sgrekhov: Thank you! Let's leave it as is.

PR #4201, opened 2026-08-31
#4173. Add more tests for UP function

sgrekhov:


TypeSystem/upper-lower-bounds/up_A39_t01.dart
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// @assertion We define the upper bound of two types `T1` and `T2` to be
+/// UP(`T1`,`T2`) as follows.
+/// ...
+/// - UP(`T1`, `T2`) = `T2` if `T1` <: `T2`
+///   Note that both types must be class types at this point.
```
eernstg, line 9: This means class types, mixin types, mixin class types, enum types, extension type types. However, we're just testing `UP`, not `<:`, so it's good that we don't include all those other cases.

eernstg (approved):
LGTM

PR #4206, opened 2026-09-02
#4173. Add test for distinct interface types

sgrekhov:


eernstg (approved):
LGTM.

We should also have some cases where the shared supertype graph contains some types that aren't alone at their level, but that'll be another PR.

sgrekhov:
> We should also have some cases where the shared supertype graph contains some types that aren't alone at their level, but that'll be another PR.

Please clarify. What types do you mean?

eernstg:
For example,

```dart
class A;
class I implements A;
class J implements A;
class B1 implements I, J;
class B2 implements I, J;

void test(B1 b1, B2 b2) {
  var v = 1 > 2 ? b1 : b2;
  v.expectStaticType<Exactly<A>>;
}
```
In this example, the `UP` of `B1` and `B2` cannot be `I` and it cannot be `J` because they are both at level 3 (`A` is at level 2, `Object` at 1, `Object?` at 0, that is, the greatest number of nominal subtype steps needed to reach `Obejct?`). In general, we take the deepest level which is a singleton; in this case it's `A`.

PR #4207, opened 2026-09-02
#3182. Update tests for augmenting enums

sgrekhov:


LanguageFeatures/Augmentations/augmenting_enums_A05_t01.dart
```diff
@@ -2,9 +2,14 @@
...
+/// introductory and complete declarations of:
+/// - `int get index`
+/// - `int get hashCode`
+/// - `bool operator ==(Object)`
+/// - `static const List<E> values;` where `E` is the enum type.
+/// For ordering purposes, these implicit declarations are before any members
+/// declared in the declaration.
 ///
 /// @description Checks that it is a compile-time error to augment enum's
 /// `values` getter.
```
eernstg: I don't think that's true any more, it just can't have a complete augmenting declaration because the implicitly induced one is already complete.

LanguageFeatures/Augmentations/augmenting_enums_A05_t01.dart
eernstg: This would indeed be an error, but the test should probably have two cases (or we should have two tests): One where `values` is augmented, which is fine because all augmenting declarations are incomplete, and another one which has an error because `values` is augmented, and one of the augmenting declarations is complete. Two complete augmenting declarations is not interesting because this means that we have three of them in total, and that would be tested elsewhere, or perhaps not at all because we never went so deep&mdash;which is fine because we do need to stop at some point.

LanguageFeatures/Augmentations/augmenting_enums_A05_t03.dart
```diff
@@ -2,12 +2,17 @@
...
+/// - `int get hashCode`
+/// - `bool operator ==(Object)`
+/// - `static const List<E> values;` where `E` is the enum type.
+/// For ordering purposes, these implicit declarations are before any members
+/// declared in the declaration.
 ///
-/// @description Checks that it is a compile-time error if an enum's
-/// augmentation declares `index`, `hashCode` or `==` members.
+/// @description Checks that it is a compile-time error if augmentations of
+/// enum's `index`, `hashCode` or `==` members are complete.
```
eernstg: Same test as LanguageFeatures/Augmentations/augmenting_enums_A05_t02.dart? However, the actual declarations are not augmenting, so maybe the description should instead be adjusted to say that we can't have a second introductory declaration.

It's useful to test this specifically for the `index` etc. of an enum declaration because it's a specialized language behavior that `index` etc are implicitly induced. In other words, it's not a duplication of other tests where we test that it's an error to have >1 complete declaration in a regular (explicitly declared) augmentation chain.

LanguageFeatures/Augmentations/augmenting_enums_A05_t01.dart
eernstg: So this one should not be an error. (Unless the location of the error message is really confusing and it's actually complaining about line 38 but showing this complaint at line 26).

LanguageFeatures/Augmentations/augmenting_enums_A05_t04.dart
eernstg, line 46: This is not new, but it's still a good catch! ;-)

LanguageFeatures/Augmentations/augmenting_enums_A05_t01.dart
eernstg: Just saw LanguageFeatures/Augmentations/augmenting_enums_A05_t06.dart which is the test where the augmentations are all incomplete. So there's no need to add such a test.

However, the description in this test would still need to be clarified, probably by adding "something something complete".

LanguageFeatures/Augmentations/augmenting_enums_A05_t01.dart
eernstg: Is it really true that the tools report an error for this declaration? Seems unnecessary...

LanguageFeatures/Augmentations/augmenting_enums_A05_t07.dart
```diff
@@ -0,0 +1,85 @@
...
+/// @author sgrekhov22@gmail.com
+
+// SharedOptions=--enable-experiment=augmentations
+
+enum E1 {
+  e0;
+}
+
+augment enum E1 {
+  values;
```
eernstg, line 25: This is not an instance member. I think it would be fine to just mention the static case in the description as well as the instance case.

eernstg (commented):
Added a few comments.

LanguageFeatures/Augmentations/augmenting_enums_A05_t03.dart
sgrekhov: Description updated.

LanguageFeatures/Augmentations/augmenting_enums_A05_t07.dart
sgrekhov: Description updated.

LanguageFeatures/Augmentations/augmenting_enums_A05_t01.dart
sgrekhov: Removed this test add all. Added testing of `values` to other tests where we are testing `index`, `hashCode` and `==` members.

sgrekhov (commented):
Thank you. Updated. PTAL.

LanguageFeatures/Augmentations/augmenting_enums_A05_t01.dart
eernstg: Sounds good!

eernstg: (The test is gone.)

LanguageFeatures/Augmentations/augmenting_enums_A05_t07.dart
```diff
@@ -0,0 +1,85 @@
...
+// [cfe] unspecified
+}
+
+enum E5 {
+  e0;
+}
+
+augment enum E5 {
+  ;
+  void set values(int _) {}
```
eernstg, line 73: But it should be OK to add a static setter!

eernstg (commented):
Looks good, but I think we should include the static setter case in A05_t07.

sgrekhov:
A separate test added. PTAL.

eernstg (approved):
LGTM, thanks!

Issue #4203, opened 2026-09-01
Remove `primary-constructors` experimental flag


sgrekhov:
Primary constructors were shipped in 3.13. This experimental flag is not needed any longer.

PR #4210, opened 2026-09-04
#4203. Add tests for DOWN(T, T). Fix UP(T, T) tests

sgrekhov:


TypeSystem/upper-lower-bounds/down_A01_t01.dart
```diff
@@ -0,0 +1,125 @@
...
+import 'dart:async';
+
+import '../../Utils/expect.dart' show Expect;
+import '../../Utils/static_type_helper.dart';
+import 'up_lib.dart';
+
+void f1(void Function(Object x) v1, void Function(Object y) v2) {
+  var v = (1 > 2) ? v1 : v2;
+  // Type of `v` is `UP(void, void) Function(DOWN(Object, Object))`
+  v.expectStaticType<Exactly<void Function(Object)>>();
```
eernstg, line 21: It would be great if we could confirm the OBJECT type as `Object`. I think the following should work:
```dart
  v = (o) => o = confirmObjectContext();
```
I know, we don't dive into structures under test in order to test something about the smaller parts that they are built from, but this is special because we are not interested in the function type for its own sake, we're just wrapping the interesting type in a function type because this allows us to trigger DOWN.

TypeSystem/upper-lower-bounds/down_A01_t01.dart
```diff
@@ -0,0 +1,125 @@
...
+}
+
+void f2(void Function(dynamic x) v1, void Function(dynamic y) v2) {
+  var v = (1 > 2) ? v1 : v2;
+  v.expectStaticType<Exactly<void Function(dynamic)>>();
```
eernstg, line 26: Similarly:
```dart
  v = (d) => 1 > 2 ? d.doesNotExist : null;
```

TypeSystem/upper-lower-bounds/down_A01_t01.dart
eernstg: And so on, for types of the form `void Funcion(S)` where `S` is TOP/OBJECT/NULL/BOTTOM. This would give us a similar level of testing as the level we have for `UP`.

eernstg: Thinking again, we also haven't done this for `UP(T, T)`, so we probably shouldn't do it here.

However, there will soon be other tests involving TOP/OBJECT/NULL/BOTTOM types, and they can do it, just like `UP`. This would make the same kind of testing here redundant.

TypeSystem/upper-lower-bounds/up_A01_t01.dart
```diff
@@ -15,11 +15,6 @@ import '../../Utils/expect.dart' show Expect;
 import '../../Utils/static_type_helper.dart';
 import 'up_lib.dart';
 
-void f1<X extends num, Y extends num>(X x, Y y) {
```
eernstg, line 18: Of course! `X` and `Y` are not the same type at all! ;-D

But shouldn't we have a test case with `X` and `X`?

eernstg (commented):
Looks good!

I wrote a comment about testing group types (TOP/OBJECT/NULL/BOTTOM) for the specific member, but then understood that we shouldn't do that in the `DOWN(T, T)` test because it would be redundant with the tests specifically about those group types. So there's no need to change anything about A01_t01, but I still kept the comments because the reasoning will apply to the tests about those groups.

So there's just one thing: `up_A01_t01.dart` removes a type variable test, but I think we should have a type variable test in that library, it should just be slightly different.

sgrekhov:
Thank you. Updated. `X` and `X` tests added for the both UP and DOWN tests. PTAL.

eernstg (approved):
LGTM

PR #4211, opened 2026-09-07
#4173. Add tests for DOWN and top types

sgrekhov:


TypeSystem/upper-lower-bounds/down_A02_t02.dart
```diff
@@ -0,0 +1,85 @@
...
+void f4(
+  void Function(FutureOr<void>) v1,
+  void Function(FutureOr<Object?>) v2,
+) {
+  // DOWN(FutureOr<void>, FutureOr<Object?>) = FutureOr<Object?>
+  // because MORETOP(FutureOr<Object?>, FutureOr<void>) = false
+  var v = (1 > 2) ? v1 : v2;
+  v.expectStaticType<Exactly<void Function(FutureOr<Object?>)>>();
+  v = (o) {
+    var x = nonNull(o);
```
eernstg: I think it should be possible to see before line 72 that it is indeed `FutureOr<Object?>`. Otherwise it could be a number of things, just not `Object?`. Perhaps a test for `Object?` on `await o` would suffice?

TypeSystem/upper-lower-bounds/down_A02_t03.dart
```diff
@@ -0,0 +1,91 @@
...
+  };
+}
+
+void f4(void Function(void) v1, void Function(FutureOr<Object?>) v2) {
+  // DOWN(void, FutureOr<Object?>) = FutureOr<Object?>
+  // because MORETOP(FutureOr<Object?>, void) = false
+  var v = (1 > 2) ? v1 : v2;
+  v.expectStaticType<Exactly<void Function(FutureOr<Object?>)>>();
+  v = (o) {
+    var x = nonNull(o);
```
eernstg: Similar to TypeSystem/upper-lower-bounds/down_A02_t02.dart.

TypeSystem/upper-lower-bounds/down_A02_t05.dart
```diff
@@ -0,0 +1,91 @@
...
+  };
+}
+
+void f3(void Function(FutureOr<Object?>) v1, void Function(dynamic) v2) {
+  // DOWN(FutureOr<Object?>, dynamic) = FutureOr<Object?>
+  // because MORETOP(dynamic, FutureOr<Object?>) = true
+  var v = (1 > 2) ? v1 : v2;
+  v.expectStaticType<Exactly<void Function(FutureOr<Object?>)>>();
+  v = (o) {
+    var x = nonNull(o);
```
eernstg: Similar to TypeSystem/upper-lower-bounds/down_A02_t02.dart again.

TypeSystem/upper-lower-bounds/down_A02_t05.dart
```diff
@@ -0,0 +1,91 @@
...
+void f4(
+  void Function(FutureOr<Object?>) v1,
+  void Function(FutureOr<dynamic>) v2,
+) {
+  // DOWN(FutureOr<Object?>, FutureOr<dynamic>) = FutureOr<Object?>
+  // because MORETOP(FutureOr<dynamic>, FutureOr<Object?>) = true
+  var v = (1 > 2) ? v1 : v2;
+  v.expectStaticType<Exactly<void Function(FutureOr<Object?>)>>();
+  v = (o) {
+    var x = nonNull(o);
```
eernstg: And again.

TypeSystem/upper-lower-bounds/down_A02_t06.dart
```diff
@@ -0,0 +1,110 @@
...
+import 'up_lib.dart';
+
+void f1(void Function(dynamic) v1, void Function(Object?) v2) {
+  // DOWN(dynamic, Object?) = Object? because MORETOP(Object?, dynamic) = false
+  var v = (1 > 2) ? v1 : v2;
+  v.expectStaticType<Exactly<void Function(Object?)>>();
+  v = (o) async {
+    var x = nonNull(o);
+    Object y = x; // ignore: unused_local_variable
+    v.checkNotDynamic;
```
eernstg: Would this be `o.checkNotDynamic;`? Several cases below.

TypeSystem/upper-lower-bounds/down_A02_t06.dart
```diff
@@ -0,0 +1,110 @@
...
+//    ^^^^^^^^^^^^^^^
+// [analyzer] unspecified
+// [cfe] unspecified
+    (await v).checkNotFutureOrDynamic;
```
eernstg: If this should be `(await o).checkNotFutureOrDynamic` then we will know (presumably, haven't searched for `checkNotFutureOrDynamic`)  that `o` doesn't have static type `FutureOr<FutureOr<dynamic>>`. Wouldn't the most simple test be to ensure that `o` doesn't have static type `FutureOr<dynamic>`? It seems likely that we'd just check `(await v).checkNotDynamic`.

eernstg (commented):

TypeSystem/upper-lower-bounds/down_A02_t06.dart
sgrekhov: Thank you! Fixed here and in other places.

TypeSystem/upper-lower-bounds/down_A02_t06.dart
sgrekhov: Fixed.

TypeSystem/upper-lower-bounds/down_A02_t02.dart
sgrekhov: >  Perhaps a test for Object? on await o would suffice?

I added here and in the ohter places `(await o).expectStaticType<Exactly<Object?>>();`. Please correct me if you meant ahother check for `Object?`.

sgrekhov (commented):
Updated. PTAL. I added check for `Object?` using `expectStaticType<Exactly<Object?>>();`. Probably you meant another one, please let me know then.

TypeSystem/upper-lower-bounds/down_A02_t02.dart
eernstg: Tricky stuff, as always in this domain! ;-)

`(await o)` could be a top type in many cases where `o` is a top type, which makes the test `(await o).expectStaticType<Exactly<Object?>>()` rather weak. We could keep an eye on warnings about "await only futures", but we don't usually do that.

I think it's better to use the `Object?` specific test: Strip off the `?` and then test for `Object`. If we've ruled out `dynamic` and `void` for other reasons then the top types are `T?` where `T` is OBJECT or TOP, plus `FutureOr<T>` where `T` is a top type. Only `T?` where `T` is OBJECT will succeed the object test (this allows `Object?` and `FutureOr<Object>?` and `FutureOr<FutureOr<Object>>?` etc).

Here's an example of testing code that recognizes `FutureOr<Object?>` and rejects at least some similar types:

```dart
X nonNull<X>(X? x) => x as X;

void main() async {
  FutureOr<Object?> ok = 1 as dynamic;
  FutureOr<FutureOr<Object?>> bad1 = 1 as dynamic;
  Object? bad2 = 1 as dynamic;

  // Confirm TOP.
  ok.expectStaticTOP;

  // Check that `ok`'s type is not of the form `T?`.
  var x1 = nonNull(ok);
  x1.expectStaticTOP;
  
  // Check that `ok`'s type is of the form `FutureOr<T?>` where `OBJECT(T)`.
  var x2 = await ok;
  var x3 = nonNull(x2);
  x3.expectStaticOBJECT;

  // ----------------------------------------------------------------------
  // Now the bad cases.

  bad1.expectStaticTOP; // OK.
  bad2.expectStaticTOP; // OK.

  var y1 = nonNull(bad1); // `FutureOr<FutureOr<Object?>>`.
  y1.expectStaticTOP; // OK.
  var y2 = nonNull(bad2); // `Object`.
  y2.expectStaticTOP; // Error. Caught `bad2`!

  var y3 = await bad1; // `FutureOr<Object?>`.
  var y4 = nonNull(y3); // `FutureOr<Object?>`.
  y4.expectStaticOBJECT; // Error. Caught `bad1`!
  var y5 = await bad2; // `Object?`.
  var y6 = nonNull(y5); // `Object?`.
  y6.expectStaticOBJECT; // OK.
}
```

We had some cases using `confirm...`, but I'm a little worried about those because they require execution to detect a failure (I think). So we aren't testing the analyzer with those techniques. So I'm suggesting that we don't use `confirm...` style methods now (and revisit the places where it is being used already).

TypeSystem/upper-lower-bounds/down_A02_t04.dart
```diff
@@ -0,0 +1,85 @@
...
+void f4(
+  void Function(FutureOr<Object?>) v1,
+  void Function(FutureOr<void>) v2,
+) {
+  // DOWN(FutureOr<Object?>, FutureOr<void>) = FutureOr<Object?>
+  // because MORETOP(FutureOr<void>, FutureOr<Object?>) = true
+  var v = (1 > 2) ? v1 : v2;
+  v.expectStaticType<Exactly<void Function(FutureOr<Object?>)>>();
+  v = (o) {
+    var x = nonNull(o);
```
eernstg: Similar again.

TypeSystem/upper-lower-bounds/down_A02_t06.dart
```diff
@@ -0,0 +1,110 @@
...
+  };
+}
+
+void f3(void Function(dynamic) v1, void Function(FutureOr<Object?>) v2) {
+  // DOWN(dynamic, FutureOr<Object?>) = FutureOr<Object?>
+  // because MORETOP(FutureOr<Object?>, dynamic) = false
+  var v = (1 > 2) ? v1 : v2;
+  v.expectStaticType<Exactly<void Function(FutureOr<Object?>)>>();
+  v = (o) async {
+    var x = nonNull(o);
```
eernstg: Yet again.

TypeSystem/upper-lower-bounds/down_A02_t06.dart
```diff
@@ -0,0 +1,110 @@
...
+void f4(
+  void Function(FutureOr<dynamic>) v1,
+  void Function(FutureOr<Object?>) v2,
+) {
+  // DOWN(FutureOr<dynamic>, FutureOr<Object?>) = FutureOr<Object?>
+  // because MORETOP(FutureOr<Object?>, FutureOr<dynamic>) = false
+  var v = (1 > 2) ? v1 : v2;
+  v.expectStaticType<Exactly<void Function(FutureOr<Object?>)>>();
+  v = (o) async {
+    var x = nonNull(o);
```
eernstg: And one more tim.

eernstg (commented):
I think the `(await o).expectStaticType<Exactly<Object?>>` test is too permissive. I've suggested a sequence of statements which can be used to test that a type is actually `FutureOr<Object?>` and not one of the other TOP types. I don't think it's complete (surely I forgot something ;-), but I do think it discriminates more precisely than the current test.

TypeSystem/upper-lower-bounds/down_A02_t02.dart
sgrekhov: Thank you. Seems this reliable. `expectStaticTOP` doesn't detect extension types which are not subtypes of `Object?` and type variables without border but the third check from the above produces compile-time errors for `ET? bad2 = ...` and for `void test<T>(T? bad2) async {...`.

I added to the `expectStaticTOP` description information that it doesn't work for extension types and type variables.

Updated.

TypeSystem/upper-lower-bounds/down_A02_t03.dart
sgrekhov: Updated.

TypeSystem/upper-lower-bounds/down_A02_t04.dart
sgrekhov: Fixed.

TypeSystem/upper-lower-bounds/down_A02_t05.dart
sgrekhov: Fixed.

TypeSystem/upper-lower-bounds/down_A02_t05.dart
sgrekhov: Fixed.

sgrekhov (commented):
Thank you. Updated. The suggested schema works with extension types that don't implement `Object?` and type variables with no borders as well. I added to `expectStaticTOP` info that extension types and type variables are exceptions. PTAL.

If this PR is OK I'll update the places which currently use `confirm...`.

TypeSystem/upper-lower-bounds/down_A02_t03.dart
```diff
@@ -0,0 +1,100 @@
...
+void f4(void Function(void) v1, void Function(FutureOr<Object?>) v2) {
+  // DOWN(void, FutureOr<Object?>) = FutureOr<Object?>
+  // because MORETOP(FutureOr<Object?>, void) = false
+  var v = (1 > 2) ? v1 : v2;
+  v.expectStaticType<Exactly<void Function(FutureOr<Object?>)>>();
+  v = (o) async {
+    // Now check that `o` is really `FutureOr<Object?>` not, say, `Object?`.
+    // Object? and FutureOr<Object?> cannot be distinguished using `expectStaticType`.
+
+    // Confirm TOP type
```
eernstg, line 66: Thanks for repeating these comments. This kind of checking is so subtle that future readers will surely be happy about having them!

eernstg (approved):
LGTM, thanks!

TypeSystem/upper-lower-bounds/down_A02_t02.dart
eernstg: Sounds good!

sgrekhov: > We had some cases using confirm..., but I'm a little worried about those because they require execution to detect a failure (I think). So we aren't testing the analyzer with those techniques. So I'm suggesting that we don't use confirm... style methods now (and revisit the places where it is being used already).

@eernstg I tried to replace confirm... with static extensions, but there is a problem. We need to somehow distinguish between `Object` and `FutureOr<Object>`. For now, `confirm...` is the only way to do this. I’ve tried different ways to do it statically, but none of them have worked. If you have any advice, please let me know.

PR #4216, opened 2026-09-10
#4173. Static detection of TOP and OBJECT types

sgrekhov:
This is an example of a solution for statically distinguishing between different TOP and OBJECT types. The key to this solution is the table in README.md, which defines how many checks we should perform. Please review it. If this solution is acceptable, I'll update all our UP and DOWN tests to use it.


eernstg (approved):
Cool!

LGTM

